PR #10, opened 2015-02-15
Improve best_version description

yaccz:
with a meaningful one


zmedico:
We should also sync up best_version docstring inside bin/portageq.

dol-sen:
Sorry, I somehow was not watching this repo...

I've updated both phase-helpers.sh and portageq with a common description.
https://github.com/gentoo/portage/commit/36f6378522faff389eb94ac1111936e3328a6688

Since it is not an atom we went with different wording.  It should be clear now.

Thank you

PR #14, opened 2015-08-10
repoman commit: Prefix commit message template with cat/PN

mgorny:
Tried to make it clean yet helpful. Long story short:

* if committing inside package, prepends `app-foo/bar:`,
* if committing inside category, prepends `app-foo:`,
* if committing the whole repo, doesn't add any prefix.

I've also made it abort the commit if user leaves commit message unchanged (consisting only of the prefix). And I've added extra empty line for readability.


mgorny:
It's merged already, you blind github.

PR #17, opened 2015-10-15
Add a deprecation warning for ruby 1.9

mrueg:
Ruby 1.9 was package masked and will be removed soon. Let's extend repoman's deprecation warning to it.


mgorny:
Please submit to gentoo-portage-dev@lists.gentoo.org. Thanks ;-P.

dol-sen:
Seems this patch never made it to the list.

Applied in commit: be57961c1115cf4281a00503fa6a9711f474c38a

PR #18, opened 2015-11-03
update egencache outdated header

den4ix:
egencache --update-changelogs writes into
the ChangeLog header the text $Header: $
instead of $Id$, should be updated


zmedico:
We've had similar patch submitted:
https://archives.gentoo.org/gentoo-portage-dev/message/d3ecb70fa86193b74321690f84eaaffb

mgorny:
And Infra doesn't want `$Id$` there since the file is not going to be VCS-controlled.

PR #23, opened 2016-01-24
Add sync-git-clone-extra-opts and sync-git-pull-extra-opts to repos.conf

konsolebox:
This feature would allow adding options to 'git clone' and 'git pull' when synchronizing a portage repo.


konsolebox:
Looks like this was already merged.

PR #27, opened 2016-05-18
install-qa-check.d/60openrc: make sure initscrips aren't using #!/sbin/runscript

austin987:
Now that everything in the tree uses openrc-run, block using runscript

### DIFF
--- a/bin/install-qa-check.d/60openrc
+++ b/bin/install-qa-check.d/60openrc
@@ -12,6 +12,12 @@ openrc_check() {
 			if [[ ${d} == /etc/init.d && ${i} != *.sh ]] ; then
 				# skip non-shell-script for bug #451386
 				[[ $(head -n1 "${i}") =~ ^#!.*[[:space:]/](runscript|sh)$ ]] || continue
+				if [[ $(head -n1 "${i}") == '#!/sbin/runscript' ]] ; then
+					eqawarn "QA Notice: #!/sbin/runscript is deprecated, use #!/sbin/openrc-run instead:"
+					while read -r ;
+						do eqawarn "   ${REPLY}"
+					done <<<  ${i#${ED/%//}}
+				fi
 			fi
 			bash -n "${i}" || die "The init.d file has syntax errors: ${i}"
 		done


bin/install-qa-check.d/60openrc
```diff
@@ -33,6 +33,21 @@ openrc_check() {
 			done
 		done
 	fi
+
+	# Make sure deprecated #!/sbin/runscript isn't being used:
+	if [[ ! -d ${ED}/etc/init.d ]] ; then
+		continue
+	fi
+
+	for x in ${ED}/etc/init.d/* ; do
```
mgorny: Quote ED here.

bin/install-qa-check.d/60openrc
```diff
@@ -33,6 +33,21 @@ openrc_check() {
...
+		if [[ -f ${ED}/etc/init.d/* ]] ; then
+			pass
```
mgorny: bash: pass: command not found

;-)

bin/install-qa-check.d/60openrc
```diff
@@ -33,6 +33,21 @@ openrc_check() {
...
+		fi
+
+		if [[ "$(head -n1 "${x}")" =~ "#!/sbin/runscript" ]] ; then
```
mgorny: btw you could use `read < ...` to avoid subshelling and calling external tools.

bin/install-qa-check.d/60openrc
```diff
@@ -33,6 +33,21 @@ openrc_check() {
...
+			die "#!/sbin/runscript is deprecated, use #!/sbin/openrc-run instead"
```
mgorny: Do you really think users should be disallowed to install any ebuild that uses runscript? I don't think having it fatal is a good idea.

bin/install-qa-check.d/60openrc
austin987: Fixed, thanks.

bin/install-qa-check.d/60openrc
austin987: This is why I shouldn't code at 2:30 AM ;). Fixed.

bin/install-qa-check.d/60openrc
austin987: I went for consistency with the earlier runscript header check (and IMHO it's less readable for a single if statement versus a loop statement.

bin/install-qa-check.d/60openrc
austin987: I meant to change that to a warning before submitting, again, thanks for pointing that out. Fixed.

austin987:
Updated @mgorny

bin/install-qa-check.d/60openrc
```diff
@@ -33,6 +33,21 @@ openrc_check() {
 			done
 		done
 	fi
+
+	# Make sure deprecated #!/sbin/runscript isn't being used:
+	if [[ ! -d "${ED}"/etc/init.d ]] ; then
```
mgorny: No quoting needed inside [[ ]].

bin/install-qa-check.d/60openrc
```diff
@@ -33,6 +33,21 @@ openrc_check() {
...
+		continue
+	fi
+
+	for x in "${ED}"/etc/init.d/* ; do
+		if [[ -f "${ED}"/etc/init.d/* ]] ; then
+			:
```
mgorny: This is still a meaningless statement. I think you really want to use nullglob there.

bin/install-qa-check.d/60openrc
austin987: Again, it's done earlier in the file (and more clear IMO to match other uses of the same variable).

austin987:
@mgorny updated

austin987:
@mgorny ping

bin/install-qa-check.d/60openrc
```diff
@@ -33,6 +33,23 @@ openrc_check() {
 			done
 		done
 	fi
+
+	# Make sure deprecated #!/sbin/runscript isn't being used:
+	if [[ ! -d "${ED}"/etc/init.d ]] ; then
+		continue
```
zmedico: It looks like you meant use to use "return" instead of "continue" here? I think it's nicer to just nest your other code inside like:
```
if [[ -d ${ED}/etc/init.d ]] ; then
    shopt -s nullglob
    for x in "${ED}"/etc/init.d/* ; do
    ...
fi
```

bin/install-qa-check.d/60openrc
```diff
@@ -33,6 +33,23 @@ openrc_check() {
...
+	fi
+
+	shopt -s nullglob
+	for x in "${ED}"/etc/init.d/* ; do
+		if [[ -f ${x} ]] ; then
+			:
+		fi
+
+		if [[ $(head -n1 ${x}) =~ ^#!/sbin/runscript$ ]] ; then
```
zmedico: It's nicer to quote ${x} I think, even though nothing should contain spaces, it could in theory.

zmedico: Actually, ${x} includes ${ED}, so quoting is even more proper.

bin/install-qa-check.d/60openrc
```diff
@@ -33,6 +33,23 @@ openrc_check() {
...
+			eqawarn "${x}: #!/sbin/runscript is deprecated, use #!/sbin/openrc-run instead"
```
zmedico: You can use ${x#${ED%/}} here, so the path isn't unnecessarily long.

bin/install-qa-check.d/60openrc
```diff
@@ -33,6 +33,23 @@ openrc_check() {
...
+	# Make sure deprecated #!/sbin/runscript isn't being used:
+	if [[ ! -d "${ED}"/etc/init.d ]] ; then
+		continue
+	fi
+
+	shopt -s nullglob
+	for x in "${ED}"/etc/init.d/* ; do
+		if [[ -f ${x} ]] ; then
+			:
+		fi
```
zmedico: The above if/then/fi doesn't actually do anything.

bin/install-qa-check.d/60openrc
austin987: Good point, thanks.

bin/install-qa-check.d/60openrc
austin987: Debugging leftover, removed.

bin/install-qa-check.d/60openrc
austin987: OK.

bin/install-qa-check.d/60openrc
austin987: Sure.

austin987:
@zmedico updated.

zmedico:
lgtm

bin/install-qa-check.d/60openrc
```diff
@@ -33,6 +33,18 @@ openrc_check() {
 			done
 		done
 	fi
+
+	# Make sure deprecated #!/sbin/runscript isn't being used:
+	# Note: $ED contains trailing slash
+	if [[ -d "${ED}"etc/init.d ]] ; then
+		shopt -s nullglob
+		for x in "${ED}"etc/init.d/* ; do
```
zmedico: Instead of x, we can re-use the local i variable used earlier in this function.

bin/install-qa-check.d/60openrc
```diff
@@ -33,6 +33,18 @@ openrc_check() {
...
+			if [[ $(head -n1 "${x}") =~ ^#!/sbin/runscript$ ]] ; then
```
zmedico: Maybe it makes sense to use plain == comparison here, since the regex doesn't really gain us anything at this time.

bin/install-qa-check.d/60openrc
```diff
@@ -33,6 +33,18 @@ openrc_check() {
...
+				eqawarn "${x#${ED%/}}: #!/sbin/runscript is deprecated, use #!/sbin/openrc-run instead"
```
zmedico: For consistency,  should probably change this message to have a "QA Notice:" prefix. We could format it like the others, where there's a banner followed by a list of files.

bin/install-qa-check.d/60openrc
zmedico: Also, we do this like the other init.d loop above:
```
[[ -e ${i} ]] || continue
[[ -L ${i} ]] && continue
```

austin987:
Updated.

zmedico:
Merged in 6296c4363abf36276bc49a28cd4635437757aa6e. Thanks!

PR #38, opened 2016-07-28
New changes metamanifest

aeroniero33:
This PR is for a GSoC project based on [GLEP:58](https://wiki.gentoo.org/wiki/GLEP:58)

Added more `MANIFEST_IDENTIFIERS` in `pym/portage/const.py` according to [GLEP:60](https://wiki.gentoo.org/wiki/GLEP:60)

Modified `guessManifestType()` for the new `MANIFEST_IDENTIFIERS` and reduced the return statements

Updated some variable names in `/pym/portage/manifest.py`

Added a new file `/pym/portage/meta-manifest.py` which includes a subclass of Manifest for manifest files outside of the packages (ex. /profiles, categories etc)

Added 4 `create_` functions for it, `create_cat()` that creates a metamanifest for categories that includes the manifest files of all packages after verifying them, `create_profile` and `create_eclass` that create a metamanifest for the `/profiles` and `/eclass` directories, and `create_master` that creates a metamanifest file that includes all the metamanifests above and all files in the root directory.

Added signature and manifest verification to the metamanifest files which verifies package manifests only if a boolean is passed and ignores missing categories or packages if the whole directory is missing.

Modified the `checkFileHashes()` method in the subclass to ignore `DIST` files because otherwise package manifests wouldn't be verifiable. 

Added some methods to help me identify things like the repolevel, reporoot and repotype.

Modified the `create()` function to choose which `create_` function to use based on the repotype.

Instructions on how to use the script: https://goo.gl/i0LIay
Test cases: https://wiki.gentoo.org/wiki/User:Aeroniero

### DIFF
--- a/pym/portage/manifest.py
+++ b/pym/portage/manifest.py
@@ -57,18 +57,27 @@ def manifest2AuxfileFilter(filename):
 def manifest2MiscfileFilter(filename):
 	return not (filename == "Manifest" or filename.endswith(".ebuild"))
 
-def guessManifestFileType(filename):
+def guessManifestFileType(filename, is_pkg = True):
 	""" Perform a best effort guess of which type the given filename is, avoid using this if possible """
 	if filename.startswith("files" + os.sep + "digest-"):
-		return None
+		ftype = None
 	if filename.startswith("files" + os.sep):
-		return "AUX"
+		ftype = "AUX"
 	elif filename.endswith(".ebuild"):
-		return "EBUILD"
+		ftype = "EBUILD"
+	elif filename.endswith("Manifest"):
+		ftype = "MANIFEST"
+	elif filename.endswith(".eclass"):
+		ftype = "ECLASS"
 	elif filename in ["ChangeLog", "metadata.xml"]:
-		return "MISC"
+		ftype = "MISC"
+	elif filename.endswith(".sh"):
+		ftype = "EXEC"
+	elif is_pkg:
+		ftype = "DIST"
 	else:
-		return "DIST"
+		ftype = "OTHER"
+	return ftype
 
 def guessThinManifestFileType(filename):
 	type = guessManifestFileType(filename)
--- a/pym/portage/manifest.py
+++ b/pym/portage/manifest.py
@@ -166,8 +166,8 @@ class Manifest(object):
 			list(self.hashes) if hashname not in hashfunc_map)
 		self.hashes.add("size")
 		self.hashes.add(MANIFEST2_REQUIRED_HASH)
-		for t in MANIFEST2_IDENTIFIERS:
-			self.fhashdict[t] = {}
+		for ftype in MANIFEST2_IDENTIFIERS:
+			self.fhashdict[ftype] = {}
 		if not from_scratch:
 			self._read()
 		if fetchlist_dict != None:
@@ -458,6 +458,7 @@ class Manifest(object):
 		distfiles to raise a FileNotFound exception for (if no file or existing
 		checksums are available), and defaults to all distfiles when not
 		specified."""
+
 		if not self.allow_create:
 			return
 		if checkExisting:
--- a/pym/portage/meta-manifest.py
+++ b/pym/portage/meta-manifest.py
@@ -164,11 +164,10 @@ class MetaManifest(manifest.Manifest):
 				except UnicodeDecodeError:
                         	        continue
 				fpath = os.path.join(eclass_dir, f)
-				ftype = guessManifestFileType(fpath)
+				ftype = guessManifestFileType(fpath, is_pkg=False)
 				f = fpath.replace(self.pkgdir, "")
-				if not f.endswith("MetaManifest"):
+				if not f.endswith("Manifest"):
 					self.fhashdict[ftype][f] = perform_multiple_checksums(fpath, self.hashes)
-		print(self.fhashdict, 10)
 
 	def create_cat(self):
 		'''Creates a MetaManifest file in the selected category'''
--- a/pym/portage/meta-manifest.py
+++ b/pym/portage/meta-manifest.py
@@ -172,19 +172,19 @@ class MetaManifest(manifest.Manifest):
 	def create_cat(self):
 		'''Creates a MetaManifest file in the selected category'''
 		catdir = self.pkgdir
-		self.fhashdict = {}
-		self.fhashdict["MANIFEST"] = {}
-		for catdir, catdir_dir, pkg_files in os.walk(catdir):
-			for f in pkg_files:
-				try:
-					f = _unicode_decode(f,encoding=_encodings['fs'], errors='strict')
-					catdir = _unicode_decode(catdir,encoding=_encodings['fs'], errors='strict')
-				except UnicodeDecodeError:
-					continue
-				if f == "Manifest" :
-					fpath = os.path.join(catdir, f)
-					f = fpath.replace(self.pkgdir, "")
-					self.fhashdict["MANIFEST"][f] = perform_multiple_checksums(fpath, self.hashes) 
+		for ftype in MANIFEST2_IDENTIFIERS:
+                        self.fhashdict[ftype] = {}
+		for pkg_dir in os.listdir(catdir):
+			pkg_dir = os.path.join(catdir, pkg_dir)
+			if os.path.isdir(pkg_dir):
+				for file in os.listdir(pkg_dir):
+					fpath = os.path.join(pkg_dir, file)
+					if os.path.isfile(fpath) and fpath.endswith('Manifest'):
+						new_mf = MetaManifest(fpath.replace('Manifest', ""))
+						new_mf.verify_manifest()
+						f = fpath.replace(self.pkgdir, "")
+						self.fhashdict["MANIFEST"][f] = perform_multiple_checksums(fpath, self.hashes)
+
 	def create_profile(self):
 		'''Creates a MetaManifest file in the profiles directory'''
 		profiledir = self.pkgdir
--- a/pym/portage/meta-manifest.py
+++ b/pym/portage/meta-manifest.py
@@ -1,0 +1,185 @@
+
+from __future__ import unicode_literals
+
+import argparse
+import errno
+import io
+import logging
+import re
+import stat
+import subprocess
+import sys
+import warnings
+
+import portage
+portage.proxy.lazyimport.lazyimport(globals(),
+        'portage.checksum:hashfunc_map,perform_multiple_checksums,' + \
+                'verify_all,_apply_hash_filter,_filter_unaccelarated_hashes',
+        'portage.repository.config:_find_invalid_path_char',
+        'portage.util:write_atomic,writemsg_level',
+)
+
+from portage import os
+from portage import _encodings
+from portage import _unicode_decode
+from portage import _unicode_encode
+import manifest
+from portage.exception import DigestException, FileNotFound, \
+        InvalidDataType, MissingParameter, PermissionDenied, \
+        PortageException, PortagePackageException
+from const import (MANIFEST1_HASH_FUNCTIONS, MANIFEST2_HASH_DEFAULTS,
+        MANIFEST2_HASH_FUNCTIONS, MANIFEST2_IDENTIFIERS, MANIFEST2_REQUIRED_HASH)
+from portage.localization import _
+
+if sys.hexversion >= 0x3000000:
+	# pylint: disable=W0622
+	_unicode = str
+	basestring = str
+else:
+	_unicode = unicode
+
+def guessManifestFileType(filename):
+        """ Perform a best effort guess of which type the given filename is, avoid using this if possible """
+        if filename.startswith("files" + os.sep + "digest-"):
+                return None
+        elif filename.startswith("files" + os.sep):
+                return "AUX"
+        elif filename.endswith(".ebuild"):
+                return "EBUILD"
+        elif filename == "Manifest" or filename == "MetaManifest":
+                return "MANIFEST"
+        elif filename.endswith(".eclass"):
+                return "ECLASS"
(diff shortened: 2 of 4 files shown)


dol-sen:
the test failure looks like was only the lack of the new gkeys dep... somehting that should be ready by end of GSOC... there are some other changes to add for using it to come.

aeroniero33:
@dastergon

pym/portage/meta-manifest.py
```diff
@@ -0,0 +1,276 @@
...
+						new_mf.verify_manifest(verify_pkgs=verify_pkgs)
+		elif mftype == 'category':
+			for f in self.fhashdict['MANIFEST']:
+				mfpath = os.path.join(mfdir, f.replace("Manifest", ""))
+				if not os.path.exists(mfpath) and self.find_repotype(f) is 'package':
+					self.checkFileHashes('MANIFEST', f, ignoreMissing=True)
+				else:
+					self.checkFileHashes('MANIFEST', f, ignoreMissing=False)
+				if verify_pkgs:
+					new_mfpath = os.path.join(mfdir, f.replace("Manifest", ""))
```
zmedico: That's kind of an odd use of `replace`. This seems more appropriate
```
new_mfpath = os.path.dirname(os.path.join(mfdir, f))
```

pym/portage/manifest.py
```diff
@@ -57,18 +57,27 @@ def manifest2AuxfileFilter(filename):
...
 	if filename.startswith("files" + os.sep + "digest-"):
-		return None
+		ftype = None
 	if filename.startswith("files" + os.sep):
-		return "AUX"
+		ftype = "AUX"
 	elif filename.endswith(".ebuild"):
-		return "EBUILD"
+		ftype = "EBUILD"
+	elif filename.endswith("Manifest"):
```
zmedico: This seems like this could be `elif filename == "Manifest":` instead.

pym/portage/meta-manifest.py
```diff
@@ -0,0 +1,276 @@
...
+		return_code = subprocess.call(gpgcmd)
+		if return_code == os.EX_OK:
+			os.rename(filename + ".asc", filename)
+		else:
+			raise portage.exception.PortageException("!!! gpg exited with '" + str(return_code) + "' status")
+
+	def find_reporoot(self):
+		'''Finds reporoot'''
+		repodir = self.pkgdir
+		while len(set(['metadata', 'eclass', 'profiles']).intersection(os.listdir(repodir))) == 0:
```
zmedico: Since os.listdir is more expensive than stat, and intersection is more expensive than any(), this would be better:
```
while not any(os.path.exists(os.path.join(repodir, name)) for name in ('metadata', 'eclass', 'profiles'))):
```

pym/portage/meta-manifest.py
```diff
@@ -0,0 +1,276 @@
...
+			if myrepo[-1] not in myrepo[:-1]:
+				repodir = repodir.replace(("/" +  myrepo[-1]), "")
+			else:
+				raise("Invalid path")
+		self.reporoot = repodir
+
+	def find_repolevel(self, path):
+		'''Finds repolevel'''
+		self.find_reporoot()
+		path = path.replace(self.reporoot, "")
```
zmedico: This is another odd use of replace. Maybe use this instead:
```
path = path[len(self.reporoot):]
```

pym/portage/meta-manifest.py
```diff
@@ -0,0 +1,276 @@
...
+		myrepo = path.split(os.path.sep)
+		repolevel = len(myrepo)
+		return repolevel
+
+	def find_repotype(self, path):
+		'''Finds the type of the repo directory'''
+		if len(set(['metadata', 'eclass', 'profiles']).intersection(os.listdir(path))) > 0:
```
zmedico:
```
if any(os.path.exists(os.path.join(path, name)) for name in ('metadata', 'eclass', 'profiles')):
```

pym/portage/manifest.py
```diff
@@ -57,18 +57,27 @@ def manifest2AuxfileFilter(filename):
 def manifest2MiscfileFilter(filename):
 	return not (filename == "Manifest" or filename.endswith(".ebuild"))
 
-def guessManifestFileType(filename):
+def guessManifestFileType(filename, is_pkg = True):
```
zmedico: I really don't like this function. Is it really necessary to classify files based on extensions like this?

pym/portage/meta-manifest.py
```diff
@@ -0,0 +1,276 @@
...
+		for file in os.listdir(myrepodir):
+			fpath = os.path.join(myrepodir, file)
+			if os.path.isfile(fpath):
+				if not fpath.endswith('Manifest'):
+					ftype = guessManifestFileType(fpath, is_pkg=False)
+					self.fhashdict[ftype][f] = perform_multiple_checksums(fpath, self.hashes)
+			else:
+				for mf in os.listdir(fpath):
+					mfpath = os.path.join(fpath, mf)
+					if mfpath.endswith('Manifest'):
```
zmedico: This is an abuse of `endswith`. Here's an alternative:
```
mf == 'Manifest':
```

pym/portage/meta-manifest.py
```diff
@@ -0,0 +1,276 @@
...
+						f = mfpath.replace(self.reporoot, "")
```
zmedico: Another abuse of replace. Here's an alternative:
```
f = mfpath[len(self.reporoot):]
```

pym/portage/meta-manifest.py
```diff
@@ -0,0 +1,276 @@
...
+		catdir = self.pkgdir
+		for ftype in MANIFEST2_IDENTIFIERS:
+                        self.fhashdict[ftype] = {}
+		for pkg_dir in os.listdir(catdir):
+			pkg_dir = os.path.join(catdir, pkg_dir)
+			if os.path.isdir(pkg_dir):
+				for file in os.listdir(pkg_dir):
+					fpath = os.path.join(pkg_dir, file)
+					if os.path.isfile(fpath) and fpath.endswith('Manifest'):
+						new_mf = MetaManifest(fpath.replace('Manifest', ""))
```
zmedico: Another abuse of `replace`.

pym/portage/meta-manifest.py
```diff
@@ -0,0 +1,276 @@
...
+		'''Creates a MetaManifest file in the selected category'''
+		catdir = self.pkgdir
+		for ftype in MANIFEST2_IDENTIFIERS:
+                        self.fhashdict[ftype] = {}
+		for pkg_dir in os.listdir(catdir):
+			pkg_dir = os.path.join(catdir, pkg_dir)
+			if os.path.isdir(pkg_dir):
+				for file in os.listdir(pkg_dir):
+					fpath = os.path.join(pkg_dir, file)
+					if os.path.isfile(fpath) and fpath.endswith('Manifest'):
```
zmedico: Another abuse of `endswith`. Alternative:
```
if file == 'Manifest' and os.path.isfile(fpath):
```
Do the string comparison first, in order to skip the stat call if the string doesn't match.

pym/portage/meta-manifest.py
```diff
@@ -0,0 +1,276 @@
...
+		mftype = self.find_repotype(mfdir)
+		if mftype == 'root':
+			for ftype in MANIFEST2_IDENTIFIERS:
+				for f in self.fhashdict[ftype]:
+					if ftype != 'MANIFEST':
+						self.checkFileHashes(ftype, f, ignoreMissing=False)
+					else:
+						mfpath = os.path.join(mfdir, f.replace("Manifest", ""))
+						missing = False
+						if not os.path.exists(mfpath) and self.find_repotype(f) is 'category':
```
zmedico: Do we really have to call this `find_repotype` method in the loop? If the result is the same for each iteration of the loop, we should call it before the loop.

pym/portage/meta-manifest.py
```diff
@@ -0,0 +1,185 @@
...
+                """ Returns the absolute path to the Manifest file for this instance """
+                return os.path.join(self.pkgdir, "MetaManifest")
+
+	def sign(self):
+		'''Signs MetaManifest file with the default PORTAGE_GPG_KEY''' 
+		filename = self.getFullname()
+		portage_settings = portage.config(clone=portage.settings)
+		portage_portdbapi = portage.portdbapi(portage_settings)
+		gpgcmd = portage_settings.get("PORTAGE_GPG_SIGNING_COMMAND")
+		if gpgcmd in [None, '']:
```
EvaSDK: note that this is equivalent to: ```if not gpgcmd:```

pym/portage/meta-manifest.py
```diff
@@ -0,0 +1,185 @@
...
+			raise portage.exception.MissingParameter(
+				"PORTAGE_GPG_SIGNING_COMMAND is unset! Is make.globals missing?")
+		if "${PORTAGE_GPG_KEY}" in gpgcmd and "PORTAGE_GPG_KEY" not in portage_settings:
+			raise portage.exception.MissingParameter("PORTAGE_GPG_KEY is unset!")
+		if "${PORTAGE_GPG_DIR}" in gpgcmd:
+			if "PORTAGE_GPG_DIR" not in portage_settings:
```
EvaSDK: Note that portage.config expose a `dict` like interface so you can just use:
```python
if "${PORTAGE_GPG_DIR}" in gpgcmd:
    portage_settings['PORTAGE_GPG_DIR'] = os.path.expanduser(
        portage_settings.get('PORTAGE_GPG_DIR', '~/.gnupg')
    )
```
instead.

pym/portage/meta-manifest.py
```diff
@@ -0,0 +1,185 @@
...
+					portage_settings["PORTAGE_GPG_DIR"])
+		gpgvars = {"FILE": filename}
+		for k in ("PORTAGE_GPG_DIR", "PORTAGE_GPG_KEY"):
+			v = portage_settings.get(k)
+			if v is not None:
+				gpgvars[k] = v
+		gpgcmd = portage.util.varexpand(gpgcmd, mydict=gpgvars)
+		gpgcmd = portage.util.shlex_split(gpgcmd)
+		gpgcmd = [portage._unicode_encode(arg, encoding=portage._encodings['fs'], errors='strict') for arg in gpgcmd]
+		return_code = subprocess.call(gpgcmd)
```
EvaSDK: This is a minor nitpick as I have no idea whether this would be preferred or not in portage code base but what about using [`subprocess.check_call`](https://docs.python.org/2/library/subprocess.html#subprocess.check_call)

pym/portage/meta_manifest.py
```diff
@@ -0,0 +1,275 @@
...
+			self.create_master()
+		else:
+			self.create_cat()
+
+	def create_master(self):
+		'''Creates a MetaManifest file from all the MetaManifest files of the repository'''
+		self.find_reporoot()
+		myrepodir = self.reporoot
+		self.pkgdir = self.reporoot
+		self.fhashdict = {}
```
EvaSDK: You can probably replace this with a `collections.defaultdict(dict)` unless you need to get `KeyError` when accessing an unknown `ftype`.
Also this particular bit is used in multiple `create_` functions, maybe you want to put this in __init__ ? If not, you probably don't want to save this as an instance variable.

pym/portage/meta_manifest.py
```diff
@@ -0,0 +1,270 @@
...
+		'''Finds repolevel'''
+		self.find_reporoot()
+		path = path[len(self.reporoot):]
+		myrepo = path.split(os.path.sep)
+		repolevel = len(myrepo)
+		return repolevel
+
+	def find_repotype(self, path):
+		'''Finds the type of the repo directory'''
+		if len(set(['metadata', 'eclass', 'profiles']).intersection(os.listdir(path))) > 0:
```
zmedico: This is faster than listdir:
`if any(os.path.isdir(os.path.join(path, name)) for name in ('metadata', 'eclass', 'profiles')):`

pym/portage/meta_manifest.py
```diff
@@ -0,0 +1,270 @@
...
+	def validate_signature(self):
+		'''Validate the Manifest's file signature.'''
+		self.gkeys = GkeysInterface('portage', '/')
+		is_good, has_sig = self.gkeys.verify_file(self.getFullname())
+		return (is_good, has_sig)
+
+	def verify_manifest(self, verify_pkgs = True):
+		'''Verifies the manifest's validity'''
+		mfdir = self.pkgdir
+		mftype = self.find_repotype(mfdir)
```
zmedico: I feel like we should be passing information into the Manifest constructor, at least the depth, in order to reduce the effort necessary for the Manifest object to orient itself withing the repository.

pym/portage/meta_manifest.py
```diff
@@ -0,0 +1,270 @@
...
+		gpgcmd = portage.util.shlex_split(gpgcmd)
+		gpgcmd = [portage._unicode_encode(arg, encoding=portage._encodings['fs'], errors='strict') for arg in gpgcmd]
+		return_code = subprocess.call(gpgcmd)
+		if return_code == os.EX_OK:
+			os.rename(filename + ".asc", filename)
+			self.signed = True
+		else:
+			raise portage.exception.PortageException("!!! gpg exited with '" + str(return_code) + "' status")
+
+	def find_reporoot(self):
```
zmedico: I think the reporoot and the repolevel should be constructor parameters, since it's very inefficient for each Manifest instance to have to figure this stuff out automatically.

zmedico:
[GLEP 58](https://www.gentoo.org/glep/glep-0058.html) has been replaced by [GLEP 74](https://www.gentoo.org/glep/glep-0074.html) which is already implemented on the infrastructure side. See [bug 638908](https://bugs.gentoo.org/638908).

PR #41, opened 2016-08-15
repoman: deduplicate dependency.syntax errors (bug 591184)

zmedico:
X-Gentoo-bug: 591184
X-Gentoo-bug-url: https://bugs.gentoo.org/show_bug.cgi?id=591184

### DIFF
--- a/repoman/pym/repoman/modules/scan/depend/_depend_checks.py
+++ b/repoman/pym/repoman/modules/scan/depend/_depend_checks.py
@@ -1,5 +1,6 @@
 # -*- coding:utf-8 -*-
 
+import collections
 
 from _emerge.Package import Package
 
@@ -160,7 +161,12 @@ def _depend_checks(ebuild, pkg, portdb, qatracker, repo_metadata):
 			check_slotop(mydepstr, pkg.iuse.is_valid_flag,
 				badsyntax, mytype, qatracker, ebuild.relative_path)
 
+	dedup = collections.defaultdict(set)
 	for m, b in badsyntax:
+		if b in dedup[m]:
+			continue
+		dedup[m].add(b)
+
 		if m.endswith("DEPEND"):
 			qacat = "dependency.syntax"
 		else:


repoman/pym/repoman/modules/scan/depend/_depend_checks.py
```diff
@@ -160,7 +161,12 @@ def _depend_checks(ebuild, pkg, portdb, qatracker, repo_metadata):
 			check_slotop(mydepstr, pkg.iuse.is_valid_flag,
 				badsyntax, mytype, qatracker, ebuild.relative_path)
 
+	dedup = collections.defaultdict(set)
 	for m, b in badsyntax:
+		if b in dedup[m]:
```
EvaSDK: This seems useless since `set` will do this for you.

zmedico: @EvaSDK it's not useless, as is solves an issue reported in [bug 591184, comment #4](https://bugs.gentoo.org/show_bug.cgi?id=591184#c4). Of course there are multiple ways to use the set could be used to fix the problem. The way the way that I've done it here has the advantage of leaving the surrounding code unchanged.

EvaSDK: Sorry, it seems I was not explicit enough. The if seems redundant to me as adding multiple times the same element to a `set` adds the item only once. This if would be needed in the `defaultdict` used a list though.

zmedico: Without the ` if`, the `continue` statement would not execute when appropriate.

EvaSDK: Ok my bad, I missed the fact that the loop extends beyond building `dedup`.

zmedico: Haha, indeed one of my goals was to make it possible for you to *mostly* neglect the code.

PR #45, opened 2016-08-31
bin/*, misc/emerge-delta-webrsync: migrate to [[.

monsieurp:
Right now, there are lots of inconsistencies in the Portage codebase. Some scripts still use the builtin `[` command for performing tests whilst some others have been migrated to its improved counterpart, the two brackets `[[`. Sometimes a script uses a mix of the two which is inconsistent. `[[` clearly outnumbers `[` in the codebase so we should migrate to the former. This PR attempts to bring coherence and converts parts still using `[` to `[[`.

See [1] to understand the difference between the two.

Please review.

[1]: http://mywiki.wooledge.org/BashFAQ/031

### DIFF
--- a/bin/misc-functions.sh
+++ b/bin/misc-functions.sh
@@ -22,16 +22,16 @@ install_symlink_html_docs() {
 	fi
 	cd "${ED}" || die "cd failed"
 	#symlink the html documentation (if DOC_SYMLINKS_DIR is set in make.conf)
-	if [ -n "${DOC_SYMLINKS_DIR}" ] ; then
+	if [[ -n "${DOC_SYMLINKS_DIR}" ]] ; then
 		local mydocdir docdir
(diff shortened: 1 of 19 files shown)


bin/misc-functions.sh
```diff
@@ -22,16 +22,16 @@ install_symlink_html_docs() {
...
-			if [ -d "usr/share/doc/${docdir}" ] ; then
+			if [[ -d "usr/share/doc/${docdir}" ]]; then
 				mydocdir="/usr/share/doc/${docdir}"
 			fi
 		done
-		if [ -n "${mydocdir}" ] ; then
+		if [[ -n "${mydocdir}" ]]; then
 			local mysympath
-			if [ -z "${SLOT}" -o "${SLOT%/*}" = "0" ] ; then
+			if [ -z "${SLOT}" -o "${SLOT%/*}" = "0" ]; then
```
zmedico: Oops, missed one.

monsieurp:
Thanks. I did miss a couple but it should be fixed now.

PR #49, opened 2016-09-16
Handle situation with invalid entries in /proc/self/mountinfo

mpkh:
That fixes Gentoo installation won WLS
Source: https://github.com/Microsoft/BashOnWindows/issues/992#issuecomment-244460439
Can reproduce on two machines

### DIFF
--- a/pym/portage/util/writeable_check.py
+++ b/pym/portage/util/writeable_check.py
@@ -58,7 +58,13 @@ def linux_ro_checker(dir_list):
 				# to the left of the ' - ', after the attr's, so split it there
 				mount = line.split(' - ', 1)
 				_dir, attr1 = mount[0].split()[4:6]
-				attr2 = mount[1].split()[2]
+				# check for situation with invalid entries for /home and /root in /proc/self/mountinfo
+				# root path is missing sometimes on WSL
+				# for example: 16 1 0:16 / /root rw,noatime - lxfs  rw
+				try:
+					attr2 = mount[1].split()[2]
+				except IndexError:
+					attr2 = mount[1].split()[1]
 				if attr1.startswith('ro') or attr2.startswith('ro'):
 					ro_filesystems.add(_dir)
 
--- a/pym/portage/util/writeable_check.py
+++ b/pym/portage/util/writeable_check.py
@@ -57,14 +57,22 @@ def linux_ro_checker(dir_list):
 				# there can be a variable number of fields
 				# to the left of the ' - ', after the attr's, so split it there
 				mount = line.split(' - ', 1)
-				_dir, attr1 = mount[0].split()[4:6]
+				try:
+					_dir, attr1 = mount[0].split()[4:6]
+				except ValueError:
+					# If it raises ValueError we can simply ignore the line.
+					continue
 				# check for situation with invalid entries for /home and /root in /proc/self/mountinfo
 				# root path is missing sometimes on WSL
 				# for example: 16 1 0:16 / /root rw,noatime - lxfs  rw
-				try:
-					attr2 = mount[1].split()[2]
-				except IndexError:
-					attr2 = mount[1].split()[1]
+				if len(mount) > 1:
+					try:
+						attr2 = mount[1].split()[2]
+					except IndexError:
+						try:
+							attr2 = mount[1].split()[1]
+						except IndexError:
+							attr2 = mount[1]
 				if attr1.startswith('ro') or attr2.startswith('ro'):
 					ro_filesystems.add(_dir)
 


pym/portage/util/writeable_check.py
```diff
@@ -58,7 +58,11 @@ def linux_ro_checker(dir_list):
 				# to the left of the ' - ', after the attr's, so split it there
 				mount = line.split(' - ', 1)
 				_dir, attr1 = mount[0].split()[4:6]
```
zmedico: This line is also fragile. We should handle `ValueError: not enough values to unpack`.

pym/portage/util/writeable_check.py
```diff
@@ -58,7 +58,11 @@ def linux_ro_checker(dir_list):
...
-				attr2 = mount[1].split()[2]
+				# check for situation with invalid entries for /home and /root in /proc/self/mountinfo
+				try:
+					attr2 = mount[1].split()[2]
+				except IndexError:
+					attr2 = mount[1].split()[1]
```
zmedico: I imagine this line could also raise IndexError, given unexpected input. We should handle that too.

mpkh: @zmedico alike
```
except IndexError:
 try:
  attr2 = mount[1].split()[1]
    except IndexError:
      attr2 = mount[1]
```
?

zmedico: Maybe a comment documenting the the WSL mountinfo format would be nice.

zmedico: The mount array might not contain 2 elements, depending on the input, so mount[1] is still unsafe.

pym/portage/util/writeable_check.py
mpkh: @zmedico yes but is there situation when it's possible and what should be taken if it's not [4:6]

pym/portage/util/writeable_check.py
mpkh: I can't find some official docs about mountinfo there but here's what I've got
```
1 1 0:1 / / rw,noatime - lxfs rootfs rw
2 1 0:2 / /data rw,noatime - lxfs data rw
3 1 0:3 / /cache rw,noatime - lxfs cache rw
4 1 0:4 / /mnt rw,noatime - lxfs mnt rw
5 1 0:5 / /sys rw,noatime - sysfs sysfs rw
6 1 0:6 / /proc rw,noatime - proc proc rw
7 1 0:7 / /acct rw,noatime - cgroup cgroup rw
8 1 0:8 / /dev rw,noatime - tmpfs none rw,mode=755
9 8 0:9 / /dev/pts rw,nosuid,noexec,noatime - devpts devpts rw
10 1 0:10 / /run rw,nosuid,noexec,noatime - tmpfs none rw,mode=755
11 10 0:11 / /run/lock rw,nosuid,nodev,noexec,noatime - tmpfs none rw
12 10 0:12 / /run/shm rw,nosuid,nodev,noatime - tmpfs none rw
13 10 0:13 / /run/user rw,nosuid,nodev,noexec,noatime - tmpfs none rw,mode=755
14 4 0:14 / /mnt/c rw,noatime - drvfs none rw
15 4 0:15 / /mnt/d rw,noatime - drvfs none rw
16 1 0:16 / /root rw,noatime - lxfs  rw
17 1 0:17 / /home rw,noatime - lxfs  rw
18 6 0:18 / /proc/sys/fs/binfmt_misc rw,nosuid,nodev,noexec,relatime - binfmt_misc binfmt_misc rw
```

pym/portage/util/writeable_check.py
zmedico: If it raises ValueError we can simply ignore the line.

pym/portage/util/writeable_check.py
mpkh: my situation is `ext3 /dev/root rw` against `lxfs  rw` where rw is last which is `[1]` is it possible that `rw` will be the only parameter there?

mpkh:
@zmedico added additional checks, review please

pym/portage/util/writeable_check.py
```diff
@@ -57,8 +57,22 @@ def linux_ro_checker(dir_list):
...
+				# root path is missing sometimes on WSL
+				# for example: 16 1 0:16 / /root rw,noatime - lxfs  rw
+				if len(mount) > 1:
+					try:
+						attr2 = mount[1].split()[2]
+					except IndexError:
+						try:
+							attr2 = mount[1].split()[1]
+						except IndexError:
+							attr2 = mount[1]
```
zmedico: In this case it would be a format that we've never seen before, so I think we should ignore the whole line in this case.

pym/portage/util/writeable_check.py
```diff
@@ -57,8 +57,22 @@ def linux_ro_checker(dir_list):
 				# there can be a variable number of fields
 				# to the left of the ' - ', after the attr's, so split it there
 				mount = line.split(' - ', 1)
-				_dir, attr1 = mount[0].split()[4:6]
-				attr2 = mount[1].split()[2]
+				try:
+					_dir, attr1 = mount[0].split()[4:6]
+				except ValueError:
+					# If it raises ValueError we can simply ignore the line.
```
zmedico: We should show a warning here too.

zmedico:
Merged with warning messages in https://github.com/gentoo/portage/pull/50.

mpkh:
@zmedico thank you and sorry for delay, I was planning to add warnings today and didn't notify here.

zmedico:
@mpkh no problem, I just wanted it to get into the 2.3.1 release.

mpkh:
@zmedico and not sure how to deal with ipc_daemon being enabled by default and broken on WSL, maybe we can just use provided there workaround until bug will be fixed on their side

zmedico:
@mpkh do you know if there is an issue filed for it yet? If not, we should create some kind of minimal test case and file an issue. We can start by checking if [IpcDaemonTestCase](https://github.com/gentoo/portage/blob/portage-2.3.1/pym/portage/tests/ebuild/test_ipc_daemon.py) fails. Run it from a checkout like this:
```
pym/portage/tests/runTests.py pym/portage/tests/ebuild/test_ipc_daemon.py
```
If that fails, we can use that as a starting point for a more minimal test case.

mpkh:
@zmedico it's here: https://github.com/Microsoft/BashOnWindows/issues/1016

zmedico:
@mpkh portage uses fifos / named pipes for "ipc", which is a different thing than System V IPC.

mpkh:
@zmedico another thing is that I guess I need to disable multilib for gcc/glibc somehow there https://bpaste.net/show/bfee6bdec14b `/build-x86-x86_64-pc-linux-gnu-nptl/sunrpc/cross-rpcgen: cannot execute binary file: Exec format error`

zmedico:
@mpkh I've seen similar issues using binfmt_misc with qemu-arm, triggered by slightly unusual ELF headers. In that case I just had to modify my binfmt_misc configuration to accept the unusual ELF header. You should get a hexdump of the ELF header, and and report it as a BashOnWindows issue. Maybe they have some way to configure it to accept that ELF header, like binfmt_misc allows you to.

PR #54, opened 2016-09-25
ebuild.sh: start phases in temporary HOME if available (bug 595028)

zmedico:
This will avoid undesirable interactions with the python sitedir,
as reported in bug 574002. Since the temporary HOME is not guaranteed
to exist for some phases, use PORTAGE_PYM_PATH as a fallback. Also,
use PORTAGE_PYM_PATH inside __dyn_clean, since HOME is removed there.

X-Gentoo-Bug: 595028
X-Gentoo-Bug-URL: https://bugs.gentoo.org/595028

@floppym

### DIFF
--- a/bin/ebuild.sh
+++ b/bin/ebuild.sh
@@ -167,9 +167,14 @@ export SANDBOX_ON=0
 
 # Ensure that $PWD is sane whenever possible, to protect against
 # exploitation of insecure search path for python -c in ebuilds.
-# See bug #239560 and bug #469338.
-cd "${PORTAGE_PYM_PATH}" || \
-	die "PORTAGE_PYM_PATH does not exist: '${PORTAGE_PYM_PATH}'"
+# See bug #239560, bug #469338, and bug #595028.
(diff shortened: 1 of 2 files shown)


bin/ebuild.sh
```diff
@@ -167,9 +167,14 @@ export SANDBOX_ON=0
 
 # Ensure that $PWD is sane whenever possible, to protect against
 # exploitation of insecure search path for python -c in ebuilds.
-# See bug #239560 and bug #469338.
-cd "${PORTAGE_PYM_PATH}" || \
-	die "PORTAGE_PYM_PATH does not exist: '${PORTAGE_PYM_PATH}'"
+# See bug #239560, bug #469338, and bug #595028.
+if [[ -d ${HOME} ]] then
```
floppym: Forgot a semicolon here.

floppym:
Other than the typo, this seems reasonable to me.

bin/ebuild.sh
zmedico: Thanks, fixed now.

PR #57, opened 2016-10-04
einstalldocs: check whether default docs are indeed files

rindeal:
X-Gentoo-Bug: 596102
X-Gentoo-Bug-URL: https://bugs.gentoo.org/show_bug.cgi?id=596102


zmedico:
Thanks, merged in 843ad4b4b9878d5d10b4e1be2229960100e482ce.

PR #58, opened 2016-10-06
repoman: Disable SRC_URI.mirror warnings when there is only 1 mirror

mgorny:
Disable SRC_URI.mirror warnings when the thirdpartymirrors entry
consists of no more than 1 mirror. In this case, the mirror:// is really
no different than direct URL, and it most likely means the entry is only
kept for backwards compatibility.


mgorny:
@gentoo/qa

zmedico:
LGTM

dol-sen:
Yeah, looks good, go ahead and push it.

PR #60, opened 2016-10-18
bin/phase-helpers.sh: Added code from eapply_user to handle multiple calls

wltjr:
Added _DEFAULT_SRC_PREPARE_CALLED and _DEFAULT_EAPPLY_CALLED to allow tests to see if either have been called, like eclasses calling default in custom src_prepare

### DIFF
--- a/bin/phase-helpers.sh
+++ b/bin/phase-helpers.sh
@@ -805,6 +805,8 @@ __eapi4_src_install() {
 }
 
 __eapi6_src_prepare() {
+	declare -r _DEFAULT_SRC_PREPARE_CALLED
+
 	if [[ $(declare -p PATCHES 2>/dev/null) == "declare -a"* ]]; then
 		[[ -n ${PATCHES[@]} ]] && eapply "${PATCHES[@]}"
 	elif [[ -n ${PATCHES} ]]; then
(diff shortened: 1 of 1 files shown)


zmedico:
Variables like these can not be relied upon in official gentoo ebuilds unless they are added as an EAPI extension.

wltjr:
Is that way eapply_user sets a file vs var? I can switch to file in place of the variable. Or are you saying in general any such change is EAPI?

wltjr:
This is for an eclass not ebuild, java-utils-2.eclass.

wltjr:
Seems like this cannot take place as it needs to happen for other package managers as well.

zmedico:
Right, for official gentoo eclasses, it has to be added as an EAPI extension.

wltjr:
That is unfortunate as EAPI=6 will create a massive mess on the Java side, till EAPI=7 it seems. May be best to avoid for now.

zmedico:
Just add a wrapper function in one of the java eclasses, like java-utils-eapply or something.

wltjr:
Nothing I can do will tell me if another eclass called eapply is the problem.

wltjr:
Unless your saying create some global wrapper that effects any eclass/ebuild call to eapply after java-utils-2 inherit.

PR #62, opened 2016-10-26
bug 598116

hackers-terabit:
https://bugs.gentoo.org/598116 "Autounmask fails when there is no file in /etc/portage/package.*/"; Portage Development, Core; UNCO; terabit.funtoo:dev-portage


hackers-terabit:
Closing this due to an unfortunate typo.

PR #66, opened 2016-11-05
sync: call git prune before shallow fetch (bug 599008)

zmedico:
This is necessary in order to avoid "There are too many unreachable
loose objects" warnings from automatic git gc calls.

X-Gentoo-Bug: 599008
X-Gentoo-Bug-URL: https://bugs.gentoo.org/show_bug.cgi?id=599008


zmedico:
Closing, since I want to revert all changes from [bug 552814](https://bugs.gentoo.org/show_bug.cgi?id=552814) for now.

PR #86, opened 2017-01-12
__multijob_init: work around Cygwin FIFO shortcoming

haubi:
Cygwin does not support multiple read-handles for one FIFO (yet).
As we really need just one readonly- and one writeonly-handle, we can
reorder to open one single readwrite- and one writeonly-handle.

X-Gentoo-Bug: 583962
X-Gentoo-Bug-URL: https://bugs.gentoo.org/show_bug.cgi?id=583962


zmedico:
Merged in 5a53e732efa0ee7e2a3f8afe90d729212ff187fb.

PR #89, opened 2017-01-17
emaint: exit with non-zero status code when module fails (bug 567478)

alexandru-elisei:
Currently module functions return a message to emaint after invocation.
Emaint then prints this message then exits normally (with a success
return code) even if the module encountered an error. This patch aims
to change this by having each module function return a tuple of
(returncode, message). Emaint will inspect the return code and will
exit with an unsucessful status code if necessary.


zmedico:
Instead of closing pull requests and creating new ones, you can use `git push --force` to overwrite the branch that the existing pull request references.

alexandru-elisei:
Noted, thank you.

PR #90, opened 2017-01-17
emaint: exit with non-zero status code when module fails (bug 567478)

alexandru-elisei:
Module functions currently return a message to emaint after invocation.
Emaint prints this message then exits normally (with a success return
code) even if the module encountered an error. This patch aims to
change this by having each module public function return a tuple of
(returncode, message), where returncode is boolean True if the function
was successful or False otherwise. Emaint will inspect the return codes
and exit unsuccessfully if necessary.

The variable PORT_LOGDIR was added to the test environment to prevent
CleanLogs.clean() from failing when the variable isn't set or not a
directory.

### DIFF
--- a/pym/portage/emaint/main.py
+++ b/pym/portage/emaint/main.py
@@ -115,6 +115,7 @@ class TaskHandler(object):
 		"""Runs the module tasks"""
 		if tasks is None or func is None:
 			return
+		returncodes = []
 		for task in tasks:
 			inst = task()
 			show_progress = self.show_progress_bar and self.isatty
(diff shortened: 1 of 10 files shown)


zmedico:
Merged in f143e58dd3fd80ab67121e7a62e8cf47151d3907.

zmedico:
This commit introduced [bug 606588](https://bugs.gentoo.org/show_bug.cgi?id=606588).

alexandru-elisei:
I am running tests on a fix.

I am very confused about something: in the working version
SyncRepos.auto_sync() would never return a message because
options.return_messages was set to false, so retvals would always be
None. Why the check for retvals though?

Here's the code I am talking about:
https://gitweb.gentoo.org/proj/portage.git/tree/pym/_emerge/actions.py#n2003

PR #92, opened 2017-01-18
Add the "ebuildshell" feature.

haubi:
X-Gentoo-Bug: 155161
X-Gentoo-Bug-URL: https://bugs.gentoo.org/show_bug.cgi?id=155161

### DIFF
--- a/bin/ebuild.sh
+++ b/bin/ebuild.sh
@@ -121,7 +121,7 @@ __qa_source() {
 __qa_call() {
 	local shopts=$(shopt) OLDIFS="$IFS"
 	local retval
-	"$@"
+	__call-ebuildshell "$@"
 	retval=$?
 	set +e
 	[[ $shopts != $(shopt) ]] &&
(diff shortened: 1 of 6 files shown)


bin/ebuild.sh
```diff
@@ -547,6 +547,150 @@ if [[ -n ${QA_INTERCEPTORS} ]] ; then
 	unset BIN_PATH BIN BODY FUNC_SRC
 fi
 
+__call-ebuildshell() {
```
zmedico: I'd like to see this function moved to a separate ebuildshell.sh file that's conditionally sourced, so that ebuild.sh is as small as possible. Also, new internal functions should start with triple underscore so that they are automatically filtered by __save_ebuild_env.

man/make.conf.5
```diff
@@ -405,6 +405,12 @@ exist). Also see the related \fIunmerge\-backup\fR feature.
 Use locks to ensure that unsandboxed ebuild phases never execute
 concurrently. Also see \fIparallel\-install\fR.
 .TP
+.B ebuildshell
```
zmedico: I think it makes more sense to make this an ebuild(1) commandline option rather than a FEATURES setting, since FEATURES settings tend to be used much differently.  An environment variable can be used to signal to various code that ebuildshell is active.

grobian:
@haubi in prefix we carry a 2.3.45 version of the patch which appears to apply on 2.3.52(.1).  Perhaps we can refresh, get zmedico's suggestions in and merge this one at last?

man/make.conf.5
haubi: Not sure if I fully understand this: Do you suggest something like `emerge --ebuildshell <atom>` instead of `FEATURES=ebuildshell emerge <atom>` for front end usage?
(I won't care, but I want to be able to get the ebuildshell prompt for each phase with the `emerge` command.)

PR #95, opened 2017-01-20
emerge: check correctly for --sync success (bug 606588)

alexandru-elisei:


zmedico:
I've pushed a minimal fix in 23f239134b0290cbe0435e13d8f1b1757c6d8524. It just returns 1 on failure. Would it really be useful to preserve the returncode from the first failed repo? Probably not.

PR #101, opened 2017-01-27
Emaint cleanups

EvaSDK:
Minor things that made me irk at the code while trying to understand why `emaint logs` does not want to clean anything on my systems.

### DIFF
--- a/pym/portage/emaint/modules/logs/logs.py
+++ b/pym/portage/emaint/modules/logs/logs.py
@@ -18,9 +18,9 @@ class CleanLogs(object):
 
 	short_desc = "Clean PORT_LOGDIR logs"
 
+	@staticmethod
 	def name():
 		return "logs"
-	name = staticmethod(name)
 
 
 	def can_progressbar(self, func):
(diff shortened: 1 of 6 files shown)


pym/portage/emaint/modules/logs/logs.py
```diff
@@ -48,16 +47,15 @@ def clean(self, **kwargs):
...
-			if not settings:
-				settings = portage.settings
-			options = kwargs.get('options', None)
-			if options:
-				num_of_days = options.get('NUM', None)
-				pretend = options.get('pretend', False)
+
+		# convoluted, I know, but portage.settings does not exist in
+		# kwargs.get() when called from _emerge.main.clean_logs()
+		settings = kwargs.get('settings', portage.settings)
```
zmedico: This reference to `portage.settings` should raise AttributeError when called by emerge, since emerge calls `portage._disable_legacy_globals()`. This should suppress the AttributeError:
```
settings = kwargs.get('settings', getattr(portage, 'settings', None))
```

EvaSDK: I used an empty `dict` instead so the following `settings.get()` stay functional.

zmedico:
Merged in fe5154cbe5300ac7f20bc6bba57dca14e6114d02 and 3627dbfb635921f285083432e8b704b31c5a688a. Thanks!

PR #102, opened 2017-01-28
tests/emerge: add test for 'emerge --sync'

alexandru-elisei:


zmedico:
Looks good, but this has be wondering why travis didn't execute testSyncLocal for f143e58dd3fd80ab67121e7a62e8cf47151d3907, since that would have caught the issue fixed in 23f239134b0290cbe0435e13d8f1b1757c6d8524:

https://travis-ci.org/gentoo/portage/builds/193511816

alexandru-elisei:
I actually have a newer version on my local machine, where I don't do
a sub on the string, but set the auto_sync to "no" directly in the
repository object (and restore the previous value after generating the
config string). I think this new approach might be better than
manipulating a string with regular expressions.

I'll take a look at testSyncLocal.

alexandru-elisei:
Now that testSyncLocal has been fixed, is my basic test still useful?

zmedico:
> Now that testSyncLocal has been fixed, is my basic test still useful?

Do you want to move your test to testSyncLocal, and add an assertion to verify that `auto-sync = no` works correctly?

alexandru-elisei:
Sure, I'll move it to testSyncLocal.

PR #103, opened 2017-01-28
tests: add missing __test__.py files

zmedico:
The missing pym/portage/tests/sync/__test__.py file is the
reason that travis did not trigger the TypeError that was
fixed in commit 23f239134b0290cbe0435e13d8f1b1757c6d8524.


zmedico:
testSyncLocal currently fails in travis because /usr/bin/emaint does not exist:
```
testSyncLocal (portage.tests.sync.test_sync_local.SyncLocalTestCase) ... /home/travis/virtualenv/python3.4.4/bin/python: can't open file '/usr/bin/emaint': [Errno 2] No such file or directory
FAIL
```

zmedico:
Added commit to fix testSyncLocal for travis.

PR #107, opened 2017-01-31
sync.py: add warning when sync-type is not set

alexandru-elisei:


zmedico:
It seems like this belongs in the _get_repos method, since we already have a repo.sync_type check there.

alexandru-elisei:
_get_repos() behaves slightly different when encountering a repo
without sync-type based on whether sync was called from emerge or from
emaint:
- if called from emerge, sync is aborted if at least one repo doesn't
have sync-type specified
- if called from emaint, the check is done in _sync() and the repos
without sync-type are skipped and the valid repos are synced.

I was trying to preserve this behavior.

I can move the sync-type check after the _filter_auto() function is
applied (if necessary) and print the warning for the remaining repos.

alexandru-elisei:
I've tried doing this check in _get_repos(), and I get two error messages at the same time when doing emaint sync -r test (where test doesn't have sync-type):
- one error message in _get_repos() where it is removed from the list of repos because it doesn't have sync-type. Here I get "!!! Missing sync-type for repo: test".
- one error message from repo() because no repos were returned from _get_repos(): "Emaint sync, the specified repos were not found: test ... returning."

I think this is confusing.

zmedico:
Well, we really want emerge --sync and emaint sync to return an unsuccessful status for all of these errors that currently only trigger warning messages in the _get_repos method. We've got open bugs:

https://bugs.gentoo.org/show_bug.cgi?id=567478
https://bugs.gentoo.org/show_bug.cgi?id=576282

alexandru-elisei:
I wasn't aware of those bug reports, I'll fix _get_repos() and _sync() to signal a failure.

PR #108, opened 2017-02-01
Emerge sync patch

alexandru-elisei:
I really need feedback on this, the patch if very large.

I was on the fence about doing the checks in _get_repos(), and in the end I chose to do them in _sync(). I felt that generating the messages and returning is best done inside the same function instead of passing messages from _get_repos(), deciding if I should return failure, calling _sync(), concatenating the messages from _sync() then returning.

If I were to do the checks in _get_repos(), then _get_repos would have to return something like (exit_with_failure, selected_repos, msgs). exit_with_failure is True only if selected_repos is empty because all repos had an error, False otherwise (repos are empty because no repos are defined/have auto-sync, or there are valid repos to sync).

all_repos(), auto_sync() and repo() will have to inspect exit_with_failure and exit with failure if need be.

If you think having the checks in _get_repos() is better, I can make the changes.

### DIFF
--- a/pym/portage/emaint/modules/sync/sync.py
+++ b/pym/portage/emaint/modules/sync/sync.py
@@ -149,38 +149,7 @@ class SyncRepos(object):
 
 	def _get_repos(self, auto_sync_only=True):
 		selected_repos = []
-		unknown_repo_names = []
-		missing_sync_type = []
-		if self.emerge_config.args:
-			for repo_name in self.emerge_config.args:
-				#print("_get_repos(): repo_name =", repo_name)
-				try:
-					repo = self.emerge_config.target_config.settings.repositories[repo_name]
-				except KeyError:
-					unknown_repo_names.append(repo_name)
-				else:
-					selected_repos.append(repo)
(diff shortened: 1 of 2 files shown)


pym/portage/emaint/modules/sync/sync.py
```diff
@@ -199,20 +167,58 @@ def _filter_auto(self, repos):
...
+		missing_sync_type = []
+		for repo in selected_repos:
+			if repo.sync_type is not None:
+				valid_repos.append(repo)
+			else:
+				missing_sync_type.append(repo.name)
+		if missing_sync_type:
+			msgs.append(warn(" * ") + "Missing sync-type for repo(s): %s" %
+				" ".join(missing_sync_type) + ", skipping...\n")
+		if not valid_repos:
```
zmedico: It only fails if valid_repos is empty? I want it to fail if there are __any__ unknown repos or missing sync types for repos that are supposed to be synced.

alexandru-elisei:
Reworked in #109.

PR #109, opened 2017-02-02
Emerge sync patch2

alexandru-elisei:
This is a rewrite of #108.

Emerge --sync fails if at least one repo name is not found.

Emerge --sync and emaint sync will fail if at least one repo doesn't have sync-type specified.

I've decided to have the checks in _get_repos() after all, I think checking the validity of the repos there makes the most sense. I also have greater flexibility in how I treat _get_repos() errors (see the repos() function).


alexandru-elisei:
Reworked, closing.

PR #110, opened 2017-02-04
emaint: specify multiple repos when doing emaint sync --repo

alexandru-elisei:
A list of space separated repo names can be specified when doing
emaint sync --repo, similar to what is possible with emerge --sync.

Sync will fail if one of the repos is missing.


alexandru-elisei:
Rewriting it, it conflicts with changes introduced by 79103f9.

PR #117, opened 2017-02-16
logs.py: print the log files in check() instead of calling clean()

alexandru-elisei:
CleanLogs.check() will print the log files instead of calling
CleanLogs.clean() with the --pretend option set.

The --pretend option removes the -delete argument from the
PORT_LOGDIR_CLEAN command. This does not work if the clean command is
not 'find ... -delete', which would lead to the clean command getting
executed when doing emaint logs --check.

Also the CleanLogs.name() function now uses the @staticmethod decorator.

### DIFF
--- a/pym/portage/emaint/modules/logs/logs.py
+++ b/pym/portage/emaint/modules/logs/logs.py
@@ -18,9 +18,9 @@ class CleanLogs(object):
 
 	short_desc = "Clean PORT_LOGDIR logs"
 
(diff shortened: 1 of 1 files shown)


zmedico:
It looks like this duplicates code from the clean method. In order to avoid duplication we should share code via private methods and/or functions.

alexandru-elisei:
Fair point, I was going to redo it anyway.

PR #119, opened 2017-02-20
emerge: sync given repos even if auto-sync is false

alexandru-elisei:
This is following our discussion on the mailing list about patch sync.py: extend the checks in _get_repos() and fail when necessary (bugs 567478, 576272, 601054).

If this gets merged, I'll make a series of patches to update the man page, remove repos that don't have sync-type specified (as per the man page) and fail if sync-uri is invalid.

### DIFF
--- a/pym/_emerge/actions.py
+++ b/pym/_emerge/actions.py
@@ -1999,9 +1999,13 @@ def action_sync(emerge_config, trees=DeprecationWarning,
 			action=action, args=[], trees=trees, opts=opts)
 
 	syncer = SyncRepos(emerge_config)
-
 	return_messages = "--quiet" not in emerge_config.opts
-	success, msgs = syncer.auto_sync(options={'return-messages': return_messages})
+	options = {'return-messages' : return_messages}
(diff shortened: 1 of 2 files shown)


pym/_emerge/actions.py
```diff
@@ -1999,9 +1999,13 @@ def action_sync(emerge_config, trees=DeprecationWarning,
...
-
 	return_messages = "--quiet" not in emerge_config.opts
-	success, msgs = syncer.auto_sync(options={'return-messages': return_messages})
+	options = {'return-messages' : return_messages}
+	if emerge_config.args:
+		options['repo'] = emerge_config.args
+		success, msgs = syncer.repo(options=options)
+	else:
+		success, msgs = syncer.auto_sync(options=options)
 	if return_messages:
```
zmedico: The above change works fine by itself, so I've merged it in 75f0936b8866b143643abcd6c302cd72fc71eef3.

The changes to sync-type and sync-uri validation can go in separately.

PR #122, opened 2017-02-21
sync.py: validate repos in _get_repos() (bugs 567478, 576282, 601054)

alexandru-elisei:
SyncRepos._get_repos() now receives an optional match argument, which
represents the names of the repositories to retrieve.

SyncRepos._get_repos() will return a tuple of (result, repos, messages).
Result is True on success and False on failure, repos is a list of
repository objects and messages is a list of errors or warnings
generated while getting the repos.

Repos given to emerge --sync are no longer treated separately.

After commit 75f0936 multiple repos can be given to SyncRepos.repo()
via emerge --sync. SyncRepos._get_repos() will fail if at least one
of these repos cannot be synced (the repo is missing or has sync
disabled).

If no repos are specified, SyncRepos._get_repos() ignores repos with
sync-type empty as per the portage.5 man file. If repos are specified,
the function will fail. This check was previously done in _sync(), which
could lead to a situation when a repo given to emerge --sync or
emaint sync --repo was not synced and the operation was considered
successful.

SyncRepos._get_repos() also validates repos by checking for a non-empty
sync-uri. Finding a repo with an empty sync-uri is considered a failure
and the appropiate message will be generated.

### DIFF
--- a/pym/portage/emaint/modules/sync/sync.py
+++ b/pym/portage/emaint/modules/sync/sync.py
@@ -89,47 +89,50 @@ class SyncRepos(object):
 	def auto_sync(self, **kwargs):
 		'''Sync auto-sync enabled repos'''
 		options = kwargs.get('options', None)
-		selected = self._get_repos(True)
 		if options:
 			return_messages = options.get('return-messages', False)
 		else:
 			return_messages = False
-		return self._sync(selected, return_messages,
-			emaint_opts=options)
+		success, repos, msgs = self._get_repos(True)
+		if not success:
+			if return_messages:
+				return (False, msgs)
+			return (False, None)
+		return self._sync(repos, return_messages, emaint_opts=options)
 
 
 	def all_repos(self, **kwargs):
 		'''Sync all repos defined in repos.conf'''
-		selected = self._get_repos(auto_sync_only=False)
 		options = kwargs.get('options', None)
 		if options:
 			return_messages = options.get('return-messages', False)
 		else:
 			return_messages = False
-		return self._sync(selected, return_messages,
-			emaint_opts=options)
+		success, repos, msgs = self._get_repos(auto_sync_only=False)
+		if not success:
+			if return_messages:
+				return (False, msgs)
+			return (False, None)
+		return self._sync(repos, return_messages, emaint_opts=options)
 
 
 	def repo(self, **kwargs):
 		'''Sync the specified repo'''
 		options = kwargs.get('options', None)
 		if options:
-			repos = options.get('repo', '')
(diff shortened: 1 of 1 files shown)


pym/portage/emaint/modules/sync/sync.py
```diff
@@ -122,9 +122,10 @@ def repo(self, **kwargs):
 			repos = repos.split()
 		available = self._get_repos(auto_sync_only=False)
 		selected = self._match_repos(repos, available)
-		if not selected:
-			msgs = [red(" * ") + "The specified repos were not found: %s" %
-				(bold(", ".join(repos))) + "\n   ...returning"]
+		if len(selected) < len(repos):
+			missing_repo_names = set(repos) - set(repo.name for repo in selected)
+			msgs = [red(" * ") + "The specified repo(s) were not found or have sync-type empty: %s" %
```
zmedico: I don't like the way the missing sync-type is combined with other problems in the error message here. We need to have specific messages, especially considering that we will likely need to add similar checks for various other kinds of problems in the future.

alexandru-elisei: I've rewritten the patch to make the checks in _get_repos(). Adding extra validation steps should be a lot easier now.

pym/portage/emaint/modules/sync/sync.py
```diff
@@ -147,44 +150,40 @@ def _match_repos(repos, available):
...
-					level=logging.INFO, noiselevel=2)
-				return []
+	def _get_repos(self, auto_sync_only=True, match_repos=None):
+		msgs = []
+		repos = self.emerge_config.target_config.settings.repositories
+		if match_repos is not None:
+			repos = self._match_repos(match_repos, repos)
+			if len(repos) < len(match_repos):
+				available = [repo.name for repo in repos]
+				missing = [repo for repo in match_repos if repo not in available]
```
zmedico: Technically, this code as well as the _match_repos method should handle repo aliases.

alexandru-elisei: Let's merge it as it is and I'll handle repo aliases in the next patch?

zmedico: Yeah, sounds good.

zmedico: I've filed a new bug for the repo alias support:

https://bugs.gentoo.org/show_bug.cgi?id=610852

zmedico:
Merged in 79103f99799c9365389290e812213c46e006c019. Thanks!

PR #131, opened 2017-03-02
movefile: support in-kernel file copying on Linux (bug 607868)

zmedico:
movefile: support in-kernel file copying on Linux (bug 607868)

Perform in-kernel file copying when possible, and also support
reflinks and sparse files. If the optimized implementation
fails at runtime, gracefully fallback to a plain read/write
loop.

Compile-time and run-time fallbacks are implemented, so that
any incompatiblities will be handled gracefully. For example,
if the code is compiled on a system that supports the
copy_file_range syscall, but at run-time an older kernel that
does not support this syscall is detected, it will be handled
gracefully. There are similar fallbacks for lack of lseek
SEEK_DATA and sendfile support.

X-Gentoo-Bug: 607868
X-Gentoo-Bug-Url: https://bugs.gentoo.org/show_bug.cgi?id=607868

### DIFF
--- a/pym/portage/util/movefile.py
+++ b/pym/portage/util/movefile.py
@@ -8,7 +8,6 @@ __all__ = ['movefile']
 import errno
 import fnmatch
 import os as _os
-import shutil as _shutil
 import stat
 import sys
 import textwrap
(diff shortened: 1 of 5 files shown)


pym/portage/util/movefile.py
```diff
@@ -114,7 +116,7 @@ def movefile(src, dest, newmtime=None, sstat=None, mysettings=None,
 		_copyfile = selinux.copyfile
 		_rename = selinux.rename
 	else:
-		_copyfile = _shutil.copyfile
```
simonvanderveldt: The shutil import at line 11 can be removed :)

zmedico: Fixed, thanks!

zmedico:
@chutz

PR #134, opened 2017-03-08
depgraph: fix backtracking for slot operator rebuilds (bug 612042)

zmedico:
Fix package selection logic to avoid pulling in undesirable
rebuilds/updates during backtracking for slot operator rebuilds.
The undesirable rebuilds/updates have sent some calculations off
course, by triggering more and more rebuilds/updates with each
backtracking run.

In order to solve the problem, make various adjustments to the
package selection logic so that installed packages are preferred
over rebuilds/updates when appropriate. Also update unit tests
to work with these adjustments.

Fixes: 5842e87872fd ("Fix slot operator handling bug")
X-Gentoo-bug: 612042
X-Gentoo-bug-url: https://bugs.gentoo.org/show_bug.cgi?id=612042

### DIFF
--- a/pym/_emerge/depgraph.py
+++ b/pym/_emerge/depgraph.py
@@ -2304,7 +2304,7 @@ class depgraph(object):
 				# Check for slot update first, since we don't want to
 				# trigger reinstall of the child package when a newer
 				# slot will be used instead.
-				if rebuild_if_new_slot:
+				if rebuild_if_new_slot and dep.want_update:
 					new_dep = self._slot_operator_update_probe(dep,
 						new_child_slot=True)
 					if new_dep is not None:
@@ -6241,7 +6241,7 @@ class depgraph(object):
 							if highest_installed is None or pkg.version > highest_installed.version:
 								highest_installed = pkg
 
-					if highest_installed:
+					if highest_installed and self._want_update_pkg(parent, highest_installed):
 						non_installed = [pkg for pkg in matched_packages \
 							if not pkg.installed and pkg.version > highest_installed.version]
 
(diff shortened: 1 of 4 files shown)


kentfredric:
On Wed, 08 Mar 2017 03:08:34 -0800
Zac Medico <notifications@github.com> wrote:

> The undesirable rebuilds/updates have sent some calculations off
> course, by triggering more and more rebuilds/updates with each
> backtracking run.
> 
> In order to solve the problem, make various adjustments to the
> package selection logic so that installed packages are preferred
> over rebuilds/updates when appropriate. Also update unit tests
> to work with these adjustments.

Clarification, does this mean that portage will be more disposed to rebuild
"absolutely everything" that slot-depends on Perl?

Or will portage be more disposed to treating those slot-depends as blockers
for reinstalling perl.

My reading suggests that portage will be more inclined to favour "installed",
which is the opposite of the desired reaction.

( Though I could be just misunderstanding what "where appropriate" means )

zmedico:
```
My reading suggests that portage will be more inclined to favour "installed",
which is the opposite of the desired reaction.
```
It will only favor "installed" when appropriate for the current combination of package arguments (such as `@world`, `--deep`, and `--update` parameters. The behavior for `emerge --update --deep --with-bdeps=y @world` remains unchanged, since that command specifies to update all packages that are not eligible for remove by `emerge --depclean`.

The change in behavior only applies to commands that specify to update fewer packages. It fixes a case where each backtracking run would trigger more and more rebuilds/updates, which ultimately lead to exhaustion of backtracking runs (and tried to update/rebuild packages which were not requested).

zmedico:
Note that the slot operator rebuild code is designed to trigger all updates/rebuilds in the initial calculation, before any backtracking occurs. This patch fixes a misbehavior, where more and more rebuilds/updates were triggered with each backtracking calculation.

zmedico:
For perl updates, rebuilds of perl packages should be triggered in the initial calculation, due to broken slot operator dependencies that the update causes.

kentfredric:
> rebuilds of perl packages should be triggered in the initial calculation

Oohh, so approximates my "Feed portage all the things" hacks instead of "Get not enough, and then progressively wind outwards till we exceed backtrack" 

I look forward to this :D

PR #137, opened 2017-03-10
econf: multijob-safe shebang tweaking

haubi:
Using econf in parallel for multiple configurations (multilib, or
ncurses' wide+narrow), both may try to fix configure's shebang.
On Cygwin at least, this may cause 'sed -i' to fail with:
 sed: cannot rename /.../work/ncurses-6.0/sedXZsjI6: Permission denied
Instead of 'sed -i', better use 'mv -f' towards the original file.

### DIFF
--- a/bin/phase-helpers.sh
+++ b/bin/phase-helpers.sh
@@ -572,14 +572,14 @@ econf() {
 	if [ -x "${ECONF_SOURCE}/configure" ]; then
 		if [[ -n $CONFIG_SHELL && \
 			"$(head -n1 "$ECONF_SOURCE/configure")" =~ ^'#!'[[:space:]]*/bin/sh([[:space:]]|$) ]] ; then
-			# preserve timestamp, see bug #440304
-			touch -r "${ECONF_SOURCE}/configure" "${ECONF_SOURCE}/configure._portage_tmp_.${pid}" || die
+			cp -p "${ECONF_SOURCE}/configure" "${ECONF_SOURCE}/configure._portage_tmp_.${pid}" || die
 			sed -i \
 				-e "1s:^#![[:space:]]*/bin/sh:#!$CONFIG_SHELL:" \
-				"${ECONF_SOURCE}/configure" \
+				"${ECONF_SOURCE}/configure._portage_tmp_.${pid}" \
 				|| die "Substition of shebang in '${ECONF_SOURCE}/configure' failed"
-			touch -r "${ECONF_SOURCE}/configure._portage_tmp_.${pid}" "${ECONF_SOURCE}/configure" || die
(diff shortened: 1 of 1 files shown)


mgorny:
1. Running econf inside multijob thingy is disallowed because it violates `die` rules.
2. Making Portage account more for broken `sed` just fixes one of thousands of uses...

zmedico:
> 1. Running econf inside multijob thingy is disallowed because it violates `die` rules.
> 2. Making Portage account more for broken `sed` just fixes one of thousands of uses...

That's fair enough, but regardless, it's always nice to have atomic file operations.

bin/phase-helpers.sh
```diff
@@ -573,13 +573,13 @@ econf() {
 		if [[ -n $CONFIG_SHELL && \
 			"$(head -n1 "$ECONF_SOURCE/configure")" =~ ^'#!'[[:space:]]*/bin/sh([[:space:]]|$) ]] ; then
 			# preserve timestamp, see bug #440304
-			touch -r "${ECONF_SOURCE}/configure" "${ECONF_SOURCE}/configure._portage_tmp_.${pid}" || die
+			cp "${ECONF_SOURCE}/configure" "${ECONF_SOURCE}/configure._portage_tmp_.${pid}" || die
```
zmedico: Aren't we going to lose the executable bit in this step?

haubi: Huh? Have never seen cp dropping mode bits that are possible to be preserved.
And indeed, man cp(1) tells for `--preserve[=ATTR_LIST]`:
`preserve the specified attributes (default: mode,ownership,timestamps), ...`

mgorny: Just FYI, that 'default' notion is for `--preserve` without parameter. Obviously `cp` doesn't preserve ownership 'by default'.

mgorny: That said, per http://pubs.opengroup.org/onlinepubs/9699919799/utilities/cp.html:

> [3.b.] If dest_file does not exist, a file descriptor shall be obtained by performing actions equivalent to the open() function defined in the System Interfaces volume of POSIX.1-2008 called using dest_file as the path argument, and the bitwise-inclusive OR of O_WRONLY and O_CREAT as the oflag argument. The file permission bits of source_file shall be the mode argument.

So if I'm reading this right, `cp` should preserve original permissions & ~umask.

zmedico: In practice, on Linux it doesn't reliably preserve the executable bit, so we should use `cp -p`.

haubi: Although wondering about the scenarios where Linux is unreliable here, this is nothing I really care about - patch updated.

zmedico:
Merged in f4f1caaeb37825c01a2092d11bec4f932772bd38. Thanks~

PR #138, opened 2017-03-14
Fix broken fetch after b33c965283

Coacher:
valid_hashes must be a set, not frozenset.

Gentoo-Bug: 612578


EvaSDK:
I wonder what is the actual error and how your changes fixes it as running in an interactive shell the code logic seems fine (for both python 2 and 3):
```
In [2]: a = frozenset({1, 4, 9, 2, 169, 400})

In [3]: sorted(a)
Out[3]: [1, 2, 4, 9, 169, 400]
```

Coacher:
frozenset is immutable and doesn't have discard method. This is the real fix.
Instead, you are looking at now unneeded cast to set() I removed for clarity.

EvaSDK:
Indeed, missed that.

PR #140, opened 2017-03-16
portage.package.ebuild: Use a fake FILESDIR to catch invalid accesses

mgorny:
Use a model of fake FILESDIR path to ensure that invalid accesses to
FILESDIR will result in failures rather than being silently allowed by
Portage. This mostly involves accesses in the global scope and pkg_*
phases, although the current model does not cover the latter completely
(i.e. does not guarantee that the directory is removed post src_*).

This model aims to follow PMS wording quite precisely. The value of
FILESDIR is meant to be stable throughout the build process, and it is
reliably set to a temporary directory path. However, since the path is
not guaranteed to be present outside src_*, the directory symlink is not
actually created before src_* phases.


mgorny:
From a quick test run, the following breaks:
* sys-devel/autoconf in depend+ since it sources code from FILESDIR in global scope;
* sys-libs/glibc partially since it sources code from FILESDIR in pkg_* conditional to whether it exists (i.e. skips a few phases),
* sys-kernel/mips-sources since it sources a lot of code from FILESDIR in pkg_setup.

I will run the patch for a while to see if anything else breaks. The listed ebuilds have a bug open each for at least half a year with hard deadline next week, so we can disregard them.

toralf:
no issues here at the tinderbox so far

zmedico:
Merged in 0637c95545ab7dc0bb5d091de221b9e99e2cf5b2, and fixed up in 6affafbd59262760091b3b7c2d63ff5c96e43807.

PR #157, opened 2017-04-22
Closes bug #616262

fusion809:
Per bug [616262](https://bugs.gentoo.org/show_bug.cgi?id=616262) quickpkg now takes atom input, not `/var/db/pkg/<category>/<package>`. I only reported that bug before filing this pull request because someone has neglected to fix this line from the [Project:Portage](https://wiki.gentoo.org/index.php?title=Project:Portage&oldid=553258) Gentoo Wiki:

> We also have a GitHub mirror, but note that this is just a mirror, so do not file bugs nor pull requests there.

Please someone fix it.

### DIFF
--- a/man/emerge.1
+++ b/man/emerge.1
@@ -50,7 +50,7 @@ so this syntax shouldn't be used.
 .BR tbz2file
 A \fItbz2file\fR must be a valid .tbz2 created with \fBebuild
 <package>\-<version>.ebuild package\fR or \fBemerge \-\-buildpkg
-[category/]<package>\fR or \fBquickpkg /var/db/pkg/<category>/<package>\fR.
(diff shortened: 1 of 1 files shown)


zmedico:
Thanks, merged in 8889360554e80e2ca461ee67bbbc11de8a9a5a1a.

> I only reported that bug before filing this pull request because someone has neglected to fix this line from the Project:Portage Gentoo Wiki:
>>  We also have a GitHub mirror, but note that this is just a mirror, so do not file bugs nor pull requests there.

Yeah, we'll have to re-evaluate that policy. We may have some team members that prefer to avoid using a "proprietary" platform, since Gentoo hosts its own infrastructure using entirely on free software. Within that context, patch submission and review is done via Gentoo mailing lists.

PR #159, opened 2017-05-03
copyfile: fix compiling against musl libc

pks-t:
The native C module for copying files directly in-kernel is using the
`loff_t` data type, which is defined in fcntl.h. While the file seems to
be indirectly included when building with glibc, it breaks compiling
with musl libc.

Fix the issue by directly including fcntl.h. Define GNU_SOURCE as the
typedef is guarded by this preprocessor macro.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -676,7 +676,8 @@ setup(
 		['$sysconfdir/portage/repo.postsync.d', ['cnf/repo.postsync.d/example']],
 	],
 
-	ext_modules = [Extension(name=n, sources=m) for n, m in x_c_helpers.items()],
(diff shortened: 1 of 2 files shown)


zmedico:
Can we just replace loff_t with off64_t, so that we can keep things more portable by avoiding the _GNU_SOURCE define?

blueness:
Yes, that's an acceptable solution.  I tested to make sure.

zmedico:
I'm thinking with should use off_t along with the Large File Support macros (-D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE) in extra_compile_args.

pks-t:
Just did some quick tests with `off_t`, works for me. I'm closing this PR in favor of yours. Thanks!

PR #165, opened 2017-05-26
GitSync: Add support for setting the environment

mrueg:
This can be used to provide private SSH keys to portage in order to
clone repositories from a non-public repository.

An exemplary usage would be:
sync-git-pull-env = GIT_SSH_COMMAND=ssh -i /etc/portage/.ssh/id_rsa -o UserKnownHostsFile=/etc/portage/.ssh/known_hosts
sync-git-clone-env = GIT_SSH_COMMAND=ssh -i /etc/portage/.ssh/id_rsa -o UserKnownHostsFile=/etc/portage/.ssh/known_hosts

### DIFF
--- a/man/portage.5
+++ b/man/portage.5
@@ -1,4 +1,4 @@
-.TH "PORTAGE" "5" "Jan 2017" "Portage VERSION" "Portage"
+.TH "PORTAGE" "31" "May 2017" "Portage VERSION" "Portage"
 .SH NAME
 portage \- the heart of Gentoo
 .SH "DESCRIPTION"
@@ -979,8 +979,30 @@ Specifies CVS repository.
 .B sync\-depth
 This is a deprecated alias for the \fBclone\-depth\fR option.
 .TP
+.B sync\-git\-clone\-env
+Set environment variables for git when cloning repository (git clone).
+This will override settings from sync-git-env.
+.RS
+.TP
+.I Example:
+sync-git-clone-env="VAR1=word1 word2" VAR2=word3 "VAR3=$word 5 6"
+.br
+Gives three variables "VAR1", "VAR2", "VAR3" with the values "word1 word2",
+"word3", "$word 5 6".
+.RE
+.TP
 .B sync\-git\-clone\-extra\-opts
 Extra options to give to git when cloning repository (git clone).
+.TP
+.B sync\-git\-env
+Set environment variables for git when cloning or pulling the repository.
+These will be overridden by setting them again in sync-git-clone-env and sync-git-pull-env.
+See also example for sync-git-clone-env.
+.TP
+.B sync\-git\-pull\-env
+Set environment variables for git when updating repository (git pull).
+This will override settings from sync-git-env.
+See also example for sync-git-clone-env.
 .TP
 .B sync\-git\-pull\-extra\-opts
 Extra options to give to git when updating repository (git pull).
--- a/pym/portage/sync/modules/git/git.py
+++ b/pym/portage/sync/modules/git/git.py
@@ -1,4 +1,4 @@
-# Copyright 2005-2015 Gentoo Foundation
+# Copyright 2005-2017 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 
 import logging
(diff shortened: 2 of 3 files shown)


zmedico:
Does it allow setting multiple variables? We should document it in `man/portage.5`.

mrueg:
No it currently doesn't. I'll need to have a look what's wrong. I suppose there's some escaping done when repos.conf gets read, that I need unescape.

mrueg:
@zmedico Added doc in manpage. Changed it to support multiple variables by adding a newline character to the conf.

Tested setting:
sync-git-pull-env = GIT_SSH_COMMAND=ssh -i /etc/portage/.ssh/id_rsa -o UserKnownHostsFile=/etc/portage/.ssh/known_hosts\n GIT_TRACE=true

zmedico:
Rather than invent a new way of embedding environment variables in ini files, I think it would be nice to emulate the way that [systemd embeds environment settings in ini files](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Environment=):
```
Example:

Environment="VAR1=word1 word2" VAR2=word3 "VAR3=$word 5 6"
gives three variables "VAR1", "VAR2", "VAR3" with the values "word1 word2", "word3", "$word 5 6".
```
In our context, it would be something like this:
```
sync-git-pull-env = "GIT_SSH_COMMAND=ssh -i /etc/portage/.ssh/id_rsa -o UserKnownHostsFile=/etc/portage/.ssh/known_hosts" GIT_TRACE=true
```

mrueg:
@zmedico Okay, replaced ConfigParser with shlex_split.

pym/portage/sync/modules/git/git.py
```diff
@@ -50,6 +50,11 @@ def new(self, **kwargs):
 			sync_uri = sync_uri[6:]
 
 		git_cmd_opts = ""
+		if self.repo.module_specific_options.get('sync-git-clone-env'):
+			shlexed_env = shlex_split(self.repo.module_specific_options['sync-git-clone-env'])
+			clone_env = {k: v for [k, v] in [ assignment.split('=', 1) for assignment in shlexed_env]}
```
zmedico: This will raise ValueError if any of the tokens is missing an `=` character. Here is a way to silently ignore errors:
```
dict((k, v) for k, __, v in (assignment.partition('=') for assignment in shlexed_env) if k)
```

mrueg: Ah haven't thought about that one. Thanks for providing a solution.

zmedico:
I guess we can copy [the example from the systemd.exec](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Environment=) man page into our man page.

mrueg:
@zmedico Added the example.

zmedico:
We should make it possible to set environment variables that apply to both pull and clone commands, without having to specify them separately in repos.conf. I suppose this could be accomplished using an Environment setting that is used for all git commands? Should sync-git-pull-env and sync-git-clone-env inherit and override this shared Environment setting?

mrueg:
We could add a general setting to make.conf and allow it to be overwritten by sync-git-{clone,pull}-env
Maybe call it PORTAGE_GIT_ENV (as there's already PORTAGE_RSYNC_OPTS)?

zmedico:
Any common variables can be set directly in make.conf, so there's no need for a separate PORTAGE_GIT_ENV variable. However, it's useful to be have to have per-repo environment variables, so I think we need something like the repos.conf Environment setting that I suggested earlier.

mrueg:
There's a sync-git-env variable now. s-g-clone-env and s-g-pull-env can be used to override individual environment variables.

man/portage.5
```diff
@@ -979,9 +979,43 @@ Specifies CVS repository.
...
+.B sync\-git\-env
+Set environment variables for git when cloning or pulling the repository.
+These will be overridden by setting them again in sync-git-clone-env and sync-git-pull-env.
+.RS
+.TP
+.I Example:
+sync-git-env="VAR1=word1 word2" VAR2=word3 "VAR3=$word 5 6"
+.br
+Gives three variables "VAR1", "VAR2", "VAR3" with the values "word1 word2",
+"word3", "$word 5 6".
```
zmedico: The other two can refer to this example, so that we don't have to duplicate it.

mrueg: Okay, referred to it.

PR #170, opened 2017-05-31
Allow to choose compression method for binpkg

mrueg:
Allow to set a different compression method than bzip2 for binpackages and allow to set a compression level.

This is still WIP, it might not work 100% yet and lacks documentation.

Added two new settings for make.conf
```
BINPKG_COMPRESSION="zstd"
BINPKG_COMPRESSION_ARGS="-l5"
```
Open for better names here.


dol-sen:
Manuel, what about the possibility to use my pydecomp lib instead. just keep a small backup types in portage, then depend on it for others.

https://github.com/dol-sen/pyDeComp

I created it to clean up the mess in catalyst.  it is also easily extended without high level API changes required by consumer apps.

bin/quickpkg
```diff
@@ -43,6 +46,19 @@ def quickpkg_atom(options, infos, arg, eout):
...
 
 	try:
+		binpkg_compression = settings.get("BINPKG_COMPRESSION", "bzip2")
+		compression = _compressors[binpkg_compression]
+	except KeyError as e:
+		eout.eerror("Unsupported compression method: %s" % e.args[0])
+		del e
+		return
+
+	if subprocess.call(["type", compression["decompress"]], shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE) != 0:
```
zmedico: Normally we use portage.process.find_binary() for this, because calling a shell is overkill (and python2.7 does not have shutil.which()).

bin/quickpkg
```diff
@@ -43,6 +46,19 @@ def quickpkg_atom(options, infos, arg, eout):
...
+		missing_package = compression["package"]
+		eout.eerror("File compression unsupported %s. Missing package: %s" % (binpkg_compression, missing_package))
+		return
```
zmedico: We should make quickpkg return an unsuccessful exit status in this case.

bin/quickpkg
```diff
@@ -132,14 +148,14 @@ def quickpkg_atom(options, infos, arg, eout):
...
+				cpv + ".tar." + compression["suffix"] + "." + str(os.getpid()))
 			ensure_dirs(os.path.dirname(binpkg_tmpfile))
 			# The tarfile module will write pax headers holding the
 			# xattrs only if PAX_FORMAT is specified here.
-			tar = tarfile.open(binpkg_tmpfile, "w:bz2",
-				format=tarfile.PAX_FORMAT if xattrs else tarfile.DEFAULT_FORMAT)
-			tar_contents(contents, root, tar, protect=protect, xattrs=xattrs)
-			tar.close()
+			with open(binpkg_tmpfile, "wb") as fp:
+				proc = subprocess.Popen([compression["compress"], "-"], stdin=subprocess.PIPE, stdout=fp, shell=True)
```
zmedico: We should call `proc.stdin.close()` and `proc.wait()` after the tar_contents() call. We should also raise an error if `proc.wait()` returns an unsuccessful status.

zmedico: Also, generally, we never use `shell=True` for subprocess calls.

mrueg: I know shell is bad, especially as external input is parsed.
Should we remove support for ${PORTAGE_BZIP2_COMMAND}?

bin/quickpkg
mrueg: That needs to propagate back, as quickpkg_atom gets called via different other methods.

bin/quickpkg
mrueg: Okay, I'll have a look at that one. I was aware of shutil.which() but not of find_binary(). I was considering distutils.spawn.find_executable().
Do you think assuming compression["decompress"].partition(' ')[0] is a safe assumption for the binaryname?

pym/_emerge/BinpkgExtractorAsync.py
```diff
@@ -38,6 +39,15 @@ def _start(self):
 			self.returncode = 1
 			self._async_wait()
 			return
+		if subprocess.call("type" + decomp_cmd, shell=True,
```
zmedico: The calling code is asynchronous, so this subprocess call needs to be asynchronous as well. In fact, BinpkgExtractorAsync inheris from SpawnProcess specifically for this purpose. You should assign self.args and then call SpawnProcess._start(self) like this code in this file that immediately follows.

mrueg:
@dol-sen I wasn't that pyDeComp exists. This would add a python package as a dependency to portage. I thought this is something we might want to avoid. We could incorporate it in portage.utils because catalyst will use portage anyway(?).

pym/portage/dbapi/bintree.py
```diff
@@ -568,7 +579,7 @@ def _populate(self, getbinpkgs=False, getbinpkg_refresh=True):
 				metadata[_instance_key(cpv)] = d
 				path = d.get("PATH")
 				if not path:
-					path = cpv + ".tbz2"
+					path = cpv + "." + self.compression["suffix"]
```
zmedico: My plan was to use the `.xpak` suffix for all types of compression. The `.xpak` extension is already used when `FEATURES=binpkg-multi-instance` is enabled. We can use it here too.

bin/quickpkg
zmedico: We can use portage.util.varexpand() to support ${PORTAGE_BZIP2_COMMAND} without using a shell.

bin/quickpkg
zmedico: Generally, use `portage.util.shlex_split(compression["decompress"])[0]`.

bin/quickpkg
zmedico: We can propagate it by raising an exception.

zmedico:
Does pyDeComp support configuring the number of decompression threads? We definitely want control over decompression threads (eventually at least).

bin/quickpkg
mrueg: It seems like varexpand does not support :- syntax.
```
from portage.util import varexpand
varexpand("${PORTAGE_BZIP2_COMMAND:-bzip2}")
```
Results in
```
unexpected EOF while looking for matching `}'
''
```

bin/quickpkg
mrueg: I can probably leave it out of quickpkg, as binarytree.__init__ will raise an exception.

bin/quickpkg
zmedico: Right, so we'll have to do the equivalent in python, for example:
```
dcomp_command = settings.get('PORTAGE_BZIP2_COMMAND', 'bzip2')
```

bin/quickpkg
zmedico: As long as quickpkg exits unsuccessfully when it fails, and we don't leave any garbage files in $PKGDIR, then we're good.

bin/quickpkg
mrueg: Okay, this is going to work as soon as varexpand() supports :- syntax

bin/quickpkg
mrueg: There's a ValueError raised from ```binarytree.__exit__``` and I removed the duplicated code in quickpkg

bin/quickpkg
mrueg: What about this case?
```decompress": "${PORTAGE_BUNZIP2_COMMAND:-${PORTAGE_BZIP2_COMMAND}} -d"```

File: pym/portage/dbapi/bintree.py (hunk 5 above)
Comment (mrueg) [inline pym/portage/dbapi/bintree.py]: I can change it to xpak. That will also make sure there's only a single file for a PVR available. We might loose some visibility about the used compression format.

File: bin/quickpkg (hunk 3 above)
Comment (zmedico) [inline bin/quickpkg]: We could change that to be a tuple of things for try, which will work with varexpand:
```
decompress: ("${PORTAGE_BUNZIP2_COMMAND}", "${PORTAGE_BZIP2_COMMAND} -d")
```

**Comment by dol-sen** [comment]
no, not yet, but likely not hard to add that option.  Actually the options need to be made more configurable.

File: pym/portage/dbapi/bintree.py [hunk 6]
```diff
@@ -141,7 +148,7 @@ def aux_get(self, mycpv, wants, myrepo=None):
 				return [aux_cache.get(x, "") for x in wants]
 		mysplit = mycpv.split("/")
 		mylist = []
-		tbz2name = mysplit[1]+".tbz2"
+		tbz2name = mysplit[1] + "." + self.bintree.compression["suffix"]
```
zmedico: Seems tbz2name is actually unused in this function, so we can just remove it.

pym/portage/dbapi/bintree.py
zmedico: For code like this we should also support "tbz2" for backward compatibility. Using xpak for all compression types is good, since compression_probe allows us to quickly determine the compression whenever we need to.

pym/portage/dbapi/bintree.py
mrueg: Thanks, removed.

pym/portage/dbapi/bintree.py
```diff
@@ -373,6 +382,17 @@ def __init__(self, _unused=DeprecationWarning, pkgdir=None,
...
+			try:
+				binpkg_compression = settings.get("BINPKG_COMPRESSION", "bzip2")
+				self.compression = _compressors[binpkg_compression]
+			except KeyError as e:
+				raise ValueError("Invalid or unsupported compression method: %s" % e.args[0])
+
+			compression_binary = varexpand(shlex_split(self.compression["compress"])[0])
+			if find_binary(compression_binary) is None:
+				missing_package = self.compression["package"]
+				raise ValueError("File compression unsupported %s. Missing package: %s" % (binpkg_compression, missing_package))
```
zmedico: The unit tests are currently failing here in travis:
```
ValueError: File compression unsupported bzip2. Missing package: app-arch/bzip2
```

mrueg: Yep, that's because I still need to add support for ${BZIP2_COMMAND:-bzip2} or a suitable fallback.

mrueg: It seems like I have not the right idea of how to make properly use of portage.settings for this. Do I need to set portage._internal_caller = True?

zmedico: You're not allowed to us portage.settings there. Whatever code uses it those variables should have a config instance and use portage.util.varexpand() to evaluate the variables on-demand.

zmedico: The reason is that portage.settings is deprecated in most areas of the portage code. It's still used in some places, but mostly as a fallback when another config instance has not been passed in.

mrueg: Is mydict=settings the correct fix here? (it seems to work, not sure if there's any better).

zmedico: Yes, passing mydict=settings to varexpand is fine when you want access to all variables defined via make.defaults, make.conf, and package.env. For package.env support, settings.setcpv has to be called first, but that's probably already handled in the calling code.

pym/portage/dbapi/bintree.py
```diff
@@ -672,17 +693,17 @@ def _populate(self, getbinpkgs=False, getbinpkg_refresh=True):
 					multi_instance = False
 					invalid_name = False
 					build_id = None
-					if myfile.endswith(".xpak"):
+					if myfile.endswith("." + self.suffix):
```
mrueg: @zmedico If we use xpak as a general suffix, this will treat all packages as multi_instance. I assume this needs to be changed, then?

pym/portage/dbapi/bintree.py
```diff
@@ -672,17 +693,17 @@ def _populate(self, getbinpkgs=False, getbinpkg_refresh=True):
...
 						multi_instance = True
 						build_id = self._parse_build_id(myfile)
```
mrueg: _parse_build_id also seems to return strings instead of a long sometime because the method catches the ValueError. @zmedico Any advice here?

pym/portage/dbapi/bintree.py
zmedico: I feel like the self.suffix changes should be split into a separate patch. We can choose compression method with having change the suffix code at all, because compression_probe does not care about the file suffix.

pym/portage/dbapi/bintree.py
zmedico: Again, we really don't have to touch the suffix code in order to choose compression, so let's do that later in a separate patch.

mrueg:
@zmedico Removed the suffix renaming to make the patch more readable

pym/portage/dbapi/bintree.py
```diff
@@ -373,6 +380,16 @@ def __init__(self, _unused=DeprecationWarning, pkgdir=None,
...
+			try:
+				binpkg_compression = settings.get("BINPKG_COMPRESSION", "bzip2")
+				self.compression = _compressors[binpkg_compression]
+			except KeyError as e:
+				raise ValueError("Invalid or unsupported compression method: %s" % e.args[0])
+
+			compression_binary = varexpand(shlex_split(self.compression["compress"])[0], mydict=settings)
+			if find_binary(compression_binary) is None:
+				missing_package = self.compression["package"]
+				raise ValueError("File compression unsupported %s. Missing package: %s" % (binpkg_compression, missing_package))
```
zmedico: We'll need to handle these exceptions gracefully. We can add a method to validate the BINPKG_COMPRESSION when necessary. I really only needs to be validated by emerge commands that build packages (including FEATURES=unmerge-backup) and quickpkg commands.

pym/_emerge/BinpkgExtractorAsync.py
```diff
@@ -39,6 +46,21 @@ def _start(self):
 			self._async_wait()
 			return
 
+		decompression_binary = varexpand(shlex_split(decomp_cmd)[0], mydict=self.env)
```
zmedico: This `shlex_split(decomp_cmd)[0]` part looks like it might raise IndexError when PORTAGE_BUNZIP2_COMMAND is undefined?

mrueg: Good catch, i thought it was defined in actions.py as well, but there's only a fallback for PORTAGE_BZIP2_COMMAND.

pym/portage/dbapi/bintree.py
mrueg: So how about a BinpkgCompError, raise it here and catch it in emerge and quickpkg when instantiating the binarytree class?

zmedico: That might work. Mainly, I don't want essential commands like `portageq` and `emerge --sync` to fail because of this.

zmedico: If you raise an exception unconditionally, it might be hard to update all the code to handle it, so that's why it might be better to have a method that we call to validate it when needed.

pym/_emerge/BinpkgExtractorAsync.py
mrueg: Actually the indexing needs to happen outside the varexpand and there I need to catch if its an empty one.

zmedico: This could be used as-is, but you would need to check for empty decompression_binary afterwards.

zmedico: Well, I guess it's best to split after the varexpand call, in case PORTAGE_BZIP2_COMMAND contains multiple tokens.

pym/portage/dbapi/bintree.py
mrueg: Okay, moved it back to quickpkg

pym/_emerge/BinpkgExtractorAsync.py
mrueg: Catching the index error now.

mrueg:
In the last revision I added a fallback so the compression package will always be using bzip2 in order to avoid a chicken and egg problem for a client of a binhost.

bin/quickpkg
```diff
@@ -134,12 +138,36 @@ def quickpkg_atom(options, infos, arg, eout):
...
-			tar = tarfile.open(binpkg_tmpfile, "w:bz2",
-				format=tarfile.PAX_FORMAT if xattrs else tarfile.DEFAULT_FORMAT)
-			tar_contents(contents, root, tar, protect=protect, xattrs=xattrs)
-			tar.close()
+			binpkg_compression = settings.get("BINPKG_COMPRESSION", "bzip2")
+			try:
+				compression = _compressors[binpkg_compression]
+			except KeyError as e:
+				eout.error("Invalid or unsupported compression method: %s" % e.args[0])
+				return False
```
zmedico: Nothing checks the quickpkg_atom return value. We want quickpkg to abort an return unsuccessful status in this case.

zmedico:
We need support in the dyn_package function inside bin/misc-functions.sh. Maybe it's easiest to use a helper script, like it uses `"$PORTAGE_BIN_PATH"/xpak-helper.py` now.

mrueg:
Okay, can add it. Just curious, what's using misc-functions.sh? I didn't find any traces for __dyn_package in portage.

zmedico:
The __dyn_package function is called for the `package` phase, which is how binary packages are usually created (except from quickpkg). It's called by emerge via the EbuildBinpkg class.

mrueg:
For some reason the PR didn't update anymore. https://github.com/mrueg/portage/commit/8c3cafae09ee840d2d5ab390672a4361e56c9b38 *sigh*

bin/misc-functions.sh
```diff
@@ -478,8 +478,11 @@ __dyn_package() {
 	[ -z "${PORTAGE_BINPKG_TMPFILE}" ] && \
 		die "PORTAGE_BINPKG_TMPFILE is unset"
 	mkdir -p "${PORTAGE_BINPKG_TMPFILE%/*}" || die "mkdir failed"
+	COMPRESSION_COMMAND=$(PYTHONPATH=${PORTAGE_PYTHONPATH:-${PORTAGE_PYM_PATH}} \
```
zmedico: If all that we really need is COMPRESSION_COMMAND, then the helper script is overkill. We should just pass a variable from doebuild_environment function. The variable name must be prefixed with `PORTAGE_`.

mrueg: COMPRESSION_COMMAND is depending on the PN to be package, so I guess it makes sense.

zmedico:
If the PORTAGE_COMPRESSION_COMMAND setting in doebuild_environment works now, I guess we're about ready to merge this.

pym/portage/package/ebuild/doebuild.py
```diff
@@ -518,6 +526,23 @@ def doebuild_environment(myebuild, mydo, myroot=None, settings=None,
 				mysettings["KV"] = ""
 			mysettings.backup_changes("KV")
 
+	binpkg_compression = mysettings.get("BINPKG_COMPRESSION", "bzip2")
+	try:
+		compression = _compressors[binpkg_compression]
+	except KeyError as e:
+		raise TypeError("Invalid or unsupported compression method: %s" % e.args[0])
```
zmedico: If it's possible for the user to trigger this somehow, then we need to handle the exception somewhere and make emerge bail out before dependency calculations.

The BINPKG_COMPRESSION code should all go under the `if mydo != "depend":` conditional, since we want to minimize the code executed for the "depend" phase.

zmedico: We want emerge to bail out somewhere in  pym/_emerge/actions.py, either in the run_action or action_build function. Actions like emerge --info and emerge --sync should still succeed when the binpkg compression configuration is invalid.

pym/portage/package/ebuild/doebuild.py
```diff
@@ -518,6 +526,23 @@ def doebuild_environment(myebuild, mydo, myroot=None, settings=None,
...
+		binpkg_compression = mysettings.get("BINPKG_COMPRESSION", "bzip2")
+		try:
+			compression = _compressors[binpkg_compression]
+		except KeyError as e:
+			writemsg("Warning: Invalid or unsupported compression method: %s" % e.args[0])
+		try:
+			compression_binary = shlex_split(varexpand(compression["compress"], mydict=settings))[0]
+		except IndexError as e:
+			writemsg("Warning: Invalid or unsupported compression method: %s" % e.args[0])
+		if find_binary(compression_binary) is None:
```
zmedico: Something like this:
```python
try:
	compression = _compressors[binpkg_compression]
except KeyError as e:
	writemsg("Warning: Invalid or unsupported compression method: %s" % e.args[0])
else:
	try:
		compression_binary = shlex_split(varexpand(compression["compress"], mydict=settings))[0]
	except IndexError as e:
		writemsg("Warning: Invalid or unsupported compression method: %s" % e.args[0])
	else:
		if find_binary(compression_binary) is None:
```

mrueg:
Patch available as: https://github.com/gentoo/portage/commit/cff2c0149142843316e1851c2e73bcec30f08471

Thanks for the reviews!

PR #171, opened 2017-06-01
depgraph: prune unnecessary rebuilds for --autounmask-continue (bug 619626)

zmedico:
When there are autounmask USE changes, avoid unnecessary rebuilds
by accepting binary packages that were rejected due to the preexisting
USE configuration. This reuses the prune_rebuilds backtracker support
which was added for bug 439688.

X-Gentoo-bug: 619626
X-Gentoo-bug-url: https://bugs.gentoo.org/show_bug.cgi?id=619626

### DIFF
--- a/pym/_emerge/depgraph.py
+++ b/pym/_emerge/depgraph.py
@@ -5,6 +5,7 @@ from __future__ import division, print_function, unicode_literals
 
 import collections
 import errno
+import functools
 import io
 import logging
 import stat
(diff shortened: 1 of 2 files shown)


pym/_emerge/depgraph.py
```diff
@@ -885,10 +880,70 @@ def _show_ignored_binaries(self):
...
+			can be avoided by backtracking
+		"""
+		if not self._dynamic_config._allow_backtracking:
+			return False
+
+		if not self._dynamic_config._needed_use_config_changes:
+			return False
+
+		if not self._dynamic_config.ignored_binaries:
+			return False
```
dol-sen: are these already evaluated? not properties...  Then this could be shortened to:

if False in [ self._dynamic_config._allow_backtracking,
                  self._dynamic_config._needed_use_config_changes,
 		  self._dynamic_config.ignored_binaries]:
			return False

zmedico: Thanks! Updated as follows:
```
if not all([
	self._dynamic_config._allow_backtracking,
	self._dynamic_config._needed_use_config_changes,
	self._dynamic_config.ignored_binaries]):
	return False
```

PR #173, opened 2017-06-16
emerge: Add head commit per repo to --info

mrueg:
This adds more information to emerge --info output if timestamp.commit exists.
``` Head commit of repository gentoo: 0518b330edac963f54f98df33391b8e7b9eaee4c ```

### DIFF
--- a/pym/_emerge/actions.py
+++ b/pym/_emerge/actions.py
@@ -1644,8 +1644,18 @@ def action_info(settings, trees, myopts, myfiles):
 
 	for repo in repos:
 		last_sync = portage.grabfile(os.path.join(repo.location, "metadata", "timestamp.chk"))
+		head_commit = None
 		if last_sync:
 			append("Timestamp of repository %s: %s" % (repo.name, last_sync[0]))
+		if repo.sync_type:
+			sync = portage.sync.module_controller.get_class(repo.sync_type)()
+			options = { 'repo': repo }
+			try:
+				head_commit = sync.retrieve_head(options=options)
+			except NotImplementedError:
+				head_commit = (1, False)
+		if head_commit and head_commit[0] == os.EX_OK:
+			append("Head commit of repository %s: %s" % (repo.name, head_commit[1]))
 
 	# Searching contents for the /bin/sh provider is somewhat
 	# slow. Therefore, use the basename of the symlink target
--- a/pym/portage/sync/modules/git/git.py
+++ b/pym/portage/sync/modules/git/git.py
@@ -130,3 +130,15 @@ class GitSync(NewBase):
 			cwd=portage._unicode_encode(self.repo.location))
 
 		return (os.EX_OK, current_rev != previous_rev)
+
+	def retrieve_head(self, **kwargs):
+		'''Get information about the head commit'''
+		if kwargs:
+			self._kwargs(kwargs)
+		rev_cmd = [self.bin_command, "rev-list", "--max-count=1", "HEAD"]
+		try:
+			ret = (os.EX_OK, subprocess.check_output(rev_cmd,
+				cwd=portage._unicode_encode(self.repo.location)))
+		except CalledProcessError:
+			ret = (1, False)
+		return ret
--- a/pym/portage/sync/modules/rsync/rsync.py
+++ b/pym/portage/sync/modules/rsync/rsync.py
@@ -303,6 +303,18 @@ class RsyncSync(NewBase):
 			return (1, False)
 		return self.update()
 
+	def retrieve_head(self, **kwargs):
+		'''Get information about the head commit'''
+		if kwargs:
+			self._kwargs(kwargs)
+		last_sync = portage.grabfile(os.path.join(self.repo.location, "metadata", "timestamp.commit"))
+		ret = (1, False)
+		if last_sync:
+			try:
+				ret = (os.EX_OK, last_sync[0].split()[0])
+			except IndexError:
+				pass
+		return ret
 
 	def _set_rsync_defaults(self):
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: pym/_emerge/actions.py [hunk 1]
```diff
@@ -1646,6 +1646,10 @@ def action_info(settings, trees, myopts, myfiles):
 		last_sync = portage.grabfile(os.path.join(repo.location, "metadata", "timestamp.chk"))
 		if last_sync:
 			append("Timestamp of repository %s: %s" % (repo.name, last_sync[0]))
+		last_sync = portage.grabfile(os.path.join(repo.location, "metadata", "timestamp.commit"))
+		if last_sync:
+			head_commit = last_sync[0].split()[0]
```
zmedico: This will raise IndexError if the first line is empty for some reason, so we should handle that.

mrueg: Yeah I guess you're right. I guess I shouldn't assume the checkout to be in a sane state. ;)

pym/_emerge/actions.py
```diff
@@ -1646,6 +1646,13 @@ def action_info(settings, trees, myopts, myfiles):
 		last_sync = portage.grabfile(os.path.join(repo.location, "metadata", "timestamp.chk"))
 		if last_sync:
 			append("Timestamp of repository %s: %s" % (repo.name, last_sync[0]))
+		last_sync = portage.grabfile(os.path.join(repo.location, "metadata", "timestamp.commit"))
+		if last_sync:
+			try:
+				head_commit = last_sync[0].split()[0]
+				append("Head commit of repository %s: %s" % (repo.name, head_commit))
+			except IndexError:
+				pass
```
zmedico: In order to avoid swallowing unintended exceptions, I prefer to limit the scope of the exception handler as much as possible, for example:
```
	try:
		head_commit = last_sync[0].split()[0]
 	except IndexError:
		pass
	else:
		append("Head commit of repository %s: %s" % (repo.name, head_commit))
```
This also makes it easier for the reader to discern where the exceptions are expected to be raised from.

mrueg: I also forgot to reset head_commit in the loop. Next revision includes support to provide the same information for sync_type git as well.

pym/_emerge/actions.py
```diff
@@ -1644,8 +1644,30 @@ def action_info(settings, trees, myopts, myfiles):
...
+		head_commit = None
 		if last_sync:
 			append("Timestamp of repository %s: %s" % (repo.name, last_sync[0]))
+		last_sync = portage.grabfile(os.path.join(repo.location, "metadata", "timestamp.commit"))
+		if last_sync:
+			try:
+				head_commit = last_sync[0].split()[0]
+			except IndexError:
+				pass
+		elif repo.sync_type == "git":
```
zmedico: Maybe this can be a method of GitSync and we can load the sync plugins to get find this info.

zmedico: SyncManager calls `portage.sync.module_controller.get_class(repo.sync_type)`, and the class constructor takes zero arguments. The new method will have to call `SyncBase._kwargs` to initialize self.repo.

mrueg: Created a method in GitSync and RsyncSync for it called retrieve_head. Tested with SVNsyncer that it'll properly raise and catch a NotImplementedError.
Please have a look at the code for _kwargs by providing a options dict to the retrieve_head function, I'm not sure if this matches with what you had in mind there.

pym/portage/sync/modules/rsync/rsync.py
```diff
@@ -303,6 +303,17 @@ def new(self, **kwargs):
...
+		'''Get information about the head commit'''
+		if kwargs:
+			self._kwargs(kwargs)
+		last_sync = portage.grabfile(os.path.join(self.repo.location, "metadata", "timestamp.commit"))
+		if last_sync:
+			try:
+				ret = (os.EX_OK, last_sync[0].split()[0])
+			except IndexError:
+				ret = (1, False)
+		return ret
```
zmedico: This will raise NameError for ret if last_sync is empty.

pym/portage/sync/modules/git/git.py
```diff
@@ -130,3 +130,15 @@ def update(self):
 			cwd=portage._unicode_encode(self.repo.location))
 
 		return (os.EX_OK, current_rev != previous_rev)
+
+	def retrieve_head(self, **kwargs):
+		'''Get information about the head commit'''
+		if kwargs:
+			self._kwargs(kwargs)
```
zmedico: I guess we don't really have to call self._kwargs if we use the repo keyword argument directly, but self._kwargs could still be useful if we needed to call any other methods that access self.repo.

zmedico: Well, self._kwargs is also useful if some other method wants to call retrieve_head after it has already called self._kwargs earlier.

pym/portage/sync/modules/rsync/rsync.py
mrueg: Thanks, fixed.

pym/portage/sync/modules/git/git.py
mrueg: Okay, sounds good to me. I was relating to https://github.com/gentoo/portage/pull/173/files#diff-290c5cc3f4bd746508c90fd4d6e5c052R1654 if options is set up correctly of if there's a different way to do it.

mrueg:
https://github.com/gentoo/portage/commit/0e1699ad6b3f8eec56fbd6dd6255ed1145e89dd5

PR #177, opened 2017-07-04
.travis.yml: Add pypy3, nightly (3.7-dev) support (allow-failures)

mrueg:
X-Gentoo-Bug: https://bugs.gentoo.org/show_bug.cgi?id=575840


mrueg:
py3 unfortunately fails.  Upgrading to a later version requires the container-based trusty image, which sets sudo: False.
Need to figure out howto replace  `sudo rsync -a /tmp/install-root/. /` before upgrading.

mgorny:
@mrueg, can you reproduce it succeeding with a newer pypy3 locally?

mrueg:
@mgorny That's why I set travis it to "allow-failures" for pypy3 and python-nightly, so the button being green or red does not depend on those jobs. ;-)

I haven't tried pypy3 though locally.

zmedico:
The lxml dependency is annoying. I suppose we could revert 633cabfb9215633c554fd967b9875310be3718bd.

PR #180, opened 2017-07-25
PEP8 checks

elacheche:


mrueg:
LGTM, but I would prefer squashing them into a single commit

elacheche:
@mrueg Don't bother merging this.. This was my 1st attempt, since this commit, I knew that:

1. It's better to do a commit per file
2. I should commit this to the repoman branch instead of master
3. MY commits comments should be "repoman: filename: desc"
4. That the repoman branch build tests still need some work before I can commit my PEP8 checks (see #181 & #182)

I think I should close the PR! Because the master branch don't need it.. Thanks :)

PR #185, opened 2017-08-04
enable allow_build_id=True for user config keywords and use

tclh123:
Hello,

I need use build-id feature in /etc/portage/{package.keywords, package.use}. And I think the long ago patches http://comments.gmane.org/gmane.linux.gentoo.portage.devel/5239 missing this.


zmedico:
Merged in 9c819f72399d4fdae731883d9d915aa824d30fcc, thanks!

PR #186, opened 2017-08-06
eapply_user: combine sort for all dirs (bug 608880)

zmedico:
Combine the patch basenames from all matched directories into a
list, and apply them in POSIX sorted order.  This allows patches in
more-specific directories to override patches of the same basename found
in less-specific directories. An empty patch (or /dev/null symlink)
negates a patch with the same basename found in a less-specific
directory.

This behavior is much more flexible and intuitive than the previous one,
while remaining backward-compatible to some extent.

NOTE: The implementation uses an associative array, which requires bash
version 4 or later.

X-Gentoo-bug: 608880
X-Gentoo-bug-url: https://bugs.gentoo.org/608880

### DIFF
--- a/bin/phase-helpers.sh
+++ b/bin/phase-helpers.sh
@@ -1094,22 +1094,43 @@ if ___eapi_has_eapply_user; then
 
 		local basedir=${PORTAGE_CONFIGROOT%/}/etc/portage/patches
 
-		local d applied
+		local applied d f
+		local -A _eapply_user_patches
 		local prev_shopt=$(shopt -p nullglob)
 		shopt -s nullglob
 
-		# possibilities:
+		# Patches from all matched directories are combined into a
+		# sorted (POSIX order) list of the patch basenames. Patches
+		# in more-specific directories override patches of the same
+		# basename found in less-specific directories. An empty patch
+		# (or /dev/null symlink) negates a patch with the same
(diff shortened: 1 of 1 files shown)


bin/phase-helpers.sh
```diff
@@ -1094,23 +1094,44 @@ if ___eapi_has_eapply_user; then
...
 		done
 
+		if [[ ${#_eapply_user_patches[@]} -gt 0 ]]; then
+			while read -r -d '' f; do
+				f=${_eapply_user_patches[${f}]}
+				if [[ -s ${f} ]]; then
+					eapply "${f}"
+					applied=1
+				fi
+			done < <(printf -- "%s\0" "${!_eapply_user_patches[@]}" |
```
mrueg: shellcheck:  ^-- SC1117: Backslash is literal in "\0". Prefer explicit escaping: "\\0".
https://github.com/koalaman/shellcheck/wiki/SC1117

zmedico: Thanks, fixed now.

mgorny: I'd dare say a more readable fix would be to use single quotes, i.e. `'%s\0'`.

zmedico: Yeah that's better. Done now.

PR #187, opened 2017-08-11
Add emerge option to auto-unmask USE flags (and LICENSE) changes only.

zmedico:
https://bugs.gentoo.org/show_bug.cgi?id=622480

### DIFF
--- a/pym/_emerge/main.py
+++ b/pym/_emerge/main.py
@@ -131,6 +131,7 @@ def insert_optional_args(args):
 		'--autounmask-only'      : y_or_n,
 		'--autounmask-keep-masks': y_or_n,
 		'--autounmask-unrestricted-atoms' : y_or_n,
+		'--autounmask-use-only'  : y_or_n,
 		'--autounmask-write'     : y_or_n,
 		'--binpkg-changed-deps'  : y_or_n,
 		'--buildpkg'             : y_or_n,
@@ -355,6 +356,11 @@ def parse_opts(tmpcmdline, silent=False):
 
 		"--autounmask-write": {
 			"help"    : "write changes made by --autounmask to disk",
+			"choices" : true_y_or_n
+		},
+
+		"--autounmask-use-only": {
+			"help"    : "only calculate required USE flag changes",
 			"choices" : true_y_or_n
 		},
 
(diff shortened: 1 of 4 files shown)


zmedico:
Hmm, --autounmask-use-only seems like a misnomer given that it also allows licence changes. I'm thinking we should have separate --autounmask-use and --autounmask-licence options that are enabled by default.

zmedico:
Or rename this options to --autounmask-keep-keywords.

zmedico:
Soliciting feedback here:

https://archives.gentoo.org/gentoo-portage-dev/message/0b3693ce4c73031524dae92dbb5e085e

pym/_emerge/main.py
```diff
@@ -358,6 +359,11 @@ def parse_opts(tmpcmdline, silent=False):
 			"choices" : true_y_or_n
 		},
 
+		"--autounmask-use-only": {
+			"help"    : "only calculate required USE flag changes",
```
mrueg: Can this description be improved a bit? Maybe "write autounmask changes for USE flag changes only" (this would match --autounmask-unrestricted-atoms description)

pym/portage/tests/resolver/test_autounmask_use_only.py
```diff
@@ -0,0 +1,68 @@
...
+				unstable_keywords={'app-misc/B-1'},
+			),
+			# --autounmask-use-only prefers app-misc/A-1 because
+			# it can be installed without accepting unstable
+			# keyworkds
+			ResolverPlaygroundTestCase(
+				['app-misc/A'],
+				success = False,
+				options = {
+					"--autounmask-use-only": True,
```
mrueg: nitpick: should this option be 'y' instead of True to match with the previous test which uses 'n' instead of False?

zmedico: Fixed in https://github.com/gentoo/portage/pull/188.

zmedico:
https://github.com/gentoo/portage/pull/188 adds a --autounmask-keep-keywords option instead.

zmedico:
Closed in favor of https://github.com/gentoo/portage/pull/188.

PR #199, opened 2017-08-31
ebuild.sh: Completely ban external commands in global scope

mgorny:
Set PATH to /dev/null when sourcing the ebuild for dependency resolution
in order to prevent shell from finding external commands via PATH
lookup. While this does not prevent executing programs via full path, it
should catch the majority of accidental uses.

// Note: this can't be merged right now since we still have ebuilds
// calling external commands; see:
// https://bugs.gentoo.org/show_bug.cgi?id=629222

### DIFF
--- a/bin/ebuild.sh
+++ b/bin/ebuild.sh
@@ -80,8 +80,12 @@ else
 	done
 	unset funcs x
 
+	# prevent the shell from finding external executables
+	# note: we can't use empty because it implies current directory
+	_PORTAGE_ORIG_PATH=${PATH}
+	export PATH=/dev/null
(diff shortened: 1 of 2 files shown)


zmedico:
Does PMS say anything about this? Otherwise, looks fine.

mgorny:
https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-12400011.3

> When an ebuild is being sourced for metadata querying rather than for a build (that is to say, when none of the src_ or pkg_ functions are to be called), no external command may be executed. The package manager may take steps to enforce this.

PR #205, opened 2017-09-17
Fix emerge --info when using webrsync (bug 630538)

Nicop06:
When calling retrieve_head from a SyncBase object, it is expected to
raise a NotImplementedError. However, all classes that do not inherit
from NewBase will raise an AttributeError which is not caught.


zmedico:
Merged in 0d1d2b8dccdd5ce9f99358fe842fb968a0423a25. Thanks!

PR #207, opened 2017-09-19
postinst-qa-check.d: fix [[ ${files[@]} ]] logic in for loops

zmedico:
Use a separate all_files array to accumulate the results
from all loops, so that [[ ${files[@]} ]] only checks for
files found during the current loop.

### DIFF
--- a/bin/postinst-qa-check.d/50gnome2-utils
+++ b/bin/postinst-qa-check.d/50gnome2-utils
@@ -1,12 +1,12 @@
 # check for missing calls to gnome2-utils regen functions
 
 gnome2_icon_cache_check() {
-	local d f files=() find_args
+	local d f all_files=() find_args
 	for d in usr/share/icons/*/; do
 		# gnome2_icon_cache_update updates only themes with an index
 		[[ -f ${d}/index.theme ]] || continue
 
-		find_args=(
+		local files=() find_args=(
 			# gtk-update-icon-cache supports only specific file
 			# suffixes; match that to avoid false positives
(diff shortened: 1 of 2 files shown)


zmedico:
There's a fixup in 743bedc636671e1307735e82e49b0488f1433809.

bin/postinst-qa-check.d/50gnome2-utils
```diff
@@ -1,12 +1,12 @@
 # check for missing calls to gnome2-utils regen functions
 
 gnome2_icon_cache_check() {
-	local d f files=() find_args
+	local d f all_files=() find_args
```
mgorny: There's no point in including `find_args` here if you `local` it later.

bin/postinst-qa-check.d/50gnome2-utils
```diff
@@ -27,15 +27,16 @@ gnome2_icon_cache_check() {
 		# (note: yes, it will eagerly repeat the update for next dirs
```
mgorny: This comment is no longer valid now.

bin/postinst-qa-check.d/50gnome2-utils
zmedico: FIxed in 5fc7281657e2a5a4b22e969853364b142a37d05f, thanks.

bin/postinst-qa-check.d/50gnome2-utils
zmedico: Also fixed in 5fc7281657e2a5a4b22e969853364b142a37d05f.

PR #209, opened 2017-09-24
depgraph: make _minimize_children deterministic (bug 631894)

zmedico:
In order for the eliminate_pkg loop to produce deterministic results,
the order of the pkgs list must not be random. Prefer to eliminate
installed packages first, in case rebuilds are needed, and also sort
in ascending order so that older versions are eliminated first.

X-Gentoo-bug: 631894
X-Gentoo-bug-url: https://bugs.gentoo.org/631894

@danielrobbins

### DIFF
--- a/pym/_emerge/depgraph.py
+++ b/pym/_emerge/depgraph.py
@@ -3584,6 +3584,15 @@ class depgraph(object):
 						if atom_set.findAtomForPackage(pkg2, modified_use=self._pkg_use_enabled(pkg2)):
 							atom_pkg_graph.add(pkg2, atom)
 
+			# In order for the following eliminate_pkg loop to produce
+			# deterministic results, the order of the pkgs list must
+			# not be random (bug 631894). Prefer to eliminate installed
+			# packages first, in case rebuilds are needed, and also sort
+			# in ascending order so that older versions are eliminated
+			# first.
+			pkgs = (sorted(pkg for pkg in pkgs if pkg.installed) +
+				sorted(pkg for pkg in pkgs if not pkg.installed))
+
 			for pkg in pkgs:
 				eliminate_pkg = True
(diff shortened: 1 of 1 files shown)


pym/_emerge/depgraph.py
```diff
@@ -3584,6 +3584,15 @@ def _minimize_children(self, parent, priority, root_config, atoms):
 						if atom_set.findAtomForPackage(pkg2, modified_use=self._pkg_use_enabled(pkg2)):
 							atom_pkg_graph.add(pkg2, atom)
 
+			# In order for the following eliminate_pkg loop to produce
+			# deterministic results, the order of the pkgs list must
+			# not be random (bug 631894). Prefer to eliminate installed
+			# packages first, in case rebuilds are needed, and also sort
+			# in ascending order so that older versions are eliminated
+			# first.
+			pkgs = (sorted(pkg for pkg in pkgs if pkg.installed) +
```
mrueg: How big is the list going to be? Maybe it makes sense to bisect.insort here and not touch every element twice.
```
installed = []
not_installed = []
for pkg in pkgs:
  if pkg.installed:
    bisect.insort(installed, pkg)
  else:
    bisect.insort(not_installed,pkg)

pkgs = installed + not_installed
```

zmedico: Typically we're only dealing with 2 packages, so there's not much cost here to optimize away. Also, insort would be less efficient than something like this:
```python
installed = []
not_installed = []
for pkg in sorted(pkgs):
   if pkg.installed:
     installed.append(pkg)
   else:
     not_installed.append(pkg)
```
And that's going to have a negligible performance difference compared to my 2-line version that's easier to read.

mrueg: okay, if it's just a small list then nevermind my comment 👍

PR #210, opened 2017-09-27
Remove trailer when decompressing binary packages

matijaskala:
xz and zstd don't support trailing data
and fail upon encountering xpak trailer


zmedico:
There's a fixup for the PIPESTATUS handling in 26c0a0e31b5150f7ef23b2156b87863e47dd5996.

PR #215, opened 2017-10-03
man/portage.5: Update example, fix a typo

mrueg:
As some users might just copy the example for their overlays.


mrueg:
Applied as https://github.com/gentoo/portage/commit/44437b45ed34a158a2c4b50bc0ca71988af6ef05

PR #216, opened 2017-10-03
new docs for PackageTracker. Attempting to make them Sphinx-compatible.

danielrobbins:


pym/_emerge/resolver/package_tracker.py
```diff
@@ -31,41 +31,94 @@ def __len__(self):
...
-		@param soname_deps: enable soname match support
-		@type soname_deps: bool
+		:param soname_deps bool: Determines whether support for soname deps should be enabled or not.
+		:var _cp_pkg_map: The collection of to-be-installed (not yet merged) packages. We care about conflicts in these
+		 packages.
+		:var _cp_vdb_pkg_map: The collection of already-installed packages.
+		:var _multi_pkgs: A list of keys in ``self._cp_pkg_map`` that have potential slot and cpv conflicts.
+		:var _replacing: The mechanism by which ``PackageTracker`` records to-be-installed packages that 'cover up'
+		 already-installed packages. ``self._replacing[cp_key] = [ new_pkg_that_replaced_cp_key... ]``.
+		:var _replaced_by: ``self.replaced_by[cp_key] == [ replaced_pkg_1, replaced_pkg_2 ]
```
zmedico: The instance attributes should be documented as shown in the example here:

http://www.sphinx-doc.org/en/stable/ext/autodoc.html#directive-autoattribute

zmedico: Maybe [info fields](http://www.sphinx-doc.org/en/stable/domains.html#info-field-lists) are more appropriate, as suggested here:

https://stackoverflow.com/a/8659919

zmedico: I converted them to info fields in 8fba660fb08ddd468d934c7478dcfc8015330e9b.

pym/_emerge/resolver/package_tracker.py
```diff
@@ -31,41 +31,94 @@ def __len__(self):
...
 	"""
 
 	def __init__(self, soname_deps=False):
+
 		"""
-		@param soname_deps: enable soname match support
-		@type soname_deps: bool
+		:param soname_deps bool: Determines whether support for soname deps should be enabled or not.
+		:var _cp_pkg_map: The collection of to-be-installed (not yet merged) packages. We care about conflicts in these
+		 packages.
```
zmedico: I just checked, and installed packages are also added to _cp_pkg_map when they are added to the depgraph.

PR #218, opened 2017-10-13
Better cache implementation. Passing all tests.

danielrobbins:


pym/portage/dbapi/porttree.py
```diff
@@ -255,6 +255,48 @@ def __init__(self, _unused_param=DeprecationWarning, mysettings=None):
...
+		# This information can then be used by other functions like aux_get and findname2 to avoid
+		# doing an exhaustive search of repositories for ebuilds. Without this tweak, Portage will
+		# get slower and slower as more overlays are added.
+		#
+		# Also note that it is OK if this cache has some 'false positive' catpkgs in it. We use it
+		# to search for specific catpkgs listed in ebuilds. The likelihood of a false positive catpkg
+		# in our cache causing a problem is extremely low. Thus, the code below is optimized for
+		# speed rather than correctness.
+
+		self.better_cache = {}
```
zmedico: Let's name it `_better_cache` since we can consider it an implementation detail. If we do this then the nested lists will be created on demand:
```python
self._better_cache = collections.defaultdict(list)
```

pym/portage/dbapi/porttree.py
```diff
@@ -255,6 +255,48 @@ def __init__(self, _unused_param=DeprecationWarning, mysettings=None):
...
+
+		for repo_loc in reversed(self.porttrees):
+			repo = self.repositories.get_repo_for_location(repo_loc)
+			try:
+				for d in os.listdir(repo_loc):
+					if d[0] == "." or d[0] == "-":
+						continue
+					cat_dir = repo_loc + "/" + d
+					if not os.path.isdir(cat_dir):
```
zmedico: We could eliminate the stat call by doing something like this:
```python
try:
    cat_dir_list = os.listddir(cat_dir)
except OSError as e:
    if e.errno != errno.ENOTDIR:
        raise
    continue
```

pym/portage/dbapi/porttree.py
```diff
@@ -255,6 +255,48 @@ def __init__(self, _unused_param=DeprecationWarning, mysettings=None):
...
+						continue
+					try:
+						for p in os.listdir(cat_dir):
+							catpkg_dir = cat_dir + "/" + p
+							if not os.path.isdir(catpkg_dir):
+								continue
+							catpkg = d + "/" + p
+							if not catpkg in self.better_cache:
+								self.better_cache[catpkg] = []
```
zmedico: `self._better_cache = collections.defaultdict(list)` eliminates the above 2 lines.

pym/portage/dbapi/porttree.py
```diff
@@ -374,29 +416,29 @@ def findname2(self, mycpv, mytree=None, myrepo = None):
...
 			raise InvalidPackageName(mycpv)
 
+		cp = mysplit[0] + "/" + psplit[0]
+		if cp in self.better_cache:
+			for repo in self.better_cache[cp]:
+				if mytree and mytree != repo.location:
+					continue
+				elif (not myrepo) or \
+					(myrepo == repo.name) or \
+					(repo.aliases != None and myrepo in repo.aliases):
```
zmedico: Use `self.repoositories.get_location_for_name(myrepo)` here, it handles aliases.

zmedico:
It would be neat to make this on-demand, processing one category at a time. For example, when a list of repositories containing sys-apps/portage is requested, it would cache the sys-apps category results for all repositories. The category list for each repository could be cached as a frozenset, and we'd only call listdir for the sys-apps category on a particular repository if its cached category set contains sys-apps.

danielrobbins:
I like that idea. Will work on it.

zmedico:
According to PMS, for each repository we should use `${repo}/profiles/categories` to obtain the set of categories. Since portage currently isn't strict about this, there may be some non-conforming repositories, but it's easy enough to fix that.

danielrobbins:
Actually, PMS doesn't define how one must obtain a set of categories. It just defines the specification for categories and for the metadata/categories file. For the purposes of the cache, we are not obligated to perform one approach or another for the sake of PMS compliance. In fact it is fine if _better_cache has a false positive category; it causes no harm. Problems occur only if it accidentally omits a catpkg that is actually there.

zmedico:
True, the spec doesn't say that we have to use the `profiles/categories` file. However, the alternative is to treat directories like `licenses`, `metadata`, and `profiles` as though they are category directories, or maintain a blacklist.

danielrobbins:
It causes no harm to have false positives in better_cache, but we can check against the categories list to be more correct if you want. It should have a minimal impact on performance.

danielrobbins:
Remember that the way better cache is used is as follows: someone is going to ask it "Where does a catpkg of foo/bar exist?" As long as the caller ensures that it is using valid categories, we are fine. The caller (and consider that this is an internal API) is not going to be say "does metadata/mypkg exist?"

zmedico:
If we implement the on-demand cache, then we're going to need a cached set of categories for each repository. Having false-positives in that set will have a negligible impact, since those directories won't even be listed unless a dependency atom references the category.

Except, emerge arguments are allowed to omit the category. So in order to evaluate emerge arguments, we do need to exhaustively list all category directories.

zmedico:
So, in the absence of a package names cache, an on-demand _better_cache is not going to be tremendously helpful.

PR #228, opened 2017-11-14
repoman/setup.py: create pym -> lib symlink for unit tests (bug 637460)

zmedico:
Various code involving the .repoman_not_installed file assumes
that the 'pym' directory should be added to sys.path, therefore
create a pym -> lib symlink during unit tests.

Bug: https://bugs.gentoo.org/637460

### DIFF
--- a/repoman/setup.py
+++ b/repoman/setup.py
@@ -147,6 +147,11 @@ class x_clean(clean):
 			print('removing %s symlink' % repr(conf_dir))
 			os.unlink(conf_dir)
 
+		pym_dir = os.path.join(top_dir, 'pym')
+		if os.path.islink(pym_dir):
+			print('removing %s sylink' % repr(pym_dir))
+			os.unlink(pym_dir)
+
 		pni_file = os.path.join(top_dir, '.repoman_not_installed')
 		if os.path.exists(pni_file):
 			print('removing %s' % repr(pni_file))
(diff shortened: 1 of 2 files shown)


repoman/setup.py
```diff
@@ -393,6 +398,17 @@ def run(self):
...
 		os.symlink(conf_src, conf_dir)
 
+		# symlink 'pym' directory
+		pym_dir = os.path.join(self.top_dir, 'pym')
+		if os.path.exists(pym_dir):
+			if not os.path.islink(pym_dir):
+				raise SystemError('%s exists and is not a symlink (collision)'
+					% repr(pym_dir))
+			os.unlink(pym_dir)
+		pym_src = 'lib'
```
mgorny: But why `lib`? Can't we either fix the stuff that requires it to use `pym`, or finally drop the senseless `pym` if we need to have `lib` anyway?

zmedico: Yeah, it doesn't matter what the name of the directory is, because the code doing the sys.path insert has direct access to the actual directory name and therefor it doesn't need to make an assumption.

zmedico: I forgot that the files in the executable are not nested in the module structure, so they really don't have a way of knowing pym vs. lib. I'd really like to eliminate the sys.path inserts and just use PYTHONPATH when when tweaks are needed.

zmedico: In fact, PYTHONPATH is perfectly reliable for the purposes of unit tests, since the unit test code can easily export PYTHONPATH for use by any executables called in subprocesses.

PR #230, opened 2017-12-03
depgraph: eval disjunctive build deps earlier (bug 639346)

zmedico:
Since built-time deps tend to be a superset of run-time deps, evaluate
disjunctive build-time deps before run-time deps, so that choices for
build-time deps influence choices for run-time deps.

Also, fix OnlydepsMinimalTestCase to specify ambiguous_merge_order,
since the merge order is affected by the order of evaluation.

Bug: https://bugs.gentoo.org/639346

@danielrobbins


danielrobbins:
Thanks for this :) :+1:

PR #232, opened 2017-12-10
Fix a misprint for a SANDBOX_ON variable condition check

ahlinc:
While studying the portage source code I accidentally found a typo in a condition that checks the SANDBOX_ON variable state and currently has no sense as it always returns the true.
Also checked the code for similar misprints with the following command:
```grep -HPrn '\[ +-[nz] (\\"|")?[^$`\\][^$`]' .``` but there are no more ones :smiley_cat:


zmedico:
Fixes 94015a2fe3b2d886b71e874c10bdd2504e54a133.

zmedico:
Thanks!

PR #233, opened 2017-12-10
bin/doins.py: remove file before creating symlink (bug 640376)

zmedico:
Fixes: d9522ba661b5 ("Rewrite doins in python (bug 624526)")
Bug: https://bugs.gentoo.org/640376


zmedico:
Merged in 992b4ffdcbf469db543c9cab90a5f7a3a4125c97.

PR #236, opened 2017-12-11
Apply shellcheck recommendations

mrueg:
WIP, let's see if stuff breaks.
```
shellcheck bashrc-functions.sh 

In bashrc-functions.sh line 7:
        for x in $* ; do
                 ^-- SC2048: Use "$@" (with quotes) to prevent whitespace problems.


In bashrc-functions.sh line 8:
                has $x $EBUILD_DEATH_HOOKS || \
                    ^-- SC2086: Double quote to prevent globbing and word splitting.
                       ^-- SC2086: Double quote to prevent globbing and word splitting.


In bashrc-functions.sh line 15:
        for x in $* ; do
                 ^-- SC2048: Use "$@" (with quotes) to prevent whitespace problems.


In bashrc-functions.sh line 16:
                has $x $EBUILD_SUCCESS_HOOKS || \
                    ^-- SC2086: Double quote to prevent globbing and word splitting.
                       ^-- SC2086: Double quote to prevent globbing and word splitting.


In bashrc-functions.sh line 34:
        local KV=$@
                 ^-- SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.


In bashrc-functions.sh line 41:
        local KV=$@
                 ^-- SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.


In bashrc-functions.sh line 49:
        local KV=$@
                 ^-- SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.


In bashrc-functions.sh line 57:
        local KV_MAJOR=$(KV_major "$1")
              ^-- SC2155: Declare and assign separately to avoid masking return values.


In bashrc-functions.sh line 58:
        local KV_MINOR=$(KV_minor "$1")
              ^-- SC2155: Declare and assign separately to avoid masking return values.


In bashrc-functions.sh line 59:
        local KV_MICRO=$(KV_micro "$1")
              ^-- SC2155: Declare and assign separately to avoid masking return values.


In bashrc-functions.sh line 77:
        echo $(KV_to_int "${_RC_GET_KV_CACHE}")
             ^-- SC2046: Quote this to prevent word splitting.
             ^-- SC2005: Useless echo? Instead of 'echo $(cmd)', just use 'cmd'.
```

### DIFF
--- a/bin/bashrc-functions.sh
+++ b/bin/bashrc-functions.sh
@@ -4,16 +4,16 @@
 
 register_die_hook() {
 	local x
-	for x in $* ; do
-		has $x $EBUILD_DEATH_HOOKS || \
+	for x in "$@" ; do
+		has "$x" "$EBUILD_DEATH_HOOKS" || \
 			export EBUILD_DEATH_HOOKS="$EBUILD_DEATH_HOOKS $x"
 	done
 }
 
 register_success_hook() {
 	local x
-	for x in $* ; do
-		has $x $EBUILD_SUCCESS_HOOKS || \
+	for x in "$@" ; do
+		has "$x" "$EBUILD_SUCCESS_HOOKS" || \
 			export EBUILD_SUCCESS_HOOKS="$EBUILD_SUCCESS_HOOKS $x"
 	done
 }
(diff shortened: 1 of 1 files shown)


bin/bashrc-functions.sh
```diff
@@ -4,16 +4,16 @@
 
 register_die_hook() {
 	local x
-	for x in $* ; do
-		has $x $EBUILD_DEATH_HOOKS || \
+	for x in "$@" ; do
+		has "$x" "$EBUILD_DEATH_HOOKS" || \
```
zmedico: That won't work because EBUILD_DEATH_HOOKS is a space separated list, and has() expects a separate argument for each element.

PR #237, opened 2017-12-15
Remove old code.

antarus12345:
Most of these functions have not been called in years and have been "deprecated" for at least 3 years. In the case of autouse; longer.


pym/portage/dep/__init__.py
```diff
@@ -239,147 +238,6 @@ def strip_empty(myarr):
 		('portage.dep.strip_empty',), DeprecationWarning, stacklevel=2)
 	return [x for x in myarr if x]
 
-def paren_reduce(mystr, _deprecation_warn=True):
```
zmedico: This one's still used by gentoolkit. If we really want to remove something then we need to make the deprecation warning unconditional (it currently only triggers for internal callers).

mgorny: I've seen it in qa-scripts as well. @gentoo/qa.

PR #238, opened 2017-12-15
Disable dynamic-deps by default

mgorny:
We have prepared for this for quite a while and it's time to pull
the plug. Disable dynamic-deps by default and restore the standard
PMS behavior. This will cause some one-time pain but eventually will
result in improvement of ebuild quality, especially when developers
start experiencing the need for revbumps first hand.

### DIFF
--- a/man/emerge.1
+++ b/man/emerge.1
@@ -547,7 +547,7 @@ In dependency calculations, substitute the dependencies of installed
 packages with the dependencies of corresponding unbuilt ebuilds from
 source repositories. This causes the effective dependencies of
 installed packages to vary dynamically when source ebuild dependencies
-are modified. This option is enabled by default.
+are modified. This option is disabled by default.
 
 \fBWARNING:\fR
 If you want to disable \-\-dynamic\-deps, then it may be necessary to
--- a/pym/_emerge/FakeVartree.py
+++ b/pym/_emerge/FakeVartree.py
@@ -54,7 +54,7 @@ class FakeVartree(vartree):
 	is not a matching ebuild in the tree). Instances of this class are not
 	populated until the sync() method is called."""
 	def __init__(self, root_config, pkg_cache=None, pkg_root_config=None,
-		dynamic_deps=True, ignore_built_slot_operator_deps=False,
+		dynamic_deps=False, ignore_built_slot_operator_deps=False,
 		soname_deps=False):
 		self._root_config = root_config
 		self._dynamic_deps = dynamic_deps
--- a/pym/_emerge/Scheduler.py
+++ b/pym/_emerge/Scheduler.py
@@ -352,7 +352,7 @@ class Scheduler(PollScheduler):
 		"""
 		self._set_graph_config(graph_config)
 		self._blocker_db = {}
-		dynamic_deps = self.myopts.get("--dynamic-deps", "y") != "n"
+		dynamic_deps = self.myopts.get("--dynamic-deps", "n") != "n"
 		ignore_built_slot_operator_deps = self.myopts.get(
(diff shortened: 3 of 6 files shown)


mgorny:
@zmedico, could you look at the test failures Travis going to spew in a while? I'm not sure if they aren't actual bugs in dynamic-deps=n code path.

```
======================================================================

FAIL: testMultirepo (portage.tests.resolver.test_multirepo.MultirepoTestCase)

----------------------------------------------------------------------

  File "/home/travis/build/mgorny/portage/build/lib.linux-x86_64-2.7/portage/tests/__init__.py", line 222, in run

    testMethod()

  File "/home/travis/build/mgorny/portage/build/lib.linux-x86_64-2.7/portage/tests/resolver/test_multirepo.py", line 234, in testMultirepo

    self.assertEqual(test_case.test_success, True, test_case.fail_msg)

AssertionError: atoms: (dev-libs/H), key: mergelist, expected: ['dev-libs/I-2::repo2'], got: []


======================================================================

FAIL: testMultirepoUserConfig (portage.tests.resolver.test_multirepo.MultirepoTestCase)

----------------------------------------------------------------------

  File "/home/travis/build/mgorny/portage/build/lib.linux-x86_64-2.7/portage/tests/__init__.py", line 222, in run

    testMethod()

  File "/home/travis/build/mgorny/portage/build/lib.linux-x86_64-2.7/portage/tests/resolver/test_multirepo.py", line 396, in testMultirepoUserConfig

    self.assertEqual(test_case.test_success, True, test_case.fail_msg)

AssertionError: atoms: (dev-libs/J), key: mergelist, expected: [u'dev-libs/I-2'], got: []
```

kensington:
Has the 20140826 Council question/decision been addressed yet?
```
"The council asks the Portage team to first outline their long-term
  plan regarding removal or replacement of dynamic dependencies,
  before they remove this feature. In particular, tree policies and
  the handling of eclasses and virtuals need to be clarified."
  Accepted unanimously.
```

zmedico:
MultirepoTestCase is fixed by b47daf4d12ad54668c75f53661da4b252e665c22.

mgorny:
@kensington:
1. We are not removing it.
2. Tree policy on revbumps has been accepted lately (gentoo/devmanual#67).

kensington:
@mgorny you're sidestepping the decision. Please provide a direct response to the information requested by Council.

mgorny:
Then you have to be more specific.

kensington:
Please outline your long-term plan regarding the disabling of dynamic dependencies. In particular, please clarify the handling of eclasses and virtuals.

mgorny:
This is not specific at all.

Long-term plan: today we disable them. Enjoy!

Policy problems have been solved in the past. Almost all developers (with a single exception) don't have problems following the new policies. I'm using no-dyndeps for over 3 years already, and hadn't had more than a few issues which developers had no problem fixing. Of course, as they upgrade to no-dyndeps, they would notice the problems immediately and wouldn't need me to report them.

Eclasses are covered by the same policy as ebuilds inheriting them. If the change is major and requires revbump, you get to revbump all the ebuilds. Which usually indicates that you should think twice before doing it in place anyway.

I don't really get what you want to know about virtuals. They are normal packages for years now, and I don't see why they would need any special treatment. Especially that they are almost zero-cost to rebuild, so you can revbump them often without major consequences.

kensington:
It's good to know that you think you're exempt from Council decisions, thanks.

mgorny:
Merged as 2905e1c2c28db495b74fde7d9b3bfdd4e7515dbd.

PR #239, opened 2017-12-28
multilib_category.py: add ALTERA_NIOS2 ELF type

trofi:
Discovered as missing ELF as:

 * QA Notice: Unrecognized ELF file(s):
 *      ALTERA_NIOS2;/lib/libz.so.1.2.11;libz.so.1;;libc.so.6;

Signed-off-by: Sergei Trofimovich <slyfox@gentoo.org>

### DIFF
--- a/pym/portage/dep/soname/multilib_category.py
+++ b/pym/portage/dep/soname/multilib_category.py
@@ -35,7 +35,8 @@ from __future__ import unicode_literals
 
 from portage.util.elf.constants import (
 	EF_MIPS_ABI, EF_MIPS_ABI2, ELFCLASS32, ELFCLASS64,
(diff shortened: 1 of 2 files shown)


zmedico:
Are there 32 bit and 64 bit variants? Can you verify that compute_multilib_category would work correctly for all variants?

Note that compute_multilib_category has some special logic for x32, and we need to know whether nios2 supports any special variants like that.

trofi:
AFAIU nios2 has only one 32-bit ABI. No 64-bit variants, not other 32-bit variants.

Currently `binutils` supports `nios2elf` (bare metal) and `nios2linux` (linux userspace target).

https://www.altera.com/content/dam/altera-www/global/en_US/pdfs/literature/hb/nios2/n2cpu_nii51016.pdf

zmedico:
Looks good. Thanks!

PR #251, opened 2018-02-04
Fix misc. bash errors.

R030t1:
On line [699](https://github.com/gentoo/portage/compare/master...R030t1:master#diff-bf3d975db19a8df2c968da7bc7616aadL699), I [suspect the test may erroneously expand](https://github.com/koalaman/shellcheck/wiki/SC2144) into:

```if [ -f ./Makefile ./makefile -o -f ./GNUmakefile ] ; then```

If both files are present.

The other changes are almost stylistic.

### DIFF
--- a/bin/phase-helpers.sh
+++ b/bin/phase-helpers.sh
@@ -696,7 +696,7 @@ einstall() {
 		unset EI_DESTLIBDIR
 	fi
 
-	if [ -f ./[mM]akefile -o -f ./GNUmakefile ] ; then
+	if [ -f ./Makefile -o -f ./makefile -o -f ./GNUmakefile ] ; then
 		if [ "${PORTAGE_DEBUG}" == "1" ]; then
 			${MAKE:-make} -n prefix="${ED}usr" \
 				datadir="${ED}usr/share" \
@@ -810,7 +810,7 @@ __eapi4_src_install() {
 
 __eapi6_src_prepare() {
 	if [[ $(declare -p PATCHES 2>/dev/null) == "declare -a"* ]]; then
-		[[ -n ${PATCHES[@]} ]] && eapply "${PATCHES[@]}"
+		[[ -n ${PATCHES[*]} ]] && eapply "${PATCHES[@]}"
 	elif [[ -n ${PATCHES} ]]; then
 		eapply ${PATCHES}
 	fi
@@ -965,7 +965,7 @@ if ___eapi_has_einstalldocs; then
 					[[ -f ${d} && -s ${d} ]] && docinto / && dodoc "${d}"
 				done
 			elif [[ $(declare -p DOCS) == "declare -a"* ]] ; then
-				[[ ${DOCS[@]} ]] && docinto / && dodoc -r "${DOCS[@]}"
+				[[ ${DOCS[*]} ]] && docinto / && dodoc -r "${DOCS[@]}"
 			else
 				[[ ${DOCS} ]] && docinto / && dodoc -r ${DOCS}
 			fi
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: bin/phase-helpers.sh [hunk 1]
```diff
@@ -696,7 +696,7 @@ einstall() {
 		unset EI_DESTLIBDIR
 	fi
 
-	if [ -f ./[mM]akefile -o -f ./GNUmakefile ] ; then
+	if [ -f ./Makefile -o -f ./makefile -o -f ./GNUmakefile ] ; then
```
mgorny: Don't use `[ ... ]` in bash, it's the source of issues like the one you're trying to fix.

bin/phase-helpers.sh
```diff
@@ -810,7 +810,7 @@ __eapi4_src_install() {
 
 __eapi6_src_prepare() {
 	if [[ $(declare -p PATCHES 2>/dev/null) == "declare -a"* ]]; then
-		[[ -n ${PATCHES[@]} ]] && eapply "${PATCHES[@]}"
+		[[ -n ${PATCHES[*]} ]] && eapply "${PATCHES[@]}"
```
mgorny: This is a meaningless change from my coding style to yours. No value for anyone, especially given that `[@]` is used on the same line.

bin/phase-helpers.sh
```diff
@@ -1014,6 +1014,7 @@ if ___eapi_has_eapply; then
 		local i found_doublehyphen
 		# first, try to split on --
 		for (( i = 1; i <= ${#@}; ++i )); do
+			# shellcheck disable=SC2199
```
mgorny: Why not use `${@[i]}` instead?

bin/phase-helpers.sh
R030t1: The use of `[` or `[[` doesn't seem to matter. I can't get a similar test to work:
```sh
touch a ab ac
[ -f "./a[bc]" ] && echo 'true'
[[ -f "./a[bc]" ]] && echo 'true'
```
Produces no output.

The test can be changed to `[[` for consistency. I left it as it was because I suspected it was chosen for a reason. That reason may have been the test did not obviously fail, per the `shellcheck` Wiki.

bin/phase-helpers.sh
R030t1: The reasoning is that `[@]` actually means "expand as separate words" while `[*]` means "expand as a single word." Consequently, [the semantics of this expansion may be misleading to some readers](https://github.com/koalaman/shellcheck/wiki/SC2199).

bin/phase-helpers.sh
R030t1: Are referring to the line below, containing `[[ ${@:i:1} == -- ]]`? It seems like `shellcheck` is is not parsing the script properly.

bin/phase-helpers.sh
zmedico: I've just tested with with app-shells/bash-4.4_p12, and `[ -f ./[mM]akefile ]` works with the single brackets, while `[[ -f ./[mM]akefile ]]` does not. So that explains the use of single brackets.

However, if both files exist then `[ -f ./[mM]akefile ]` fails:
```shell
$ [ -f ./[mM]akefile ]
bash: [: ./makefile: binary operator expected
$ echo $?
2
```

zmedico: Fixed in 46324633cbcf50fa6e09bedfe317f088de50fc08.

bin/phase-helpers.sh
zmedico: I feel like the shellcheck authors might be willing to accept `[[ -n ${PATCHES[@]} ]]` or `[[ ${PATCHES[@]} ]]` as correct, since the "Problematic code" that they cite involves an incorrect comparison:

https://github.com/koalaman/shellcheck/wiki/SC2199#problematic-code

zmedico: Fixed in e42e9b61024d752a3dfb56ab8e8ea6936333ec58 to use the array length.

bin/phase-helpers.sh
zmedico: Actually `${*:i:1}` will work here, as shellcheck suggests (and `${@[i]}` is invalid syntax).

I suspect that the shellcheck authors would agree that `${@:i:1}` is valid in this context, since the length of the slice is exactly 1, so there's no implicit concatenation happening.

floppym:
This PR no longer applies cleanly, and there have been no updates in years. Closing.

PR #254, opened 2018-02-04
PMS-compliant USE_EXPAND handling

mgorny:
Now possible to merge since LINGUAS is no more.


pym/portage/package/ebuild/config.py
```diff
@@ -1775,26 +1735,17 @@ def setcpv(self, mycpv, use_cache=None, mydb=None):
...
 				continue
 			for i in range(len(x_split) - 1):
 				k = '_'.join(x_split[:i+1])
 				if k in use_expand_split:
-					v = use_expand_iuses.get(k)
-					if v is None:
-						v = set()
-						use_expand_iuses[k] = v
-					v.add(x)
+					v = use_expand_iuses[k].add(x)
```
zmedico: Unnecessary assignment of `v` here.

mgorny: Yeah, looks like I forgot to remove it when I modified the `v=` line ;-).

PR #258, opened 2018-02-21
portage.process.spawn: default close_fds=False (bug 648432)

zmedico:
For python3.4 and later, default to close_fds=False, since file
descriptors are non-inheritable by default due to PEP 446. This solves
a performance problem on systems like FreeBSD, where our get_open_fds
function returns all possible file descriptor values (including those
that are not open).

Bug: https://bugs.gentoo.org/648432
See: https://www.python.org/dev/peps/pep-0446/

### DIFF
--- a/bin/ebuild
+++ b/bin/ebuild
@@ -58,6 +58,8 @@ import portage.util
 from _emerge.Package import Package
 from _emerge.RootConfig import RootConfig
 
+portage.process.sanitize_fds()
+
 description = "See the ebuild(1) man page for more info"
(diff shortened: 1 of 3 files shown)


mgorny:
Weird. This doesn't seem to fix the bug for me at all ;-f. Maybe something else is closing all those fds? (using py3.5 here)

zmedico:
If you see something else closing fds, maybe it's the subprocess module, since subprocess.Popen defaults to close_fds=True.

PR #261, opened 2018-02-26
EAPI 7 clean branch (for merging)

mgorny:
This branch contains EAPI 7 commits that are ready for review & merge.

### DIFF
--- a/pym/portage/dep/__init__.py
+++ b/pym/portage/dep/__init__.py
@@ -1,5 +1,5 @@
 # deps.py -- Portage dependency resolution functions
-# Copyright 2003-2014 Gentoo Foundation
+# Copyright 2003-2018 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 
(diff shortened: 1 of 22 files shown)


pym/portage/dep/__init__.py
```diff
@@ -562,6 +562,9 @@ def missing_white_space_check(token, pos):
 					basestring):
 					if stack[level][-1] == "||" and not l:
 						#Optimize: || ( ) -> .
+						if not eapi_attrs.empty_groups_always_true:
+							# in EAPI 7+, we need to fail here
+							l.append(token_class("__const__/empty-any-of"))
```
zmedico: We should update the token_default to  `token_class=_unicode`, for calls that do not specify token_class like in pym/portage/package/ebuild/_config/LicenseManager.py. Ultimately we'll need some special user interface handling, but this is fine for now I guess.

mgorny: And done!

PR #264, opened 2018-03-06
EAPI 7, part two

mgorny:
Another batch of relatively safe patches.


bin/ebuild-helpers/nonfatal
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# Copyright 1999-2017 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+source "${PORTAGE_BIN_PATH}"/isolated-functions.sh || exit 1
+
+if ! ___eapi_has_nonfatal_as_executable; then
+	die "$FUNCNAME() not supported as fallback helper in this EAPI"
```
Arfrever: This is not in a function, so `$FUNCNAME` is always empty.

bin/ebuild-helpers/nonfatal
```diff
@@ -0,0 +1,14 @@
...
+fi
+if [[ $# -lt 1 ]]; then
+	die "$FUNCNAME(): Missing argument"
```
Arfrever: And here.

bin/ebuild.sh
```diff
@@ -137,6 +137,11 @@ shift $#
 
 # Unset some variables that break things.
 unset GZIP BZIP BZIP2 CDPATH GREP_OPTIONS GREP_COLOR GLOBIGNORE
+if ___eapi_has_ENV_UNSET; then
+	for x in ${ENV_UNSET}; do
```
Arfrever: Add `unset x` after this loop.

bin/phase-helpers.sh
```diff
@@ -1172,6 +1172,10 @@ if ___eapi_has_in_iuse; then
 	}
 fi
 
+if ___eapi_has_version_functions; then
+	source "${PORTAGE_BIN_PATH}/eapi7-ver-funcs.sh" || exit 1
```
Arfrever: This is not sufficient to allow to use these new functions in **global scope** of ebuilds.
`ebuild.sh` sources `phase-helpers.sh` only when `[[ $EBUILD_PHASE != depend ]]`.
Probably this code should be added to `ebuild.sh` instead of `phase-helpers.sh`.

(For better performance of generation of metadata cache, it would be even better to inline these new functions in `ebuild.sh` so that a separate file does not need to be sourced...)

mgorny:
All fixed.

pym/portage/package/ebuild/doebuild.py
```diff
@@ -446,6 +447,11 @@ def doebuild_environment(myebuild, mydo, myroot=None, settings=None,
 	if eapi_exports_REPOSITORY(eapi) and "PORTAGE_REPO_NAME" in mysettings.configdict["pkg"]:
 		mysettings.configdict["pkg"]["REPOSITORY"] = mysettings.configdict["pkg"]["PORTAGE_REPO_NAME"]
 
+	if not eapi_exports_PORTDIR(eapi):
+		del mysettings["PORTDIR"]
+	if not eapi_exports_ECLASSDIR(eapi):
+		del mysettings["ECLASSDIR"]
```
zmedico: For filtering variables like this, I'd prefer that we use the `config.environ()` method. The `__delitem__` method is messy because it exhaustively deletes from all of the dictionaries in the config's lookuplist (separate dictionaries for make.conf, make.defaults, etc).

mgorny: Updated. Please disregard the first failure from travis, I have fixed that already.

PR #265, opened 2018-03-11
EAPI 7: D/ROOT trailing slash

mgorny:


bin/ebuild-helpers/prepstrip
```diff
@@ -143,7 +143,7 @@ save_elf_debug() {
 	local x=$1
 	local inode_debug=$2
 	local splitdebug=$3
-	local y=${ED}usr/lib/debug/${x:${#D}}.debug
+	local y=${ED%/}/usr/lib/debug/${x:${#D}}.debug
```
zmedico: We need to fix `${x:${#D}}` to account for `${D}` without a trailing slash, maybe use `local d=${D%/}` and `${#d}`.

bin/install-qa-check.d/80libraries
```diff
@@ -6,7 +6,7 @@ scanelf_lib_check() {
 	# Check for shared libraries lacking SONAMEs
 	local qa_var="QA_SONAME_${ARCH/-/_}"
 	eval "[[ -n \${!qa_var} ]] && QA_SONAME=(\"\${${qa_var}[@]}\")"
-	f=$(scanelf -ByF '%S %p' "${ED}"{,usr/}lib*/lib*.so* | awk '$2 == "" { print }' | sed -e "s:^[[:space:]]${ED}:/:")
+	f=$(scanelf -ByF '%S %p' "${ED%/}"/{,usr/}lib*/lib*.so* | awk '$2 == "" { print }' | sed -e "s:^[[:space:]]${ED%/}:/:")
```
zmedico: For the sed argument, it should be `${ED%/}/` in order to preserve behavior.

bin/install-qa-check.d/80libraries
```diff
@@ -40,7 +40,7 @@ scanelf_lib_check() {
 	# Check for shared libraries lacking NEEDED entries
 	qa_var="QA_DT_NEEDED_${ARCH/-/_}"
 	eval "[[ -n \${!qa_var} ]] && QA_DT_NEEDED=(\"\${${qa_var}[@]}\")"
-	f=$(scanelf -ByF '%n %p' "${ED}"{,usr/}lib*/lib*.so* | awk '$2 == "" { print }' | sed -e "s:^[[:space:]]${ED}:/:")
+	f=$(scanelf -ByF '%n %p' "${ED%/}"/{,usr/}lib*/lib*.so* | awk '$2 == "" { print }' | sed -e "s:^[[:space:]]${ED%/}:/:")
```
zmedico: `${ED%/}/` for sed argument again.

bin/ebuild-helpers/prepstrip
mgorny: Ai, my grep didn't catch `${#D` etc.

mgorny:
Done. I've named the var `d_noslash` to avoid confusion in the future.

PR #268, opened 2018-03-16
Support BDEPEND (CBUILD dependencies) in EAPI 7

mgorny:
Bug: https://bugs.gentoo.org/317337

CC @chewi


pym/portage/dbapi/bintree.py
```diff
@@ -330,6 +330,7 @@ def __init__(self, _unused=DeprecationWarning, pkgdir=None,
 				"USE_EXPAND", "USE_EXPAND_HIDDEN", "USE_EXPAND_IMPLICIT",
 				"USE_EXPAND_UNPREFIXED"])
 			self._pkgindex_default_pkg_data = {
+				"BDEPEND" : "",
```
Arfrever: Internal indentation is inconsistent with the following lines.

mgorny:
Thanks. I'm going to fix that alignment and push.

pym/portage/dbapi/bintree.py
mgorny: But it's consistent with the shorter entries following them.

PR #271, opened 2018-03-16
repoman: Update exception message

arkamar:
Closes: https://bugs.gentoo.org/574130

### DIFF
--- a/repoman/bin/repoman
+++ b/repoman/bin/repoman
@@ -44,7 +44,7 @@ try:
 	repoman_main(sys.argv[1:])
 except IOError as e:
 	if e.errno == errno.EACCES:
-		print("\nRepoman: Need user access")
+		print("\nRepoman: You don't have access permition to do this operation")
 		sys.exit(1)
 	else:
 		raise
--- a/repoman/pym/repoman/metadata.py
+++ b/repoman/pym/repoman/metadata.py
@@ -74,9 +74,13 @@ def fetch_metadata_xsd(metadata_xsd, repoman_settings):
 				return False
 
 		destdir = repoman_settings["DISTDIR"]
-		fd, metadata_xsd_tmp = tempfile.mkstemp(
-			prefix='metadata.xsd.', dir=destdir)
-		os.close(fd)
+		try:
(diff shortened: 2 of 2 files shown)


repoman/bin/repoman
```diff
@@ -44,7 +44,7 @@ try:
 	repoman_main(sys.argv[1:])
 except IOError as e:
 	if e.errno == errno.EACCES:
-		print("\nRepoman: Need user access")
+		print("\nRepoman: You don't have access permition to do this operation")
```
zmedico: This message is too vague, we really need to handle the error closer to where it occurs. We can simply raise the IOError here, since it's a bug if it's not handled closer to where it occurs.

arkamar:
Well, I tried to handle it closer to where it occurs, but I am not sure if it is a good solution. Btw, `fetch_metadata_xsd` returns `True` or `False` but the `get_metadata_xsd` function never checks it.

repoman/pym/repoman/metadata.py
```diff
@@ -74,9 +74,13 @@ def fetch_metadata_xsd(metadata_xsd, repoman_settings):
...
-		fd, metadata_xsd_tmp = tempfile.mkstemp(
-			prefix='metadata.xsd.', dir=destdir)
-		os.close(fd)
+		try:
+			fd, metadata_xsd_tmp = tempfile.mkstemp(
+				prefix='metadata.xsd.', dir=destdir)
+			os.close(fd)
+		except OSError:
+			print('Could not write file to %s' % destdir)
+			raise
```
zmedico: Better to raise `portage.exception.PermissionDenied(destdir)` here.

zmedico:
```python
except OSError as e:
    if e.errno == PermissionDenied.errno:
        raise portage.exception.PermissionDenied(destdir)
    else
        raise
```

repoman/bin/repoman
zmedico: We can simply handle `portage.exception.PermissionDenied` here.

zmedico: Allowing IOError to raise will be a good thing, since it will expose other areas that need attention.

PR #272, opened 2018-03-17
Extremely small while-to-dowhile-loop change 

hugbubby:
The loop checked the "unchecked" stack twice, which it doesn't need to. Thought that was a little cheeky so I implemmented a do-while loop using break.

### DIFF
--- a/pym/_emerge/depgraph.py
+++ b/pym/_emerge/depgraph.py
@@ -1539,9 +1539,11 @@ class depgraph(object):
 		explored_nodes = set()
 
 		while unexplored:
-			# Handle all unexplored packages.
-			while unexplored:
-				node = unexplored.pop()
+			while True:
+				try:
+				    node = unexplored.pop()
+				except IndexError:
+					break
 				for child in conflict_graph.child_nodes(node):
 					# Don't explore a node more than once, in order
 					# to avoid infinite recursion. The forced set


pym/_emerge/depgraph.py
```diff
@@ -1555,6 +1555,8 @@ def __str__(self):
 						unexplored.add(child)
 					else:
 						unexplored_tuples.add(child)
+                                if not unexplored:
+                                    break
```
zmedico: install app-vim/editorconfig-vim then the .editorconfig will configure you for tab indents

zmedico: Sometimes it's necessary to exit a while loop with a break, but in this case I think it makes the code less readable. The cost of checking the size of `unexplored` should be negligible.

hugbubby:
Made the changes and installed editorconfig.

Is it better to check a variable twice and end up using two fewer lines of code? Is it better to implemment a do-while loop at the slight expense of readability? The pressure is on zmedico. Time is ticking. Or not. The PR will kinda just stay here actually.

pym/_emerge/depgraph.py
```diff
@@ -1539,8 +1539,7 @@ def __str__(self):
 		explored_nodes = set()
 
 		while unexplored:
-			# Handle all unexplored packages.
-			while unexplored:
+			while True: # Do {see below}
 				node = unexplored.pop()
```
zmedico: Maybe more readable like this:
```
try:
    node = unexplored.pop()
except IndexError:
    break
```

hugbubby: https://pixel.nymag.com/imgs/daily/selectall/2017/02/13/wojak_01.nocrop.w710.h2147483647.jpg

zmedico:
Looks good now! Please squash this into a single commit.

hugbubby:
> Out of work experience: develop the gentoo package manager

zmedico:
Thanks! Merged in c3b98e6b66364a08d1dadd3df7d84ad5f133fb1c.

PR #274, opened 2018-03-18
INSTALL_MASK reform, next version

mgorny:


bin/phase-functions.sh
```diff
@@ -666,7 +666,8 @@ __dyn_install() {
 		ASFLAGS CBUILD CC CFLAGS CHOST CTARGET CXX \
 		CXXFLAGS EXTRA_ECONF EXTRA_EINSTALL EXTRA_MAKE \
 		LDFLAGS LIBCFLAGS LIBCXXFLAGS QA_CONFIGURE_OPTIONS \
-		QA_DESKTOP_FILE QA_PREBUILT PROVIDES_EXCLUDE REQUIRES_EXCLUDE ; do
+		QA_DESKTOP_FILE QA_PREBUILT PROVIDES_EXCLUDE REQUIRES_EXCLUDE \
+		PKG_INSTALL_MASK ; do
```
zmedico: Looks like we're not reading back this PKG_INSTALL_MASK value at uninstall time yet?

zmedico: Nevermind, PKG_INSTALL_MASK filters the file set, so we don't need to account for it at uninstall time.

pym/portage/dbapi/vartree.py
```diff
@@ -3851,6 +3881,17 @@ def eerror(lines):
...
+		try:
+			print('!!!\nRead INSTALL_MASK from %s\n!!!' % os.path.join(inforoot, "INSTALL_MASK"))
+			with io.open(_unicode_encode(
+						os.path.join(inforoot, "INSTALL_MASK"),
+						encoding=_encodings['fs'], errors='strict'),
+					mode='r', encoding=_encodings['repo.content'],
+					errors='replace') as f:
+				install_mask = f.read().split()
+		except EnvironmentError as e:
+			install_mask = self.settings.get('INSTALL_MASK', '').split()
```
zmedico: This INSTALL_MASK fallback means that a new INSTALL_MASK setting can create orphan files during unmerge!

zmedico: maybe create an placeholder build-info/INSTALL_MASK file so when know when the fallback is not needed, like we do for DEFINED_PHASES

bin/misc-functions.sh
```diff
@@ -383,12 +383,13 @@ preinst_mask() {
...
-	# remove share dir if unnessesary
-	if has nodoc $FEATURES || has noman $FEATURES || has noinfo $FEATURES; then
-		rmdir "${ED%/}/usr/share" &> /dev/null
-	fi
+	# Store the final value of INSTALL_MASK in build-info
+	local x
+	set -f
+	local IFS=$' \t\n\r'
+	x=$(echo ${INSTALL_MASK})
+	[[ -n $x ]] && echo "$x" > "${PORTAGE_BUILDDIR}"/build-info/INSTALL_MASK
```
zmedico: If INSTALL_MASK is empty, we can do like for DEFINED_PHASES here, and put `-` in the file. Then the unmerge code will know whether or not to fall back to `self.settings['INSTALL_MASK']` here (which we don't want to do if it could create orphans).

pym/portage/dbapi/vartree.py
mgorny: Done. Also removed that debug print above.

bin/misc-functions.sh
mgorny: I don't think there's a point in putting `-` when we can just print an empty string.

pym/portage/dbapi/vartree.py
```diff
@@ -2491,7 +2498,10 @@ def unlink(file_name, lstatobj):
 								(statobj.st_dev, statobj.st_ino),
 								[]).append(relative_path)
 
-					if is_owned:
+					# if the file was replaced by new version, keep it
+					# unless it is also INSTALL_MASK-ed in the new version
+					# (then it was not really installed and we want to clean it)
+					if is_owned and not self._is_install_masked(relative_path[1:], install_mask):
```
zmedico: Since contents contains files matches by INSTALL_MASK, we should also not unmerge files that were matched by INSTALL_MASK at install time for the package that is currently being unmerged.

mgorny: Wrong. Then you will be not remove files that were installed previously but now are masked.

zmedico: Wow this is confusing. Suppose that nothing is masked now. We don't want to unmerge files that were masked at install time for the package that is currently being unmerged, because those files don't really belong to this package.

zmedico: Maybe it's a big mistake to record masked files in CONTENTS. Why are we doing that? If we need that info, maybe a separate file from CONTENTS is more appropriate?

mgorny:
Updated to read old INSTALL_MASK as well. Now files that were masked before are not unmerged with `--- masked`

pym/portage/dbapi/vartree.py
```diff
@@ -3922,6 +3979,10 @@ def onerror(e):
 
 					relative_path = fpath[srcroot_len:]
 
+					# filter on INSTALL_MASK
+					if self._is_install_masked(relative_path, install_mask):
+						continue
```
zmedico: This filters the `_collision_protect` file list for the current package. However, it's asymmetric, because the following `_collision_protect` code is still going to report collisions if the current package installs a non-masked file that is install-masked by another package (isowner will return True even for install-masked files because they are still listed in CONTENTS):
```python
				for ver in mypkglist:
					if ver.isowner(f):
						isowned = True
						break```

Comment (mgorny) [reply pym/portage/dbapi/vartree.py]: Is that bad? We don't allow two packages to own the same file. We don't have `protect-owned` to ignore collisions just because some developer is using INSTALL_MASK. As far as I'm concerned, we can remove this filter altogether from collision checks.

Comment (zmedico) [reply pym/portage/dbapi/vartree.py]: It would be a change in behavior, which means a potential source of breakage for current users of INSTALL_MASK. In fact, at my workplace, it's very common to resolve file collisions via INSTALL_MASK. Dropping support for that could upset some people. If we don't plan to allow resolution of file collisions via INSTALL_MASK, we can advise users to implement an alternative to INSTALL_MASK via bashrc post_src_install hooks.

Comment (mgorny) [reply pym/portage/dbapi/vartree.py]: Or teach them to use features specifically designed for this purpose such as `COLLISION_IGNORE`.

Comment (mgorny) [reply pym/portage/dbapi/vartree.py]: Updated to cover this yet another awful obscure corner case of this idiotic package manager. Hope that's all and we can finally get this merged after 2 years of complete ignorance and then few days of torture on me.

Comment (zmedico) [reply pym/portage/dbapi/vartree.py]: COLLISION_IGNORE doesn't resolve collisions, it ignores them. In many cases it's preferable to prevent collisions from occurring, since collisions can have negative consequences.

Comment (zmedico) [reply pym/portage/dbapi/vartree.py]: Updated to cover this yet another awful obscure corner case of this idiotic package manager.

Collision protection is an extremely important feature, and any bugs that we introduce in collision protection could cause severe problems.

    Hope that's all and we can finally get this merged after 2 years of complete ignorance and then few days of torture on me.

I'm not trying to torture you, I just want to eliminate any bugs that I forsee.

File: pym/portage/dbapi/vartree.py [hunk 6]
```diff
@@ -3522,6 +3552,8 @@ def _collision_protect(self, srcroot, destroot, mypkglist,
 						if fnmatch.fnmatch(f_match, pattern):
 							stopmerge = False
 							break
+					if self._is_install_masked(relative_path, install_mask):
+						stopmerge = False
```
zmedico: This case will never trigger because treewalk has already filtered any files matched by install_mask. If install-masked files are included in CONTENTS, then the _collision_protect isowner calls can return incorrect results:
```python
				for ver in mypkglist:
					if ver.isowner(f):
						isowned = True
						break
```
Also, this code in treewalk can return incorrect results:
```python
					owners = self.vartree.dbapi._owners.get_owners(collisions)
```

mgorny:
You're giving me problems with no solution when you know that I can't figure out a solution myself.

zmedico:
A very important detail, which is mentioned nowhere in this pull request, is that a reason to include masked files in CONTENTS is so that we can detect collisions that would have occurred had people *not* been using INSTALL_MASK. This is a crucial detail, and I will now have to reevaluate everything from this perspective. It seems like will need a FEATURES setting for this, since people can use INSTALL_MASK to intentionally avoid collisions (as they do at my workplace).

mgorny:
I've revised this for 5 or 6 times already. I'm not interested in spending more time on it. If you want to support more corner cases, please implement them yourself.

zmedico:
Speaking of corner cases, your intention to detect collisions involving INSTALL_MASKed files was the source of most of the complexity, it was asymmetrically implemented and completely undocumented.

zmedico:
In https://github.com/gentoo/portage/pull/274, I have specifically excluded changes related to [bug 364633](https://bugs.gentoo.org/364633), since CONTENTS modifications add a significant amount of complexity which has nothing to do with exclusion support.

PR #275, opened 2018-03-18
Remaining EAPI 7 cross-compile bits boilerplate

mgorny:
This is just a boilerplate that satisfies EAPI 7 without actually changing much from the existing code. Cross people will probably want to work on it further later on, e.g. to bring full SYSROOT support in Portage.

### DIFF
--- a/pym/portage/eapi.py
+++ b/pym/portage/eapi.py
@@ -119,12 +119,16 @@ def eapi_path_variables_end_with_trailing_slash(eapi):
 	return eapi in ("0", "1", "2", "3", "4", "4-python", "4-slot-abi",
 			"5", "5-progress", "6")
 
+def eapi_has_broot(eapi):
+	return eapi not in ("0", "1", "2", "3", "4", "4-python", "4-slot-abi",
+			"5", "5-progress", "6")
+
 def eapi_has_sysroot(eapi):
 	return eapi not in ("0", "1", "2", "3", "4", "4-python", "4-slot-abi",
 			"5", "5-progress", "6")
 
 _eapi_attrs = collections.namedtuple('_eapi_attrs',
-	'bdepend dots_in_PN dots_in_use_flags exports_EBUILD_PHASE_FUNC '
+	'bdepend broot dots_in_PN dots_in_use_flags exports_EBUILD_PHASE_FUNC '
 	'exports_PORTDIR exports_ECLASSDIR '
 	'feature_flag_test feature_flag_targetroot '
 	'hdepend iuse_defaults iuse_effective posixish_locale '
@@ -153,6 +157,7 @@ def _get_eapi_attrs(eapi):
 
 	eapi_attrs = _eapi_attrs(
 		bdepend = (eapi is not None and eapi_has_bdepend(eapi)),
+		broot = (eapi is None or eapi_has_broot(eapi)),
 		dots_in_PN = (eapi is None or eapi_allows_dots_in_PN(eapi)),
 		dots_in_use_flags = (eapi is None or eapi_allows_dots_in_use_flags(eapi)),
 		empty_groups_always_true = (eapi is not None and eapi_empty_groups_always_true(eapi)),
--- a/pym/portage/eapi.py
+++ b/pym/portage/eapi.py
@@ -119,6 +119,10 @@ def eapi_path_variables_end_with_trailing_slash(eapi):
 	return eapi in ("0", "1", "2", "3", "4", "4-python", "4-slot-abi",
 			"5", "5-progress", "6")
 
+def eapi_has_sysroot(eapi):
+	return eapi not in ("0", "1", "2", "3", "4", "4-python", "4-slot-abi",
+			"5", "5-progress", "6")
+
 _eapi_attrs = collections.namedtuple('_eapi_attrs',
 	'bdepend dots_in_PN dots_in_use_flags exports_EBUILD_PHASE_FUNC '
 	'exports_PORTDIR exports_ECLASSDIR '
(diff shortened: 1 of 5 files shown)


pym/portage/package/ebuild/config.py
```diff
@@ -2736,6 +2736,8 @@ def environ(self):
 			mydict["ESYSROOT"] = (
 					mydict["SYSROOT"].rstrip(os.path.sep) + self.get("EPREFIX", "")
 					+ os.path.sep)
+		if eapi_attrs.broot and _phase_func_map.get(phase, '').startswith('src_'):
+			mydict["BROOT"] = self.get("EPREFIX", "") + os.path.sep
```
chewi: I believe disjoint prefix values should be possible now, look at how I implemented it in chewi/portage@6a2f06ec076c47e105292126636d626c63434172, specifically this line in LocationsManager.py. Was there anything wrong with this?

```python
self.broot = portage.const.EPREFIX
```

pym/portage/package/ebuild/config.py
```diff
@@ -2729,6 +2729,14 @@ def environ(self):
 		if not eapi_exports_merge_type(eapi):
 			mydict.pop("MERGE_TYPE", None)
 
+		if eapi_attrs.sysroot and _phase_func_map.get(phase, '').startswith('src_'):
+			# TODO: implement Portage-wide support for custom SYSROOT
+			mydict["SYSROOT"] = (
+					self.get('SYSROOT', '').rstrip(os.path.sep) + os.path.sep)
+			mydict["ESYSROOT"] = (
+					mydict["SYSROOT"].rstrip(os.path.sep) + self.get("EPREFIX", "")
+					+ os.path.sep)
```
zmedico: It feels awkward having a trailing slash here (since we have just removed if from ${D} and ${ED}). If we really need the trailing slash, if would we nice to have an explanation somewhere.

zmedico: Okay, good.

pym/portage/package/ebuild/config.py
zmedico: `self.get("EPREFIX", "")` will correspond to the cross-prefix, which is different from portage.const.EPREFIX when cross compiling.

zmedico: The difference is particularly visible in implementation of has_version and best_version in bin/phase-helpers.sh:
```bash
	if ___eapi_has_prefix_variables; then
		# [[ ${root} == / ]] would be ambiguous here,
		# since both prefixes can share root=/ while
		# having different EPREFIX offsets.
		if ${host_root} ; then
			eroot=${root%/}${PORTAGE_OVERRIDE_EPREFIX}/
		else
			eroot=${root%/}${EPREFIX}/
		fi
	else
```

pym/portage/package/ebuild/config.py
```diff
@@ -2729,6 +2729,16 @@ def environ(self):
...
+		if eapi_attrs.broot and _phase_func_map.get(phase, '').startswith('src_'):
+			mydict["BROOT"] = self.get("EPREFIX", "") + os.path.sep
```
zmedico: It seems like BROOT should begin with the value of `${ROOT}` here, since we join EPREFIX with ROOT for cross-prefix support.

pym/portage/package/ebuild/config.py
chewi: @zmedico, right but to be clear, are you agreeging that `portage.const.EPREFIX` is the correct value to use in this instance? Also is it better to setup this variable in config.py as @mgorny did or in LocationsManager.py as I did?

pym/portage/package/ebuild/config.py
chewi: No, BROOT is supposed to refer to /, not ROOT.

zmedico: Okay, then we should use portage.const.EPREFIX (actually better to refer to it indirectly as PORTAGE_OVERRIDE_EPREFIX) here.

zmedico: I'd like to eliminate `portage.const.EPREFIX`, see https://bugs.gentoo.org/643920, so I prefer indirect reference via `self['PORTAGE_OVERRIDE_EPREFIX']`.

pym/portage/package/ebuild/config.py
zmedico: Yes, `portage.const.EPREFIX` is correct, but I prefer indirect reference via `PORTAGE_OVERRIDE_EPREFIX` due to https://bugs.gentoo.org/643920. It does seem like a good idea to add a broot attribute to LocationsManager. It can refer to `portage.const.EPREFIX` for now, but ultimately I would like to pass it in via the the constructor in order to solve https://bugs.gentoo.org/643920.

pym/portage/eapi.py
```diff
@@ -119,15 +119,23 @@ def eapi_path_variables_end_with_trailing_slash(eapi):
 	return eapi in ("0", "1", "2", "3", "4", "4-python", "4-slot-abi",
 			"5", "5-progress", "6")
 
+def eapi_has_broot(eapi):
+	return eapi not in ("0", "1", "2", "3", "4", "4-python", "4-slot-abi",
+			"5", "5-progress", "6")
+
+def eapi_has_sysroot(eapi):
+	return eapi not in ("0", "1", "2", "3", "4", "4-python", "4-slot-abi",
+			"5", "5-progress", "6")
```
chewi: Since we're not killing off 5-hdepend, do we want BROOT and SYSROOT available to it? You haven't included them here.

pym/portage/package/ebuild/config.py
chewi: I did look at passing it via the constructor but `PORTAGE_OVERRIDE_EPREFIX` is currently set more than 300 lines after the constructor is used. It's in the same very long clause but I don't want to move it in case it breaks something. I'll leave it to you.

PR #276, opened 2018-03-19
Remaining EAPI 7 cross-compile bits boilerplate v2

chewi:
This is my version of #275. You may want to diff the two to see what's really changed. Admittedly it's barely tested because it's late and I'm tired. I tried to do `BROOT` in the way that was requested with `PORTAGE_OVERRIDE_EPREFIX` while keeping a `broot` variable in `LocationsManager`, currently defined with `portage.const.EPREFIX` until that can be fixed properly. @zmedico, I hope that's what you wanted.

### DIFF
--- a/bin/phase-helpers.sh
+++ b/bin/phase-helpers.sh
@@ -878,46 +878,55 @@ __eapi6_src_install() {
 	einstalldocs
 }
 
-# @FUNCTION: has_version
-# @USAGE: [--host-root] <DEPEND ATOM>
-# @DESCRIPTION:
-# Return true if given package is installed. Otherwise return false.
-# Callers may override the ROOT variable in order to match packages from an
-# alternative ROOT.
-has_version() {
-
-	local atom eroot host_root=false root=${ROOT}
-	if [[ $1 == --host-root ]] ; then
-		host_root=true
-		shift
-	fi
+___best_version_and_has_version_common() {
+	local atom root root_arg
+	case $1 in
+		--host-root|-r|-d|-b)
+			root_arg=$1
+			shift ;;
+	esac
 	atom=$1
 	shift
-	[ $# -gt 0 ] && die "${FUNCNAME[0]}: unused argument(s): $*"
+	[ $# -gt 0 ] && die "${FUNCNAME[1]}: unused argument(s): $*"
 
-	if ${host_root} ; then
-		if ! ___eapi_best_version_and_has_version_support_--host-root; then
-			die "${FUNCNAME[0]}: option --host-root is not supported with EAPI ${EAPI}"
-		fi
-		root=/
-	fi
+	case ${root_arg} in
+		"") if ___eapi_has_prefix_variables; then
+				root=${EROOT}
+			else
+				root=${ROOT}
+			fi ;;
+		--host-root)
+			if ! ___eapi_best_version_and_has_version_support_--host-root; then
+				die "${FUNCNAME[1]}: option ${root_arg} is not supported with EAPI ${EAPI}"
+			fi
+			if ___eapi_has_prefix_variables; then
+				root=/${PORTAGE_OVERRIDE_EPREFIX#/}
+			else
+				root=/
+			fi ;;
+		-r|-d|-b)
+			if ! ___eapi_best_version_and_has_version_support_-b_-d_-r; then
+				die "${FUNCNAME[1]}: option ${root_arg} is not supported with EAPI ${EAPI}"
+			fi
+			if ___eapi_has_prefix_variables; then
+				case ${root_arg} in
+					-r) root=${EROOT} ;;
+					-d) root=${ESYSROOT} ;;
+					-b) root=${BROOT:-/} ;;
+				esac
+			else
+				case ${root_arg} in
+					-r) root=${ROOT} ;;
+					-d) root=${SYSROOT} ;;
+					-b) root=/ ;;
+				esac
+			fi ;;
+	esac
 
-	if ___eapi_has_prefix_variables; then
-		# [[ ${root} == / ]] would be ambiguous here,
-		# since both prefixes can share root=/ while
-		# having different EPREFIX offsets.
-		if ${host_root} ; then
-			eroot=${root%/}${PORTAGE_OVERRIDE_EPREFIX}/
-		else
-			eroot=${root%/}${EPREFIX}/
-		fi
-	else
-		eroot=${root}
-	fi
 	if [[ -n $PORTAGE_IPC_DAEMON ]] ; then
-		"$PORTAGE_BIN_PATH"/ebuild-ipc has_version "${eroot}" "${atom}"
+		"${PORTAGE_BIN_PATH}"/ebuild-ipc "${FUNCNAME[1]}" "${root}" "${atom}"
 	else
-		"${PORTAGE_BIN_PATH}/ebuild-helpers/portageq" has_version "${eroot}" "${atom}"
+		"${PORTAGE_BIN_PATH}"/ebuild-helpers/portageq "${FUNCNAME[1]}" "${root}" "${atom}"
 	fi
 	local retval=$?
 	case "${retval}" in
@@ -925,75 +934,36 @@ has_version() {
 			return ${retval}
 			;;
 		2)
-			die "${FUNCNAME[0]}: invalid atom: ${atom}"
+			die "${FUNCNAME[1]}: invalid atom: ${atom}"
 			;;
 		*)
 			if [[ -n ${PORTAGE_IPC_DAEMON} ]]; then
-				die "${FUNCNAME[0]}: unexpected ebuild-ipc exit code: ${retval}"
+				die "${FUNCNAME[1]}: unexpected ebuild-ipc exit code: ${retval}"
 			else
-				die "${FUNCNAME[0]}: unexpected portageq exit code: ${retval}"
+				die "${FUNCNAME[1]}: unexpected portageq exit code: ${retval}"
 			fi
 			;;
 	esac
 }
 
+# @FUNCTION: has_version
+# @USAGE: [--host-root|-r|-d|-b] <DEPEND ATOM>
+# @DESCRIPTION:
+# Return true if given package is installed. Otherwise return false.
+# Callers may override the ROOT variable in order to match packages from an
+# alternative ROOT.
+has_version() {
+	___best_version_and_has_version_common "$@"
+}
+
 # @FUNCTION: best_version
-# @USAGE: [--host-root] <DEPEND ATOM>
+# @USAGE: [--host-root|-r|-d|-b] <DEPEND ATOM>
 # @DESCRIPTION:
 # Returns highest installed matching category/package-version (without .ebuild).
 # Callers may override the ROOT variable in order to match packages from an
 # alternative ROOT.
 best_version() {
-
-	local atom eroot host_root=false root=${ROOT}
-	if [[ $1 == --host-root ]] ; then
-		host_root=true
-		shift
-	fi
-	atom=$1
-	shift
-	[ $# -gt 0 ] && die "${FUNCNAME[0]}: unused argument(s): $*"
-
-	if ${host_root} ; then
-		if ! ___eapi_best_version_and_has_version_support_--host-root; then
-			die "${FUNCNAME[0]}: option --host-root is not supported with EAPI ${EAPI}"
-		fi
-		root=/
-	fi
-
-	if ___eapi_has_prefix_variables; then
-		# [[ ${root} == / ]] would be ambiguous here,
-		# since both prefixes can share root=/ while
-		# having different EPREFIX offsets.
-		if ${host_root} ; then
-			eroot=${root%/}${PORTAGE_OVERRIDE_EPREFIX}/
-		else
-			eroot=${root%/}${EPREFIX}/
-		fi
-	else
-		eroot=${root}
-	fi
-	if [[ -n $PORTAGE_IPC_DAEMON ]] ; then
-		"$PORTAGE_BIN_PATH"/ebuild-ipc best_version "${eroot}" "${atom}"
-	else
-		"${PORTAGE_BIN_PATH}/ebuild-helpers/portageq" best_version "${eroot}" "${atom}"
-	fi
-	local retval=$?
-	case "${retval}" in
-		0|1)
-			return ${retval}
-			;;
-		2)
-			die "${FUNCNAME[0]}: invalid atom: ${atom}"
-			;;
-		*)
-			if [[ -n ${PORTAGE_IPC_DAEMON} ]]; then
-				die "${FUNCNAME[0]}: unexpected ebuild-ipc exit code: ${retval}"
-			else
-				die "${FUNCNAME[0]}: unexpected portageq exit code: ${retval}"
-			fi
-			;;
-	esac
+	___best_version_and_has_version_common "$@"
 }
 
 if ___eapi_has_get_libdir; then
--- a/bin/phase-helpers.sh
+++ b/bin/phase-helpers.sh
@@ -640,7 +640,7 @@ econf() {
 		fi
 
 		local conf_args=()
-		if ___eapi_econf_passes_--disable-dependency-tracking || ___eapi_econf_passes_--disable-silent-rules || ___eapi_econf_passes_--docdir_and_--htmldir; then
+		if ___eapi_econf_passes_--disable-dependency-tracking || ___eapi_econf_passes_--disable-silent-rules || ___eapi_econf_passes_--docdir_and_--htmldir || ___eapi_econf_passes_--with-sysroot; then
 			local conf_help=$("${ECONF_SOURCE}/configure" --help 2>/dev/null)
 
 			if ___eapi_econf_passes_--disable-dependency-tracking; then
@@ -662,6 +662,12 @@ econf() {
 
 				if [[ ${conf_help} == *--htmldir* ]]; then
 					conf_args+=( --htmldir="${EPREFIX}"/usr/share/doc/${PF}/html )
+				fi
+			fi
+
+			if ___eapi_econf_passes_--with-sysroot; then
+				if [[ ${conf_help} == *--with-sysroot* ]]; then
+					conf_args+=( --with-sysroot="${ESYSROOT:-/}" )
 				fi
 			fi
 		fi
--- a/pym/_emerge/depgraph.py
+++ b/pym/_emerge/depgraph.py
@@ -3322,8 +3322,10 @@ class depgraph(object):
 		if removal_action:
 			depend_root = myroot
 		else:
-			if eapi_attrs.bdepend or eapi_attrs.hdepend:
+			if eapi_attrs.hdepend:
 				depend_root = myroot
+			elif eapi_attrs.bdepend:
+				depend_root = pkg.root_config.settings["ESYSROOT"]
 			else:
 				depend_root = self._frozen_config._running_root.root
 				root_deps = self._frozen_config.myopts.get("--root-deps")
--- a/pym/portage/__init__.py
+++ b/pym/portage/__init__.py
@@ -525,7 +525,7 @@ class _trees_dict(dict):
 		self._target_eroot = None
 
 def create_trees(config_root=None, target_root=None, trees=None, env=None,
-	eprefix=None):
+	sysroot=None, eprefix=None):
 
 	if trees is None:
 		trees = _trees_dict()
@@ -538,7 +538,7 @@ def create_trees(config_root=None, target_root=None, trees=None, env=None,
 		env = os.environ
 
 	settings = config(config_root=config_root, target_root=target_root,
-		env=env, eprefix=eprefix)
+		env=env, sysroot=sysroot, eprefix=eprefix)
 	settings.lock()
 
 	depcachedir = settings.get('PORTAGE_DEPCACHEDIR')
@@ -562,7 +562,7 @@ def create_trees(config_root=None, target_root=None, trees=None, env=None,
 		if depcachedir is not None:
 			clean_env['PORTAGE_DEPCACHEDIR'] = depcachedir
 		settings = config(config_root=None, target_root="/",
-			env=clean_env, eprefix=None)
+			env=clean_env, sysroot="/", eprefix=None)
 		settings.lock()
 		trees._running_eroot = settings['EROOT']
 		myroots.append((settings['EROOT'], settings))
--- a/pym/portage/package/ebuild/_config/LocationsManager.py
+++ b/pym/portage/package/ebuild/_config/LocationsManager.py
@@ -16,7 +16,7 @@ from portage import os, eapi_is_supported, _encodings, _unicode_encode
 from portage.const import CUSTOM_PROFILE_PATH, GLOBAL_CONFIG_PATH, \
 	PROFILE_PATH, USER_CONFIG_PATH
 from portage.eapi import eapi_allows_directories_on_profile_level_and_repository_level
-from portage.exception import DirectoryNotFound, ParseError
+from portage.exception import DirectoryNotFound, InvalidLocation, ParseError
 from portage.localization import _
 from portage.util import ensure_dirs, grabfile, \
 	normalize_path, read_corresponding_eapi_file, shlex_split, writemsg
@@ -40,12 +40,13 @@ _allow_parent_colon = frozenset(
 class LocationsManager(object):
 
 	def __init__(self, config_root=None, eprefix=None, config_profile_path=None, local_config=True, \
-		target_root=None):
+		target_root=None, sysroot=None):
 		self.user_profile_dir = None
 		self._local_repo_conf_path = None
 		self.eprefix = eprefix
 		self.config_root = config_root
 		self.target_root = target_root
+		self.sysroot = sysroot
 		self._user_config = local_config
 
 		if self.eprefix is None:
@@ -64,6 +65,13 @@ class LocationsManager(object):
 		self._check_var_directory("PORTAGE_CONFIGROOT", self.config_root)
 		self.abs_user_config = os.path.join(self.config_root, USER_CONFIG_PATH)
 		self.config_profile_path = config_profile_path
+
+		if self.sysroot is None:
+			self.sysroot = "/"
+		else:
+			self.sysroot = normalize_path(os.path.abspath(self.sysroot)).rstrip(os.sep) + os.sep
+
+		self.esysroot = self.sysroot.rstrip(os.sep) + self.eprefix + os.sep
 
 	def load_profiles(self, repositories, known_repository_paths):
 		known_repository_paths = set(os.path.realpath(x)
@@ -297,6 +305,13 @@ class LocationsManager(object):
 
 		self.target_root = normalize_path(os.path.abspath(
 			self.target_root)).rstrip(os.path.sep) + os.path.sep
+
+		if self.sysroot != "/" and self.sysroot != self.target_root:
+			writemsg(_("!!! Error: SYSROOT (currently %s) must "
+				"equal / or ROOT (currently %s).\n") %
+				(self.sysroot, self.target_root),
+				noiselevel=-1)
+			raise InvalidLocation(self.sysroot)
 
 		ensure_dirs(self.target_root)
 		self._check_var_directory("ROOT", self.target_root)
--- a/pym/portage/package/ebuild/_config/LocationsManager.py
+++ b/pym/portage/package/ebuild/_config/LocationsManager.py
@@ -73,6 +73,10 @@ class LocationsManager(object):
 
 		self.esysroot = self.sysroot.rstrip(os.sep) + self.eprefix + os.sep
 
+		# TODO: Set this via the constructor using
+		# PORTAGE_OVERRIDE_EPREFIX.
+		self.broot = portage.const.EPREFIX
+
 	def load_profiles(self, repositories, known_repository_paths):
 		known_repository_paths = set(os.path.realpath(x)
 			for x in known_repository_paths)
--- a/pym/portage/package/ebuild/config.py
+++ b/pym/portage/package/ebuild/config.py
@@ -175,7 +175,7 @@ class config(object):
 
 	def __init__(self, clone=None, mycpv=None, config_profile_path=None,
 		config_incrementals=None, config_root=None, target_root=None,
-		eprefix=None, local_config=True, env=None,
+		sysroot=None, eprefix=None, local_config=True, env=None,
 		_unmatched_removal=False, repositories=None):
 		"""
 		@param clone: If provided, init will use deepcopy to copy by value the instance.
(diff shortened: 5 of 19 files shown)


bin/phase-helpers.sh
```diff
@@ -843,122 +849,92 @@ __eapi6_src_install() {
...
+				root=/
+			fi ;;
+		-r|-d|-b)
+			if ! ___eapi_best_version_and_has_version_support_-b_-d_-r; then
+				die "${FUNCNAME[1]}: option ${root_arg} is not supported with EAPI ${EAPI}"
+			fi
+			if ___eapi_has_prefix_variables; then
+				case ${root_arg} in
+					-r) root=${EROOT} ;;
+					-d) root=${ESYSROOT} ;;
```
zmedico: How do we know that `${ESYSROOT}` will be recognized as a valid root inside `pym/portage/package/ebuild/_ipc/QueryCommand.py`, where this code might reject it:
```
		if root not in db:
			return ('', '%s: Invalid ROOT: %s\n' % (cmd, root), 3)
```

pym/portage/package/ebuild/_config/LocationsManager.py
```diff
@@ -302,6 +302,7 @@ def set_root_override(self, root_overwrite=None):
 		self._check_var_directory("ROOT", self.target_root)
 
 		self.eroot = self.target_root.rstrip(os.sep) + self.eprefix + os.sep
+		self.broot = portage.const.EPREFIX
```
mgorny: But you're not using this at all?

chewi: I'm not but @zmedico said it would be good to have it and we would later set it via the constructor.

mgorny: Maybe leave a TODO somewhere, so that people don't get too excited about using it.

Also, could you add something similar for sysroot in LocationsManager?

bin/phase-helpers.sh
chewi: At this point, it only makes sense for SYSROOT to equal / or ROOT. Until Portage supports a separate third location, setting SYSROOT to anything else is only going to break things anyway. If we accept this then the existing code should be sufficient. I could add explicitly add code to support a different location but I'm struggling to work out what that would look like.

zmedico: Ok, we should die up front if we detect an unsupported SYSROOT configuration. The `profile_check` function inside `pym/_emerge/main.py` would be a reasonable place to report the problem and exit unsuccessfully.

zmedico: I wonder if it will break people's existing SYSROOT usage if emerge dies early though, since I suspect that things like chromium os builds use a SYSROOT variable.

chewi: @mgorny was trying not to break existing systems so there is a conflict here. On the other hand, it was @vapier who really wanted to called it SYSROOT in the first place.

chewi: I tried to come up with some code anyway. How about this? I think `settings` is only defining `SYSROOT` in `environ` but it would need to be defined in the constructor for this to work?

```python
if settings['SYSROOT'] != '/' and settings['SYSROOT'] != settings['ROOT']:
	mysettings = config(config_root=config_root, target_root=settings['SYSROOT'],
		env=env, eprefix=eprefix)
	mysettings.lock()
	myroots.append((mysettings['EROOT'], mysettings))
```

zmedico: Well, there aren't any references to SYSROOT in the python code, so maybe adding support in create_trees is overkill. Do we intend for emerge to check install things into SYSROOT, or use packages from SYSROOT to satisify dependencies? If not, we could simply have QueryCommand dynamically instantiate a vardbapi instance in order to handle has_version and best_version calls that refer to SYSROOT.

mgorny: Yes, the final goal would be to install DEPEND into SYSROOT.

chewi: Wouldn't it be better to let these query commands just blow up if SYSROOT is set to some other location? I mean SYSROOT will only be checked if the -d option is given and this is new in EAPI 7 so it won't break any existing ebuilds. If some new ebuild does use it then the query couldn't possibly return any meaningful answer. We may never implement the additional location (it'd be nice but I don't plan to) and we'll need users to fix their SYSROOT value at some point if we don't want to be in this situation forever.

That's assuming anyone has even set it inappropriately. I'm not familiar with Chromium OS but I've had a look at their wiki. They use the --root-deps option, which implies that they have set SYSROOT to ROOT. They seem to emulate the additional "staging" location by building binary packages of whatever is installed to ROOT and then reinstalling those without build-time dependencies into a final "build_image".

pym/portage/package/ebuild/_config/LocationsManager.py
chewi: I would add esysroot but until I don't know where to get the value from as the environment variables used in config.py (i.e. `mydict["ESYSROOT"]`) aren't available in this context. It may be easier when this gets rejigged around the constructor.

chewi:
Added the TODO in LocationsManager.py.

I still need to fix `pym/_emerge/depgraph.py` because BDEPEND has been copied from HDEPEND when they're not exactly the same with regards to the --root-deps option as my man page change explains. I'll look at this tomorrow, shouldn't take long.

bin/phase-helpers.sh
zmedico: Yes, blow up for currently unsupported value is reasonable, but it would be nice if config normalized the value of SYSROOT in order to ensure that QueryCommand recognizes it. For example, the ROOT value is currently normalized in the LocationsManager set_root_override method:
```python
		self.target_root = normalize_path(os.path.abspath(
			self.target_root)).rstrip(os.path.sep) + os.path.sep
```

pym/_emerge/depgraph.py
```diff
@@ -3300,13 +3300,13 @@ def _add_pkg_deps(self, pkg, allow_unsatisfied=False):
...
-			if eapi_attrs.bdepend or eapi_attrs.hdepend:
+			if eapi_attrs.hdepend:
 				depend_root = myroot
 			else:
 				depend_root = self._frozen_config._running_root.root
 				root_deps = self._frozen_config.myopts.get("--root-deps")
 				if root_deps is not None:
-					if root_deps is True:
+					if root_deps is True or eapi_attrs.bdepend:
 						depend_root = myroot
```
zmedico: My hope was the the `--root-deps` option would become obsolete, since it was intended only as a workaround in the absence of native cross-compilation support in the EAPI. This is why `eapi_attrs.hdepend` behavior was independent of the `--root-deps` option. If `eapi_attrs.bdepend` needs to interact with the `--root-deps` option, then I'd like some kind of explanation.

chewi: I explained what the option would do against this EAPI in 0fc532a4524b48a62f2178fa7a2298312a4e81a4. As for why, we're allowing users to choose between building against headers/libraries in / and ROOT. Both are valid use cases although only the latter is an option for cross-compiling. 5-hdepend forgot to consider the bigger picture and this would have annoyed plenty of users.

I initially didn't think we'd use the --root-deps option for this and expected to just take the value of SYSROOT alone. Given that --root-deps already exists, this might be a confusing inconsistency. crossdev already uses --root-deps=rdeps by default, which is useful if we ignore the rdeps part for EAPI 7.

zmedico: I see the commit message mentions something about the `ROOT!=/ SYSROOT=/ use case`, so I'd need some clarification on why that introduces a need for `--root-deps`. My expectation was that we would be using ebuild variables to specify where dependencies are installed, rather than emerge options.

chewi: For the `ROOT!=/ SYSROOT=/`  use case, you wouldn't give --root-deps. The problem is that 5-hdepend effectively enabled --root-deps all the time and that implied SYSROOT=ROOT. There was no way to prevent that so 7 gives you back that control.

zmedico: Based on your explanation, I think that for eapi_attrs.bdepend, depend_root should be purely a function of the SYSROOT (and independent of the `--root-deps` option, as I have explained earlier).

chewi: I'm okay with that but if `SYSROOT` is set to some other location then I think we'll just have to bail out instead of trying to accommodate existing setups.

crossdev's `cross-emerge` currently sets `SYSROOT=${EPREFIX}/usr/${CHOST}`, which is the same as the default `ROOT` value but if you change `ROOT` to something else then `SYSROOT` will not change accordingly. Perhaps this could be fixed.

You probably have the best idea of what it would take to actually support a third location for `SYSROOT`. Is it as much work as I suspect it would be? It would share `PORTAGE_CONFIGROOT` with `ROOT` so it wouldn't be an entirely separate configuration.

zmedico: We can have backward-compatible SYSROOT pass-through as the default behavior, and add a `--sysroot=SYSROOT` emerge option that explicitly enables new behavior for EAPI 7.

zmedico: Let's document each of the configurations that can exist simultaneously. Prior to EAPI 7, there were a maximum of 2 configurations existing simultaneously, each having independent sets of variables:

running root: ROOT=/ + PORTAGE_CONFIGROOT=/
target root: ROOT=${FOO} + PORTAGE_CONFIGROOT=${BAR}

For EAPI 7, the configuration for the "running root" should not export SYSROOT to the ebuild environment, since its purpose is to build packages for `ROOT=/` in the classic manner. The configuration for the "target root" should export SYSROOT to the ebuild environment. For EAPI 7, we can introduce a third configuration with `ROOT=${SYSROOT}` and `PORTAGE_CONFIGROOT=${BAZ}` (maybe add a `--sysroot-configroot` option if necessary).

chewi: I hadn't even considered `ROOT=/ SYSROOT!=/` but I think you're right, that would be madness so let's avoid it.

On your last point, my brain is currently doing backflips so I can't work out if you're wrong, I'm wrong, or I've just misunderstood what you said. Last night's visit to the hospital at 3am probably isn't helping. I'll come back after more sleep.

zmedico: The gist of it is, we've got at least 2 separate configurations to deal with, each with mostly independent environment variables. We need to carefully define which variables are available to each configuration. For example, `pym/portage/__init__.py` we have a clean_env dictionary which contains variables that are are shared by all of the configurations. It may be that only one of the configurations really needs a SYSROOT variable where SYSROOT != ROOT, and another one might set `ROOT=${SYSROOT}`.

chewi: Sleep has helped. I think that even under EAPI 7, this boils down to 2 situations when `ROOT!=/`.

## SYSROOT=/
In this situation, the same PORTAGE_CONFIGROOT must be applied to BDEPEND (/), DEPEND (/) and RDEPEND (ROOT). Why? BDEPEND and DEPEND must have matching configurations if they're both installing into the same location. RDEPEND and DEPEND must also have matching configurations because you can't, for instance, build against OpenSSL, and then run against LibreSSL. You also have to consider existing packages on / that may conflict.

```
$ emerge -pv socat
[ebuild   R    ] net-misc/socat-1.7.3.2::gentoo  USE="ipv6 readline ssl -bindist -libressl -tcpd" 0 KiB

$ ROOT=/tmp/foo USE=libressl emerge -pv socat
[ebuild  N     ] app-misc/c_rehash-1.7-r1::gentoo to /tmp/foo/ 0 KiB
[ebuild  N     ] sys-apps/debianutils-4.8.4::gentoo to /tmp/foo/ USE="installkernel -static" 0 KiB
[ebuild  N     ] dev-libs/libressl-2.6.4:0/44::gentoo  USE="asm -static-libs" ABI_X86="32 (64) (-x32)" 3,153 KiB
[ebuild  N     ] dev-libs/libressl-2.6.4:0/44::gentoo to /tmp/foo/ USE="asm -static-libs" ABI_X86="32 (64) (-x32)" 0 KiB
[ebuild  N     ] app-misc/ca-certificates-20170717.3.36::gentoo to /tmp/foo/ USE="-cacert -insecure_certs" 22,486 KiB
[ebuild  N     ] sys-libs/ncurses-6.1-r2:0/6::gentoo to /tmp/foo/ USE="cxx gpm threads unicode -ada -debug -doc -minimal -profile -static-libs {-test} -tinfo -trace" ABI_X86="32 (64) (-x32)" 3,287 KiB
[ebuild  N     ] sys-libs/gpm-1.20.7-r2::gentoo to /tmp/foo/ USE="(-selinux) -static-libs" ABI_X86="32 (64) (-x32)" 0 KiB
[ebuild  N     ] sys-libs/readline-7.0_p3:0/7::gentoo to /tmp/foo/ USE="-static-libs -utils" ABI_X86="32 (64) (-x32)" 0 KiB
[ebuild  N     ] net-misc/socat-1.7.3.2::gentoo to /tmp/foo/ USE="ipv6 libressl readline ssl -bindist -tcpd" 0 KiB
[blocks B      ] dev-libs/openssl:0 ("dev-libs/openssl:0" is blocking dev-libs/libressl-2.6.4)
```

These are just some of the many things that could go wrong with differing configurations. You can already fall into this trap right now, it's not a new problem arising from EAPI 7.

## SYSROOT!=/
Here, the same PORTAGE_CONFIGROOT must be applied to DEPEND (SYSROOT) and RDEPEND (ROOT) for the same reason given above but BDEPEND (/) may be different. Because we only have one PORTAGE_CONFIGROOT variable, we assume the configuration for BDEPEND to be in /.

## Conclusions
Thinking about all this has made me realise a few things.

Your suggestion of `--sysroot-configroot` doesn't fit anywhere. In fact, the whole notion of a hypothetical third location for SYSROOT doesn't seem all that useful because with the introduction of BDEPEND, DEPEND will equal RDEPEND in most cases anyway. If we did support it, you certainly wouldn't want to build these packages twice. Perhaps Portage could automatically install binary packages created from SYSROOT into ROOT but what would be the point? Sure, it would avoid the odd static library or header from being put into ROOT but embedded system builders can trivially remove these, along with all the other headers that will inevitably end up in ROOT.

I noticed that PORTAGE_CONFIGROOT is effective even when `ROOT=/`. This is confusing because as soon as you set ROOT to be something else, it applies to that location instead of /. Sure, this is true for most environment variables like USE and CFLAGS but unlike those, you obviously can't set PORTAGE_CONFIGROOT in make.conf.

When I consider that, plus the restrictions on PORTAGE_CONFIGROOT that I have outlined above, I start to wonder whether it would be simpler to just do away with PORTAGE_CONFIGROOT entirely and use SYSROOT for this purpose instead. In all cases, the configuration needs to follow SYSROOT. The ability to put the configuration in some arbitrary location doesn't seem that useful and just complicated matters for both us and the users. You might not want /etc/portage in your final system but just like headers, this can trivially be removed.

What do you think?

zmedico: > RDEPEND (ROOT) and DEPEND (/)  must also have matching configurations because you can't, for instance, build against OpenSSL, and then run against LibreSSL.

In order to keep things simple, that's a fair assumption to make for now. However, they only need to be compatible, not necessarily identical. Our code should try to be flexible enough to allow some deviation in the configurations, though we don't necessarily have to expose this kind of flexibility to emerge users.

> Because we only have one PORTAGE_CONFIGROOT variable, we assume the configuration for BDEPEND to be in /.

This is also a user interface decision. It's fine to do it that way for now, but internally our code should try to be flexible enough to allow for deviation from that.

> I start to wonder whether it would be simpler to just do away with PORTAGE_CONFIGROOT entirely and use SYSROOT for this purpose instead.

That's fine, but in order to maintain backward compatibility I'd like to hide any breaking changes behind commandline options. For example, we can make usage of a new --sysroot=SYSROOT option be incompatible with --config-root=PORTAGE_CONFIGROOT.

chewi: Thanks, that helps a lot. I may still need a hand with the code as you're hard to please but I'll give it my best shot. :wink:

chewi:
Rebased and reworked the man page changes in line with what was discussed above. That should give some idea about where I'll be going with the code. Still aiming to finish this week.

chewi:
Just an update to say that I am still trying to put this together but while I'm happy with the concepts, I'm really struggling with the code. Haven't quite given up yet. No doubt you understand how bewildering this codebase is.

chewi:
It's starting to come together now... I think.

chewi:
Sorry, the past week has been heavy for me but I should be able to progress this week. At least all the tests should pass now! This might actually be all that's needed, just have to actually try it out and clean it up a bit.

chewi:
Slight tweak. It seems easier to have the underlying SYSROOT value end in a trailing slash as we need to compare against ROOT in some places.

I've done some preliminary testing around the flex and pam packages to see if BDEPEND and SYSROOT behave properly in terms of where Portage tries to install things. Looking good so far. More tomorrow.

chewi:
@zmedico, just thought I should mention that while I've added a separate SYSROOT configuration as you requested, its potential usefulness is immediately lost because we throw all the configurations into a dict using the paths for the keys. Given that SYSROOT will always be equal to either / or ROOT, one configuration will always be overwritten. It works anyway but if you want this to make sense, I think we should refer to these configurations using their variable names, i.e. `/`, `ROOT`, and `SYSROOT`.

bin/phase-helpers.sh
```diff
@@ -872,122 +878,92 @@ __eapi6_src_install() {
...
-# Callers may override the ROOT variable in order to match packages from an
-# alternative ROOT.
-has_version() {
-
-	local atom eroot host_root=false root=${ROOT}
-	if [[ $1 == --host-root ]] ; then
-		host_root=true
-		shift
-	fi
+__best_version_and_has_version_common() {
```
zmedico: It's best to use the tripple underscore namespace for this function, which is automatically filtered from the saved environment since 29f3837382babcac38d4c0e6498620b0dbc6df67.

pym/portage/__init__.py
```diff
@@ -561,11 +561,20 @@ def create_trees(config_root=None, target_root=None, trees=None, env=None,
...
+		myroots.append((bsettings['EROOT'], bsettings))
+
+		if settings['SYSROOT'] == '/':
+			dsettings = config(clone=bsettings)
+			dsettings.lock()
+			myroots.append((dsettings['EROOT'], dsettings))
+		elif settings['SYSROOT'] == settings['ROOT']:
+			dsettings = config(clone=settings)
+			dsettings.lock()
+			myroots.append((dsettings['EROOT'], dsettings))
```
zmedico: Thanks to this assertion, it seems to me that we only need 2 config instances (just like before SYSROOT):
```python
		if self.sysroot != "/" and self.sysroot != self.target_root:
			writemsg(_("!!! Error: SYSROOT (currently %s) must "
				"equal / or ROOT (currently %s).\n") %
				(self.sysroot, self.target_root),
				noiselevel=-1)
			raise InvalidLocation(self.sysroot)
```
Maybe this is all we need:
```diff
diff --git a/pym/portage/__init__.py b/pym/portage/__init__.py
index 137351a69..4f3159524 100644
--- a/pym/portage/__init__.py
+++ b/pym/portage/__init__.py
@@ -562,7 +562,7 @@ def create_trees(config_root=None, target_root=None, trees=None, env=None,
 		if depcachedir is not None:
 			clean_env['PORTAGE_DEPCACHEDIR'] = depcachedir
 		settings = config(config_root=None, target_root="/",
-			env=clean_env, eprefix=None)
+			env=clean_env, sysroot="/", eprefix=None)
 		settings.lock()
 		trees._running_eroot = settings['EROOT']
 		myroots.append((settings['EROOT'], settings))
```

pym/portage/package/ebuild/config.py
```diff
@@ -499,8 +505,10 @@ def __init__(self, clone=None, mycpv=None, config_profile_path=None,
 			#Loading Repositories
 			self["PORTAGE_CONFIGROOT"] = config_root
 			self["ROOT"] = target_root
+			self["SYSROOT"] = sysroot
 			self["EPREFIX"] = eprefix
 			self["EROOT"] = eroot
+			self["ESYSROOT"] = esysroot
```
zmedico: We need BROOT here too, right?

chewi: Yep. Must have missed that when I reworked things. While we're here, could you explain what the `backup_changes` stuff below is for? Do I need to add BROOT there too?

zmedico: Yes, BROOT needs the `backup_changes` thing too. All that `backup_changes`  does is store the variable in self.backupenv, which causes it to persist through the mutations that occur when we call the `setcpv` method to configure settings for a specific package.

pym/portage/package/ebuild/_config/LocationsManager.py
```diff
@@ -298,6 +310,13 @@ def set_root_override(self, root_overwrite=None):
 		self.target_root = normalize_path(os.path.abspath(
 			self.target_root)).rstrip(os.path.sep) + os.path.sep
 
+		if self.sysroot != "/" and self.sysroot != self.target_root:
```
zmedico: Even if this assertion break things for existing build systems (like chromium os), I think it's acceptable for now. If necessary, we can find some way to accommodate them in the future.

bin/phase-helpers.sh
hanetzer: Just thought I'd put in my two cents; I've been testing this locally with the `testpath` script, with
a modified autotools.eclass to use has_version -b in `autotools_env_setup()` and its blowing up
with `has_version: Invalid ROOT: ./`; unsure where the dotslash is coming from following it up
the chain.

chewi: Thanks for testing, I haven't got very far with mine yet but was about to continue now that I've cleaned things up a bit. I'll certainly look into that.

hanetzer: https://gist.github.com/hanetzer/ec9467127f895841c3cd8936288a380c
These are the modifications I've made to the gentoo tree I'm working against.
gentoo/gentoo@96d2e7109eb672f699c50fb2827a29ca1729bb76

chewi:
@hanetzer, fixed your issue. BROOT was possibly empty to start with so I ensured it has a trailing slash. On top of that, the new EAPI 7 feature to strip trailing slashes was making it empty again! We should obviously not strip slashes when the value is just `/`.

hanetzer:
Further investigation and experimentation has lead me to another interesting issue: that of SYSROOT
for cmake-utils.eclass; when called as:
```
PORTAGE_CONFIGROOT=${HOME}/mingw-w64 emerge --root=${HOME}/mingw-w64 --sysroot=${HOME}/mingw-w64 -1v =media-libs/openal-1.18.2-r1
```
with [these](https://gist.github.com/hanetzer/fd2cacf124d2bcb70c64418b057b27d2) changes to gentoo/gentoo@96d2e7109eb672f699c50fb2827a29ca1729bb76 results the following gentoo_toolchain.cmake
file:
```cmake
SET (CMAKE_ASM_COMPILER "x86_64-w64-mingw32-gcc")
SET (CMAKE_ASM-ATT_COMPILER "x86_64-w64-mingw32-gcc")
SET (CMAKE_C_COMPILER "x86_64-w64-mingw32-gcc")
SET (CMAKE_CXX_COMPILER "x86_64-w64-mingw32-g++")
SET (CMAKE_Fortran_COMPILER "x86_64-w64-mingw32-gfortran")
SET (CMAKE_AR /usr/bin/x86_64-w64-mingw32-ar CACHE FILEPATH "Archive manager" FORCE)
SET (CMAKE_RANLIB /usr/bin/x86_64-w64-mingw32-ranlib CACHE FILEPATH "Archive index generator" FORCE)
SET (CMAKE_SYSTEM_PROCESSOR "x86_64")
SET (CMAKE_RC_COMPILER x86_64-w64-mingw32-windres)
SET (CMAKE_SYSTEM_NAME "Windows")
```
which, as you can see, lacks the sysroot logic from cmake-utils.eclass:~570 and as such grabs
headers from BROOT or /, which causes build failures of course.

mgorny:
All those variables are empty when `/` by design and specification.

mgorny:
i.e. you do `${BROOT}/usr/bin/foo`

hanetzer:
@mgorny what I'm saying is, should not setting --root and --sysroot with emerge set the needed
stuff to activate the lines of cmake-utils.eclass around L570? because its not, so either I'm doing
it wrong or the code is (maybe by design for now).

mgorny:
> @mgorny what I'm saying is, should not setting --root and --sysroot with emerge set the needed
stuff to activate the lines of cmake-utils.eclass around L570? because its not, so either I'm doing
it wrong or the code is (maybe by design for now).

I don't see a problem with that code. FWICS, it accounts both for empty SYSROOT and equal to `/`.

hanetzer:
@mgorny yes, but as I've said, I'm explicitly using --sysroot=~/mingw-w64, which should be
setting SYSROOT=~/mingw-w64, and its not doing so, or something which makes the code
in cmake-utils.eclass not activate.

hanetzer:
Further, adding an `elog "SYSROOT=${SYSROOT}"` before the `if [ "${SYSROOT:-/}" != "/" ] ; then`
check in cmake-utils shows `SYSROOT=`, so something is going wrong down the pipeline.

chewi:
@mgorny, you're right, I'll fix this in the helpers instead.

@hanetzer, SYSROOT should be coming through there so I'll find out why it isn't. I don't know if CMake respects SYSROOT or requires an argument or entry in the toolchain file. We have handled autotools by passing `--with-sysroot` to `configure` in `econf`. In the past, I have added toolchain wrappers to the PATH to ensure that `--sysroot` is always added. I'm not proposing to add this to Portage but I'm not sure how easily we'll be able to cover the various build systems, including static Makefiles, either. Perhaps we could embed the `--sysroot` argument in CC like we do for multilib and `-m32`. Or we may just have to say that we only officially support the default SYSROOT given to the toolchain by crossdev (i.e. `/usr/${CHOST}`) and if you need anything else then you're on your own. It isn't just about the toolchain SYSROOT though, other things like libtool also need to know about it.

hanetzer:
@chewi cmake can handle it, and the checks [here](https://github.com/gentoo/gentoo/blob/cfa7dd0e726345dcce77e2cb197b0f5f92af0dee/eclass/cmake-utils.eclass#L553-L582) should inject some `SYSROOT` stuff into the
`gentoo_toolchain.cmake` file, which would accomplish the same thing as the `--with-sysroot` in
econf.

chewi:
@hanetzer, tried this myself and it works. Your problem is that you haven't actually told Portage that you're cross-compiling. Normally you'd use `mingw-w64-emerge` as set up by crossdev. You have set `PORTAGE_CONFIGROOT` so I guess that would set `CHOST` but you still need to set `CBUILD`.

Strictly speaking though, changing SYSROOT isn't exclusive to cross-compiling and we should modify cmake-utils.eclass to handle this even when not cross-compiling.

hanetzer:
@chewi the `make.conf` in `PORTAGE_CONFIGROOT` has the following contents:
```bash
CHOST=x86_64-w64-mingw32
CBUILD=x86_64-pc-linux-gnu
ARCH=amd64
LIBDIR_amd64="lib"

HOSTCC=x86_64-pc-linux-gnu-gcc

ROOT=/home/hanetzer/mingw-w64/

ACCEPT_KEYWORDS="amd64 ~amd64"

USE="${ARCH} -pam"

CFLAGS="-O2 -pipe -fomit-frame-pointer"
CXXFLAGS="${CFLAGS}"

FEATURES="-collision-protect sandbox buildpkg noman noinfo nodoc -news"
# Be sure we dont overwrite pkgs from another repo..
PKGDIR=${ROOT}packages/
PORTAGE_TMPDIR=${ROOT}tmp/
DISTDIR=/var/cache/portage/distfiles
ELIBC="mingw"

PKG_CONFIG_PATH="${ROOT}usr/lib/pkgconfig/"
#PORTDIR_OVERLAY="/usr/portage/local/"
EMERGE_LOG_DIR=${ROOT}/var/log

# vim: ft=sh
```

chewi:
Hmm, true enough, crossdev does put `CBUILD` in make.conf and it is enough for me to just set `PORTAGE_CONFIGROOT` and `SYSROOT`. I also tried with `--root` and `--sysroot` as you have done and it still works. I don't think the changes I have just made will help but try again anyway.

hanetzer:
@chewi no dice. the elog still shows SYSROOT as null, and I've even added an explicit SYSROOT to the make.conf in ~/mingw-w64

mgorny:
Ping. Is this 'good enough' or do we need further changes to start EAPI 7, at least for non-cross users?

chewi:
A bit more testing would be nice but issues can always be fixed later. I haven't managed to reproduce @hanetzer's issue. I would be testing right now but I'm currently away at a Gentoo meet.

chewi:
@hanetzer, I was able to reproduce your issue but only when changing openal's EAPI to 7. When leaving it at 6, I can still see a difference as the path written to the toolchain file has a trailing /, which isn't there if you use the latest Portage release. I'll try to find out what's wrong tomorrow.

zmedico:
Merged in 716f69482ffabf1a8e0e3496942a00ee5cdfa229 so this won't hold back EAPI 7 for non-cross users. Thanks!

bin/phase-helpers.sh
zmedico: > Just thought I'd put in my two cents; I've been testing this locally with the testpath script, with
a modified autotools.eclass to use has_version -b in autotools_env_setup() and its blowing up
with has_version: Invalid ROOT: ./; unsure where the dotslash is coming from following it up
the chain.

There's an `Invalid ROOT: ./` fix in 107dca9a3171e4870038769aef66612209ecd423.

zmedico:
There are fixes in 107dca9a3171e4870038769aef66612209ecd423 and a0ac6e6727abec8d2482c95b1e84d8df24d78619.

zmedico:
> @chewi no dice. the elog still shows SYSROOT as null, and I've even added an explicit SYSROOT to the make.conf in ~/mingw-w64

In #gentoo-portage @chewi has reported that this issue is caused by SYSYROOT being filtered by the `config.environ()` method because it's not in the `environ_whitelist`. He plans to send a PR to fix it.

zmedico:
>> @chewi no dice. the elog still shows SYSROOT as null, and I've even added an explicit SYSROOT to the make.conf in ~/mingw-w64

>In #gentoo-portage @chewi has reported that this issue is caused by SYSYROOT being filtered by the `config.environ()` method because it's not in the `environ_whitelist`.

Fixed in cd7364fbbb048564c2714d8a08df1ecbdc9cfe6e.

zmedico:
In b69a756099973abbf9719717ea3726519b32ce60 there's a fix involving normalization of an empty SYSROOT setting.

chewi:
This stuff sure is finicky.

zmedico:
Yes, it's very finicky. ;-) We have another fixup in 9abebd3e6aae049d2eb9c8039453d7f187bca00f.

PR #278, opened 2018-03-24
portageq repos_config: fix <eroot> parameter (bug 648062)

zmedico:
The `<eroot>` parameter is ineffective for commands that query
configuration, since the PORTAGE_CONFIGROOT variable controls
the location of configuration files. Therefore, for portageq
repos_config, implicitly set PORTAGE_CONFIGROOT equal to the
value of the `<eroot>` parameter. Note that this works correctly
for both prefix and non-prefix systems, because both EROOT and
PORTAGE_CONFIGROOT are supposed to include the EPREFIX offset.

Bug: https://bugs.gentoo.org/648062

### DIFF
--- a/bin/portageq
+++ b/bin/portageq
@@ -1,5 +1,5 @@
 #!/usr/bin/python -b
-# Copyright 1999-2016 Gentoo Foundation
+# Copyright 1999-2018 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 
 from __future__ import print_function, unicode_literals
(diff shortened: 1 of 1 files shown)


bin/portageq
```diff
@@ -1425,6 +1433,14 @@ def main(argv):
 
 		os.environ["ROOT"] = root
 
+		if getattr(function, "uses_configroot", False):
+			os.environ["PORTAGE_CONFIGROOT"] = eroot
+			# Disable RepoConfigLoader location validation, allowing raw
```
Arfrever: Maybe it would be better:
```
if eroot != "/":
	# Disable RepoConfigLoader location validation...
	portage._sync_mode = True
```

zmedico: I think it is potentially more valuable to return the "raw" data in any case. I think the main reason that RepoConfigLoader checks for existence is to help it portage cope with broken repos.conf, but portageq currently has no need for coping mechanisms here, and we can introduce that kind of thing later if the need arises.

PR #281, opened 2018-03-26
Add dostrip for EAPI 7

zmedico:
This patch includes the essential parts of the dostrip implementation
from portage-mgorny. All of the prepstrip code has moved to
bin/estrip, without any changes except the addition of argument
parsing for estrip --ignore, --queue, and --deque modes which are
equivalent to the corresponding ecompressdir modes.

Due to overlap, also ban the non-standard prepstrip and prepallstrip
helpers in EAPI 7, with a die message suggesting to use 'dostrip'
instead. Also ignore the non-standard STRIP_MASK variable for EAPI 7.

Bug: https://bugs.gentoo.org/203891


bin/ebuild-helpers/prepall
```diff
@@ -19,7 +19,7 @@ fi
 prepallman
```
mgorny: On a related note, you probably want to ban all of `prep*` in EAPI 7.

bin/estrip
```diff
@@ -0,0 +1,458 @@
...
+else
+	get_inode_number() { stat -c '%i' "$1"; }
+fi
+cd "${tmpdir}/inodes" || die "cd failed unexpectedly"
+while read -r x ; do
+	inode_link=$(get_inode_number "${x}") || die "stat failed unexpectedly"
+	echo "${x}" >> "${inode_link}" || die "echo failed unexpectedly"
+done < <(
+	# Use sort -u to eliminate duplicates for bug #445336.
+	(
```
mgorny: But now you're not using the results of `--queue` and instead stripping all files again…

zmedico: Thanks, updated to use this code conditionally:
```bash
while IFS= read -d '' -r x ; do
	rm -f "${x}" || die
	inode_link=$(get_inode_number "${x%.estrip}") || die "stat failed unexpectedly"
	echo "${x%.estrip}" >> "${inode_link}" || die "echo failed unexpectedly"
done < <(find "${ED}" -name '*.estrip' -print0)
```

bin/ebuild-helpers/prepall
zmedico: Sure, we can do that in a separate patch. I this patch, I'm only concerned with *strip commands due to overlap.

bin/estrip
```diff
@@ -0,0 +1,470 @@
...
+		banner=true
+	fi
+
+	(
+	__multijob_child_init
+	f=$(file "${x}") || exit 0
+	[[ -z ${f} ]] && exit 0
+
+	if ! ${prepstrip}; then
+		strip_this=true
```
zmedico: Updated to ignore STRIP_MASK here for EAPI 7.

zmedico: For clarity, I've split the changes into a separate commit from the prepstrip -> bin/estrip rename.

PR #282, opened 2018-03-26
Disallow package.provided in EAPI 7

zmedico:
This is also implemented in portage-mgorny by the following commit:

https://github.com/mgorny/portage/commit/386cdb131c99b01541d53c8c894b2ec6534b6dea

Bug: https://bugs.gentoo.org/568884

### DIFF
--- a/pym/portage/eapi.py
+++ b/pym/portage/eapi.py
@@ -104,6 +104,10 @@ def eapi_has_automatic_unpack_dependencies(eapi):
 def eapi_has_hdepend(eapi):
 	return eapi in ("5-hdepend",)
 
+def eapi_allows_package_provided(eapi):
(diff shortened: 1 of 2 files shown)


mgorny:
EAPI 7 says it's banned, not ignored silently.

zmedico:
> EAPI 7 says it's banned, not ignored silently.

Thanks, I've updated it to use the `!!! package.provided not allowed` message from https://github.com/mgorny/portage/commit/386cdb131c99b01541d53c8c894b2ec6534b6dea.

mgorny:
It looks like 80% my commit except without attribution.

zmedico:
> It looks like 80% my commit except without attribution.

Good point, I've now split it into 2 separate commits for purposes of attribution.

PR #290, opened 2018-04-06
repoman: Obey to GLEP66

mrueg:
As of https://www.gentoo.org/glep/glep-0066.html
"The summary line must not exceed 69 characters, and must not be
wrapped."


mgorny:
This would make it practically impossible to commit to some packages. I've added the exception to avoid insanity. However, if you are able to achieve shorter package names, I'm all for it.

mrueg:
I prefer to fix the spec to be adjusted to cat/packagename: + 50 instead of hard coding 69 chars? It doesn't make sense to write a spec and then ignore it.

PR #294, opened 2018-04-07
add antergos in the list of arch derivatives supported

kewlfft:
manjaro was already defined but antergos is missing in the list


zmedico:
Merged in c26bcc33c40e2259ae71e8f2666e29e087de420e. Thanks!

PR #300, opened 2018-04-13
rsync: default to sync-rsync-verify-jobs = 1 (bug 650696)

zmedico:
Some users have reported that using all processors to verify
manifests results in overloading, therefore default to using
a single processor. On modern hardware, verification of the
gentoo repository completes in less than 20 seconds, so using
multiple processors is not really necessary. Also, gemato-13.0
disables parallel verification due to the unresolved deadlock
issue reported in bug 647964, so this brings the default
portage configuration into alignment with current gemato
behavior.

Bug: https://bugs.gentoo.org/650696
Bug: https://bugs.gentoo.org/647964

### DIFF
--- a/cnf/repos.conf
+++ b/cnf/repos.conf
@@ -6,6 +6,7 @@ location = /usr/portage
 sync-type = rsync
 sync-uri = rsync://rsync.gentoo.org/gentoo-portage
 auto-sync = yes
+sync-rsync-verify-jobs = 1
 sync-rsync-verify-metamanifest = yes
 sync-rsync-verify-max-age = 24
 sync-openpgp-key-path = /var/lib/gentoo/gkeys/keyrings/gentoo/release/pubring.gpg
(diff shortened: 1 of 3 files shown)


mgorny:
There's no point in changing the option if it doesn't do anything.

zmedico:
It's a documentation update, so documentation reflects the current state of affairs. Also it makes the default behavior independent of the version of gemato that's installed.

mgorny:
But for users who want parallel gemato, downgrading gemato is the simplest solution. Also `-j1` is not equivalent to the current 'no multiprocessing at all' state.

zmedico:
> But for users who want parallel gemato, downgrading gemato is the simplest solution.

Well, downgrading packages is not the simplest way to configure things, and I really don't want to suggest that in the documentation.

 > Also `-j1` is not equivalent to the current 'no multiprocessing at all' state.

Maybe gemato should make it so?

PR #308, opened 2018-04-20
emerge --config: return pkg_config exit status (bug 653638)

zmedico:
Fix emerge --config exit status so that it will return 1 if
the pkg_config function calls die.

Bug: https://bugs.gentoo.org/653638


mrueg:
Looks good to me!

PR #313, opened 2018-04-24
ManifestScheduler: async fetchlist_dict (bug 653946)

zmedico:
In order to avoid event loop recursion, pass fetchlist_dict to
ManifestTask as a Future.

Bug: https://bugs.gentoo.org/653946

### DIFF
--- a/pym/portage/package/ebuild/_parallel_manifest/ManifestScheduler.py
+++ b/pym/portage/package/ebuild/_parallel_manifest/ManifestScheduler.py
@@ -1,10 +1,10 @@
-# Copyright 2012-2013 Gentoo Foundation
+# Copyright 2012-2018 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 
 import portage
 from portage import os
+from portage.dbapi.porttree import _async_manifest_fetchlist
 from portage.dep import _repo_separator
-from portage.exception import InvalidDependString
 from portage.localization import _
 from portage.util._async.AsyncScheduler import AsyncScheduler
 from .ManifestTask import ManifestTask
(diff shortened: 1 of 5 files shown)


pym/portage/package/ebuild/_parallel_manifest/ManifestScheduler.py
```diff
@@ -91,3 +81,55 @@ def _task_exit(self, task):
...
+	@type cpv_list: list
+	@return: a Future resulting in a Mapping compatible with FetchlistDict
+	@rtype: asyncio.Future (or compatible)
+	"""
+	loop = getattr(loop, '_asyncio_wrapper', loop)
+	result = loop.create_future()
+	futures = {}
+	for cpv in cpv_list:
+		futures[cpv] = portdb.async_fetch_map(
+			cpv, mytree=repo_config.location, loop=loop)
```
zmedico: We need so use something like `async_iter_completed(max_jobs=1)` to limit the number of concurrent async_fetch_map calls here, otherwise the only limit is the number of ebuilds in the directory.

zmedico: Fixed to use `async_iter_completed(max_jobs=1)`.

pym/portage/package/ebuild/_parallel_manifest/ManifestScheduler.py
```diff
@@ -91,3 +84,90 @@ def _task_exit(self, task):
 					noiselevel=-1)
 
 		AsyncScheduler._task_exit(self, task)
+
+
+def _future_fetchlist(loop, portdb, repo_config, cp, cpv_list,
+	max_jobs=None, max_load=None):
+	"""
+	Asynchronous form of FetchlistDict, with max_jobs and max_load
+	parameters in order to control async_aux_get concurrency.
```
zmedico: I'd like to split out a utility function that's like [asyncio.gather](https://docs.python.org/3/library/asyncio-task.html#asyncio.gather) with `max_jobs` and `max_load` parameters.

zmedico: I've split out an `iter_gather` function.

PR #317, opened 2018-05-03
config.environ:	always strip slash from	SYSROOT	(bug 654600)

zmedico:
Since SYSROOT=/ interacts badly with autotools.eclass (bug 654600),
and no EAPI expects SYSROOT to have a trailing slash, always strip
the trailing slash from SYSROOT.

Bug: https://bugs.gentoo.org/654600
Fixes: a41dacf7926c ("Export SYSROOT and ESYSROOT in ebuild env in EAPI 7")

@chewi

### DIFF
--- a/pym/portage/package/ebuild/config.py
+++ b/pym/portage/package/ebuild/config.py
@@ -2813,10 +2813,15 @@ class config(object):
 			mydict.pop("ECLASSDIR", None)
 
 		if not eapi_attrs.path_variables_end_with_trailing_slash:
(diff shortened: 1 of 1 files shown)


chewi:
Fair enough. @mgorny might want to add his 2¢ here as he added the stripping.

mgorny:
Well, I just wanted to have the conditions consistent. Given that there are no EAPIs that define SYSROOT and have trailing slash, it is not strictly necessary to do it conditionally.

PR #318, opened 2018-05-03
config: Whitelist SYSROOT and ESYSROOT

chewi:
Forgot to add these earlier.


chewi:
Looks like you committed the change yourself before you saw this. Oh well!

zmedico:
Whoops, yeah I was just about ready to commit when I went off to lunch, then came back from lunch and missed this PR. Thanks anyway!

PR #326, opened 2018-06-20
repoman: add a check for unsorted KEYWORDS

sbraz:
I thought @SoapGentoo would enjoy that kind of check :)

### DIFF
--- a/repoman/pym/repoman/modules/scan/keywords/keywords.py
+++ b/repoman/pym/repoman/modules/scan/keywords/keywords.py
@@ -55,6 +55,8 @@ class KeywordChecks(ScanBase):
 
 		self._checkForMaskLikeKeywords(xpkg, y_ebuild, ebuild.keywords)
 
+		self._checkForUnsortedKeywords(ebuild, xpkg, y_ebuild)
+
 		self.slot_keywords[pkg.slot].update(ebuild.archs)
 		return False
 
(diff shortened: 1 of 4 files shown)


dol-sen:
hmm, this is good, but I am wondering if repoman shouldn't just auto-sort the entries like it does the copyright update.   Make it a warn that it found unsorted keywords (and fixed).

repoman/pym/repoman/modules/scan/keywords/keywords.py
```diff
@@ -133,6 +135,16 @@ def _checkForMaskLikeKeywords(self, xpkg, y_ebuild, keywords):
 				self.qatracker.add_error("KEYWORDS.stupid",
 					"%s/%s.ebuild" % (xpkg, y_ebuild))
 
+	def _checkForUnsortedKeywords(self, ebuild, xpkg, y_ebuild):
+		'''Ebuilds that contain KEYWORDS
+		which are not sorted alphabetically.'''
+		sorted_keywords = sorted(ebuild.keywords,
+			key=lambda x: x.lstrip("-~"))
```
Arfrever: This is broken implementation of sorting.

`ekeyword` tool uses https://gitweb.gentoo.org/proj/gentoolkit.git/tree/pym/gentoolkit/ekeyword/ekeyword.py?id=bb585671418fa51682a7e1dbfef5914e26261adf#n82 which places `*-*` keywords after other keywords, so e.g. `~amd64 ~x86 ~amd64-fbsd ~x86-fbsd ~amd64-linux ~x86-linux` is properly sorted value.

SoapGentoo: Then send in a fix instead of just peanut gallery'ing

sbraz:
@Arfrever fixed. This will behave like ekeyword.

sbraz:
@Arfrever re-pushed and this time it matches the behaviour of ekeyword, can you confirm?

Arfrever:
It seems to work.

zmedico:
Merged in f4faf6d7950150a3068f92a3f3615b884db897d6, thanks!

PR #329, opened 2018-06-25
portage.data: add build user (bug 600804)

zmedico:
Add attributes and override variables for a separate build
user, with defaults corresponding to the status quo. This
reuses the existing PORTAGE_BUILD_USER and PORTAGE_BUILD_GROUP
variables that were already exported to the ebuild environment.

When PORTAGE_BUILD_USER and PORTAGE_BUILD_GROUP are unset,
the build user can still be overridden by the existing
PORTAGE_USERNAME and PORTAGE_GRPNAME variables.

This sets the stage for migration of internal code to use
a separate build user, as required to solve bug 600804 in a
manner that complies with PMS.

Bug: https://bugs.gentoo.org/600804

### DIFF
--- a/lib/portage/data.py
+++ b/lib/portage/data.py
@@ -125,6 +125,15 @@ try:
 except KeyError:
     pass
 
+_attr_env_vars = {
+    # TODO: Migrate all internals to use the build user/group
+    # instead of the "portage" user/group where appropriate.
+    "_build_group": "PORTAGE_BUILD_GROUP",
+    "_build_user": "PORTAGE_BUILD_USER",
+    "_portage_grpname": "PORTAGE_GRPNAME",
+    "_portage_username": "PORTAGE_USERNAME",
+}
+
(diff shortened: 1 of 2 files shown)


pym/portage/data.py
```diff
@@ -142,6 +151,32 @@ def _get_global(k):
 		elif _get_global('portage_gid') in os.getgroups():
 			v = 1
 
+	elif k in ('_build_gid', '_build_uid'):
+		keyerror = False
+		try:
+			build_uid = pwd.getpwnam(_get_global('_build_user')).pw_uid
+		except KeyError:
+			writemsg('!!! PORTAGE_BUILD_USER refers to non-existant user "%s"\n'
```
Arfrever: s/non-exist**a**nt/nonexist**e**nt/
(And 7 lines below)

zmedico: Thanks, fixed.

PR #332, opened 2018-07-05
GitSync: abort checkout for signature problem (bug 660372)

zmedico:
Fetch the upstream remote and use git merge to update the checkout
only after successful verification of the upstream head.

Suggested-by: Richard Freeman <rich0@gentoo.org>
Bug: https://bugs.gentoo.org/660372

### DIFF
--- a/pym/portage/sync/modules/git/git.py
+++ b/pym/portage/sync/modules/git/git.py
@@ -109,6 +109,7 @@ class GitSync(NewBase):
 		if not self.has_bin:
 			return (1, False)
 		git_cmd_opts = ""
+		quiet = self.settings.get("PORTAGE_QUIET") == "1"
 		if self.repo.module_specific_options.get('sync-git-env'):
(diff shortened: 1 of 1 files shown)


pym/portage/sync/modules/git/git.py
```diff
@@ -133,20 +148,29 @@ def update(self):
...
+				return (1, False)
+
+			merge_cmd = [self.bin_command, 'merge', 'refs/remotes/%s' % remote_branch]
+			if quiet:
+				merge_cmd.append('--quiet')
+			exitcode = subprocess.call(merge_cmd,
+				cwd=portage._unicode_encode(self.repo.location))
+
 		if exitcode != os.EX_OK:
 			msg = "!!! git pull error in %s" % self.repo.location
```
Arfrever: Maybe this should be changed to `msg = "!!! git fetch error ...`

zmedico: Ok, updated to show separate `fetch` and `merge` error messages.

PR #345, opened 2018-07-25
install-qa-check.d/60pngfix: parallel support (bug 630292)

zmedico:
If xargs supports the --max-procs option then use the makeopts_jobs
function from helper-functions.sh to generate a --max-procs argument.
Use xargs `-L 1` to limit the number of png files per pngfix process,
in order to ensure that enough processes are spawned, since otherwise
xargs minimizes the number of processes spawned. A benchmark with
flightgear-data-2018.2.1 shows that larger values of -L only decrease
performance.

Bug: https://bugs.gentoo.org/630292
Reviewed-by: Lars Wendler <polynomial-c@gentoo.org>
Reviewed-by: Mike Gilbert <floppym@gentoo.org>

### DIFF
--- a/bin/install-qa-check.d/60pngfix
+++ b/bin/install-qa-check.d/60pngfix
@@ -1,7 +1,17 @@
 # Check for issues with PNG files
 
+source "${PORTAGE_BIN_PATH}/helper-functions.sh" || exit 1
+
 pngfix_check() {
-	local pngfix=$(type -P pngfix)
+	local chunksize=1 jobs pngfix=$(type -P pngfix) xargs=(${XARGS})
+
+	if "${xargs[@]}" --help | grep -q -- --max-procs=; then
+		jobs=$(makeopts_jobs)
+		if [[ ${jobs} -gt 1 ]]; then
+			xargs+=("--max-procs=${jobs}" -L "${chunksize}")
+		fi
+	fi
+
 	if [[ -n ${pngfix} ]] ; then
 		local pngout=()
 		local next
@@ -25,7 +35,7 @@ pngfix_check() {
 				fi
 				eqawarn "   ${pngout[@]:7}: ${error}"
 			fi
-		done < <(find "${ED}" -type f -name '*.png' -exec "${pngfix}" {} +)
+		done < <(find "${ED}" -type f -name '*.png' -print0 | "${xargs[@]}" -0 "${pngfix}")
 	fi
 }
 


Polynomial-C:
Thank you very much Zac.

While this patch improves situation quite a bit, I think there is still room for improvement.

But first some numbers.
My chosen reference package without your patch:
```
     Mon Jun  4 08:36:17 2018 >>> games-simulation/flightgear-data-2018.2.1
       merge time: 20 minutes and 55 seconds.
```

And the very same package with your patch:
```
     Thu Jul 26 11:07:54 2018 >>> games-simulation/flightgear-data-2018.2.1
       merge time: 9 minutes and 52 seconds.
```

Please only look at the time delta. I have `buildpkg` in FEATURES and my dev machine usually has the `ondemand` CPU frequency gouvernor in charge which means CPU frequency is usually at lowest when being in idle and only gets raised once there's load on the CPU cores. So I could improve these merge times quite a bit but I decided to go with my usual setup.

So build/merge time was dropped by about 50% which is really good in general. But since this package does exactly zero compiling, I think it is still not optimal.

While I was watching the merge, portage spawned xargs like this:

  `xargs -r --max-procs=16 -0 /usr/bin/pngfix`

Which is the correct number of CPUs on my dev machine. But having htop running, the highest number of concurrent pngfix processes I saw in htop were five which then dropped from three to only one remaining pngfix process for at least half of the merge time.

Looking at the xargs man page checking for the `--max-procs` option, I found this:
```
  Use the -n option or the -L option with -P; otherwise
  chances are that only one exec will be  done.
```

So I change the following line in your patch
```
+			xargs+=(--max-procs=${jobs})
```
to
```
+			xargs+=(--max-procs=${jobs} -L ${jobs})
```
and this was the result:
```
     Thu Jul 26 11:27:19 2018 >>> games-simulation/flightgear-data-2018.2.1
       merge time: 5 minutes and 26 seconds.
```
which is a 75% decrease in time consumption only for the pngfix part. And htop showed 16 pngfix processes for the most time.
What do you think?

floppym:
Ideally, you would want to count the number of total files, divide that by the number of cores, and pass that as `--max-args`.

In practice, I imagine that setting `--max-args` to some reasonably low number would be fine. The lower the number, the better the parallelism, but that needs to be balanced against the overhead of calling fork/exec multiple times.

zmedico:
I did some benchmarks with flightgear-data-2018.2.1, and `-L 1` was optimal, larger values only decreased performance. Updated to use `-L 1`.

floppym:
Ah, I failed to account for the fact that some png files take longer to process than others, so you could end up with lopsided buckets if they are split into larger groups. I guess the syscall overhead is not significant enough to outweigh that.

PR #347, opened 2018-07-30
Replace implicit {FEATURES->USE}=test forcing with USE default

mgorny:
Use an explicit USE_ORDER entry to control mapping FEATURES=test into
default-enabled USE=test, rather than forcing/masking it depending
on the state of FEATURES.

This makes it possible for users to enable (or disable) USE=test
independently of FEATURES.  An example use case is installing test
dependencies and building test cases without actually running tests
at a particular moment which is something I've been doing quite
frequently with LLVM.

### DIFF
--- a/lib/portage/package/ebuild/config.py
+++ b/lib/portage/package/ebuild/config.py
@@ -294,6 +294,7 @@ class config(object):
 			self.configlist = [
 				self.configdict['env.d'],
 				self.configdict['repo'],
+				self.configdict['features'],
 				self.configdict['pkginternal'],
 				self.configdict['globals'],
 				self.configdict['defaults'],
@@ -461,12 +462,15 @@ class config(object):
 			# back up our incremental variables:
 			self.configdict={}
 			self._use_expand_dict = {}
-			# configlist will contain: [ env.d, globals, defaults, conf, pkg, backupenv, env ]
+			# configlist will contain: [ env.d, globals, features, defaults, conf, pkg, backupenv, env ]
 			self.configlist.append({})
 			self.configdict["env.d"] = self.configlist[-1]
 
 			self.configlist.append({})
 			self.configdict["repo"] = self.configlist[-1]
+
+			self.configlist.append({})
+			self.configdict["features"] = self.configlist[-1]
 
 			self.configlist.append({})
 			self.configdict["pkginternal"] = self.configlist[-1]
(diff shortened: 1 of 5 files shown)


lib/portage/package/ebuild/config.py
```diff
@@ -1677,11 +1700,8 @@ def setcpv(self, mycpv, use_cache=None, mydb=None):
 		# package has different IUSE.
 		use = set(self["USE"].split())
 		unfiltered_use = frozenset(use)
-		if explicit_iuse is None:
```
mgorny: Note: this is not removed, it's moved higher up.

lib/portage/package/ebuild/config.py
```diff
@@ -1650,6 +1656,23 @@ def setcpv(self, mycpv, use_cache=None, mydb=None):
...
+		if explicit_iuse is None:
+			explicit_iuse = frozenset(x.lstrip("+-") for x in iuse.split())
+		if eapi_attrs.iuse_effective:
+			iuse_implicit_match = self._iuse_effective_match
+		else:
+			iuse_implicit_match = self._iuse_implicit_match
+
+		if "test" in explicit_iuse or iuse_implicit_match("test"):
+			if "test" in self.features:
+				feature_use += "test"
```
zmedico: I know this works fine but the string append without space here *looks* like a bug, so it would be nice if we could change that. Maybe something like this:
```
feature_use += (" " if feature_use else "") + "test"
```

mgorny:
W dniu pon, 30.07.2018 o godzinie 14∶41 -0700, użytkownik Zac Medico
napisał:
> zmedico commented on this pull request.
> 
> 
> 
> > @@ -1650,6 +1656,23 @@ def setcpv(self, mycpv, use_cache=None, mydb=None):
> 
>  		if has_changed:
>  			self.reset(keeping_pkg=1)
>  
> +		if explicit_iuse is None:
> +			explicit_iuse = frozenset(x.lstrip("+-") for x in iuse.split())
> +		if eapi_attrs.iuse_effective:
> +			iuse_implicit_match = self._iuse_effective_match
> +		else:
> +			iuse_implicit_match = self._iuse_implicit_match
> +
> +		if "test" in explicit_iuse or iuse_implicit_match("test"):
> +			if "test" in self.features:
> +				feature_use += "test"
> 
> I know this works fine but the string append without space here *looks* like a bug, so it would be nice if we could change that. Maybe something like this:
> ```
> feature_use += (" " if feature_use else "") + "test"
> ```
> 

I was wondering how to handle that sanely.  Maybe instead I should make
it a list first, then join before comparing?

-- 
Best regards,
Michał Górny

zmedico:
On Mon, Jul 30, 2018 at 10:27 PM Michał Górny <notifications@github.com>
wrote:

> I was wondering how to handle that sanely. Maybe instead I should make
> it a list first, then join before comparing?
>

Yeah, sounds good.
-- 
Thanks,
Zac

mgorny:
Thanks!

PR #348, opened 2018-08-01
Support tox

mgorny:
tox is the common Pythonic tool for running tests against multiple interpreters.  It integrates well with virtualenv, making testing trivial on practically any system.  Add a tox.ini file so users can take advantage of it, and make travis use it.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,16 @@
+[tox]
+envlist = py27,py34,py35,py36,pypy,pypy3
+skipsdist = True
+
+[testenv]
+deps =
+	lxml!=4.2.0
+	pygost
+	pyyaml
+	py27,py34,py35,pypy: pyblake2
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -0,0 +1,16 @@
...
+	lxml!=4.2.0
+	pygost
+	pyyaml
+	py27,py34,py35,pypy: pyblake2
+	py27,py34,py35,pypy: pysha3
+setenv =
+	PYTHONPATH={toxinidir}/lib
+commands =
+	python setup.py test
+	python repoman/setup.py test
```
zmedico: How about if we use `python -b` to enable the BytesWarning when comparing str and bytes, like .travis.yml did before?

mgorny: Done. Also added `-Wd`.

PR #349, opened 2018-08-01
adding arch32 os release id

kewlfft:
adding arch32 to the list for the 32-bit users

### DIFF
--- a/bin/etc-update
+++ b/bin/etc-update
@@ -37,7 +37,7 @@ OS_RELEASE_ID=$(cat /etc/os-release 2>/dev/null | grep '^ID=' | cut -d'=' -f2 | 
 case $OS_RELEASE_ID in
 	suse|opensuse|opensuse-leap|opensuse-tumbleweed) OS_FAMILY='rpm' ;;
 	fedora|rhel) OS_FAMILY='rpm' ;;
-	arch|archarm|manjaro|antergos) OS_FAMILY='arch' NEW_EXT='pacnew';;
+	arch|archarm|arch32|manjaro|antergos) OS_FAMILY='arch' NEW_EXT='pacnew';;
 	*) OS_FAMILY='gentoo' ;;
 esac
 


zmedico:
Merged in 50232e5d8c153e5500c97be29d6347f960e01760, thanks!

eli-schwartz:
IMHO the correct solution here is to gain support for `os-release(5)` "ID_LIKE", which should be "arch" for any Arch Linux derivative and means you only need to check for actual Arch Linux, Manjaro (which doesn't conform), and anything else which declares itself "like arch".

BTW the os-release spec states that a shell-compatible parser for key=val assignment should be used, which needs to support either single or double quotes as well as "Shell special characters ("$", quotes, backslash, backtick) must be escaped with backslashes,"

This grep doesn't work:
 `OS_RELEASE_ID=$(cat /etc/os-release 2>/dev/null | grep '^ID=' | cut -d'=' -f2 | sed -e 's/"//g')`
(Also useless use of `cat | grep`)

Consider using instead,
 `OS_RELEASE_ID=$(source /etc/os-release >/dev/null 2>&1; echo $ID)`

kewlfft:
Thanks very much @eli-schwartz
`ID_LIKE` would make more sense, I understand Manjaro does not use it and it will not be compatible with it anymore. I don't understand why they don't use it, it may serve a purpose and I didn't want to break something I don't fully understand. Then if you confirm it makes sense to move forward I'm happy to do so.

eli-schwartz:
I'm not saying drop Manjaro support. I'm saying, use `ID_LIKE` as a fallback if `ID` cannot be detected, and then, remove the hardcoded support for archarm, arch32, and antergos since they'll be caught anyway by searching for `ID_LIKE=arch`

Also +1 on implementing the other part of my suggestion in #371

kewlfft:
see [gitlab.manjaro](https://gitlab.manjaro.org/packages/core/filesystem/commit/be254c117db9e715cad0f0f35487fb2c007a407b): things are progressing.

kewlfft:
@eli-schwartz implementing your suggestion in [#372](https://github.com/gentoo/portage/pull/372)

PR #359, opened 2018-08-11
Support !test? conditionals in RESTRICT (bug 663278)

zmedico:
Since RESTRICT="!test? ( test )" can be very useful within the context
of bug 663278, pass an appropriate uselist parameter to the RESTRICT
use_reduce call. Make self.configdict["features"]["USE"] independent
of IUSE and RESTRICT, so that the same value can be shared between
packages with different settings, which is important when evaluating
USE conditional RESTRICT. When the evaluated value of RESTRICT contains
"test", handle it like IUSE="-test", since features USE is independent
of RESTRICT.

Bug: https://bugs.gentoo.org/663278

### DIFF
--- a/lib/portage/package/ebuild/config.py
+++ b/lib/portage/package/ebuild/config.py
@@ -1457,6 +1457,7 @@ class config(object):
 		cp = cpv_getkey(mycpv)
 		cpv_slot = self.mycpv
 		pkginternaluse = ""
+		pkginternaluse_list = []
 		feature_use = []
 		iuse = ""
 		pkg_configdict = self.configdict["pkg"]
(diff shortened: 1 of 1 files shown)


zmedico:
Conditional RESTRICT has never been handled inside setcpv, and we need to handle it here for correct behavior within the context of bug 663278.

lib/portage/package/ebuild/config.py
```diff
@@ -1675,13 +1675,10 @@ def setcpv(self, mycpv, use_cache=None, mydb=None):
...
 						uselist=built_use, flat=True)
 				else:
-					# Use matchnone=True to ignore USE conditional parts
-					# of RESTRICT, since we want to know whether to mask
-					# the "test" flag _before_ we know the USE values
-					# that would be needed to evaluate the USE
-					# conditionals (see bug #273272).
 					restrict = use_reduce(raw_restrict,
-						matchnone=True, flat=True)
+						uselist=frozenset(x for x in self['USE'].split()
```
zmedico: Hmm, I've just realize that a `self.configdict["features"]["USE"]` setting from an earlier package could corrupt `self['USE']` here...

zmedico: I've solved this by making `self.configdict["features"]["USE"]` independent of IUSE, and by resetting `self.configdict["features"]["USE"]` when the previous package had a FEATURES setting from package.env.

zmedico: However, `self.configdict["features"]["USE"]` is still dependent on RESTRICT, so I need to eliminate that dependency or handle it somehow.

zmedico: I've solve this by making features USE independent of RESTRICT. When the evaluated value of RESTRICT contains "test", it's now handled like IUSE="-test".

PR #371, opened 2018-09-29
improve OS_RELEASE_ID definition

kewlfft:
adopt the [proposal](https://github.com/gentoo/portage/pull/349#issuecomment-425483117) fixing `OS_RELEASE_ID`

### DIFF
--- a/bin/etc-update
+++ b/bin/etc-update
@@ -32,7 +32,8 @@ get_config() {
 		"${PORTAGE_CONFIGROOT}"etc/etc-update.conf)
 }
 
-OS_RELEASE_ID=$(cat /etc/os-release 2>/dev/null | grep '^ID=' | cut -d'=' -f2 | sed -e 's/"//g')
+OS_RELEASE_ID=$(source /etc/os-release >/dev/null 2>&1; echo "${ID}")
+
 
 case $OS_RELEASE_ID in
 	suse|opensuse|opensuse-leap|opensuse-tumbleweed) OS_FAMILY='rpm' ;;
--- a/bin/etc-update
+++ b/bin/etc-update
@@ -34,7 +34,6 @@ get_config() {
 
 OS_RELEASE_ID=$(source /etc/os-release >/dev/null 2>&1; echo "${ID}")
 
-
 case $OS_RELEASE_ID in
 	suse|opensuse|opensuse-leap|opensuse-tumbleweed) OS_FAMILY='rpm' ;;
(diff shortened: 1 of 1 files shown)


kewlfft:
It is fixing the case for example: 
ID=arch\"007\"
The actual ID is arch"007" and not arch\007\

zmedico:
According to https://www.freedesktop.org/software/systemd/man/os-release.html, it's supposed to contain "shell-compatible variable assignments", so using `source` appears to be the right thing to do.

bin/etc-update
```diff
@@ -32,7 +32,7 @@ get_config() {
 		"${PORTAGE_CONFIGROOT}"etc/etc-update.conf)
 }
 
-OS_RELEASE_ID=$(cat /etc/os-release 2>/dev/null | grep '^ID=' | cut -d'=' -f2 | sed -e 's/"//g')
+OS_RELEASE_ID=$(source /etc/os-release >/dev/null 2>&1; echo $ID)
```
zmedico: Just in case $ID contains special shell characters, we should use `echo "$ID"`.

kewlfft: Ok then the `OS_RELEASE_ID` variable is not really necessary and we could simply use `ID`

zmedico: Well, the isolation provided by the subshell is useful, since lots of things could go wrong when a file is sourced.

kewlfft: yes sure, agreed.

mgorny: Also `${ID}`, please, for consistency with what we tend to use everywhere.

kewlfft:
Signed-off-by: Xrjy SSG xrjy@nygb.rh.bet (rot13)

bin/etc-update
eli-schwartz: When echoing a variable, shell characters in the variable definition shouldn't be reinterpreted? Quotes should only really matter for word splitting, which as a matter of good practice should still be quoted on those grounds alone.

The actual practical effect would be that 
```
$ ID="foo    bar"
$ echo $ID
foo bar
$ echo "$ID"
foo    bar
$ IFS=f
$ echo $ID
 oo    bar
$ echo "$ID"
foo    bar
```
Failure to quote, means the shell collapses whitespace (usually) due to reinterpreting it by way of `$IFS`.

kewlfft:
Can I close this PR as it is included in #372 which is being discussed ?

zmedico:
> Can I close this PR as it is included in #372 which is being discussed ?

Yes, please do.

PR #376, opened 2018-10-09
Repoman copyright changes

dol-sen:
@zmedico @ulm

### DIFF
--- a/repoman/man/repoman.1
+++ b/repoman/man/repoman.1
@@ -45,6 +45,18 @@ due to the automatic GitHub mirroring.
 
 Other platforms using the \fBCloses\fR tag include GitLab and Bitbucket.
 .TP
+\fB\-\-copyright\fR
+Update the copright header to include the \fICOPYRIGHT_OWNER\fR value if
+defined in \fBmake.conf\fR(5) or the default "Gentoo Authors".  This is not
+normally required for submissions from unpaid (volunteer) individuals.  It may
+however be a requirement of paid employee submissions from their complany
+sponsored Gentoo work.  This should only be used if there are significant
+changes (>=30%) being made to an existing or copied ebuild or vice-versa.
+ie: a Gentoo developer makes significant changes to an ebuild submitted with a
+different \fICOPYRIGHT_OWNER\fR than "Gentoo Authors".  Please see:
+.br
+https://www.gentoo.org/glep/glep-0076.html for further details.
+.TP
 \fB\-\-digest=<y|n>\fR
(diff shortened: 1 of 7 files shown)


ulm:
Not sure if that is a good change. We should strongly prefer the simplified notice with "Gentoo Authors" in ebuilds. IMHO it is not repoman's task to update any copyright lines. Especially:
- If there is a "Gentoo Authors" notice, don't change it and don't add anything else to it. (The idea is that "Gentoo Authors" is a catch-all.)
- If the notice contains "and others" (as policy requires if there is more than one contributor), don't add additional contributors to it. (Again, the idea is that "and others" is a catch-all, because tracing contributors may be difficult for ebuilds.)
- If the main contributor has changed, it could be updated. However, I don't see any way how this could be done in an automatic fashion. Consensus in discussions was that normally "_original author_ and others" should be kept.

dol-sen:
Well, this code actually takes Glep 76 into account more fully than just the "Gentoo Authors" simplified form.
This allows for company sponsored Gentoo work which may require a company copyright holder notice.  It is this reason, why this patch exists.  The "and others" may not be good enough for company sponsored Gentoo work.  Plus, it was my employer that sponsored this patch in anticipation of their copyright approval from the legal department.  I have done dual Signed-off-by lines so far (subject to change) while awaiting our companies legal department decision.

Plus, it makes repoman code more downstream friendly for many of the "Gentoo based" distributions or company repositories.

As for your last point.  It is not a fully automated process.  The change "addition of another copyright holder" only happens when the --copyright option is used by the committing developer.  It is the committing developer that must decide when to enable this option.  This code does not remove a copyright holder ever.   The closest thing to that is the conversion from "Gentoo Foundation" to "Gentoo Authors" code which pre-exists in the code.  It is just in a different form and more visible in this patch. please see this patch for those details: https://github.com/gentoo/portage/commit/9f089d270503d18ee8800fe97f26c6644f20cac9

repoman/man/repoman.1
```diff
@@ -45,6 +45,18 @@ due to the automatic GitHub mirroring.
 
 Other platforms using the \fBCloses\fR tag include GitLab and Bitbucket.
 .TP
+\fB\-\-copyright\fR
+Update the copright header to include the \fICOPYRIGHT_OWNER\fR value if
+defined in \fBmake.conf\fR(5) or the default "Gentoo Authors".  This is not
+normally required for submissions from unpaid (volunteer) individuals.  It may
+however be a requirement of paid employee submissions from their complany
```
kentfredric: typo: "complany"

PR #377, opened 2018-10-09
In case ID_LIKE contains space separated identifiers

kewlfft:
As per follow up discussion in https://github.com/gentoo/portage/pull/372#discussion_r223737087

### DIFF
--- a/bin/etc-update
+++ b/bin/etc-update
@@ -32,7 +32,7 @@ get_config() {
 		"${PORTAGE_CONFIGROOT}"etc/etc-update.conf)
 }
 
-OS_RELEASE_POSSIBLE_IDS=$(source /etc/os-release >/dev/null 2>&1; echo ":${ID}:${ID_LIKE}:")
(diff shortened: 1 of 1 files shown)


kewlfft:
Hopefully this should be a straightforward fix, I am waiting for it to update my Arch user package and avoid any bad surprise in case of ID_LIKE with multiple OS names

bin/etc-update
```diff
@@ -32,7 +32,7 @@ get_config() {
 		"${PORTAGE_CONFIGROOT}"etc/etc-update.conf)
 }
 
-OS_RELEASE_POSSIBLE_IDS=$(source /etc/os-release >/dev/null 2>&1; echo ":${ID}:${ID_LIKE}:")
+OS_RELEASE_POSSIBLE_IDS=$(source /etc/os-release >/dev/null 2>&1; echo ":${ID}:${ID_LIKE// /:}:")
```
zmedico: We can use `${ID_LIKE//[[:space:]]/:}` here to handle all whitespace characters (including tabs).

kewlfft:
@zmedico I have implemented your proposal.

kewlfft:
squashed as requested

zmedico:
Merged in 7a3d898b33d6c13f184d1328278e1ddaed368cf4, thanks!

PR #381, opened 2018-11-19
manjaro now uses arch ID_LIKE

kewlfft:
Signed-off-by: Kewl Fft <kewl@alto.eu.org>

See [os-release](https://gitlab.manjaro.org/packages/core/filesystem/blob/master/os-release)


zmedico:
Merged in cd3effcd48ca83057ca49d6e2727df931bfad4a0, thanks!

PR #387, opened 2018-12-13
do not use host PATH by default and prepend EPREFIX PATH.

heroxbd:
EPREFIX could be overridden in cross-eprefix, in that case tools
  inside EPREFIX should be prioritized.

  Link necessary binary tools to the test environment.

Closes: https://bugs.gentoo.org/585986


heroxbd:
Closes: https://github.com/gentoo/portage/pull/385

PR #389, opened 2018-12-20
bin/ebuild-helpers/portageq: match against portage/bin/ebuild-helpers

heroxbd:
In case PORTAGE_OVERRIDE_EPREFIX is modified in the environment,
  the actual ebuild-helpers PATH will not match PORTAGE_OVERRIDE_EPREFIX.

### DIFF
--- a/bin/ebuild-helpers/portageq
+++ b/bin/ebuild-helpers/portageq
@@ -13,7 +13,7 @@ set -f # in case ${PATH} contains any shell glob characters
 
 for path in ${PATH}; do
 	[[ -x ${path}/${scriptname} ]] || continue
-	[[ ${path} == ${PORTAGE_OVERRIDE_EPREFIX}/usr/lib*/portage/* ]] && continue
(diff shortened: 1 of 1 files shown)


bin/ebuild-helpers/portageq
```diff
@@ -13,7 +13,7 @@ set -f # in case ${PATH} contains any shell glob characters
 
 for path in ${PATH}; do
 	[[ -x ${path}/${scriptname} ]] || continue
-	[[ ${path} == ${PORTAGE_OVERRIDE_EPREFIX}/usr/lib*/portage/* ]] && continue
+	[[ ${path} == */portage/bin/ebuild-helpers* ]] && continue
```
zmedico: Generally the path looks like `/usr/lib/portage/python3.7/ebuild-helpers`, so maybe use `*/portage/*/ebuild-helpers/*`.

zmedico:
Merged in daeb75b345c4433218ab9e7a5319e8914092f048, thanks!

PR #392, opened 2019-01-02
file_copy: handle EUCLEAN from copy_file_range (bug 674332)

zmedico:
EXT4 can set the errno to EUCLEAN for copy_file_range.

Bug: https://bugs.gentoo.org/674332


zmedico:
The EUCLEAN might be triggered by a kernel bug, so I don't want to merge a patch until we have more info.

PR #400, opened 2019-01-17
fix try_source qa condition test

junwei-qu:


zmedico:
Merged in 0180b4e4a4499878a01005b317c8cea431bbb5fa, thanks!

PR #403, opened 2019-01-22
inherit: use local -x ECLASS (bug 676014)

zmedico:
Use local -x ECLASS, and also make __export_funcs_var a local variable,
since this is simpler than managing these variables as globals
(eliminates PECLASS and prev_export_funcs_var variables). As an
additional benefit, this prevents interference from ebuild variables
as reported in bug 656354 for the ECLASS variable.

Bug: https://bugs.gentoo.org/656354
Bug: https://bugs.gentoo.org/676014

### DIFF
--- a/bin/ebuild.sh
+++ b/bin/ebuild.sh
@@ -240,23 +240,22 @@ inherit() {
 	ECLASS_DEPTH=$(($ECLASS_DEPTH + 1))
 	if [[ ${ECLASS_DEPTH} -gt 1 ]]; then
 		debug-print "*** Multiple Inheritence (Level: ${ECLASS_DEPTH})"
+
+		# Since ECLASS_DEPTH > 1, the following variables are locals from the
+		# previous inherit call in the call stack.
+		if [[ -n ${ECLASS} && -n ${!__export_funcs_var} ]] ; then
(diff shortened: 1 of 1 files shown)


bin/ebuild.sh
```diff
@@ -242,34 +242,32 @@ inherit() {
...
 	local B_RDEPEND
 	local B_PDEPEND
 	local B_HDEPEND
 	local B_BDEPEND
+
+	if [[ -n ${ECLASS} && -n ${!__export_funcs_var} ]] ; then
+		echo "QA Notice: EXPORT_FUNCTIONS is called before inherit in" \
+			"${ECLASS}.eclass. For compatibility with <=portage-2.1.6.7," \
+			"only call EXPORT_FUNCTIONS after inherit(s)." \
+			| fmt -w 75 | while read -r ; do eqawarn "${REPLY}" ; done
```
Arfrever: This code was just moved in this commit, but I will note that it does not work since commit fb2459330cf226ee34d3875a1143531bd109aaf2 :) .
`fmt` is `/usr/bin/fmt`...

zmedico: Thanks, fixed in 2d37467170737bf055a058322c543f313737b983.

zmedico: Yes it would be nice to drop the warning, but the current  definition of EXPORT_FUNCTIONS in PMS is not really specific about this case.

PR #414, opened 2019-04-07
Change repoman exit code when nothing changed

Zlogene:
Signed-off-by: Mikle Kolyada <zlogene@gentoo.org>

Would be nice to get a code other than 1 on this, for better filtering the action in tools, this is not an error or a QA violation, so separate exit code should be useful.

### DIFF
--- a/repoman/lib/repoman/actions.py
+++ b/repoman/lib/repoman/actions.py
@@ -388,7 +388,7 @@ the whole commit message to abort.
 			print()
 			print("(Didn't find any changed files...)")
 			print()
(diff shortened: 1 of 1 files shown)


zmedico:
Maybe a different exit code would be better, since 127 (command not found) doesn't quite fit.

Also, we could add an `EXIT VALUES` section to [repoman/man/repoman.1](https://github.com/gentoo/portage/blob/master/repoman/man/repoman.1) to document the codes that we define.

floppym:
repoman is gone.

PR #422, opened 2019-04-29
autounmask: prefer || choices with existing packages (bug 327177)

zmedico:
When autounmask is enabled, make depgraph retry _select_atoms calls when
necessary to prefer || choices with existing packages. For example,
if package `C` is masked and package `B` does not exist, then autounmask
should choose `C` when given the choice `|| ( B C )`, as shown in the
included unit tests. The unit tests also show that autounmask still
prefers choices containing packages that are not masked (if available).

Bug: https://bugs.gentoo.org/327177

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -365,6 +365,14 @@ class _use_changes(tuple):
 		return obj
 
 
+_select_atoms_deps = collections.namedtuple('_select_atoms_deps', (
+	'normal_deps',
+	'virt_deps',
+	'ignored_deps',
+	'unsatisfied_deps',
+))
+
+
 class _dynamic_depgraph_config(object):
 
 	"""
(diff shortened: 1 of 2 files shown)


lib/_emerge/depgraph.py
```diff
@@ -3686,22 +3673,90 @@ def _wrapped_add_pkg_dep_string(self, pkg, dep_root, dep_priority,
...
+				break
+		else:
+			# If all choices have unsatisfied deps, fall back to default
+			# autounmask=False behavior.
+			choice = choices[0]
+
+			if autounmask:
+				# An autounmask choice has been rejected, so clear its
+				# package selections from the cache.
+				self._dynamic_config._filtered_trees[dep_root]["porttree"].dbapi._clear_cache()
```
zmedico: In the above case we actually need to revert any autounmask settings that might have been created.

zmedico: Also, when autounmask succeeds, it could create extraneous configuration changes. So, there has to be some way to extract only the changes that are actually needed.

PR #423, opened 2019-04-30
install: print category along package name

bjacquin:
It is more common in gentoo to see ${CATEGORY}/${PF} rather than ${PF}
followed by ${CATEGORY}


zmedico:
Merged in 445784f2cd9529ad03712d8d4b85784401d8b863. Thanks!

PR #424, opened 2019-05-04
fix some documentation in the dep API

DerDakon:
- fix bad parameter names
 - fix duplicate @ param that should be @ type
 - move "Example usage:" below the parameter list


zmedico:
Merged in 7f2c4b0d402be190a9575346a76e3ee9361179ac. Thanks!

DerDakon:
Would be nice if you can regenerate the docs on dev.gentoo.org then so they become better readable ;)

zmedico:
Thanks, updated:

https://dev.gentoo.org/~zmedico/portage/doc/api/portage.dep-module.html

PR #426, opened 2019-05-09
really fix some dep API documentation

DerDakon:
- interface description needs to be last for epydoc, even if it works at some places when it is not
 - fix some other minor things like wrong parameter names and missing types

Looks like #424 actually made some things worse than before.


zmedico:
Merged in 75dfea84dea28ca2e14b6f1cb7bc3ca1c2f87a0c. Thanks again! Rendered here:

https://dev.gentoo.org/~zmedico/portage/doc/api/portage.dep-module.html

PR #434, opened 2019-07-07
emerge: Don't treat empty EPREFIX or PORTAGE_CONFIGROOT as unset

chewi:
If a prefix user wanted to build within a `ROOT` but without a prefix,
they previously had to set `EPREFIX=/` rather than `EPREFIX=` as the
latter was simply treated as unset.

Also applies to `ROOT` and `SYSROOT` but probably makes no difference to
these as they are blank by default. This should be safe to do as all
these variables get normalised anyway.

Bug: https://bugs.gentoo.org/642604
Signed-off-by: James Le Cuirot <chewi@gentoo.org>


zmedico:
Merged in cfc3ca6647235d8a3fbe719b9d447863a75deeb6. Thanks!

PR #435, opened 2019-07-08
WIP: Cgroup Freezer Support

aidanharris:
This implements support for using the [cgroup-v1 freezer subsystem](https://raw.githubusercontent.com/torvalds/linux/v5.2/Documentation/cgroup-v1/freezer-subsystem.txt) and is largely a copy/paste job of the existing cgroup code which is only used for process tracking. This allows us to reliably pause/resume long-running `emerge` jobs (e.g `echo 'FROZEN' > /sys/fs/cgroup/freezer/portage/www-client:chromium*/freezer.state`)

The code works but I doubt is something you'd want to merge just yet. I'm posting this to start a discussion over whether or not this is something you'd be interested in. When writing this I largely duplicated a lot of existing boilerplate code, it made me wonder if there's perhaps a more generic way to handle this? Maybe we also want to limit CPU usage or memory usage using cgroups, etc.

### DIFF
--- a/bin/cgroup-freezer-release-agent
+++ b/bin/cgroup-freezer-release-agent
@@ -1,0 +1,6 @@
+#!/bin/sh
+# notify_on_release triggers for /sys/fs/freezer/portage too
+# We never want to call `rmdir` on this directory so if the
+# length of $1 is 8 (the same length as `/portage`)
+# skip the call to rmdir
+[ "${#1}" -gt 8 ] && exec rmdir "/sys/fs/cgroup/freezer/${1}"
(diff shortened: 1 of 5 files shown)


mgorny:
It would certainly be useful, especially with ninja not handling SIGTSTP correctly. Maybe we should even go further and map SIGTSTP/SIGCONT into freezing them.

aidanharris:
First attempt at signal handling. This works surprisingly well:

https://asciinema.org/a/rjkjKayiiTDgPoP4onKmDnSVn

PR #439, opened 2019-07-23
Fix emaint -c binhost support for binpkg-multi-instance

doctaweeks:
The check function should operate the same way as the fix function when
using FEATURES=binpkg-multi-instance (using the correct _instance_key)
otherwise packages will be erroneously reported as missing.

Bug: https://bugs.gentoo.org/690446
Signed-off-by: Daniel M. Weeks <dan@danweeks.net>


zmedico:
Merged in d107da65c3234b3e73d3ec296a5f843bfde97260. Thanks!

PR #447, opened 2019-08-19
Add FEATURES=pkgdir-index-trusted

zmedico:
Trust that the PKGDIR index file is valid, meaning that no packages
have been manually added or removed since the last call to emaint --fix
binhost. This feature eliminates overhead involved with detection of
packages that have been manually added or removed, which significantly
improves performance in some cases, such as when PKGDIR resides on
a high-latency network file system.

Bug: https://bugs.gentoo.org/688902

@mattst88


mattst88:
This works great. 2 minutes for 'time emerge world -vekp' down to 17 seconds on my system with 100ms of latency to the binpkg directory (NFS).

PR #452, opened 2019-08-30
emerge: deprecate --changelog option (bug 693096)

zmedico:
The emerge --changelog option is not very useful since the
gentoo repository no longer includes ChangeLog files.

Bug: https://bugs.gentoo.org/693096


mrueg:
Should it be highlighted on the man-page as well, that it's deprecated?

zmedico:
> Should it be highlighted on the man-page as well, that it's deprecated?

Yes, now I've added a note to the man-page.

PR #462, opened 2019-10-03
fetch: Support GLEP 75 mirror structure

mgorny:
Add a support for the subset of GLEP 75 needed by Gentoo Infra.  This
includes fetching and parsing layout.conf, and support for flat layout
and filename-hash layout with cutoffs being multiplies of 4.

// I'll send to the ml once I get green light from travis.

### DIFF
--- a/lib/portage/package/ebuild/fetch.py
+++ b/lib/portage/package/ebuild/fetch.py
@@ -6,13 +6,17 @@ from __future__ import print_function
 __all__ = ['fetch']
 
 import errno
+import functools
 import io
+import itertools
+import json
 import logging
 import random
 import re
 import stat
 import sys
 import tempfile
+import time
 
 from collections import OrderedDict
 
@@ -27,12 +31,17 @@ portage.proxy.lazyimport.lazyimport(globals(),
 	'portage.package.ebuild.doebuild:doebuild_environment,' + \
 		'_doebuild_spawn',
 	'portage.package.ebuild.prepare_build_dirs:prepare_build_dirs',
+	'portage.util:atomic_ofstream',
+	'portage.util.configparser:SafeConfigParser,read_configs,' +
+		'NoOptionError,ConfigParserError',
+	'portage.util._urlopen:urlopen',
 )
 
 from portage import os, selinux, shutil, _encodings, \
 	_movefile, _shell_quote, _unicode_encode
 from portage.checksum import (get_valid_checksum_keys, perform_md5, verify_all,
-	_filter_unaccelarated_hashes, _hash_filter, _apply_hash_filter)
+	_filter_unaccelarated_hashes, _hash_filter, _apply_hash_filter,
+	checksum_str)
 from portage.const import BASH_BINARY, CUSTOM_MIRRORS_FILE, \
 	GLOBAL_CONFIG_PATH
 from portage.data import portage_gid, portage_uid, secpass, userpriv_groups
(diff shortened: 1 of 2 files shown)


lib/portage/package/ebuild/fetch.py
```diff
@@ -434,8 +564,9 @@ def fetch(myuris, mysettings, listonly=0, fetchonly=0,
 	for myfile, myuri in file_uri_tuples:
 		if myfile not in filedict:
 			filedict[myfile]=[]
-			for y in range(0,len(locations)):
-				filedict[myfile].append(locations[y]+"/distfiles/"+myfile)
+			for l in locations:
+				filedict[myfile].append(get_mirror_url(l, myfile,
+						mysettings["EROOT"]))
```
zmedico: Let's make this lazy by appending `functools.partial(get_mirror_url, l, myfile, mysettings["EROOT"])`, and later evaluate it like this:
```diff
--- a/lib/portage/package/ebuild/fetch.py
+++ b/lib/portage/package/ebuild/fetch.py
@@ -895,6 +895,8 @@ def fetch(myuris, mysettings, listonly=0, fetchonly=0,
                        tried_locations = set()
                        while uri_list:
                                loc = uri_list.pop()
+                               if isinstance(loc, functools.partial):
+                                       loc = loc()
                                # Eliminate duplicates here in case we've switched to
                                # "primaryuri" mode on the fly due to a checksum failure.
                                if loc in tried_locations:

```

mgorny: Done and pushed.

lib/portage/package/ebuild/fetch.py
```diff
@@ -253,6 +260,130 @@ def _check_distfile(filename, digests, eout, show_errors=1, hash_filter=None):
...
+	"""
+	Class to read layout.conf from a mirror.
+	"""
+
+	def __init__(self):
+		self.structure = ()
+
+	def read_from_file(self, f):
+		cp = SafeConfigParser()
+		read_configs(cp, [f])
```
zmedico: We need to handle `configparser.ParsingError` somehow.

lib/portage/package/ebuild/fetch.py
```diff
@@ -253,6 +260,130 @@ def _check_distfile(filename, digests, eout, show_errors=1, hash_filter=None):
...
+			finally:
+				f.close()
+
+			mirror_conf.read_from_file(data)
+		except IOError:
+			pass
+
+		cache[mirror_url] = (time.time(), mirror_conf.serialize())
+		with open(cache_file, 'w') as f:
+			json.dump(cache, f)
```
zmedico: Unless we provide locking here, we should atomically update the file using `portage.util.atomic_ofstream` for example.

lib/portage/package/ebuild/fetch.py
mgorny: Hmm, I've did a quick test earlier and presumed it won't fail. But now I managed to mess up the config file enough for it to fail ;-). So I'll handle it.

mgorny: I went for letting the caller care for `ConfigParserException`. In emirrordist, we will want to know when mirror config is broken.

lib/portage/package/ebuild/fetch.py
mgorny: Done.

lib/portage/package/ebuild/fetch.py
```diff
@@ -253,6 +262,144 @@ def _check_distfile(filename, digests, eout, show_errors=1, hash_filter=None):
...
+
+	@param mirror_url: Base URL to the mirror (without '/distfiles')
+	@param filename: Filename to fetch
+	@param eroot: EROOT to use for the cache file
+	@return: Full URL to fetch
+	"""
+
+	mirror_conf = MirrorLayoutConfig()
+
+	cache_file = os.path.join(eroot, CACHE_PATH, 'mirror-metadata.json')
```
zmedico: This path currently does not have write access for the portage group, so it will cause the fetch function to fail with EACCES for repoman/ebuild(1) users in the portage group. However, DISTDIR is writable by these users.

zmedico: Also, for nfs root_squash support we could drop privileges for the entire fetch function (https://bugs.gentoo.org/601252).

mgorny: I suppose you're suggesting moving this file to DISTDIR.

zmedico: Yeah, for example, DISTDIR is also where repoman caches metadata.xsd.

mgorny:
Cache file now in DISTDIR, also made the path passed as argument.

mgorny:
Thanks, will do.

PR #463, opened 2019-10-06
[wip] emirrordist: mirror layout.conf support

mgorny:
@zmedico, how does this approach look? Note I've only implemented layouts for fetching new distfiles. I suppose I need to implement cleanup as well. Not sure if we want to explicitly support creating symlinks for previously fetched distfiles or if we want to do that by hand (I suppose it may have major performance impact since we'd have to test all distfiles).

### DIFF
--- a/lib/portage/_emirrordist/DeletionIterator.py
+++ b/lib/portage/_emirrordist/DeletionIterator.py
@@ -1,4 +1,4 @@
-# Copyright 2013 Gentoo Foundation
+# Copyright 2013-2019 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 import logging
@@ -20,17 +20,28 @@ class DeletionIterator(object):
 		deletion_db = self._config.deletion_db
 		deletion_delay = self._config.options.deletion_delay
 		start_time = self._config.start_time
-		distfiles_set = set(os.listdir(self._config.options.distfiles))
+		distfiles_set = set()
+		for layout in self._config.layouts:
+			distfiles_set.update(layout.get_filenames(distdir))
 		for filename in distfiles_set:
-			try:
-				st = os.stat(os.path.join(distdir, filename))
-			except OSError as e:
-				logging.error("stat failed on '%s' in distfiles: %s\n" %
-					(filename, e))
+			# require at least one successful stat()
+			first_exception = None
+			for layout in reversed(self._config.layouts):
+				try:
+					st = os.stat(
+							os.path.join(distdir, layout.get_path(filename)))
+				except OSError as e:
+					first_exception = e
+				else:
+					if stat.S_ISREG(st.st_mode):
+						break
+			else:
+				if first_exception is not None:
+					logging.error("stat failed on '%s' in distfiles: %s\n" %
+						(filename, first_exception))
 				continue
-			if not stat.S_ISREG(st.st_mode):
-				continue
-			elif filename in file_owners:
+
+			if filename in file_owners:
 				if deletion_db is not None:
 					try:
 						del deletion_db[filename]
--- a/lib/portage/_emirrordist/FetchTask.py
+++ b/lib/portage/_emirrordist/FetchTask.py
@@ -1,4 +1,4 @@
-# Copyright 2013-2014 Gentoo Foundation
+# Copyright 2013-2019 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 from __future__ import division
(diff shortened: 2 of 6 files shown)


lib/portage/_emirrordist/FetchTask.py
```diff
@@ -551,6 +552,23 @@ def _fetch_copier_exit(self, copier):
...
+		for layout in self.config.layouts[1:]:
+			if dist_path is None:
+				dist_path = os.path.join(self.config.options.distfiles,
+						self.config.layouts[0].get_path(self.distfile))
+			sym_path = os.path.join(self.config.options.distfiles,
+					layout.get_path(self.distfile))
+			ensure_dirs(os.path.dirname(sym_path))
+			# TODO: how to handle errors?
+			os.symlink(os.path.relpath(dist_path, os.path.dirname(sym_path)),
+					sym_path)
```
zmedico: Using symlinks rather than hardlinks so that mirrors can avoid the overhead of rsync --hard-links? When eliminating the flat structure, mirrors will have to re-transfer all files at that point?

mgorny: More like because we don't know what mirror admins are using. `-l` is part of `-a` (but `-H` is not), so I think there's a better chance of it actually working.

And yes, I presume that eliminating that structure will have major impact. That's why my plan involves eliminating it a few directories at a time, to reduce the change that users will have major problems using the mirrors while they update.

Of course the final decision is up to @gentoo/infra. I wonder if we should have a command-line option to decide between symlinks and hardlinks.

mgorny:
Ok, added support for deleting and I think that's all we need for now. I've also switched to hardlinks since they're easier to support.

lib/portage/package/ebuild/fetch.py
```diff
@@ -267,6 +268,9 @@ class FlatLayout(object):
 	def get_path(self, filename):
 		return filename
 
+	def get_filenames(self, distdir):
+		return os.listdir(distdir)
```
zmedico: Let's return `iter(os.listdir(distdir))` so it returns an iterator like the FilenameHashLayout method.

lib/portage/_emirrordist/FetchTask.py
```diff
@@ -551,6 +552,38 @@ def _fetch_copier_exit(self, copier):
...
+						self.config.layouts[0].get_path(self.distfile))
+			link_path = os.path.join(self.config.options.distfiles,
+					layout.get_path(self.distfile))
+			ensure_dirs(os.path.dirname(link_path))
+			try:
+				os.link(dist_path, link_path)
+			except OSError as e:
+				success = False
+				msg = ("%s -> %s link failed unexpectedly" %
+						(link_path, dist_path))
```
zmedico: We can include the exception message to make this more informative.

lib/portage/_emirrordist/DeletionIterator.py
```diff
@@ -20,17 +20,25 @@ def __iter__(self):
...
-				continue
-			if not stat.S_ISREG(st.st_mode):
+			# require at least one successful stat()
+			for layout in self._config.layouts:
+				try:
+					st = os.stat(
+							os.path.join(distdir, layout.get_path(filename)))
+				except OSError as e:
+					logging.error("stat failed on '%s' in distfiles: %s\n" %
+						(filename, e))
```
zmedico: This will produce some spurious error messages if all layouts do not contain exactly the same files. Do we expect all layouts to be consistent with each other at all times?

mgorny: I was wondering about it but couldn't come up with a good resolution. But yes, we expect them to be consistent except for a short initial propagation period.

What do you suggest? Storing the exception for the first layout, and logging it only if none of the files stat()ed correctly?

zmedico: Yeah, storing first exception sounds good.

lib/portage/_emirrordist/FetchTask.py
```diff
@@ -551,6 +552,49 @@ def _fetch_copier_exit(self, copier):
...
+			link_path = os.path.join(self.config.options.distfiles,
+					layout.get_path(self.distfile))
+			ensure_dirs(os.path.dirname(link_path))
+
+			if self.config.options.symlinks:
+				try:
+					os.symlink(
+							os.path.relpath(dist_path,
+								os.path.dirname(link_path)),
+							link_path)
```
zmedico: For completeness, we probably should support replacing an existing hardlink or symlink here. I suppose we could extend the _hardlink_atomic method to support symlink creation.

mgorny: …aaand done.

PR #465, opened 2019-10-11
emirrordist: Support specifying custom --layout-conf

mgorny:
Support specifying custom layout.conf file to use.  This makes it
possible to start propagating new mirror layout without exposing it
to users.


mgorny:
Indeed. I'll send that as the next patch.

mgorny:
Manpages changes sent to gentoo-portage-dev@. I suppose there's nothing to CI there, so I'll just skip pull request.

PR #466, opened 2019-10-12
emirrordist: Refetch only if none of the layouts have the file

mgorny:
Check all layouts for a file before refetching it.  This is helpful
when combined with --layout-conf that specifies the new layout
as primary to cover new files but user-oriented layout.conf still uses
old layout.

### DIFF
--- a/lib/portage/_emirrordist/FetchTask.py
+++ b/lib/portage/_emirrordist/FetchTask.py
@@ -61,23 +61,24 @@ class FetchTask(CompositeTask):
 			self._async_wait()
 			return
 
-		distfile_path = os.path.join(
-			self.config.options.distfiles,
-			self.config.layouts[0].get_path(self.distfile))
-
 		st = None
-		size_ok = False
-		try:
-			st = os.stat(distfile_path)
-		except OSError as e:
-			if e.errno not in (errno.ENOENT, errno.ESTALE):
-				msg = "%s stat failed in %s: %s" % \
-					(self.distfile, "distfiles", e)
-				self.scheduler.output(msg + '\n', background=True,
-					log_path=self._log_path)
-				logging.error(msg)
-		else:
-			size_ok = st.st_size == self.digests["size"]
+		for layout in self.config.layouts:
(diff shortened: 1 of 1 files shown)


lib/portage/_emirrordist/FetchTask.py
```diff
@@ -61,23 +61,24 @@ def _start(self):
...
+				layout.get_path(self.distfile))
+			try:
+				st = os.stat(distfile_path)
+			except OSError as e:
+				if e.errno not in (errno.ENOENT, errno.ESTALE):
+					msg = "%s stat failed in %s: %s" % \
+						(self.distfile, "distfiles", e)
+					self.scheduler.output(msg + '\n', background=True,
+						log_path=self._log_path)
+					logging.error(msg)
```
zmedico: We probably want the first_exception thing here again, like in 0d34d89d5028ac24cb45ed43ee1eed9cc9836e0c for DeletionIterator.

zmedico: Actually, it's fine as is since we ignore ENOENT.

mgorny: That's exactly what I thought.

lib/portage/_emirrordist/FetchTask.py
```diff
@@ -88,16 +89,20 @@ def _start(self):
...
+					if self._unlink_file(unlink_path, "distfiles"):
+						if st is not None:
+							logging.debug(("delete '%s' with "
+								"wrong size from distfiles") % (self.distfile,))
+					else:
+						self.config.log_failure("%s\t%s\t%s" %
+							(self.cpv, self.distfile, "unlink failed in distfiles"))
+						self.returncode = os.EX_OK
+						self._async_wait()
+						return
```
zmedico: It returns before we've looped through all of the layouts here.

mgorny: I don't think it really matters since it won't do anything useful anyway but sure, changed that.

PR #469, opened 2019-10-18
fetch: Use distfile fetching method to get layout.conf

mgorny:
Rewrite the layout.conf getter to reuse the standard fetch() method
rather than using urlopen().  While at it, fix negative cache elision
to apply to memory cache as well (and not get written to disk if next
mirror was fine).

Most importantly, this ensures that we respect FETCHCOMMAND while
fetching layout.conf, and so layout.conf is fetched the same way normal
distfiles are.  With some uncommon configurations, the previous disjoint
logic might have resulted in one of the fetches failing while the other
succeeded.

This also adds some nice verbosity.  If mirror connection takes a while,
the user sees that rather than having Portage wait silently.

Bug: https://bugs.gentoo.org/697566

//
@zmedico, I think this is gonna break that read-only fetching. Do you have a suggestion how to handle it? Or even preferably, could you fix it for me? I'm thinking we'd use a split-out function that figures out the filename for RO case.

### DIFF
--- a/lib/portage/package/ebuild/fetch.py
+++ b/lib/portage/package/ebuild/fetch.py
@@ -382,7 +382,7 @@ class MirrorLayoutConfig(object):
 		return ret
 
 
-def get_mirror_url(mirror_url, filename, cache_path=None):
+def get_mirror_url(mirror_url, filename, mysettings, cache_path=None):
 	"""
 	Get correct fetch URL for a given file, accounting for mirror
 	layout configuration.
@@ -408,23 +408,22 @@ def get_mirror_url(mirror_url, filename, cache_path=None):
 	if ts >= time.time() - 86400:
 		mirror_conf.deserialize(data)
 	else:
+		tmpfile = '.layout.conf.%s' % urlparse(mirror_url).hostname
 		try:
-			f = urlopen(mirror_url + '/distfiles/layout.conf')
-			try:
-				data = io.StringIO(f.read().decode('utf8'))
-			finally:
-				f.close()
-
-			mirror_conf.read_from_file(data)
+			if fetch({tmpfile: (mirror_url + '/distfiles/layout.conf',)},
+					mysettings, force=1, try_mirrors=0):
+				tmpfile = os.path.join(mysettings['DISTDIR'], tmpfile)
+				mirror_conf.read_from_file(tmpfile)
+			else:
+				raise IOError()
 		except (ConfigParserError, IOError, UnicodeDecodeError):
-			# Do not cache negative results.
-			cache_path = None
-
-		cache[mirror_url] = (time.time(), mirror_conf.serialize())
-		if cache_path is not None:
-			f = atomic_ofstream(cache_path, 'w')
-			json.dump(cache, f)
-			f.close()
+			pass
+		else:
+			cache[mirror_url] = (time.time(), mirror_conf.serialize())
+			if cache_path is not None:
+				f = atomic_ofstream(cache_path, 'w')
+				json.dump(cache, f)
+				f.close()
 
 	return (mirror_url + "/distfiles/" +
 			mirror_conf.get_best_supported_layout().get_path(filename))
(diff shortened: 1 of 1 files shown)


mgorny:
s/filename/file path/

lib/portage/package/ebuild/fetch.py
```diff
@@ -408,23 +408,25 @@ def get_mirror_url(mirror_url, filename, cache_path=None):
 	if ts >= time.time() - 86400:
 		mirror_conf.deserialize(data)
 	else:
+		tmpfile = '.layout.conf.%d' % time.time()
```
zmedico: We really need to use mkstemp to solve races here.

zmedico: And for FEATURES=skiprocheck read-only fetching it's good enough to use the existing urlopen code.

zmedico: In practice, we could probably get away with generating a random string instead of using mkstemp (and a random string would work for FEATURES=skiprocheck). But technically, mkstemp is the only truly reliable approach.

zmedico: Using a constant path containing the mirror name could work if we add a fetch parameter to force fetch even when the file exists already. We'd simply leave the file in DISTDIR, and stipulate that FEATURES=skiprocheck users should have a FETCHCOMMAND that replaces the file atomically.

mgorny: I've rewritten it to use `NamedTemporaryFile` but having the file exist makes Portage rename it to `_checksum_failure_` blah blah. Which of course reintroduces the race.

mgorny: I see `repoman` is also written with the brilliant idea of using `mkstemp()` and immediately discarding the file, rendering `mkstemp()` totally useless.

zmedico: Let's use a constant path containing the mirror name, assume that fetch updates it atomically, and don't remove it from DISTDIR. This approach will work with FEATURES=skiprocheck.

mgorny: Do you have any particular method of transforming mirror URI into safe filename component in mind?

zmedico: I imagined that we could use the hostname (fqdn), since it doesn't seem like there's a reason to have multiple mirrors per hostname.

mgorny: WFM.

mgorny:
Updated to append hostname to filename.

lib/portage/package/ebuild/fetch.py
```diff
@@ -408,23 +408,22 @@ def get_mirror_url(mirror_url, filename, cache_path=None):
...
 		try:
-			f = urlopen(mirror_url + '/distfiles/layout.conf')
-			try:
-				data = io.StringIO(f.read().decode('utf8'))
-			finally:
-				f.close()
-
-			mirror_conf.read_from_file(data)
+			if fetch({tmpfile: (mirror_url + '/distfiles/layout.conf',)},
+					mysettings, fetchonly=1, try_mirrors=0):
```
zmedico: We need to add a fetch parameter to make it fetch and atomically replace the file even if it happens to exist already.

zmedico: Also, the fetchonly parameter triggers non-blocking mode, but we really need to block here in order to ensure that the file is available when the function returns. In 491b3ff4d35194765c8ee93c6a438097551fd07e I've added a demonstration of the behavior for stale files when no digests are given. I'm going to work on adding a parameter to force fetch now.

zmedico: https://github.com/gentoo/portage/pull/470 implements a suitable fetch force parameter.

mgorny: Thanks! Updated now.

zmedico:
Merged in 9115e1f6ba35cdcd85f2292dc293f0696caa8f12.

PR #470, opened 2019-10-20
fetch: add force parameter (bug 697566)

zmedico:
Add a force parameter which forces download even when a file already
exists in DISTDIR (and no digests are available to verify it). This
avoids the need to remove the existing file in advance, which makes
it possible to atomically replace the file and avoid interference
with concurrent processes. This is useful when using FETCHCOMMAND to
fetch a mirror's layout.conf file, for the purposes of bug 697566.

Bug: https://bugs.gentoo.org/697566

### DIFF
--- a/lib/portage/package/ebuild/fetch.py
+++ b/lib/portage/package/ebuild/fetch.py
@@ -432,7 +432,7 @@ def get_mirror_url(mirror_url, filename, cache_path=None):
 
 def fetch(myuris, mysettings, listonly=0, fetchonly=0,
 	locks_in_subdir=".locks", use_locks=1, try_mirrors=1, digests=None,
-	allow_missing_digests=True):
+	allow_missing_digests=True, force=False):
 	"""
 	Fetch files to DISTDIR and also verify digests if they are available.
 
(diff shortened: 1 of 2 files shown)


lib/portage/package/ebuild/fetch.py
```diff
@@ -878,7 +886,7 @@ def onerror(e):
 				eout.quiet = mysettings.get("PORTAGE_QUIET") == "1"
 				match, mystat = _check_distfile(
 					myfile_path, pruned_digests, eout, hash_filter=hash_filter)
-				if match:
+				if match and not (force and not orig_digests):
```
mgorny: Wouldn't be it more logical for `force` to force even when digests are present? Or am I reading this condition wrong?

zmedico: You are correct, it would be more logical in a sense. However, if there are digests that don't match then it's going to fetch anyway.

mgorny: Could you add an explanatory comment there? Also avoiding the nested negation could be better for the reader. If my memory serves me right, `and (not force or orig_digests)` should be equivalent.

zmedico: The current logic allows `force=True` to be specified without forcing an unnecessary fetch when the digests match.

mgorny: That sounds illogical. Maybe it'd be better to error out if `force` is passed and digests are present at the same time.

zmedico: Ok, I've removed the orig_digests logic and added this condition at the beginning of the function:
```python
if force and digests:
	# Since the force parameter can trigger unnecessary fetch when the
	# digests match, do not allow force=True when digests are provided.
	raise PortageException(_('fetch: force=True is not allowed when digests are provided'))
```

PR #476, opened 2019-11-07
emirrordist: fix DeletionTask to handle broken symlink

zmedico:
After rename of a nonexistent symlink target fails, detect a
broken symlink and simply delete it.

Fixes: 0d34d89d5028 ("emirrordist: Implement mirror layout.conf support")
Bug: https://bugs.gentoo.org/699400


zmedico:
Actually, I withdraw this patch since we shouldn't have a symlink here
since self.distfile_path was added in af9c77bd037694164755d06034a74456ec45e3ab.

PR #480, opened 2019-11-26
emerge --buildpkgonly: respect buildtime hard blockers

zmedico:
Bug: https://bugs.gentoo.org/689226


mgorny:
I don't know the code but I guess it's fine. Do you need me to test it live?

zmedico:
> I don't know the code but I guess it's fine. Do you need me to test it live?

The unit tests should cover all the relevant cases I think, so it's probably fine if you don't test it live (maybe review the unit test though since they should be understandable).

PR #481, opened 2019-11-27
eapply: Output verbosely only if patch fails to apply with -F0

mgorny:
12d0c48ad disabled silent output for eapply, in order to obtain fuzz
factors in build logs.  However, this also causes eapply to report all
patched files which can make logs unreadable when there are no fuzz
factors to be reported.  Instead, use verbose output only when applying
the patch with -F0 fails.

To achieve that, attempt to apply each patch with -F0 --dry-run first.
If this succeeds, just silently apply the patch for real.  If it
doesn't, output an explicit eqawarn that the patch does not apply
cleanly and retry with the default fuzz factor and verbose output.
Non-silenced output applies both to successful application with fuzz
and to failure.

### DIFF
--- a/bin/phase-helpers.sh
+++ b/bin/phase-helpers.sh
@@ -995,8 +995,20 @@ if ___eapi_has_eapply; then
 			# -f to avoid interactivity
 			# -g0 to guarantee no VCS interaction
 			# --no-backup-if-mismatch not to pollute the sources
-			${patch_cmd} -p1 -f -g0 --no-backup-if-mismatch \
-				"${patch_options[@]}" < "${f}"
+			local all_opts=(
+				-p1 -f -g0 --no-backup-if-mismatch
+				"${patch_options[@]}"
+			)
+			# try applying with -F0 first, output a verbose warning
+			# if fuzz factor is necessary
+			if ${patch_cmd} "${all_opts[@]}" --dry-run -s -F0 \
+					< "${f}" &>/dev/null; then
+				all_opts+=( -s -F0 )
+			else
+				eqawarn "    ${f}: patch failed to apply without a fuzz factor, please rebase"
+			fi
+
+			${patch_cmd} "${all_opts[@]}" < "${f}"
 			failed=${?}
 			if ! eend "${failed}"; then
 				__helpers_die "patch -p1 ${patch_options[*]} failed with ${f}"


zmedico (approved):

temporaryrespite:
This fuzz 0 is a great idea. I only say that because I got bitten once on Archlinux by applying a patch with default fuzz(=2) and it was applied wrongly!

oh I found the note in my `patch` wrapper:
> #XXX unconditionally use -F 0 aka no fuzz! because the default of 2 fuzz applied a cargo patch wrongly and yielded a compilation error! the code was in a different function!!

so the patch added a change in a completely different (Rust)function. I wish I had saved the patch and the .rs file or something, but apparently I didn't.

bin/phase-helpers.sh
```diff
@@ -995,8 +995,20 @@ if ___eapi_has_eapply; then
...
+				-p1 -f -g0 --no-backup-if-mismatch
+				"${patch_options[@]}"
+			)
+			# try applying with -F0 first, output a verbose warning
+			# if fuzz factor is necessary
+			if ${patch_cmd} "${all_opts[@]}" --dry-run -s -F0 \
+					< "${f}" &>/dev/null; then
+				all_opts+=( -s -F0 )
+			else
+				eqawarn "    ${f}: patch failed to apply without a fuzz factor, please rebase"
```
temporaryrespite: I've just noticed that this PR already exists within portage 2.3.87 except the `else eqawarn` lines, which if would've existed, I would've noticed something... that applying my openssh patches for clock_nanosleep on top of the already applied same-patches that the ebuild had(the `https://dev.gentoo.org/~chutzpah/dist/openssh/${P}-glibc-2.31-patches.tar.xz` line has them as 0001 and 0002, from [this openssh commit](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=1aac6323197fdf15ee5d8ace28d31883a2099c9b)), just added one more clock_nanosleep block (with fuzz 2), ie. I would've seen this:

```
 * Applying 10_b1c82f4b8adf3f42476d8a1f292df33fb7aa1a56.patch ...
 *     /etc/portage/patches/net-misc/openssh/10_b1c82f4b8adf3f42476d8a1f292df33fb7aa1a56.patch: patch failed to apply without a fuzz factor, please rebase
patching file sandbox-seccomp-filter.c
Hunk #1 succeeded at 266 with fuzz 2 (offset 24 lines).   
```
instead of just [this](https://paste.debian.net/plain/1129367):
```
 * Applying 10_b1c82f4b8adf3f42476d8a1f292df33fb7aa1a56.patch ...
patching file sandbox-seccomp-filter.c
Hunk #1 succeeded at 266 with fuzz 2 (offset 24 lines).   
```

and thus unknowingly having two of these blocks in file `sandbox-seccomp-filter.c`:
```c
#ifdef __NR_clock_nanosleep
  SC_ALLOW(__NR_clock_nanosleep),
#endif
```

Sure, it doesn't hurt in this case, I guess.
But lesson learned, I'll be applying the eqawarn, at least, in my portage patches, thank you very much

PR #484, opened 2019-12-06
repoman: support profiles/package.deprecated (bug 702100)

zmedico:
Bug: https://bugs.gentoo.org/702100

### DIFF
--- a/repoman/lib/repoman/scanner.py
+++ b/repoman/lib/repoman/scanner.py
@@ -8,6 +8,7 @@ from itertools import chain
 import portage
 from portage import normalize_path
 from portage import os
+from portage._sets.base import InternalPackageSet
 from portage.output import green
 from portage.util.futures.extendedfutures import ExtendedFuture
 from repoman.metadata import get_metadata_xsd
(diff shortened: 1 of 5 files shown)


repoman/lib/repoman/scanner.py
```diff
@@ -93,6 +93,10 @@ def __init__(self, repo_settings, myreporoot, config_root, options,
 			'profile_list': profile_list,
 			'pmaskdict': global_pmaskdict,
 			'lic_deprecated': liclist_deprecated,
+			'package.deprecated': set(atom.cp for atom in
```
Arfrever: It might be useful to support entries like `x11-libs/gtk+:2` or `=x11-libs/gtk+-2*`...
(Commenting here because this `.cp` would be unnecessary.)

zmedico: That would complicate the logic in _depend_checks.py since we would have to match dependencies to dependencies (whereas dependencies normally match to packages).

zmedico: Fixed like this:
```python
all_deprecated = False
for pkg_match in portdb.xmatch("match-all", atom):
	if any(repo_metadata['package.deprecated'].iterAtomsForPackage(pkg_match)):
		all_deprecated = True
	else:
		all_deprecated = False
		break
```

PR #488, opened 2019-12-20
emerge --with-test-deps: allow circular deps

zmedico:
When USE=test is not enabled, allow circular test dependencies
by treating them like PDEPEND. When USE=test is enabled, circular
dependencies are still not allowed, as shown in unit tests.

Suggested-by: Michał Górny <mgorny@gentoo.org>
Bug: https://bugs.gentoo.org/703348


mgorny:
Hmm, so `USE=-test... --with-test-deps` installs `test?` deps. A little confusing but I suppose it makes sense. After all, you can't rely on them being installed.

PR #492, opened 2019-12-26
Initial IDEPEND support (for EAPI 8)

mgorny:
Please do an early review. If it looks reasonably good, I propose we start `eapi8` branch.


mgorny:
Found a few more places. Confirmed that `emerge -B` doesn't pull IDEPEND, `emerge -k` does.

PR #496, opened 2020-01-18
doebuild: export SANDBOX_LOG=${T}/sandbox.log (bug 704848)

zmedico:
In order to avoid a SANDBOX_LOG collision with another process
having the same pid when pid-sandbox is enabled, export a
unique SANDBOX_LOG value. The ${T} directory is a convenient
location, since it is guaranteed to exist and be writable during
relevant ebuild phases.

Bug: https://bugs.gentoo.org/704848


Whissi:
Is `$T` safe to use? Can't a malicious ebuild do `echo '' > ${T}/sandbox.log` to hide its attempt to escape sandbox?

zmedico:
The log has to be writable anyway, and we can't hid it because it's path is in the SANDBOX_LOG variable.

Whissi:
Isn't the idea that `SANDBOX_LOG` gets written by sandbox process so you are able to ensure that within sandbox itself, this path isn't within allowed path? So when you also set `SANDBOX_DENY` for the same path, file should be protected.

zmedico:
The process that violates the sandbox is the one that writes to SANDBOX_LOG, via the libsandbox.so LD_PRELOAD.

Whissi:
Ah, portage uses LD_PRELOAD and doesn't spawn sandbox process? I just tested with `SANDBOX_LOG=/tmp/foo.log SANDBOX_DENY=/tmp/foo.log sandbox` which caused an expected sandbox violation when I did something like `echo foo > /etc/test` and `/tmp/foo.log` containted that violation. I was also unable to touch that file *within* that sandbox but I was able to disable sandbox using `SANDBOX_ON=0...`

zmedico:
Portage spawns a sandbox process, and that does the export LD_PRELOAD=libsandbox.so for all child processes.

Whissi:
Am I missing something?
Test ebuild contains
```bash
src_prepare() {
        default

        echo "LOG: ${SANDBOX_LOG}"
        adddeny "${SANDBOX_LOG}"

        /tmp/test.py /etc/test1

        cat "${SANDBOX_LOG}"

        return 0
}
```

/tmp/test.py is
```Python
#!/bin/env python

import sys

def main(argv):
   outputfile = argv[0]
   if not outputfile:
      sys.exit(2)

   f = open(outputfile, 'w')
   f.write('This is a test\n')
   f.close()

if __name__ == "__main__":
   main(sys.argv[1:])
```

Output is:
```
>>> Preparing source in /var/tmp/portage/app-misc/testpkg-5.0.0/work ...
LOG: /var/log/sandbox/sandbox-4.log
 * ACCESS DENIED:  open_wr:      /etc/test1
  File "/tmp/test.py", line 15, in <module>
    main(sys.argv[1:])
  File "/tmp/test.py", line 10, in main
    f = open(outputfile, 'w')
PermissionError: [Errno 13] Permission denied: '/etc/test1'
 * ACCESS DENIED:  open_rd:      /var/log/sandbox/sandbox-4.log
cat: /var/log/sandbox/sandbox-4.log: Permission denied
>>> Source prepared.
 * --------------------------- ACCESS VIOLATION SUMMARY ---------------------------
 * LOG FILE: "/var/log/sandbox/sandbox-4.log"
 *
VERSION 1.0
FORMAT: F - Function called
FORMAT: S - Access Status
FORMAT: P - Path as passed to function
FORMAT: A - Absolute Path (not canonical)
FORMAT: R - Canonical Path
FORMAT: C - Command Line

F: open_wr
S: deny
P: /etc/test1
A: /etc/test1
R: /etc/test1
C: /usr/lib/python-exec/python3.8/python /tmp/test.py /etc/test1

F: open_rd
S: deny
P: /var/log/sandbox/sandbox-4.log
A: /var/log/sandbox/sandbox-4.log
R: /var/log/sandbox/sandbox-4.log
C: cat /var/log/sandbox/sandbox-4.log
 * --------------------------------------------------------------------------------
```
As you can see, sandbox was able to record both sandbox violations while access to `SANDBOX_LOG` was denied.

zmedico:
The libsanbox.so LD_PRELOAD is able to bypass the sandbox, it acts as a wrapper around the relevant C functions and bypasses the wrappers as necessary to perform its job (most of its job is to manage these wrappers).

mgorny:
@Whissi, sandbox is not a security feature. It never was one, and it will never be, unless it is reimplemented in a better way. There are easier ways to escape sandbox than by playing with logs.

PR #498, opened 2020-01-18
file_copy: handle lseek SEEK_DATA EBADFD for GlusterFS (bug 705536)

zmedico:
GlusterFS can set the errno to EBADFD for lseek SEEK_DATA.

Tested-by: Tomáš Mózes <hydrapolic@gmail.com>
Bug: https://bugs.gentoo.org/705536

### DIFF
--- a/src/portage_util_file_copy_reflink_linux.c
+++ b/src/portage_util_file_copy_reflink_linux.c
@@ -258,7 +258,7 @@ _reflink_linux_file_copy(PyObject *self, PyObject *args)
                     break;
                 } else if (len < 0) {
                     error = errno;
-                    if ((errno == EINVAL || errno == EOPNOTSUPP) && !offset_out) {
+                    if ((errno == EINVAL || errno == EOPNOTSUPP || errno == EBADFD) && !offset_out) {
                         lseek_works = 0;
                     }
                     break;


mgorny:
This seems to contradict POSIX. Was this reported as a bug upstream?

mgorny:
(sorry, clicked wrong button)

zmedico:
> This seems to contradict POSIX. Was this reported as a bug upstream?

No, and now that you mention it I suspect that it's not intended for lseek SEEK_DATA to fail with EBADFD. @hydrapolic we should create a minimal test case and report it at https://github.com/gluster/glusterfs/issues. For example, a command like `cp --sparse=always` should be able to create a sparse copy of a sparse file if lseek SEEK_DATA is working properly. Will you test that?

hydrapolic:
````
# mount | grep distfiles
server:/distfiles on /usr/portage/distfiles type fuse.glusterfs (rw,relatime,user_id=0,group_id=0,default_permissions,allow_other,max_read=131072)
# cp --sparse=always /usr/portage/distfiles/zlib-1.2.11.tar.gz /usr/portage/distfiles/zlib-1.2.11.tar.gz.copy
# echo $?
0
# md5sum /usr/portage/distfiles/zlib-1.2.11.tar.gz /usr/portage/distfiles/zlib-1.2.11.tar.gz.copy
1c9f62f0778697a09d36121ead88e08e  /usr/portage/distfiles/zlib-1.2.11.tar.gz
1c9f62f0778697a09d36121ead88e08e  /usr/portage/distfiles/zlib-1.2.11.tar.gz.copy
````

**Comment by zmedico** [comment]
@hydrapolic I've created a python script that we can use to check if lseek SEEK_DATA is working correctly for you. Here's example usage:
```
$ python3 ./lseek-seek-data.py -h
usage: lseek-seek-data.py [-h] input_file

positional arguments:
  input_file

optional arguments:
  -h, --help  show this help message and exit
$ dd if=/dev/zero of=/tmp/sparse bs=1M count=0 seek=1
0+0 records in
0+0 records out
0 bytes copied, 0.000220726 s, 0.0 kB/s
$ echo hi >> /tmp/sparse
$ python3 ./lseek-seek-data.py /tmp/sparse
1048576
```
This is the content of lseek-seek-data.py:
```python
#!/usr/bin/env python3

import argparse
import os


def main():
	parser = argparse.ArgumentParser()
	parser.add_argument('input_file')
	args = parser.parse_args()

	offset = 0

	with open(args.input_file, 'rb', 0) as input_file:
		print(os.lseek(input_file.fileno(), offset, os.SEEK_DATA))


if __name__ == '__main__':
	main()

```

hydrapolic:
````
# python3 ./lseek-seek-data.py /usr/portage/distfiles/sparse
  File "./lseek-seek-data.py", line 19, in <module>
    main()
  File "./lseek-seek-data.py", line 15, in main
    print(os.lseek(input_file.fileno(), offset, os.SEEK_DATA))
OSError: [Errno 77] File descriptor in bad state
````

zmedico:
@hydrapolic now we have a simple test case that you can report at https://github.com/gluster/glusterfs/issues. We need to know if this behavior is intentional or if it is an indication of some kind of problem that can be corrected somehow. If it indicates a problem that can be corrected then it's better if portage does not try to hide it from you.

hydrapolic:
@zmedico, sure, will do. Thanks.

hydrapolic:
https://bugzilla.redhat.com/show_bug.cgi?id=1793990

zmedico:
Supposed to be fixed by https://github.com/gluster/glusterfs/commit/9dd57ff963d4d0dde7f3ad0ff230837ff2f4b1df.

PR #501, opened 2020-01-27
dep_zapdeps: adjust || preference for slot upgrades (bug 706278)

zmedico:
Prefer choices that include a slot upgrade when appropriate, like for
the || ( llvm:10 ... llvm:7 ) case reported in bug 706278. In order
to avoid pulling in inappropriate slot upgrades, like those which
should only be pulled in with --update and --deep, add a want_update
flag to each choice which is True for choices that pull in a new slot
for which an update is desirable.

Mark the test case for bug 480736 as todo, since the "undesirable"
slot upgrade which triggers a blocker conflict in this test case is
practically indistinguishable from a desirable slot upgrade. This
particular blocker conflict is no longer relevant, since current
versions of media-libs/libpostproc are no longer compatible with
any available media-video/ffmpeg slot. In order to solve this test
case, some fancy backtracking (like for bug 382421) will be required.

Bug: https://bugs.gentoo.org/706278
Bug: https://bugs.gentoo.org/480736

### DIFF
--- a/lib/portage/dep/dep_check.py
+++ b/lib/portage/dep/dep_check.py
@@ -296,7 +296,7 @@ def dep_eval(deplist):
 
 class _dep_choice(SlotObject):
 	__slots__ = ('atoms', 'slot_map', 'cp_map', 'all_available',
-		'all_installed_slots', 'new_slot_count')
+		'all_installed_slots', 'new_slot_count', 'want_update', 'all_in_graph')
 
 def dep_zapdeps(unreduced, reduced, myroot, use_binaries=0, trees=None,
 	minimize_slots=False):
@@ -331,9 +331,9 @@ def dep_zapdeps(unreduced, reduced, myroot, use_binaries=0, trees=None,
 	# c) contains masked installed packages
 	# d) is the first item
 
-	preferred_installed = []
 	preferred_in_graph = []
-	preferred_any_slot = []
+	preferred_installed = preferred_in_graph
+	preferred_any_slot = preferred_in_graph
 	preferred_non_installed = []
 	unsat_use_in_graph = []
 	unsat_use_installed = []
@@ -347,8 +347,6 @@ def dep_zapdeps(unreduced, reduced, myroot, use_binaries=0, trees=None,
 	# for correct ordering in cases like || ( foo[a] foo[b] ).
 	choice_bins = (
 		preferred_in_graph,
-		preferred_installed,
-		preferred_any_slot,
 		preferred_non_installed,
 		unsat_use_in_graph,
 		unsat_use_installed,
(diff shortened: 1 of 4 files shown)


temporaryrespite:
this merge is causing this bug https://bugs.gentoo.org/707108

mgorny:
I don't think Portage should guess stuff based on `||` here. There's an explicit LLVM dependency below the `||` which tells it what max version to install.

lib/portage/dep/dep_check.py
```diff
@@ -694,15 +681,35 @@ def mydbapi_match_pkgs(atom):
...
 			cps = set(choice_1.cp_map)
 			for choice_2 in choices:
 				if choice_1 is choice_2:
 					# choice_1 will not be promoted, so move on
 					break
+				if (
+					# For removal actions, prefer choices where all packages
+					# have been pulled into the graph.
+					(graph_interface and graph_interface.removal_action and
+					choice_1.all_in_graph and not choice_2.all_in_graph)
```
zmedico: This must be the case that triggers that triggers removal of the latest python slot in https://bugs.gentoo.org/707108. We should be able to correct the problem be using the all_installed_slots attribute, for symmetry with the action that installs the new python slot.

zmedico:
> I don't think Portage should guess stuff based on `||` here. There's an explicit LLVM dependency below the `||` which tells it what max version to install.

Yeah, this patch prevents it from trying to make a guess. For || ( llvm:10 ... llvm:7 ), it simply picks the leftmost visible choice.

zmedico:
> this merge is causing this bug https://bugs.gentoo.org/707108#c14
> 
> ie. python 3_7 is being emerged(then depcleaned) just because it's in `PYTHON_COMPAT` of some dep's .ebuild .
> 
> I'd be happy to test any fixes, just lemme kno

Thanks! I'll start by trying to create a unit test case to reproduce the issue, and I'll let you know when I have a fix for you to test.

lib/portage/dep/dep_check.py
temporaryrespite: oh? if I understand this correctly, python won't be removed after the fix?
But I was under the impression that the issue was that python 3_7 was wrongly emerged(because all(?) python version from PYTHON_COMPAT were emerged), and correctly depcleaned, because without this commit, it wouldn't be emerged at all by portage.

zmedico: > oh? if I understand this correctly, python won't be removed after the fix?

Correct.

> But I was under the impression that the issue was that python 3_7 was wrongly emerged(because all(?) python versions from PYTHON_COMPAT were emerged), and correctly depcleaned, because without this commit, it wouldn't be emerged at all by portage.

Your impression is the same as the one expressed in https://bugs.gentoo.org/705700. Please refer to my response in https://bugs.gentoo.org/705700#c12 (and https://bugs.gentoo.org/706278 is an excellent example of what I'm talking about).

temporaryrespite: Understood. Thank you for clarifying.

I was confused because only the first of the following 3 emerges did bring in that python:
```
# emerge --deep --update --with-bdeps=y @world
# emerge --deep --update --with-bdeps=n @world
# emerge --update --with-bdeps=y @world
```
Unless I'm misunderstanding how they work(which is quite likely), after the fix, all 3 emerges above will bring python in?

zmedico: Those 3 emerge commands are all have very important differences which make it impossible to guarantee that they'll all bring in the new python slot. However, the first one is guaranteed to pull it in.

zmedico: https://github.com/gentoo/portage/pull/502 has a fix.

PR #502, opened 2020-02-01
depclean: do not eliminate upgrades (bug 707108)

zmedico:
For depclean actions, prefer choices where all packages have been
pulled into the graph, except for choices that eliminate upgrades.
This solves the test case for bug 707108, where depclean eliminated
a new slot of python that had been pulled in by a world update.
This should also prevent non-deterministic elimination of the
latest vala slot that was reported in bug 693790.

NOTE: There's a common perception (expressed in bug 705700) that
emerge is pulling in an "unecessary" python slot in cases when that
python slot is not enabled in PYTHON_TARGETS. However, the so-called
"unnecessary" slot is practically indistinguishable from a desirable
upgrade such as the missed llvm slot upgrade that was reported in
bug 706278. Therefore, be advised that emerge must pull in the
highest visible slot (regardless of PYTHON_TARGETS) in order to
ensure that a desirable upgrade is not missed.

Fixes: f7d83d75c6b0 ("dep_zapdeps: adjust || preference for slot upgrades (bug 706278)")
Bug: https://bugs.gentoo.org/707108
Bug: https://bugs.gentoo.org/706278
Bug: https://bugs.gentoo.org/705700
Bug: https://bugs.gentoo.org/693790

### DIFF
--- a/lib/portage/dep/dep_check.py
+++ b/lib/portage/dep/dep_check.py
@@ -690,17 +690,12 @@ def dep_zapdeps(unreduced, reduced, myroot, use_binaries=0, trees=None,
 					# choice_1 will not be promoted, so move on
 					break
 				if (
-					# For removal actions, prefer choices where all packages
-					# have been pulled into the graph.
-					(graph_interface and graph_interface.removal_action and
-					choice_1.all_in_graph and not choice_2.all_in_graph)
-
 					# Prefer choices where all_installed_slots is True, except
 					# in cases where we want to upgrade to a new slot as in
 					# bug 706278. Don't compare new_slot_count here since that
 					# would aggressively override the preference order defined
 					# in the ebuild, breaking the test case for bug 645002.
-					or (choice_1.all_installed_slots and
+					(choice_1.all_installed_slots and
 					not choice_2.all_installed_slots and
 					not choice_2.want_update)
 				):
@@ -711,8 +706,6 @@ def dep_zapdeps(unreduced, reduced, myroot, use_binaries=0, trees=None,
 					break
 
 				intersecting_cps = cps.intersection(choice_2.cp_map)
-				if not intersecting_cps:
-					continue
 				has_upgrade = False
 				has_downgrade = False
 				for cp in intersecting_cps:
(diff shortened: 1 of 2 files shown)


temporaryrespite:
Works as intended: python 3.7.6 is emerged but not depcleaned with this PR.
Thank you very much for your work and I really appreciate including that `NOTE` (not just in case future me forgets why that happens, but for others also)(not to imply that I've asked for it to be included - I haven't).

<details>

ie.:

python emerged via:
```
# emerge --deep --with-bdeps=y --update @world
```
but not via(which I understand that [is normal](https://github.com/gentoo/portage/pull/501#discussion_r373204323)):
```
# emerge --deep --with-bdeps=n --update @world
# emerge --with-bdeps=y --update @world
```

and python is NOT depcleaned(which is again normal(see the `NOTE` in OP, or [this](https://github.com/gentoo/portage/pull/501#discussion_r373182839)) via any of:
```
# emerge -av --depclean
# emerge --verbose --tree --deep --with-bdeps=y --changed-use --newuse --ask --complete-graph=y  --depclean
```
(the last depclean command most probably has args which are ignored, but as I am unsure, I kept it)


note to self: how I tested this PR ... the hacky way(ie. not using localrepo):

(showing what's changed)
```
# tail -n1 /etc/portage/package.accept_keywords 
=sys-apps/portage-9999 **

# regen() { pushd "/var/db/repos/gentoo/$1" && repoman -v  --digest=y  --ask manifest && popd ; };

# grep -B3 'BRANCH' /var/db/repos/gentoo/sys-apps/portage/portage-9999.ebuild
#EGIT_REPO_URI="https://anongit.gentoo.org/git/proj/portage.git
#	https://github.com/gentoo/portage.git"
EGIT_REPO_URI="https://github.com/zmedico/portage.git"
EGIT_BRANCH="bug_707108_depclean_do_not_eliminate_upgrades"

# regen sys-apps/portage
/var/db/repos/gentoo/sys-apps/portage /etc/portage
[INFO] checking package sys-apps/portage
>>> Creating Manifest for /var/db/repos/gentoo/sys-apps/portage
/etc/portage
```

get new portage:
```
# time emerge -av =sys-apps/portage-9999
```
get old portage:
```
# time emerge -av =sys-apps/portage-2.3.86
```
(switched between the two portages multiple times in order to depclean(or not) python, for tests)

</details>

PR #509, opened 2020-02-16
Support PORTAGE_LOG_FILTER_COMMAND (bug 709746)

zmedico:
This variable specifies a command that filters stdout and stderr of
build logs.


zmedico:
I want to redo the patch to support separate filter commands for file and tty output. For example we can use PORTAGE_LOG_FILTER_FILE to filter non-tty output, and PORTAGE_LOG_FILTER_TTY to filter tty output.

zmedico:
Issues with this patch:
* filters both tty and log output with the same command
* filter program has stdin to a pty device, which is unusual (program sees an EIO error when the opposite end of the pty is closed)

PR #514, opened 2020-02-26
Support PORTAGE_LOG_FILTER_FILE (bug 709746)

zmedico:
This variable specifies a command that filters build log output to a
log file. The plan is to extend this to support a separate filter for
tty output in the future.

In order to enable the EbuildPhase class to write elog messages to
the build log with PORTAGE_LOG_FILTER_FILE support, convert its _elog
method to a coroutine, and add a SchedulerInterface async_output
method for it to use.


zmedico:
Since https://github.com/gentoo/portage/pull/541 reverted this along with the async_start changes, https://github.com/gentoo/portage/pull/558 rebases it on current master branch.

PR #516, opened 2020-03-01
_GeneratorTask: throw CancelledError in cancelled coroutine (bug 711174)

zmedico:
Throw asyncio.CancelledError in a cancelled coroutine, ensuring
that the coroutine can handle this exception in order to perform
any necessary cleanup (like close the log file for bug 711174).
Note that the asyncio.CancelledError will only be thrown in the
coroutine if there's an opportunity (yield) before the generator
raises StopIteration.

Also fix the AsynchronousTask exit listener handling for
compatibility with this new behavior.

Fixes: 8074127bbc21 ("SpawnProcess: add _main coroutine")
Bug: https://bugs.gentoo.org/711174

### DIFF
--- a/lib/_emerge/AsynchronousTask.py
+++ b/lib/_emerge/AsynchronousTask.py
@@ -64,7 +64,7 @@ class AsynchronousTask(SlotObject):
 		@returns: Future, result is self.returncode
 		"""
 		waiter = self.scheduler.create_future()
-		exit_listener = lambda self: waiter.set_result(self.returncode)
+		exit_listener = lambda self: waiter.cancelled() or waiter.set_result(self.returncode)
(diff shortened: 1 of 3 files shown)


zmedico:
Since python3.8, the asyncio.CancelledError is not raised in the coroutine when _GeneratorTask tries to throw it, so it seems to be unusable for cleanup within the coroutine.

zmedico:
> Since python3.8, the asyncio.CancelledError is not raised in the coroutine when _GeneratorTask tries to throw it, so it seems to be unusable for cleanup within the coroutine.

Actually it is raised, it's just that it inherits from BaseException now. Moved to https://github.com/gentoo/portage/pull/517.

PR #520, opened 2020-03-04
SequentialTaskQueue: update bool(self) sooner (bug 711322)

zmedico:
Use addExitListener to add a _task_exit callback that will be invoked
as soon as the task exits (before the future's done callback is called).
This is required in order for bool(self) to have an updated value for
Scheduler._schedule to base assumptions upon. Delayed updates to
bool(self) is what caused Scheduler to hang as in bug 711322.

This reverts changes in SequentialTaskQueue task queue exit listener
behavior from commit c7e52d046621, so that only the changes necessary
to support async_start remain.

Fixes: c7e52d046621 ("EbuildPhase: add _async_start coroutine")
Bug: https://bugs.gentoo.org/711322

### DIFF
--- a/lib/_emerge/SequentialTaskQueue.py
+++ b/lib/_emerge/SequentialTaskQueue.py
@@ -2,7 +2,6 @@
 # Distributed under the terms of the GNU General Public License v2
 
 from collections import deque
-import functools
 import sys
 
 from portage.util.futures import asyncio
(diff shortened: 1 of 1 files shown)


ulm:
> is what caused Scheduler to hang as in bug 709746.

Shouldn't this refer to [bug 711322](https://bugs.gentoo.org/711322)?

zmedico:
> > is what caused Scheduler to hang as in bug 709746.
> 
> Shouldn't this refer to [bug 711322](https://bugs.gentoo.org/711322)?

Right, fixed. Thanks!

PR #529, opened 2020-03-09
emerge: Warn users when they install maintainer-needed pkgs

thesamesam:
Bug: https://bugs.gentoo.org/711466

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,11 @@ Bug fixes:
 
 * Allow nested ebegin again while there's a lack of consensus on banning it.
 
+* Fix use of deprecated (now removed) getchildren() call in XML parser,
+  which caused failures with gentoolkit. See bug #857537. (Note that this
+  was not originally in the release notes for 3.0.32, but the fix was in
+  3.0.32).
+
 * Fix doc generation by removing reference to removed EAPIs. See bug #857486.
 
 portage-3.0.31 (2022-07-10)


lib/_emerge/EbuildPhase.py
```diff
@@ -282,6 +286,9 @@ def _ebuild_exit_unlocked(self, ebuild_process, unlock_task=None):
 			_preinst_bsdflags(settings)
 		elif self.phase == "postinst":
 			_postinst_bsdflags(settings)
+			msg = _postinst_maintainer_nag(settings, self._get_metadata_maintainers())
+			if msg:
+				self._elog('ewarn', [msg], background=True)
```
zmedico: The _elog method is now a coroutine, so the calling function will have to be a coroutine as well. A simple fix would be to move this check to the _PostPhaseCommands _async_start coroutine like this:
```diff
diff --git a/lib/_emerge/EbuildPhase.py b/lib/_emerge/EbuildPhase.py
index fbb040ffd..ff3f93b02 100644
--- a/lib/_emerge/EbuildPhase.py
+++ b/lib/_emerge/EbuildPhase.py
@@ -442,6 +442,10 @@ class _PostPhaseCommands(CompositeTask):
 	__slots__ = ("commands", "elog", "fd_pipes", "logfile", "phase", "settings")
 
 	def _start(self):
+		self.scheduler.run_until_complete(self._async_start())
+
+	@coroutine
+	def _async_start(self):
 		if isinstance(self.commands, list):
 			cmds = [({}, self.commands)]
 		else:
@@ -451,6 +455,11 @@ class _PostPhaseCommands(CompositeTask):
 			cmds = [(kwargs, commands) for kwargs, commands in
 				cmds if not kwargs.get('selinux_only')]
 
+		if self.phase == "postinst":
+			msg = _postinst_maintainer_nag(self.settings, self._get_metadata_maintainers())
+			if msg:
+				yield self.elog('ewarn', [msg], background=True)
+
 		tasks = TaskSequence()
 		for kwargs, commands in cmds:
 			# Select args intended for MiscFunctionsProcess.

```

zmedico: The existing EbuildPhase _async_start method would also work, but anyway maybe it makes more sense to show this around the time of package.mask messages, so it shows for emerge --pretend.

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2050,6 +2050,12 @@ def _postinst_bsdflags(mysettings):
 			(_shell_quote(mysettings["ROOT"]),
 			_shell_quote(os.path.join(mysettings["T"], "bsdflags.mtree"))))
 
+def _postinst_maintainer_nag(mysettings, maintainers):
```
thesamesam: Needs to be guarded (not necessarily here) by FEATURES, e.g. maintainer-needed-nag (thanks Zac).

lib/_emerge/depgraph.py
```diff
@@ -53,6 +53,15 @@
 from portage.util._eventloop.global_event_loop import global_event_loop
 from portage.versions import _pkg_str, catpkgsplit
 
+try:
+	from portage.xml.metadata import MetaDataXML
+except (SystemExit, KeyboardInterrupt):
+	raise
+except (ImportError, SystemError, RuntimeError, Exception):
```
zmedico: I realize that this is the existing exception handling that we use in EbuildPhase since ee3bf6f9773c0afbd494c657d253241f79d98044, but with the versions of python that we support today (including 2.7) I think it should be sufficient to do it like this:
```python
try:
	from portage.xml.metadata import MetaDataXML
except Exception:
	MetaDataXML = None
```

lib/_emerge/depgraph.py
```diff
@@ -9157,6 +9166,35 @@ def display_problems(self):
 			show_mask_docs()
 			writemsg("\n", noiselevel=-1)
 
+		features = self._frozen_config.settings.features
+		if "maintainer-nag" in features:
+			# Build a list of problematic packages
+			maintainer_nag_packages = []
+
+			for pkg in self._dynamic_config.digraph:
+				if(isinstance(pkg, Package) and pkg.repo == "gentoo"):
```
zmedico: We can omit the outer pair of parenthesis here.

lib/_emerge/depgraph.py
```diff
@@ -9157,6 +9166,35 @@ def display_problems(self):
...
+					settings = pkg.root_config.settings
+					path = settings.environ()['PORTDIR'] + "/" + pkg.cp + "/metadata.xml"
```
zmedico: The PORTDIR variable is deprecated, so let's use `os.path.join(os.path.dirname(pkg.root_config.trees['porttree'].dbapi.findname(pkg.cpv, myrepo=pkg.repo)), 'metadata.xml')`.

lib/_emerge/depgraph.py
thesamesam: Yeah, I agree, I did check the bug but forgot to follow up on it. I'll fix that shortly.

lib/_emerge/depgraph.py
thesamesam: Thanks, that pointer is what I needed. Had a feeling it wasn't right!

lib/portage/repository/config.py
```diff
@@ -177,6 +179,13 @@ def __init__(self, name, repo_opts, local_config=True):
 		else:
 			masters = None
 
+		self.maintainer_nag = repo_opts.get('maintainer-nag', 'yes').lower() == 'yes'
+
+		maintainer_nag_url = repo_opts.get('maintainer-nag-url')
```
zmedico: We should also support setting maintainer-nag-url in metadata/layout.conf, which means supporting it in parse_layout_conf and getting the value from layout_data.

lib/portage/repository/config.py
```diff
@@ -177,6 +179,13 @@ def __init__(self, name, repo_opts, local_config=True):
 		else:
 			masters = None
 
+		self.maintainer_nag = repo_opts.get('maintainer-nag', 'yes').lower() == 'yes'
```
zmedico: We have to disable it by default here, since we can't assume that third-party repos will have maintainers listed in metadata.xml.

zmedico: Meanwhile, we need to make sure that we respect a default from layout.conf if it's set there.

lib/_emerge/depgraph.py
```diff
@@ -9157,6 +9162,47 @@ def display_problems(self):
...
+		for pkg in self._dynamic_config.digraph:
+			if not isinstance(pkg, Package):
+				continue
+
+			maintainer_nag = pkg.root_config.settings.repositories[pkg.repo].maintainer_nag
+			maintainer_nag_url = pkg.root_config.settings.repositories[pkg.repo].maintainer_nag_url
+			if not maintainer_nag:
+				continue
+
+			path = os.path.dirname(pkg.root_config.trees['porttree'].dbapi.findname(pkg.cpv, myrepo=pkg.repo))
```
zmedico: We need to check if `pkg.type_name == 'ebuild'` before doing this, in order to avoid the test failure seen here: 

https://travis-ci.org/github/gentoo/portage/builds/665647387

thesamesam: I noticed this and was checking the findname output for None, but this is nicer. I'll change to that.

lib/portage/repository/config.py
thesamesam: Fixed the default locally, need to figure out why I can't get it to respect layout.conf defaults at the moment.

lib/portage/repository/config.py
```diff
@@ -340,7 +351,7 @@ def __init__(self, name, repo_opts, local_config=True):
 
 			for value in ('allow-missing-manifest',
 				'cache-formats',
-				'create-manifest', 'disable-manifest', 'manifest-hashes',
+				'create-manifest', 'disable-manifest', 'manifest-hashes', 'maintainer-nag', 'maintainer-nag-url',
 				'manifest-required-hashes', 'profile-formats', 'properties-allowed', 'restrict-allowed',
 				'sign-commit', 'sign-manifest', 'thin-manifest', 'update-changelog'):
 				setattr(self, value.lower().replace("-", "_"), layout_data[value])
```
zmedico: The layout.conf attributes overwrites the repos.conf settings here. This is intentional in the sense that we don't want to allow repos.conf to override some settings, so maybe add a second layout.conf loop for settings that repos.conf is allowd to override.

lib/_emerge/depgraph.py
```diff
@@ -9157,6 +9162,48 @@ def display_problems(self):
...
+			writemsg("\n" + colorize("WARN", "!!!") + \
+				 " The following packages have no maintainer. They are at greater risk of bugs." +
+				 "\n!!! You can help to keep it in tree and fix bugs (inc. possible security issues)")
+
+			if maintainer_nag_url:
+				writemsg("\n!!! Please see: " + maintainer_nag_url)
+
+			print()
+			for pkg in maintainer_nag_packages:
+				print("- " + pkg)
```
zmedico: Since writemsg outputs to stderr, we need the rest of the output to go to stderr as well, otherwise there will be odd cases where the outputs do not appear together as they should.

lib/_emerge/depgraph.py
```diff
@@ -9157,6 +9162,49 @@ def display_problems(self):
...
+					metadata_xml = MetaDataXML(path, None)
+					maintainer = metadata_xml.format_maintainer_string()
+				except SyntaxError:
+					continue
+
+				if not maintainer:
+					maintainer_nag_packages.append(pkg.cpv + "::" + pkg.repo)
+
+		if maintainer_nag_packages:
+			writemsg("\n" + colorize("WARN", "!!!") + \
```
zmedico, line 9195: The first `!!!` is yellow, but not the other 2. Was that intentional? I think it might look better if they're all the same color.

Also, I'm a little concerned that this message might flood the terminal for a command like `emerge -e @world` for example. Maybe we should have some kind of limit on the number of packages displayed?

lib/portage/repository/config.py
```diff
@@ -338,13 +342,20 @@ def __init__(self, name, repo_opts, local_config=True):
...
 
+			# Allow repos.conf override of layout.conf for these
+			for value in ('maintainer-nag', 'maintainer-nag-url'):
+				setting = repo_opts.get(value)
+				if setting:
+					setattr(self, value.lower().replace("-", "_"), setting)
+
```
zmedico, line 358: When testing, I needed to add something like this to ensure that `maintainer-nag = no` translated to a boolean False, otherwise I was unable to disable the warning:
```python
			self.maintainer_nag = (self.maintainer_nag.lower() in ('true', 'yes')
				if isinstance(self.maintainer_nag, basestring) else self.maintainer_nag)
```

cnf/repos.conf
```diff
@@ -17,3 +17,5 @@ sync-openpgp-key-refresh-retry-delay-exp-base = 2
 sync-openpgp-key-refresh-retry-delay-max = 60
 sync-openpgp-key-refresh-retry-delay-mult = 4
 sync-webrsync-verify-signature = yes
+maintainer-nag = yes
+maintainer-nag-url = https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers
```
zmedico, line 21: We should document these settings in man/portage.5.

lib/_emerge/depgraph.py
zmedico: Also, if maintainer_nag will be enabled by default, then we probably need the message to include instructions for how to disable it maintainer_nag.

lib/_emerge/depgraph.py
```diff
@@ -9157,6 +9162,49 @@ def display_problems(self):
...
+				 " The following packages have no maintainer. They are at greater risk of bugs." +
+				 "\n!!! You can help to keep it in tree and fix bugs (inc. possible security issues)")
+
+			if maintainer_nag_url:
+				writemsg("\n!!! Please see: " + maintainer_nag_url)
```
zmedico, line 9200: There could be a different maintainer_nag_url for each repository.

mjeveritt:
@thesamesam yello .. how we lookin' here?! :stuck_out_tongue:  and hi!

lib/_emerge/depgraph.py
mjeveritt: I suggest setting a flag, and displaying a message at the end with the list of packages attached.

PR #536, opened 2020-03-22
emerge --info: Filter some vars for credentials

thesamesam:
Filters GENTOO_MIRRORS and PORTAGE_BINHOST for passwords.

Closes: https://bugs.gentoo.org/713726
Signed-off-by: Sam James (sam_c) <sam@cmpct.info>

### DIFF
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -56,6 +56,7 @@ bad = create_color_func("BAD")
 warn = create_color_func("WARN")
 from portage.package.ebuild._ipc.QueryCommand import QueryCommand
 from portage.package.ebuild.doebuild import _check_temp_dir
+from portage.package.ebuild.fetch import _hide_url_passwd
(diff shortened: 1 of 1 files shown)


lib/_emerge/actions.py
```diff
@@ -1899,6 +1902,10 @@ def action_info(settings, trees, myopts, myfiles):
 				if default is not None and \
 					default == v:
 					continue
+
+				if k in filter_vars:
+					v = re.sub(r'//(.+):.+@(.+)', r'//\1:*password*@\2', v)
+
```
zmedico: Maybe we can just import the _hide_url_passwd function and use it here, and I'm thinking we should use it on all the variables just in case.

zmedico:
This will fix _hide_url_passwd to replace all occurrences, since since PORTAGE_BINHOST may contain multiple values: https://github.com/gentoo/portage/pull/537

PR #539, opened 2020-03-27
process: Unshare UTS namespace, and set hostname to 'localhost'

mgorny:
Use UTS namespace to override hostname when network-sandbox is enabled.
Set it to 'localhost' as that has a better chance of being present
in /etc/hosts.  This fixes tests in some packages that try to connect
to localhost via hostname obtained using gethostname(), e.g. docker-py,
and suffer resolution problems due to the system hostname not being
defined in /etc/hosts.

### DIFF
--- a/lib/portage/process.py
+++ b/lib/portage/process.py
@@ -348,12 +348,14 @@ def spawn(mycommand, env=None, opt_name=None, fd_pipes=None, returnpid=False,
 	if unshare_net or unshare_ipc or unshare_mount or unshare_pid:
 		# from /usr/include/bits/sched.h
 		CLONE_NEWNS = 0x00020000
+		CLONE_NEWUTS = 0x04000000
 		CLONE_NEWIPC = 0x08000000
 		CLONE_NEWPID = 0x20000000
 		CLONE_NEWNET = 0x40000000
 
 		if unshare_net:
-			unshare_flags |= CLONE_NEWNET
+			# UTS namespace to override hostname
+			unshare_flags |= CLONE_NEWNET | CLONE_NEWUTS
 		if unshare_ipc:
 			unshare_flags |= CLONE_NEWIPC
 		if unshare_mount:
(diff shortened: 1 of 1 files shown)


mgorny:
I'll send it to the ml once travis finishes.

lib/portage/process.py
```diff
@@ -704,6 +706,8 @@ def _exec(binary, mycommand, opt_name, fd_pipes,
 									noiselevel=-1)
 								os._exit(1)
 						if unshare_net:
+							# use 'localhost' to avoid hostname resolution problems
+							socket.sethostname('localhost')
```
zmedico: Maybe we should treat an exception as a warning here.

zmedico: And if it fails, we can include in the message that it's for FEATURES=network-sandbox (like the other `"Unable to unshare: %s (for FEATURES=\"%s\")\n"` message.

lib/portage/process.py
```diff
@@ -704,6 +706,13 @@ def _exec(binary, mycommand, opt_name, fd_pipes,
 									noiselevel=-1)
 								os._exit(1)
 						if unshare_net:
+							# use 'localhost' to avoid hostname resolution problems
+							try:
+								socket.sethostname('localhost')
+							except:
```
zmedico: We better not catch BaseException, so lets `except Exception:`.

lib/portage/process.py
```diff
@@ -704,6 +706,13 @@ def _exec(binary, mycommand, opt_name, fd_pipes,
 									noiselevel=-1)
 								os._exit(1)
 						if unshare_net:
+							# use 'localhost' to avoid hostname resolution problems
+							try:
+								socket.sethostname('localhost')
+							except Exception:
+								writemsg("Unable to set hostname: %s (for FEATURES=\"network-sandbox\")\n" % (
+									errno.errorcode.get(errno_value, '?'),),
```
zmedico: Oh, we want the exception here rather than the errno thing.

PR #540, opened 2020-03-29
NeededEntry: infer implicit soname from file basename (bug 715162)

zmedico:
For dynamic libraries, infer an implicit DT_SONAME setting from the
file basename, which is consistent with dynamic linking behavior in
practice. This makes it possible to resolve soname dependencies for
musl's libc.so which lacks a DT_SONAME setting.

Bug: https://bugs.gentoo.org/715162


zmedico:
Need to fix this to distinguish PIE executables from dynamic libraries, like https://github.com/file/file/commit/03084b161cf888b5286dbbcd964c31ccad4f64d9 for example.

PR #547, opened 2020-05-06
doebuild: Use ccache/distcc/icecream only in src_* phases

mgorny:
Enable ccache/distcc/icecream only when src_* phases are executed.
There is generally little value from them in pkg_*, as these phases
use CC only to do a few tests at most.  These compilations generally
are not run in parallel and the programs are tiny, so usually
the overhead exceeds the gain.

On the other hand, running them in pkg_* phases results in files being
created with root ownership that breaks them afterwards when using
userpriv.  As a result, user not only loses the benefit of, say, distcc
but also frequently hits error messages that confuse some configure
scripts and result in miscompiled packages.

Bug: https://bugs.gentoo.org/581880

### DIFF
--- a/lib/portage/package/ebuild/doebuild.py
+++ b/lib/portage/package/ebuild/doebuild.py
@@ -486,7 +486,10 @@ def doebuild_environment(myebuild, mydo, myroot=None, settings=None,
 		distcc = "distcc" in mysettings.features
 		icecream = "icecream" in mysettings.features
 
-		if ccache or distcc or icecream:
+		# run cc thingies only in src_* phases as otherwise they might
+		# create files with incorrect ownership and trip over
+		if (ccache or distcc or icecream) and mydo in ('unpack',
+				'prepare', 'configure', 'test', 'install'):
 			libdir = None
 			default_abi = mysettings.get("DEFAULT_ABI")
 			if default_abi:


lib/portage/package/ebuild/doebuild.py
```diff
@@ -486,7 +486,10 @@ def doebuild_environment(myebuild, mydo, myroot=None, settings=None,
 		distcc = "distcc" in mysettings.features
 		icecream = "icecream" in mysettings.features
 
-		if ccache or distcc or icecream:
+		# run cc thingies only in src_* phases as otherwise they might
+		# create files with incorrect ownership and trip over
+		if (ccache or distcc or icecream) and mydo in ('unpack',
+				'prepare', 'configure', 'test', 'install'):
```
zmedico, line 492: The `install` phase is different from the others because it runs as root. Is that ok?

mgorny: I've decided to include it because 1) there are some bad packages that recompile stuff in `src_install`, and then ccache is very helpful, and 2) almost always some compilation happens before `src_install`, so permissions are set up already. Do you want me to add an explanatory comment there?

zmedico: Yeah, a comment would be great. Thanks!

PR #550, opened 2020-05-17
Enable FEATURES=parallel-install by default (bug 715110)

zmedico:
Enable FEATURES=parallel-install in order to increase IO throughput by
allowing files to be merged or unmerged for one package while merge or
unmerge ebuild phases execute for a different package.

This feature introduces a small risk of file collisions going
undetected for packages that are merged at about the same time, but
the risk is considered to be practically negligible since those file
collisions would typically be detected and prevented long before such
an event would have an opportunity to occur.

Bug: https://bugs.gentoo.org/715110


mjeveritt:
See also [bug 725112](https://bugs.gentoo.org/725112)

PR #552, opened 2020-05-26
config.environ: delay export of A and AA (bug 720180)

zmedico:
Since variables like A and AA can contain extremely large values which
may trigger E2BIG errors during attempts to execute subprocesses, delay
export until the last moment, and unexport when appropriate.

Bug: https://bugs.gentoo.org/720180

### DIFF
--- a/bin/eapi.sh
+++ b/bin/eapi.sh
@@ -26,6 +26,15 @@ ___eapi_has_S_WORKDIR_fallback() {
 
 # VARIABLES
 
+___eapi_exports_A() {
+	# https://bugs.gentoo.org/721088
+	true
+}
+
+___eapi_exports_AA() {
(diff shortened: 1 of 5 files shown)


zmedico:
> I am also not sure how the changes of this PR are supposed to help. Sorry for my ignorance. But the issue that the environment is potentially too large to spawn a process could already happen once portage fork+execs `ebuild.sh`. Therefore, unexporting `A` in `ebuild.sh` is too late. But maybe I am missing something? For newbies like me, portage's code isn't easy to understand.

The vars are written to a file so they can be sourced after exec. I'll close this PR though since it was more of a POC and I like your idea to use explicit files for this.

Flowdalic:
> I like your idea to use explicit files for this.

Glad that you like it. Posted a first PoC at https://github.com/gentoo/portage/pull/1357

PR #555, opened 2020-06-08
xtermTitle: support alacritty

kchibisov:
alacritty terminfo is included in ncurses-6.2 and so it is
picked up by default on most systems using alacritty terminal emulator,
which causes an issue of portage not updating window title.

Signed-off-by: Kirill Chibisov <contact@kchibisov.com>


zmedico:
Thanks!

PR #556, opened 2020-06-12
PipeLogger: non-blocking write to pipe (bug 709746)

zmedico:
Add support to write to a non-blocking pipe instead of a
log file. This is needed for the purposes of bug 709746,
where PipeLogger will write to a pipe that is drained
by anoher PipeLogger instance which is running in the same
process.

### DIFF
--- a/lib/portage/util/_async/PipeLogger.py
+++ b/lib/portage/util/_async/PipeLogger.py
@@ -8,6 +8,9 @@ import sys
 
 import portage
 from portage import os, _encodings, _unicode_encode
+from portage.util.futures import asyncio
+from portage.util.futures.compat_coroutine import coroutine
+from portage.util.futures.unix_events import _set_nonblocking
 from _emerge.AbstractPollTask import AbstractPollTask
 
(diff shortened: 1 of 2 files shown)


lib/portage/util/_async/PipeLogger.py
```diff
@@ -120,8 +131,32 @@ def _output_handler(self, fd):
...
 
 				if log_file is not None:
-					log_file.write(buf)
-					log_file.flush()
+					write_buf = buf
+					while write_buf:
+						try:
+							# Use os.write, since the log_file.write method
+							# looses data when an EAGAIN occurs.
+							write_buf = write_buf[os.write(log_file.fileno(), write_buf):]
```
zmedico, line 139: This broke FEATURES=compress-build-logs, but it can be fixed by offloading compression to a gzip process (https://bugs.gentoo.org/712758), so I'll look into implementing that now.

zmedico: There's a minimal fix in 54cdf7e6cf9ec2030283ab23d659e10562267c6b.

PR #559, opened 2020-06-15
Use env to find python.

R030t1:
Supports portage on alternate architectures; encountered issues using prefix.

```bash
rg -n '#!/usr/bin/python' \
    | awk -F: "/:1:/ { print \"sed -i '0,/python/{s/python/env -S python/}' \"\$1; }" \
    | xargs -0 -- sh -c
```

```
Ran 395 tests in 226.332s

OK


Summary:

| Version    | Status
|--------------------
| 2.7        | PASS
| 3.6        | PASS
| 3.7        | PASS
| 3.8        | PASS
| 3.9        | PASS
```
:sunglasses:

### DIFF
--- a/bin/archive-conf
+++ b/bin/archive-conf
@@ -1,4 +1,4 @@
-#!/usr/bin/python -b
+#!/usr/bin/env -S python -b
 # Copyright 1999-2014 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 
--- a/bin/binhost-snapshot
+++ b/bin/binhost-snapshot
@@ -1,4 +1,4 @@
-#!/usr/bin/python -b
+#!/usr/bin/env -S python -b
 # Copyright 2010-2014 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 
--- a/bin/check-implicit-pointer-usage.py
+++ b/bin/check-implicit-pointer-usage.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python -b
+#!/usr/bin/env -S python -b
 
 # Ripped from HP and updated from Debian
 # Update by Gentoo to support unicode output
--- a/bin/chmod-lite.py
+++ b/bin/chmod-lite.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python -b
+#!/usr/bin/env -S python -b
 # Copyright 2015 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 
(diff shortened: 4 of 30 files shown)


bin/archive-conf
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python -b
+#!/usr/bin/env -S python -b
```
mgorny, line 1: I'm pretty sure `env -S` is not very portable.

R030t1: If you expect portage to only run on systems where python is at a fixed location then the further assumption that `env` supports `-S` is inconsequential. Overall this should decrease the amount of work to support POSIX-compliant systems that are not Gentoo.

I spent some time trying to determine how portable this was. I do see manual pages that do not show `-S`, and sometimes invoking `env` with additional arguments behaves as if `-S` was passed. But keep in mind there are already portability problems, e.g. a system may lack `xz`.

There is the additonal concern of what the most common flag, `-b`, accomplishes; to me it seems more appropriate to leave this to a linter but I have no full replacement ready.

BoleynSu: Critical scripts for the system would better have a fixed path. Maybe one reason is for security, i.e., using env may lead to arbitrary executable to be run. I know Fedora does the same for python scripts.

R030t1: I understand your concern but if you can't trust `env` to be secure then it seems like you also can't trust `python` to be secure. The solution would be to use verity like Android does.

BoleynSu: You can trust /usr/bin/env, but you may not trust your environment. For example, I have a evil executable named `python` in $PATH that will run `rm -rf /something_important` instead of being a python interpreter. Then when I call emerge, bad things will happen. By using the fixed path, we can be sure /usr/bin/python will always be called and we can trust it because making changing to /usr/bin/python requires root privilege. That is, our thread model is that the attacker won't have root privilege. This is a very common thread model.

mgorny: No, this is security by obscurity. You kill valid use cases because something bad might happen. In the end, bad things happen because of silly tinfoil assumptions hardcoded somewhere.

BoleynSu: Quote from https://blogs.gnome.org/mcatanzaro/2018/02/16/on-python-shebangs/.

The first option seems to work in all environments, but it is banned in popular distros like Fedora (and I believe also Debian, but I can’t find a reference for this). Using env in shebangs is dangerous because it can result in system packages using non-system versions of python. python is used in so many places throughout modern systems, it’s not hard to see how using #!/usr/bin/env in an important package could badly bork users’ operating systems if they install a custom version of python in /usr/local. Don’t do this.

PR #565, opened 2020-07-03
ForkProcess: replace os.fork with multiprocessing.Process (bug 730192)

zmedico:
Replace os.fork with multiprocessing.Process, in order to leverage
any pre-fork and post-fork interpreter housekeeping that it provides,
promoting a healthy state for the forked interpreter.

Since multiprocessing.Process closes `sys.__stdin__`, fix relevant
code to use the portage._get_stdin() compatibility function.
In case there's a legitimate need to inherit stdin for things like
PROPERTIES=interactive support, create a temporary duplicate of
fd_pipes[0] when appropriate, and restore sys.stdin and `sys.__stdin__`
in the subprocess.

Bug: https://bugs.gentoo.org/730192

### DIFF
--- a/lib/portage/util/_async/ForkProcess.py
+++ b/lib/portage/util/_async/ForkProcess.py
@@ -1,37 +1,123 @@
-# Copyright 2012-2013 Gentoo Foundation
+# Copyright 2012-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+import fcntl
+import functools
+import multiprocessing
 import signal
 import sys
-import traceback
 
 import portage
 from portage import os
+from portage.util.futures import asyncio
+from portage.util.futures.compat_coroutine import coroutine
 from _emerge.SpawnProcess import SpawnProcess
 
 class ForkProcess(SpawnProcess):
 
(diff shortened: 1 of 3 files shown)


lib/portage/util/_async/ForkProcess.py
```diff
@@ -11,27 +13,100 @@
...
+		# access to the sentinel value should raise ValueError if the
+		# sentinel has been closed. In this case it's not safe to call
+		# remove_reader, since file descriptor has been closed and then
+		# reallocated to a concurrent coroutine. After proc is closed,
+		# most method calls and attribute accesses will raise ValueError,
+		# so we assume that it is ready to close here and that no other
+		# code will ever close it (the sentinel provides the only means
+		# to predict when it is safe to close).
+		self.scheduler.remove_reader(proc.sentinel)
+		proc.join()
```
zmedico, line 95: This blocking `proc.join()` is not optimal. We can add a timeout 0 argument, and retry if necessary.

zmedico: Added retry via `_proc_join` method.

PR #570, opened 2020-07-15
Eliminate basestring/long/_unicode py3 compat

mgorny:
Replace basestring and _unicode hacks with str, and long with int.

### DIFF
--- a/lib/portage/proxy/lazyimport.py
+++ b/lib/portage/proxy/lazyimport.py
@@ -1,4 +1,4 @@
-# Copyright 2009-2014 Gentoo Foundation
+# Copyright 2009-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 __all__ = ['lazyimport']
@@ -13,9 +13,6 @@ except ImportError:
 
 from portage.proxy.objectproxy import ObjectProxy
 
-if sys.hexversion >= 0x3000000:
-	# pylint: disable=W0622
-	basestring = str
 
 _module_proxies = {}
 _module_proxies_lock = threading.RLock()
@@ -169,7 +166,7 @@ def lazyimport(scope, *args):
 		if len(parts) == 1:
 			name = s
 
-			if not name or not isinstance(name, basestring):
+			if not name or not isinstance(name, str):
 				raise ValueError(name)
 
 			components = name.split('.')
--- a/bin/egencache
+++ b/bin/egencache
@@ -75,9 +75,6 @@ else:
 	else:
 		from portage.xml.metadata import parse_metadata_use
 
-if sys.hexversion >= 0x3000000:
-	# pylint: disable=W0622
-	long = int
 
 def parse_args(args):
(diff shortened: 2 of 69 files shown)


lib/portage/proxy/lazyimport.py
```diff
@@ -169,7 +166,7 @@ def lazyimport(scope, *args):
 		if len(parts) == 1:
 			name = s
 
-			if not name or not isinstance(name, basestring):
+			if not name or not isinstance(name, str):
 				raise ValueError(name)
```
zmedico, line 170: The travis test failures in lib/portage/tests/bin/test_doins.py are somehow triggered by the lazyimport.py change. This makes the issue more visible:
```diff
diff --git a/lib/portage/tests/bin/setup_env.py b/lib/portage/tests/bin/setup_env.py
index 33b167989..e1fcb271c 100644
--- a/lib/portage/tests/bin/setup_env.py
+++ b/lib/portage/tests/bin/setup_env.py
@@ -64,7 +64,8 @@ def portage_func(func, args, exit_status=0):
        # just their exit value and the state of $D
        global env
        f = open('/dev/null', 'wb')
-       fd_pipes = {0:0,1:f.fileno(),2:f.fileno()}
+       #fd_pipes = {0:0,1:f.fileno(),2:f.fileno()}
+       fd_pipes = {0:0, 1:2, 2:2}
        def pre_exec():
                os.chdir(env['S'])
        spawn([func] + args.split(), env=env,
```
Somehow, `isinstance(name, str)` returns False, causing it to raise ValueError like this:
```
testDoDirFallback (portage.tests.bin.test_doins.DoIns)
Tests with DIROPTIONS which doins.py doesn't understand. ... Traceback (most recent call last):
  File "/home/travis/build/gentoo/portage/build/bin/doins.py", line 30, in <module>
    from portage.util import movefile
  File "/home/travis/build/gentoo/portage/build/lib.linux-x86_64-3.6/portage/__init__.py", line 112, in <module>
    'time',
  File "/home/travis/build/gentoo/portage/build/lib.linux-x86_64-3.6/portage/proxy/lazyimport.py", line 170, in lazyimport
    raise ValueError(name)
ValueError: portage.checksum
doins failed
```

zmedico: I've patched some debug info into the ValueError and found `ValueError: expected type str, got unicode: u'portage.checksum'`.

zmedico: Fixed by removing all instances of `from __future__ import unicode_literals`.

zmedico: See https://github.com/gentoo/portage/pull/572.

mgorny:
Thanks, I'll do one more run. I wonder why it's happening on Travis and not for us. Is Ubuntu doing some custom patching, or shipping an old broken Python version?

zmedico:
It must be that the test executed doins.py with the wrong interpreter (/usr/bin/python rather than PORTAGE_PYTHON):

https://github.com/gentoo/portage/pull/573

PR #582, opened 2020-07-22
Resolve pylint R0205 in portage.

antarus12345:
Resolve pylint warning R0205. Its a py3k migration warning and we did a bunch of py3k refactoring recently.

R0205 basically says to not inherit from object in python3. Since we don't support python2 anymore; we don't need this inherit.

Also drop 2.7 from ./runtests as the 2.7 target will not pass after this change.


zmedico:
The commits weren't signed, so I've pushed signed commits in 41f5049b14beb11e6ae768c8571e7a8a326f701c.

PR #587, opened 2020-07-24
Fix redefined-builtin errors for pylint.

antarus12345:
Some of these are simple variable renames. A few of the uses were
refactored and some were simply disabled; often due to redefinition as
part of a function signature. I did not do any research in terms of API
changes so these were typically left as-is to avoid breaking callers.

Signed-off-by: Alec Warner <antarus@gentoo.org>


zmedico:
Thanks!

PR #589, opened 2020-07-24
Remove fallback getloadavg impl.

antarus12345:
I suspect this impl made sense when portage used python2.2 (which lacked it) but this function was added to the unix stdlib in python 2.3; so I think we can drop this now.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -1,0 +1,464 @@
+[MASTER]
+
+# Disable the message, report, category or checker with the given id(s). You
+# can either give multiple identifiers separated by comma (,) or put this
+# option multiple times (only on the command line, not in the configuration
+# file where it should appear only once). You can also use "--/
+# disable=all" to
+# disable everything first and then reenable specific checks. For example, if
+# you want to run only the similarities checker, you can use "--disable=all
+# --enable=similarities". If you want to run only the classes checker, but have
+# no Warning level messages displayed, use "--disable=all --enable=classes
+# --disable=W".
+#disable=no-absolute-import,bad-continuation,C0103,C0114,C0115,E1101,W0201,no-name-in-module
+disable=all
+enable=redefined-builtin,useless-object-inheritance,trailing-newlines
+
+# A comma-separated list of package or module names from where C extensions may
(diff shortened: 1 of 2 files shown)


zmedico:
As I recall, the getloadavg function was added in 8607a39dc9d979713f9434839ed1aef2602276a1 due to the function being missing for python built with uclibc back then. I've found that this is still an issue with current autobuilds found in http://ftp.romnet.org/gentoo/releases/amd64/autobuilds/current-stage3-amd64-uclibc-vanilla/, for example this is with stage3-amd64-uclibc-vanilla-20200619.tar.bz2:
```
Python 3.7.7 (default, May 31 2020, 21:22:59) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.getloadavg
  File "<stdin>", line 1, in <module>
AttributeError: module 'os' has no attribute 'getloadavg'
```

pylintrc
```diff
@@ -10,9 +10,9 @@
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use "--disable=all --enable=classes
 # --disable=W".
-#disable=no-absolute-import,bad-continuation,C0103,C0114,C0115,E1101,W0201,no-name-in-module
-disable=all
-enable=redefined-builtin,useless-object-inheritance,trailing-newlines
+disable=no-absolute-import,bad-continuation,C0103,C0114,C0115,E1101,W0201,no-name-in-module,W0613,C0116
+#disable=all
+#enable=redefined-builtin,useless-object-inheritance,trailing-newlines
```
zmedico, line 15: Maybe we should add missing-docstring in there because with pylint 2.3.1 I see a lot of errors like this:
```
************* Module portage.tests.util.test_file_copier
lib/portage/tests/util/test_file_copier.py:1:0: C0111: Missing module docstring (missing-docstring)
lib/portage/tests/util/test_file_copier.py:15:0: C0111: Missing class docstring (missing-docstring)
```

zmedico:
We're reconsidering this PR in https://github.com/gentoo/portage/pull/1383.

PR #593, opened 2020-07-29
Update pylint configuration

antarus12345:
Expose more warnings by using default confidence intervals.
    Change output format to be parseable.
    Describe awk command to enumerate and sort disables.
    
    Disable a bunch of warnings that have > 100 occurances.
    The rest I will burn down with more PRs.


zmedico:
We've got [pylint in tox.ini](https://github.com/gentoo/portage/pull/598) now, so you need to test your pylintrc changes with that.

PR #597, opened 2020-08-04
rsync and gpg behind proxy

fepitre:
To be merged with https://github.com/mgorny/gemato/pull/13


fepitre:
To be changed for another approach.

PR #607, opened 2020-08-17
rsync and gpg behind proxy v2

fepitre:
Pass directly needed variable to env for Gemato is enough to fetch keys behin proxy.

@mgorny: can you review this please?

### DIFF
--- a/lib/portage/dbapi/bintree.py
+++ b/lib/portage/dbapi/bintree.py
@@ -964,11 +964,17 @@ class binarytree:
 						download_timestamp + ttl > time.time():
 						raise UseCachedCopyOfRemoteIndex()
 
+				# Set proxy settings for _urlopen -> urllib_request
+				for setting in ('http_proxy', 'https_proxy'):
+					proxies = {}
+					if self.settings.get(setting, None):
+						proxies[setting.replace('_proxy', '')] = self.settings.get(setting)
+
 				# Don't use urlopen for https, unless
 				# PEP 476 is supported (bug #469888).
 				if parsed_url.scheme not in ('https',) or _have_pep_476():
 					try:
-						f = _urlopen(url, if_modified_since=local_timestamp)
+						f = _urlopen(url, if_modified_since=local_timestamp, proxies=proxies)
 						if hasattr(f, 'headers') and f.headers.get('timestamp', ''):
 							remote_timestamp = f.headers.get('timestamp')
 					except IOError as err:
@@ -1686,7 +1692,7 @@ class binarytree:
 			resume = True
 			writemsg(_("Resuming download of this tbz2, but it is possible that it is corrupt.\n"),
 				noiselevel=-1)
-		
+
 		mydest = os.path.dirname(self.getname(pkgname))
 		self._ensure_dir(mydest)
 		# urljoin doesn't work correctly with unrecognized protocols like sftp
--- a/lib/portage/dbapi/bintree.py
+++ b/lib/portage/dbapi/bintree.py
@@ -965,10 +965,11 @@ class binarytree:
 						raise UseCachedCopyOfRemoteIndex()
 
 				# Set proxy settings for _urlopen -> urllib_request
-				for setting in ('http_proxy', 'https_proxy'):
+				for proto in ('http', 'https'):
 					proxies = {}
-					if self.settings.get(setting, None):
-						proxies[setting.replace('_proxy', '')] = self.settings.get(setting)
+					value = self.settings.get(proto + '_proxy')
+					if value is not None:
+						proxies[proto] = value
 
 				# Don't use urlopen for https, unless
 				# PEP 476 is supported (bug #469888).
--- a/lib/portage/sync/modules/git/git.py
+++ b/lib/portage/sync/modules/git/git.py
@@ -207,19 +207,17 @@ class GitSync(NewBase):
 				'sync-git-verify-commit-signature', 'false').lower() not in ('true', 'yes')):
 			return True
 
-		if self.repo.sync_openpgp_key_path is not None:
-			if gemato is None:
-				writemsg_level("!!! Verifying against specified key requires gemato-11.0+ installed\n",
+		if self.repo.sync_openpgp_key_path is not None and gemato is None:
+			writemsg_level("!!! Verifying against specified key requires gemato-11.0+ installed\n",
 					level=logging.ERROR, noiselevel=-1)
-				return False
-			openpgp_env = self._get_openpgp_env()
-		else:
-			openpgp_env = None
+			return False
+
+		openpgp_env = self._get_openpgp_env(self.repo.sync_openpgp_key_path)
 
 		try:
 			out = EOutput()
 			env = None
-			if openpgp_env is not None:
+			if openpgp_env is not None and self.repo.sync_openpgp_key_path is not None:
 				try:
 					out.einfo('Using keys from %s' % (self.repo.sync_openpgp_key_path,))
 					with io.open(self.repo.sync_openpgp_key_path, 'rb') as f:
--- a/lib/portage/sync/modules/git/git.py
+++ b/lib/portage/sync/modules/git/git.py
@@ -212,7 +212,7 @@ class GitSync(NewBase):
 				writemsg_level("!!! Verifying against specified key requires gemato-11.0+ installed\n",
 					level=logging.ERROR, noiselevel=-1)
 				return False
-			openpgp_env = gemato.openpgp.OpenPGPEnvironment()
+			openpgp_env = self._get_openpgp_env()
 		else:
 			openpgp_env = None
 
--- a/lib/portage/sync/modules/rsync/rsync.py
+++ b/lib/portage/sync/modules/rsync/rsync.py
@@ -227,10 +227,7 @@ class RsyncSync(NewBase):
 			addrinfos = None
 			uris = []
 
-			rsync_proxy_conf = \
-				self.options['emerge_config'].running_config.settings.get(
-					'RSYNC_PROXY', None)
-			if not rsync_proxy_conf:
+			if 'RSYNC_PROXY' not in self.spawn_kwargs['env']:
 				try:
 					addrinfos = getaddrinfo_validate(
 						socket.getaddrinfo(
--- a/lib/portage/sync/modules/rsync/rsync.py
+++ b/lib/portage/sync/modules/rsync/rsync.py
@@ -230,15 +230,19 @@ class RsyncSync(NewBase):
 			addrinfos = None
 			uris = []
 
-			try:
-				addrinfos = getaddrinfo_validate(
-					socket.getaddrinfo(getaddrinfo_host, None,
-					family, socket.SOCK_STREAM))
-			except socket.error as e:
-				writemsg_level(
-					"!!! getaddrinfo failed for '%s': %s\n"
-					% (_unicode_decode(hostname), str(e)),
-					noiselevel=-1, level=logging.ERROR)
+			rsync_proxy_conf = \
+				self.options['emerge_config'].running_config.settings.get(
+					'RSYNC_PROXY', None)
+			if not rsync_proxy_conf:
+				try:
+					addrinfos = getaddrinfo_validate(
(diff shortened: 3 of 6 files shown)


lib/portage/sync/syncbase.py
```diff
@@ -252,6 +252,12 @@ def _refresh_keys(self, openpgp_env):
 		"""
 		out = portage.output.EOutput(quiet=('--quiet' in self.options['emerge_config'].opts))
 
+		# Pass proxy settings in emerge configuration to Gemato ephemeral env
+		running_config_settings = self.options['emerge_config'].running_config.settings
+		for setting in ('http_proxy', 'https_proxy'):
```
mgorny, line 257: Does this grab values from environment if not set in make.conf too?

lib/portage/sync/syncbase.py
```diff
@@ -252,6 +252,12 @@ def _refresh_keys(self, openpgp_env):
...
+			if running_config_settings.get(setting, None):
+				os.environ[setting] = running_config_settings.get(setting)
```
mgorny, line 259: I've been thinking about it. If we don't need a version with this stable very fast, I'll add a `proxy=` parameter to key refresh functions. This will be better for when I replace GPG with a pure Python implementation.

lib/portage/sync/syncbase.py
fepitre: Not at all. I preferred to keep explicit declaration of such parameters trough only the configuration file. Depending on the global portage policy, that could be adapted to use current env values, then if exists, the configuration ones.

lib/portage/sync/syncbase.py
fepitre: I would vote at first to have a non-blocking case like currently with this PR but that depends also on you and your time for doing your modifications. In both cases, thank you for advancing the subject.

fepitre:
@mgorny: I've adapted the code in accordance to your newer changes. Also, I've setup default use of global proxy setting if exists prior to portage conf. That can be discussed of course.

lib/portage/sync/modules/rsync/rsync.py
```diff
@@ -230,15 +230,19 @@ def update(self):
...
-					socket.getaddrinfo(getaddrinfo_host, None,
-					family, socket.SOCK_STREAM))
-			except socket.error as e:
-				writemsg_level(
-					"!!! getaddrinfo failed for '%s': %s\n"
-					% (_unicode_decode(hostname), str(e)),
-					noiselevel=-1, level=logging.ERROR)
+			rsync_proxy_conf = \
+				self.options['emerge_config'].running_config.settings.get(
+					'RSYNC_PROXY', None)
```
zmedico, line 235: The environment used to spawn rsync is `self.spawn_kwargs['env']`, so that's where the real RSYNC_PROXY value will come from.

lib/portage/sync/syncbase.py
```diff
@@ -252,6 +252,15 @@ def _refresh_keys(self, openpgp_env):
 		"""
 		out = portage.output.EOutput(quiet=('--quiet' in self.options['emerge_config'].opts))
 
+		# Set global proxy setting to Gemato ephemeral env
+		if 'http_proxy' in os.environ:
+			openpgp_env.proxy = os.environ['http_proxy']
```
mgorny, line 257: This isn't guaranteed to work. You need to pass `proxy` to the constructor, otherwise `dirmngr` may be started earlier and without your proxy.

lib/portage/sync/modules/rsync/rsync.py
fepitre: From my humble comprehension and tests I've made, here the issue for getting addr info is that your socket does not handle proxy settings. I've made a version handling proxy setting for `socket` but it's rather hacky. I use `socks.setdefaultproxy` and passing `socks.socksocket` to `socket.socket`. Then, I've made it working. But finally, failing on `getaddrinfo` is not failing whole `rsync`. Else, I'm missing some feature of `socket` library handling system-wide proxy.

lib/portage/sync/syncbase.py
fepitre: I'll adjust in the constructor then. Even if each tests I've ran are working maybe some configurations will not.

lib/portage/sync/modules/git/git.py
```diff
@@ -212,7 +212,19 @@ def verify_head(self, revision='-1'):
 				writemsg_level("!!! Verifying against specified key requires gemato-11.0+ installed\n",
 					level=logging.ERROR, noiselevel=-1)
 				return False
-			openpgp_env = gemato.openpgp.OpenPGPEnvironment()
+
+			proxy = None
+			# Set global proxy setting to Gemato ephemeral env
+			if 'http_proxy' in os.environ:
+				proxy = os.environ['http_proxy']
```
fepitre, line 219: @zmedico: should this be feed from `os.environ` or `spawn_kwargs`? I'm not sure to see if `spawn_kwargs` is initially feed by `os.environ`.

lib/portage/sync/modules/git/git.py
```diff
@@ -212,7 +212,19 @@ def verify_head(self, revision='-1'):
...
+
+			# Override global proxy setting with one provided in emerge configuration
+			running_config_settings = \
+				self.options['emerge_config'].running_config.settings
+			if 'http_proxy' in running_config_settings:
+				proxy = running_config_settings['http_proxy']
+
+			openpgp_env = gemato.openpgp.OpenPGPEnvironment(proxy=proxy)
```
fepitre, line 227: @mgorny: it's duplicated code amount the three classes. Normally that should be good?

mgorny: It's not that bad but maybe a helper method in base class could help.

lib/portage/sync/syncbase.py
```diff
@@ -303,6 +294,21 @@ def noisy_refresh_keys():
 				loop.run_until_complete(decorated_func())
 		out.eend(0)
 
+	def _set_proxy(self, openpgp_env):
```
fepitre, line 297: @mgorny what about doing this way?

lib/portage/dbapi/bintree.py
```diff
@@ -964,6 +964,11 @@ def _populate_remote(self, getbinpkg_refresh=True):
 						download_timestamp + ttl > time.time():
 						raise UseCachedCopyOfRemoteIndex()
 
+				# Set proxy settings for _urlopen -> urllib_request
```
fepitre, line 967: @mgorny, @zmedico : I've also solved an issue with BINPKG host set and proxy set.

mgorny: I suppose it would be cleaner to pass it through `_urlopen` explicitly.

lib/portage/sync/syncbase.py
mgorny: I've just told you that you're not supposed to touch the attribute on `OpenPGPEnvironment`, so rather some method that creates `OpenPGPEnvironment`. They may be some more stuff worth putting into the common method.

fepitre: > I've just told you that you're not supposed to touch the attribute on `OpenPGPEnvironment`, 

Sorry but I've probably missed that part in the understanding of 'your sentence of helper function in base class'. Also, as `proxy` is not a private attribute that is not clear. I'll redo something in this sense then.

lib/portage/sync/syncbase.py
```diff
@@ -303,6 +300,26 @@ def noisy_refresh_keys():
 				loop.run_until_complete(decorated_func())
 		out.eend(0)
 
+	def _get_openpgp_env(self, system=None):
```
fepitre, line 303: @mgorny what about this new attempt?

lib/portage/dbapi/bintree.py
fepitre: I pass proxy settings directly as parameter and parsed for the `ProxyHandler`.

lib/portage/dbapi/bintree.py
```diff
@@ -964,11 +964,17 @@ def _populate_remote(self, getbinpkg_refresh=True):
...
+				for setting in ('http_proxy', 'https_proxy'):
+					proxies = {}
+					if self.settings.get(setting, None):
+						proxies[setting.replace('_proxy', '')] = self.settings.get(setting)
```
mgorny, line 971: I think it will be more straightforward to:

```suggestion
				for proto in ('http', 'https'):
					proxies = {}
					value = self.settings.get(proto + '_proxy')
					if value is not None:
						proxies[proto] = value
```

I suppose this could also use list comprehension.

lib/portage/sync/syncbase.py
```diff
@@ -294,6 +300,26 @@ def noisy_refresh_keys():
...
+		if gemato:
+			proxy = None
+			# Set global proxy setting to Gemato ephemeral env
+			if 'http_proxy' in os.environ:
+				proxy = os.environ['http_proxy']
```
mgorny, line 308:
```suggestion
			proxy = os.environ.get('http_proxy')
```

lib/portage/sync/syncbase.py
mgorny: Instead of `system=`, you could pass the path to the key file and return system env when `None` (returning `None` vs system env is not very important). You could also put key loading and maybe refresh in the common function since it seems to be the same everywhere.

lib/portage/sync/modules/git/git.py
```diff
@@ -207,19 +207,17 @@ def verify_head(self, revision='-1'):
...
-			openpgp_env = None
+			return False
+
+		openpgp_env = self._get_openpgp_env(self.repo.sync_openpgp_key_path)
 
 		try:
 			out = EOutput()
 			env = None
-			if openpgp_env is not None:
+			if openpgp_env is not None and self.repo.sync_openpgp_key_path is not None:
```
fepitre, line 220: @mgorny I handled the case here where `openpgp_env` was `None` by checking if key is provided. By default the common function returns the system env if the key is not provided as you suggested. I don't see any other corner case where a `None` value would impact something. Do you?

lib/portage/sync/syncbase.py
```diff
@@ -294,6 +300,24 @@ def noisy_refresh_keys():
 				loop.run_until_complete(decorated_func())
 		out.eend(0)
 
+	def _get_openpgp_env(self, openpgp_key_path=None):
+		if gemato:
```
mgorny, line 304: I would prefer an explicit:

```suggestion
		if gemato is not None:
```

lib/portage/sync/syncbase.py
```diff
@@ -294,6 +300,24 @@ def noisy_refresh_keys():
...
+			# Set global proxy setting to Gemato ephemeral env
+			proxy = os.environ.get('http_proxy')
+
+			# Override global proxy setting with one provided in emerge configuration
+			running_config_settings = \
+				self.options['emerge_config'].running_config.settings
```
mgorny, line 310:
```suggestion
			running_config_settings = (
				self.options['emerge_config'].running_config.settings)
```

mgorny (commented):
Makes sense to me. I would still prefer moving more code to the helper function but I can do that once this is merged.

Time for a few style nitpicks.

fepitre:
> Makes sense to me. I would still prefer moving more code to the helper function but I can do that once this is merged.

Yes indeed. I'm seeing some part related to checking if  `gemato` is here. Probably we could do this only in the common function and raise something if key is provided. Anyway, I would let you do this as you know the project better than me.

fepitre:
@mgorny @zmedico are we good now for this PR or there is still something to modify/improve for this first proxy support?

fepitre:
@zmedico thank to new version of Gemato, there is no more issue related to the one previously explained. WKD checks is working without adding the port and host entry. From my side, I'm having another repository where the check GPG passed with refreshing key on keyserver and the current state of this PR.

zmedico:
All merged in https://github.com/gentoo/portage/compare/b851d7f8f3c451d4dd4863580e75461c511b44d0...ddbe8bd019552573b3f9c1ef2e5701df2edb4dd6. Thanks!

PR #612, opened 2020-09-01
emerge --search: auto-detect regular expressions (bug 737480)

zmedico:
Automatically detect regular expressions when the search string
contains any of these regular expression characters or character
sequences:

     ^ $ * [ ] { } | ? .+

This simplifies usage, so that users no longer have to remember
to prefix regular expressions with the % character. The new
behavior can be disabled by --regex-search-auto=n, in case the
regular expressions interpretation causes some kind of problem.

Note that fuzzy search and regular expression search are
mutually exclusive, and fuzzy search remains the default for
search strings that do not contain any regular expression
characters.

Bug: https://bugs.gentoo.org/737480


Arfrever:
Maybe also `.` (any single character) and `+` (one or more instances of previous character)?

zmedico:
> Maybe also `.` (any single character) and `+` (one or more instances of previous character)?

Yes, sounds good. Updated.

PR #625, opened 2020-09-20
emerge: enable parallel-fetch during pkg_pretend (bug 710432)

zmedico:
Execute pkg_pretend phases in a coroutine while parallel-fetch
is running concurrently. When it's time to execute the pkg_pretend
phase for a remote binary package, use a Scheduler _get_prefetcher
method to get a running prefetcher if available, and otherwise
start a new fetcher.

Since pkg_pretend phases now run inside of the --keep-going retry
loop, --keep-going is now able to recover from pkg_pretend
failures, which fixes bug 404157.

Bug: https://bugs.gentoo.org/404157
Bug: https://bugs.gentoo.org/710432

### DIFF
--- a/lib/_emerge/Scheduler.py
+++ b/lib/_emerge/Scheduler.py
@@ -25,6 +25,7 @@ from portage._sets import SETPREFIX
 from portage._sets.base import InternalPackageSet
 from portage.util import ensure_dirs, writemsg, writemsg_level
 from portage.util.futures import asyncio
+from portage.util.futures.compat_coroutine import coroutine, coroutine_return
 from portage.util.SlotObject import SlotObject
 from portage.util._async.SchedulerInterface import SchedulerInterface
 from portage.package.ebuild.digestcheck import digestcheck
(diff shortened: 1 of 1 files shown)


lib/_emerge/Scheduler.py
```diff
@@ -1390,6 +1393,22 @@ def _merge(self):
...
+			try:
+				rval = self._sched_iface.run_until_complete(self._run_pkg_pretend(loop=self._sched_iface))
+			except asyncio.CancelledError:
+				self.terminate()
+				rval = None
+			self._termination_check()
+			if self._terminated_tasks:
+				rval = 128 + signal.SIGINT
+			if rval != os.EX_OK:
+				return rval
```
zmedico, line 1410: We need to make sure that --keep-going will behave appropriately when it fails here (and this will fix [bug 404157](https://bugs.gentoo.org/404157)).

zmedico: I've fixed it to add failed packages to self._failed_pkgs, which --keep-going uses to drop those packages from the merge list.

PR #627, opened 2020-09-30
movefile: unlink before copying a large file

BoleynSu:
The free space on rootfs may be too small that it prevents `emerge` from working.
For example, `cp /tmp/100m#new /100m` will fail when we have 10m left on rootfs.
Note that this patch will not help if the large file is opened at the moment of unlinking.


BoleynSu:
Has sent to the mail list.

PR #628, opened 2020-10-03
process: pass a string to socket.sethostname instead of bytes

sbraz:
Recent PyPy3 versions expect a string instead of a bytes object, CPython seems to be happy with both.

Bug: https://bugs.gentoo.org/716998

### DIFF
--- a/lib/portage/process.py
+++ b/lib/portage/process.py
@@ -706,10 +706,10 @@ def _exec(binary, mycommand, opt_name, fd_pipes,
 						if unshare_net:
 							# use 'localhost' to avoid hostname resolution problems
 							try:
-								# pypy3 does not implement socket.sethostname()
+								# Old pypy3 versions do not implement socket.sethostname()
 								new_hostname = b'localhost'
 								if hasattr(socket, 'sethostname'):
-									socket.sethostname(new_hostname)
+									socket.sethostname(new_hostname.decode("ascii"))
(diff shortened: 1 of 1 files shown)


mgorny:
For the record, I'm planning to look into pypy3 when I have some time.

lib/portage/process.py
```diff
@@ -706,10 +706,10 @@ def _exec(binary, mycommand, opt_name, fd_pipes,
 						if unshare_net:
 							# use 'localhost' to avoid hostname resolution problems
 							try:
-								# pypy3 does not implement socket.sethostname()
+								# Old pypy3 versions do not implement socket.sethostname()
 								new_hostname = b'localhost'
 								if hasattr(socket, 'sethostname'):
-									socket.sethostname(new_hostname)
+									socket.sethostname(new_hostname.decode("ascii"))
```
zmedico, line 712: Let's use `portage._native_string(new_hostname)`, since that's our exiting convention for anything that requires a native string.

sbraz: I think we should wait for @mgorny. Apparently his PyPy patch was wrong and PyPy should take bytes as an argument too.

sbraz:
Fixed in PyPy3 and backported as gentoo/gentoo@f0e786f0096b75d886fca041c479e936fb27e0b9, no need to change portage.

PR #629, opened 2020-10-14
repoman: fix file.size-fatal to be 20 KiB

juippis:
To obey devmanual's guidance, and in addition to match pkgcheck, I'd like repoman's file size check to report any over 20KiB extra file as fatal. There aren't many of those left in the tree, but updating this check surely helps getting rid of the last ones.
Tracker bug: https://bugs.gentoo.org/748144

### DIFF
--- a/repoman/cnf/qa_data/qa_data.yaml
+++ b/repoman/cnf/qa_data/qa_data.yaml
@@ -60,7 +60,6 @@ qahelp:
     file:
         executable: "Ebuilds, digests, metadata.xml, Manifest, and ChangeLog do not need the executable bit"
         size: "Files in the files directory must be under 20 KiB"
-        size-fatal: "Files in the files directory must be under 60 KiB"
         empty: "Empty file in the files directory"
         name: "File/dir name must be composed of only the following chars: %s "
         UTF8: "File is not UTF8 compliant"
--- a/repoman/cnf/repository/qa_data.yaml
+++ b/repoman/cnf/repository/qa_data.yaml
@@ -59,7 +59,6 @@ qawarnings:
     - ebuild.badheader
     - ebuild.patches
     - file.empty
-    - file.size
     - HOMEPAGE.virtual
     - inherit.unused
     - inherit.deprecated
--- a/repoman/lib/repoman/modules/scan/fetch/fetches.py
+++ b/repoman/lib/repoman/modules/scan/fetch/fetches.py
@@ -120,12 +120,7 @@ class FetchChecks(ScanBase):
 							continue
 						filesdirlist.append(y + "/" + z)
 				# Current policy is no files over 20 KiB, these are the checks.
-				# File size between 20 KiB and 60 KiB causes a warning,
-				# while file size over 60 KiB causes an error.
-				elif mystat.st_size > 61440:
-					self.qatracker.add_error(
-						"file.size-fatal", "(%d KiB) %s/files/%s" % (
-							mystat.st_size // 1024, xpkg, y))
+				# File size over 20 KiB causes an error.
 				elif mystat.st_size > 20480:
 					self.qatracker.add_error(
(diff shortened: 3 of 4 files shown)


mgorny:
For the record, I don't think this check is very valuable. The past experiences regarding enforcing it point out that:

1. Most devs don't care, space is cheap, etc., why should we bother that every ::gentoo checkout is full of random cruft most of the users never use.
2. Per-package limit may make more sense than per-file limit, esp. that the latter is trivially avoided by splitting patch into smaller parts while technically few smaller files are worse than one big file.

juippis:
This is to raise awareness, it can't fix attitudes. `--force` will still let one commit if they think this check is useless. However looking at the amount of current issues, I'd like to believe in the good intentions of everyone.

This'll work until you work out the "per-package" limit since that's going to take time to document and implement.

zmedico:
> This is to raise awareness, it can't fix attitudes. `--force` will still let one commit if they think this check is useless. However looking at the amount of current issues, I'd like to believe in the good intentions of everyone.

They'd have to use `--force` for every commit to the offending package though, which would be absurd IMO. I agree that we should implement the hard 20k file.size-fatal limit under the condition that we first open bug reports for the existing offenders in order to get them fixed:
```
$ find */*/files -size +20k
app-i18n/libchewing/files/libchewing-0.5.1-autoconf-archive-2019.01.06.patch
app-i18n/mozc/files/mozc-2.23.2815.102-protobuf_generated_classes_no_inheritance.patch
app-i18n/scim-anthy/files/scim-anthy-1.2.7-gtk2_build.patch
app-text/fbreader/files/fbreader-0.99.4-combined.patch
app-text/gnome-doc-utils/files/gnome-doc-utils-0.20.10-python3.patch
dev-libs/libpreludedb/files/libpreludedb-5.1.0-update_m4_postgresql.patch
dev-libs/ppl/files/disable-boeing-tests.patch
dev-libs/xerces-c/files/xerces-c-3.2.2-fix-XERCESC-2161.patch
dev-php/pecl-ssh2/files/pecl-ssh2-1.1.2-fix-php7-port.patch
dev-ros/pcl_ros/files/pcl111.patch
games-emulation/advancescan/files/advancescan-1.16-gcc6.patch
media-libs/freeimage/files/freeimage-3.18.0-unbundling.patch
media-plugins/gst-plugins-lv2/files/patch-1.16.2-gcc10.patch
media-tv/mythtv/files/mythtv-30.0_p20190808-cast_constants_to_short.patch
media-video/mplayer/files/mplayer-1.3-ffmpeg4.patch
media-video/transcode/files/ffmpeg4.patch
net-libs/dleyna-renderer/files/0.6.0-gupnp-1.2.patch
net-vpn/ipsec-tools/files/ipsec-tools-add-openssl-1.1.x-support.patch
sci-chemistry/vmd/files/vmd-1.9.4-gentoo-plugins.patch
sys-auth/pam_ssh_agent_auth/files/pam_ssh_agent_auth-0.10.3-openssl-1.1.1.patch
sys-cluster/torque/files/torque-6.0.4-gcc7.patch
sys-devel/clang/files/10.0.1/0003-clang-tools-extra-Prevent-linking-to-duplicate-.a-li.patch
sys-fs/dislocker/files/dislocker-0.7.1-fix-find-ruby.patch
```

juippis:
It's already done, and the discussion in one of the related bugs drove me here to make the change visible in repoman. Please see the tracker bug I already linked, 
https://bugs.gentoo.org/748144
which should have links to every package you listed. I can run my sizeviolation.sh script to report any new ones that might've appeared recently.

zmedico:
Let's call it `file.size` rather than `file.size-fatal`, since that's how it is in the man page and none of our other qa categories have a `-fatal` suffix. In order to covert `file.size` to a fatal error, it needs to be removed from the  qawarnings list in qa_data.yml.

juippis:
Done, and tested.

PR #630, opened 2020-10-16
cnf/sets/portage.conf: add golang-rebuild set

gyakovlev:
go-built binaries may contain security
vulnerabilities if a binary built with vulnerable compiler.
go is known to embed vulnerable code to all binaries it builds, if
vulnerability was present in the compiler or one of standard libraries.

This commit adds `golang-rebuild` set, which allows easy
rebuild of most go-compiled system packages.

simple 'emerge @golang-rebuild' should rebuild everything affected.
a prompt to run this command can be added to postinst message in
`dev-lang/go` ebuild.

### DIFF
--- a/cnf/sets/portage.conf
+++ b/cnf/sets/portage.conf
@@ -103,3 +103,9 @@ class = portage.sets.dbapi.UnavailableBinaries
 # to the matching portdb entry.
 [changed-deps]
 class = portage.sets.dbapi.ChangedDepsSet
+
+# Installed packages that inherit from known go related eclasses.
+[golang-rebuild]
+class = portage.sets.dbapi.VariableSet
+variable = INHERITED
+includes = golang-base golang-build golang-vcs golang-vcs-snapshot go-module


mgorny:
Wouldn't it be more reliable to grab all packages having BDEPEND on go/rust compiler? Not sure if that's possible though.

gyakovlev:
yes will be more reliable, as I've seen go packages not using eclasses.
but go is a dumpsterfire bu design anyway and nothing can fix it =)

I haven't looked how `includes =` matching works, if it can work on patterns it could work.
also not all go eclasses/packages are `EAPI=7` and we'll have to check `DEPEND` as well.

with rust it should be reliable as is in this PR. pretty much everything uses cargo.eclass, except mozilla things, spidermonkey and rust itself. but those are `special` and do not obey standard rust rules, so no point hunting them.
I'm not actually sure rust set is needed anymore after doing some research. it all depends on crates package is built from. std bugs are rare, especially security ones. there's been single instance so far.
while with go it happens every bump pretty much.

gyakovlev:
tested DEPEND/BDEPEND matching.
match is literal, not a wildcard/pattern.
`class VariableSet` seems to accept only 1 variable, so I could not check both `DEPEND` + `BDEPEND`
each set only accept single `VariableSet` class

so without a lot of extra effort it's not possible.

also there's a bash script packaged as `app-admin/go-updater`,
but it requires more go packages to install and is m-n.

gyakovlev:
updated commit, removed rust.

zx2c4:
Shouldn't this instead happen every time go is bumped, automatically, via a subslot? Why would we want to punt this kind of thing to the user? If the Go runtime/compiler/etc is updated, we need to update the packages that use it. The fact that everything is statically compiled means we have no choice but to rebuild, in that case.

williamh:
@gyakovlev If there are go packages not using the eclasses in the tree, those need to be fixed. That would be like a python package not using the python eclasses.

zmedico (approved):

zmedico:
Released in portage-3.0.9.

PR #634, opened 2020-11-07
Typofix: Use just one definite article

contyk:
I first noticed this in make.conf(5) but it turned out there were many
more occurences of these; fixed with find & sed.

Signed-off-by: Petr Šabata <contyk@redhat.com>


zmedico:
Thank you!

PR #649, opened 2020-12-30
Test with Python 3.10

mrueg:
Runs tests with python 3.10


zmedico:
I think it's just not running because the monthly credits are all used up here:

https://travis-ci.com/organizations/gentoo/plan

PR #654, opened 2021-01-04
global_event_loop: return running loop for current thread

zmedico:
Like asyncio.get_event_loop(), return the running loop for the
current thread if there is one, and otherwise construct a new
one if needed. This allows the _safe_loop function to become
synonymous with the global_event_loop function.

For the case of "loop running in non-main thread" of API
consumer, this change makes portage compatible with PEP
492 coroutines with async and await syntax. Portage
internals can safely begin using async / await syntax instead
of compat_coroutine.

Bug: https://bugs.gentoo.org/763339


zmedico:
With the current patch, unit tests consistently hang with python3.6 and python3.7. but not for python3.8 and later. I've posted some more info in https://bugs.gentoo.org/763339#c3.

PR #670, opened 2021-02-19
make.defaults: prevent USE="${USE} ..." misbehavior

zmedico:
Discard parent profile USE from the expand_map variable before
it is used to evaluate a child profile. This prevents accidents
triggered by USE="${USE} ..." settings at the top of make.defaults
which caused parent profile USE to override parent profile
package.use settings.

Note that would be nice to guard the USE_EXPAND variables like
this too, but unfortunately USE_EXPAND is not known until after
make.defaults has been evaluated, so that will require some form
make make.defaults preprocessing.

Bug: https://bugs.gentoo.org/771549

### DIFF
--- a/lib/portage/package/ebuild/config.py
+++ b/lib/portage/package/ebuild/config.py
@@ -612,9 +612,20 @@ class config:
 
 			mygcfg = {}
 			if profiles_complex:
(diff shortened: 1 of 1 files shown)


Arfrever:
In commit message: `s/settngs/settings/`

zmedico:
> In commit message: `s/settngs/settings/`

Thanks

PR #674, opened 2021-02-23
repoman: revert preserve_old_lib deprecation (bug 480244)

zmedico:
Repoman should not report that preserve_old_lib is deprecated,
since preserve-libs is not covered by PMS.
This reverts commit 49cbc17bf7b99be586e158c1bd588cfe91dfe58c.

Bug: https://bugs.gentoo.org/480244
Bug: https://bugs.gentoo.org/692486


zmedico:
Due to controversy, I would like to withdraw this patch, and call upon interested parties to help decide an appropriate course of action or inaction.

PR #689, opened 2021-03-28
dispatch-conf: add conf-update.d hooks

Flowdalic:


thesamesam:
See https://wiki.gentoo.org/wiki/Future_EAPI/Triggers. I think it might be worth discussing with @mgorny and seeing if we can standardise this instead.

Flowdalic:
> See https://wiki.gentoo.org/wiki/Future_EAPI/Triggers. I think it might be worth discussing with @mgorny and seeing if we can standardise this instead.

The triggers in the linked Wiki page seem to be orthogonal to the ones proposed in this PR. The ones of the Wiki page are invoked when a package manager, e.g. portage, installs files, whereas the ones in this PR are invoked when the user dispatches new configuration files, e.g. by running `dispatch-conf`.

That said, I am all for standardization. The hooks in this PR are deliberately designed in a generic way, so that other tools that dispatch configuration files, besides `dispatch-conf`, can invoke them.

lib/portage/util/hooks.py
```diff
@@ -0,0 +1,32 @@
...
+	directory = os.path.join(prefix, portage.USER_CONFIG_PATH, rel_directory)
+
+	hooks = OrderedDict()
+	for filepath in _recursive_file_list(directory):
+		name = filepath.split(directory)[1].lstrip(portage.os.sep)
+		if portage.os.access(filepath, portage.os.X_OK):
+			hooks[filepath] = name
+		else:
+			writemsg_level(" %s %s hook: '%s' is not executable\n" % (warn("*"),
+																  directory,
```
mgorny, line 29: What's with the weird indent here? We generally use tab-indent in Portage atm, and not respect that PEP thing, and therefore we don't align.

bin/dispatch-conf
```diff
@@ -34,7 +34,7 @@ import portage
 portage._internal_caller = True
 from portage import os, shutil
 from portage import _encodings, _unicode_decode
-from portage.dispatch_conf import diffstatusoutput, diff_mixed_wrapper
+from portage.dispatch_conf import diffstatusoutput, diff_mixed_wrapper, perform_conf_update_hooks, perform_conf_update_session_hooks
```
mgorny, line 37: Wrap, please.

lib/portage/util/hooks.py
Flowdalic: Good question, it was either from where in portage's codebase I copied this code snipped or due my editors default python formatting, or due a combination of both. Going to fix that.

BTW is the call to `portage._unicode_decode(name)` still needed with a Python-3-only portage?

bin/dispatch-conf
```diff
@@ -34,7 +34,8 @@ import portage
 portage._internal_caller = True
 from portage import os, shutil
 from portage import _encodings, _unicode_decode
-from portage.dispatch_conf import diffstatusoutput, diff_mixed_wrapper
+from portage.dispatch_conf import diffstatusoutput, diff_mixed_wrapper, \
+	perform_conf_update_hooks, perform_conf_update_session_hooks
```
mgorny, line 38: We use

```
... import (...,
<tab>...)
```

wrapping in new code (i.e. without backslashes).

PR #692, opened 2021-03-29
SyncRepos: suppress portage update message if PYTHON_TARGETS changed

zmedico:
When PYTHON_TARGETS changed, the emerge --oneshot portage
suggestion a nuisance, therefore suppress it.

Bug: https://bugs.gentoo.org/722748


zmedico:
Merged in ba9866db877afdfe20d11768007233da6334d578.

PR #693, opened 2021-03-29
PORTAGE_NICENESS: Consider autogroup scheduling

Flowdalic:
With Linux's autogroup scheduling feature (CONFIG_SCHED_AUTOGROUP)
setting a nice value on a per-process base has only an effect for
scheduling decisions relative to the other threads in the same
session (typically: the same terminal window). See the section "The
nice value and group scheduling" in the sched(7) man page.

Basically this means that portage "just" setting the nice value, has
no effect in presence of autogroup scheduling being active (which is
probably true for most (desktop) user systems).

This commit changes emerge to set the autogroup's nice value, instead
of the processes' nice value, in case autogroups are present (detected
by the existence of /proc/self/autogroup). The tricky part about
autogroup nice values is that we want restore the orignal nice value
once we are finished. As otherwise, the session, e.g. your terminal,
would continue using this value, and so would subsequently executed
processes. For that we use Python's atexit functinaly, to register a
function that will restore the orignal nice value of the autogroup.

Bug: https://bugs.gentoo.org/777492
Signed-off-by: Florian Schmaus <flo@geekplace.eu>

### DIFF
--- a/bin/pid-ns-init
+++ b/bin/pid-ns-init
@@ -12,6 +12,7 @@ import subprocess
 import sys
 import termios
 
+from pathlib import Path
 
 KILL_SIGNALS = (
 	signal.SIGINT,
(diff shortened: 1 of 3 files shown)


zmedico:
We cannot entirely rely on the atexit handler to do it's job, so we'll have to document this lifecycle and any manual actions that will be necessary should the atexit handler fail to be invoked.

Flowdalic:
> We cannot entirely rely on the atexit handler to do it's job, so we'll have to document this lifecycle and any manual actions that will be necessary should the atexit handler fail to be invoked.

I've added a few sentences to make.conf's man page. Is that what you had in mind?

zmedico (approved):
Looks good. Thanks!

zmedico:
Merged in https://github.com/gentoo/portage/compare/a4d882964ee1931462f911d0c46a80e27e59fa48%5E...209be9a8bee13384dd04a4762436b4c2a5e35bc6.

I used a pattern like this to eliminate the autogroup_file.is_file() stat call:
```python
	autogroup_file = Path("/proc/self/autogroup")
	try:
		f = autogroup_file.open("r+")
	except EnvironmentError:
		# Autogroup scheduling is not enabled on this system.
		return

	with f:
		line = f.readline()
```

PR #694, opened 2021-04-09
lib/portage/_sets/dbapi.py: add glob support to exclude-files parameter 

xmhd:
The module-rebuild package set is currently all packages which own files in /lib/modules. This leads to the following scenario/error:

- Using gentoo-kernel{-bin} package (Gentoo dist-kernel project by mgorny)
- gentoo-kernel{-bin} is now part of module-rebuild package set as it owns the kernel module files in /lib/modules/${KV_FULL}
- emerge --ask @module-rebuild will now wrongly list gentoo-kernel{-bin} alongside zfs, nvidia etc.

As far as I'm aware, all external kernel module packages are using linux-mod.eclass, so by changing the module-rebuild set to inherited linux-mod (as opposed to owned /lib/modules), we solve the above problem and emerge @module-rebuild will now allow rebuilding of all packages inheriting linux-mod.eclass, and no longer wrongly list gentoo-kernel{-bin} package as an external kernel module.

Tested on ~amd64 with zfs and a few other external kernel module packages.

### DIFF
--- a/cnf/sets/portage.conf
+++ b/cnf/sets/portage.conf
@@ -71,6 +71,7 @@ includes = bzr cvs darcs git-2 git-r3 golang-vcs mercurial subversion
 [module-rebuild]
 class = portage.sets.dbapi.OwnerSet
 files = /lib/modules
+exclude-files = /usr/src/linux*
 
 # Installed packages that own files inside /usr/lib/xorg/modules,
 # excluding the package that owns /usr/bin/Xorg.
--- a/lib/portage/_sets/dbapi.py
+++ b/lib/portage/_sets/dbapi.py
@@ -79,6 +79,12 @@ class OwnerSet(PackageSet):
 				glob.iglob(os.path.join(eroot, p.lstrip(os.sep))))
 		paths = expanded_paths
 
+		expanded_exclude_paths = []
+		for p in exclude_paths:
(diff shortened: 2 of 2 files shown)


Arfrever:
There might be some packages which inherit `linux-mod.eclass` but build and install kernel modules conditionally on some USE flag...

xmhd:
> There might be some packages which inherit `linux-mod.eclass` but build and install kernel modules conditionally on some USE flag...

Now you mention it, I have found one or two packages which have that. It would be possible to conditionally inherit `linux-mod.eclass` if USE=modules (or whatever the conditions are).

I'm happy to change this approach if you have another solution.

zmedico:
I think `excluded-files=/usr/src` is probably the best alternative to something like PROPERTIES="kernel-module".

zmedico:
Really, slot-operator deps are the intended way to trigger this sort of rebuild.

The next best thing would be something like PROPERTIES="kernel-module".

xmhd:
as discussed with @zmedico on irc - I have filed a bug [https://bugs.gentoo.org/781617](url)

xmhd:
PR updated following discussion + guidance from @zmedico 

module-rebuild will remain as OwnerSet and glob support has been added to the exclude-files parameter.
exclude-files = /usr/src/linux-* has been added to the module-rebuild set to exclude any kernel packages from being identified by @module-rebuild.

zmedico (approved):
Thanks!

PR #700, opened 2021-04-24
[WIP] portage: Remove unicode coercions back & forth

mjsir911:
This speeds up --depclean on 245 packages from 12 min to 10 min, 18%
speedup

Probably need more tests to verify that this doesn't break anything

This also passes tests even when run in `mkdir portage-$(printf "\\xa9")-not-utf8` / `export TMPDIR=/tmp/tmpdir$(printf "\\xa9")-not-utf8`

Signed-off-by: Marco Sirabella <marco@sirabella.org>


thesamesam:
Hey, still interested in this?

zmedico:
In https://github.com/gentoo/portage/pull/1103 the unicode coercion is disabled for UTF-8 locales. In order to do this safely for non-UTF-8 locales, we need to force python into [utf8 mode](https://docs.python.org/3/library/os.html#python-utf-8-mode) either via shebang or python re-exec as discussed in https://bugs.gentoo.org/914722.

thesamesam:
This is now done with https://github.com/gentoo/portage/pull/1601. Thanks for working on it before.

PR #701, opened 2021-04-27
env-update: skip PATH in systemd user environment definition

Flowdalic:
Having PATH in the systemd user environment created by env-update
makes it impossible to use "systemclt --user import-environment PATH"
to set PATH in systemd user units according to the current value of
PATH. Using "systemclt --user import-environment PATH" is a well known
idiom, and hence should work. Therefore, we skip the creation of the
PATH entry by env-update.

Signed-off-by: Florian Schmaus <flo@geekplace.eu>


zmedico (commented):
Thanks!

PR #705, opened 2021-05-07
Sort emerge --unmerge order for determinism

fb913bf0de288ba84fe98f7a23d35edfdb22381:
This commit changes the order in which packages are uninstalled
with --unmerge to be deterministic.

Before this commit, when an atom specified with --unmerge matches
multiple versions, these versions were uninstalled in a random order.
For example, when some-pkg-1.0.0 and some-pkg-2.0.0 are installed,
then running emerge --unmerge some-pkg will unmerge 1.0.0 and 2.0.0
in a random order.

With this commit, when an atom matches multiple versions,
they are uninstalled in a sorted order. So in the above example,
1.0.0 will be unmerged before 2.0.0.

This does not affect the order specified on the command-line,
for example when running:
emerge --unmerge =some-pkg-2.0.0 =some-pkg-1.0.0
that order is respected, as before this commit.

Similarly, when uninstalling multiple different packages, e.g.:
emerge --unmerge some-pkg some-other-pkg
the sorting only applies to the versions matching each atom,
so the versions matching some-pkg are sorted separately from
the versions matching some-other-pkg, and since some-pkg
is specified before some-other-pkg, its versions will be
uninstalled first.

Motivation: When a package has a custom pkg_postrm hook, uninstalling
multiple versions in a random order can leave the filesystem in a
different state depending on which version is uninstalled last.
When running emerge as part of a larger build system such as catalyst,
this is an obstacle towards reproducible builds.

Bug: https://bugs.gentoo.org/782724

Signed-off-by: Felix Bier <felix.bier@rohde-schwarz.com>

### DIFF
--- a/lib/_emerge/unmerge.py
+++ b/lib/_emerge/unmerge.py
@@ -99,7 +99,6 @@ def _unmerge_display(root_config, myopts, unmerge_action,
 
 			global_unmerge = 1
 
-		localtree = vartree
 		# process all arguments and add all
 		# valid db entries to candidate_catpkgs
 		if global_unmerge:
@@ -199,7 +198,7 @@ def _unmerge_display(root_config, myopts, unmerge_action,
 				sys.exit(1)
 
 			if not mymatch and x[0] not in "<>=~":
-				mymatch = localtree.dep_match(x)
+				mymatch = vartree.dep_match(x)
 			if not mymatch:
 				portage.writemsg("\n--- Couldn't find '%s' to %s.\n" % \
 					(x.replace("null/", ""), unmerge_action), noiselevel=-1)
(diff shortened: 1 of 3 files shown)


lib/_emerge/unmerge.py
```diff
@@ -563,7 +563,13 @@ def unmerge(root_config, myopts, unmerge_action,
 	maxval = len(all_selected)
 
 	for x in range(len(pkgmap)):
-		for y in pkgmap[x]["selected"]:
+
+		selected = pkgmap[x]["selected"]
+
+		if ordered:
+			selected = sorted(selected)
```
Arfrever, line 570: You may need to use some Portage's helper function for comparisons of versions to avoid lexicographic comparison:
```
$ python -c 'print(sorted(["1.0", "1.1", "1.2", "1.9", "1.10", "1.11"]))'
['1.0', '1.1', '1.10', '1.11', '1.2', '1.9']
```

zmedico (approved):
Thank you!

PR #706, opened 2021-05-08
emerge: Fail with helpful message when --jobs=""

mjsir911:
Reported by Kobboi on the irc channel:

> oops, my portage is broken: TypeError: '>=' not supported between instances of 'int' and 'str'
> false alarm (though maybe using --jobs= (without value) could be handled more nicely than a python error :) )


zmedico:
Thanks! This issue is now tracked by https://bugs.gentoo.org/788967.

PR #707, opened 2021-05-09
Bugfix: emerge --fetchonly --quiet is too verbose. 

steeznson:
https://bugs.gentoo.org/285614

I noticed that EbuildFetcher does not pipe out file download progress to pty if it is instantiated with the background kwarg. Since we have access to the --quiet setting in `EbuildBuild.py` when the fetcher is instantiated, it seems to me that we can use that value to determine whether to print the fetch progress.

### DIFF
--- a/lib/_emerge/EbuildBuild.py
+++ b/lib/_emerge/EbuildBuild.py
@@ -23,6 +23,7 @@ from portage.package.ebuild.digestcheck import digestcheck
 from portage.package.ebuild.doebuild import _check_temp_dir
 from portage.package.ebuild._spawn_nofetch import SpawnNofetchWithoutBuilddir
 from portage.util._async.AsyncTaskFuture import AsyncTaskFuture
+from portage.util.path import first_existing
 
 
 class EbuildBuild(CompositeTask):
@@ -137,6 +138,15 @@ class EbuildBuild(CompositeTask):
 		pkg = self.pkg
 		settings = self.settings
 
+		quiet_setting = settings.get('PORTAGE_QUIET', False)
+		fetch_log = None
+		logwrite_access = False
+		if quiet_setting:
+			fetch_log = os.path.join(
+					_emerge.emergelog._emerge_log_dir,
+					'emerge-fetch.log')
+			logwrite_access = os.access(first_existing(fetch_log), os.W_OK)
+
 		if opts.fetchonly:
 			if opts.pretend:
 				fetcher = EbuildFetchonly(
@@ -165,11 +175,17 @@ class EbuildBuild(CompositeTask):
 				ebuild_path=self._ebuild_path,
 				fetchall=self.opts.fetch_all_uri,
 				fetchonly=self.opts.fetchonly,
-				background=False,
-				logfile=None,
+				background=quiet_setting if logwrite_access else False,
+				logfile=fetch_log if logwrite_access else None,
 				pkg=self.pkg,
 				scheduler=self.scheduler)
-			self._start_task(fetcher, self._fetchonly_exit)
+
+			if fetch_log and logwrite_access:
+				fetcher.addExitListener(self._fetchonly_exit)
+				self._task_queued(fetcher)
+				self.scheduler.fetch.schedule(fetcher, force_queue=True)
+			else:
+				self._start_task(fetcher, self._fetchonly_exit)
 			return
 
 		self._build_dir = EbuildBuildDir(
--- a/lib/_emerge/Scheduler.py
+++ b/lib/_emerge/Scheduler.py
@@ -560,7 +560,7 @@ class Scheduler(PollScheduler):
 			if pargs:
 				self.status = pargs[0]
 
-	def _schedule_fetch(self, fetcher):
+	def _schedule_fetch(self, fetcher, force_queue=False):
 		"""
 		Schedule a fetcher, in order to control the number of concurrent
 		fetchers. If self._max_jobs is greater than 1 then the fetch
@@ -571,7 +571,7 @@ class Scheduler(PollScheduler):
 		would be required before it would be possible to enable
 		concurrent fetching within the parallel-fetch queue.
 		"""
-		if self._max_jobs > 1:
+		if self._max_jobs > 1 and not force_queue:
 			fetcher.start()
 		else:
 			self._task_queues.fetch.addFront(fetcher)


steeznson:
There is an old comment on bugzilla suggesting that we redirect the fetcher output to `/var/log/emerge-fetch.log`. I wasn't sure if this was a requirement for the bug since the fetcher doesn't appear to use the log at the moment. I can update the PR to use the logfile kwarg if that is a better solution.

zmedico:
We want to output to go somewhere, since otherwise we won't be able to seen an error message if it fails.

steeznson:
I've fixed it so that the fetch log is being written out to now

lib/_emerge/EbuildBuild.py
```diff
@@ -165,8 +172,8 @@ def _prefetch_exit(self, prefetcher):
 				ebuild_path=self._ebuild_path,
 				fetchall=self.opts.fetch_all_uri,
 				fetchonly=self.opts.fetchonly,
-				background=False,
-				logfile=None,
+				background=quiet_setting,
+				logfile=fetch_log,
 				pkg=self.pkg,
 				scheduler=self.scheduler)
 			self._start_task(fetcher, self._fetchonly_exit)
```
zmedico, line 179: We should use `self.scheduler.fetch.schedule(fetcher)` to schedule this on the fetch queue, so that it has exclusive access to the fetch log. We'll need to somehow ensure that it's added to the queue here rather than started immediately, possibly by adding a method parameter that defaults to the status quo behavior:

https://github.com/gentoo/portage/blob/portage-3.0.18/lib/_emerge/Scheduler.py#L564

steeznson: I believe that my latest commit should add it to the scheduler queue without it starting immediately. I didn't end up altering the `_schedule_fetch()` method but hopefully I've been able to accomplish what you had in mind.

steeznson: I wondered if you were suggesting doing something like this with the new method param (superficially doesn't seem to behave differently from what I've committed but I might be wrong):

```
diff --git a/lib/_emerge/EbuildBuild.py b/lib/_emerge/EbuildBuild.py
index b71273dc6..58d2b9e4d 100644
--- a/lib/_emerge/EbuildBuild.py
+++ b/lib/_emerge/EbuildBuild.py
@@ -180,7 +180,7 @@ class EbuildBuild(CompositeTask):
 			if fetch_log:
 				fetcher.addExitListener(self._fetchonly_exit)
 				self._task_queued(fetcher)
-				self.scheduler.fetch.schedule(fetcher)
+				self.scheduler.fetch.schedule(fetcher, force_queue=True)
 			else:
 				self._start_task(fetcher, self._fetchonly_exit)
 			return
diff --git a/lib/_emerge/Scheduler.py b/lib/_emerge/Scheduler.py
index d3528ee68..3a9bef688 100644
--- a/lib/_emerge/Scheduler.py
+++ b/lib/_emerge/Scheduler.py
@@ -560,7 +560,7 @@ class Scheduler(PollScheduler):
 			if pargs:
 				self.status = pargs[0]
 
-	def _schedule_fetch(self, fetcher):
+	def _schedule_fetch(self, fetcher, force_queue=False):
 		"""
 		Schedule a fetcher, in order to control the number of concurrent
 		fetchers. If self._max_jobs is greater than 1 then the fetch
@@ -571,7 +571,7 @@ class Scheduler(PollScheduler):
 		would be required before it would be possible to enable
 		concurrent fetching within the parallel-fetch queue.
 		"""
-		if self._max_jobs > 1:
+		if self._max_jobs > 1 or not force_queue:
 			fetcher.start()
 		else:
 			self._task_queues.fetch.addFront(fetcher)
```

zmedico: The fetchonly flag triggers this `self._set_max_jobs(1)` call that happens to make it behave as desired:
```
lib/_emerge/Scheduler.py:383:   def _background_mode(self):
lib/_emerge/Scheduler.py:412:                                   self._set_max_jobs(1)
```
However, it's better if we go ahead and add a parameter to make this explicit. Your `force_queue` implementation loogs good, but the boolean expression needs an `and` like this:
```python
@@ -571,7 +571,7 @@ class Scheduler(PollScheduler):
 		would be required before it would be possible to enable
 		concurrent fetching within the parallel-fetch queue.
 		"""
-		if self._max_jobs > 1:
+		if self._max_jobs > 1 and not force_queue:
 			fetcher.start()
 		else:
 			self._task_queues.fetch.addFront(fetcher)
```

steeznson: That's the `force_queue` functionality added now. Thank you for taking the time to review my code! This was an excellent introduction to the portage codebase.

zmedico: You're welcome!

zmedico (approved):
Looks good!

If possible, please go ahead and add a Signed-off-by line to each commit which is required for compliance with [Gentoo's Certificate of Origin](url) policy.

lib/_emerge/EbuildBuild.py
```diff
@@ -137,6 +137,13 @@ def _prefetch_exit(self, prefetcher):
 		pkg = self.pkg
 		settings = self.settings
 
+		quiet_setting = settings.get('PORTAGE_QUIET', False)
+		fetch_log = None
+		if quiet_setting:
+			fetch_log = os.path.join(
+					_emerge.emergelog._emerge_log_dir,
+					'emerge-fetch.log')
```
zmedico, line 145: This needs to be conditional on write access to `emerge-fetch.log` or a parent directory if the file doesn't exist yet, so we need some kind of conditional logic here. We can use the `portage.util.path.first_existing` function like this to check for write access:
```python
os.access(first_existing(fetch_log), os.W_OK)
```
It there's no write access to the log, then I suppose you'll have to allow the output to flow to directly stdout.

If there is write access, and output goes to the log, then we also need to ensure that error messages appropriately refer to the log file when fetch fails.

zmedico (changes requested):
Thanks for including the Signed-off-by line, but your  Signed-off-by line line is supposed to have your name, not mine. I'll add an additional Signed-off-by line for myself when I rebase and merge the patch(es).

lib/_emerge/EbuildBuild.py
steeznson: I've implemented the permissions check as you requested. In terms of the error output pointing to the correct logfile, after digging into the code I can see that this behaviour is already implemented. `_proc_join_done()` in EbuildFetcher.py will raise an error and point to the correct logfile if we encounter an issue when prefetching.
I think this is ready for another review.

zmedico: Thanks! Merged in e26a08896f098ff93530efcd2d77d60d2eff0257.

PR #712, opened 2021-05-12
Implement PROPERTIES=test_network and ALLOW_TEST

mgorny:
Bug: https://bugs.gentoo.org/553696


mgorny:
After all, we've (me and @ulm) decided that this is better done retroactively via `PROPERTIES=network-test`.

bin/phase-functions.sh
```diff
@@ -489,7 +489,9 @@ __dyn_test() {
 		die "The source directory '${S}' doesn't exist"
 	fi
 
-	if has test ${RESTRICT} ; then
+	# elide test restriction if the ebuild has PROPERTIES=network-test
+	# and the user enabled FEATURES=network-test
+	if has test ${RESTRICT} && ! { has network-test ${PROPERTIES} && has network-test ${FEATURES}; }; then
```
zmedico, line 494: It looks like we'll have to evaluate USE conditionals in PROPERTIES prior to this. See this related RESTRICT code which has no PROPERTIES counterpart:

https://github.com/gentoo/portage/blob/portage-3.0.18/lib/portage/package/ebuild/config.py#L1757
https://github.com/gentoo/portage/blob/portage-3.0.18/bin/ebuild.sh#L621

mgorny: Done!

zmedico:
I think it's probably better if we preserve the notion that `RESTRICT="test"` disables `src_test`, since otherwise we've created a special case that will probably lead to confusion.

For ebuilds that have `PROPERTIES="network-test"`, why don't we remove the need to set `RESTRICT="test"`, and only execute `src_test` for these ebuilds when the user has `FEATURES="network-test"` enabled?

mgorny:
Updated to use `RESTRICT_TEST_ELIDE` as discussed on IRC. Tested with `package.env`, `make.conf` and environment setting.

The extra advantage is that you can do the following now to easily ignore RESTRICT=test in ebuild:

```
RESTRICT_TEST_ELIDE=all emerge -1v broken-test-package

zmedico (approved):
I love it! Please merge!

ulm:
PROPERTIES="test_network" should also be mentioned in the ebuild(5) man page.

zmedico:
Merged in a88f4996a9c20a638c8bf9a42bcbbc28ce0cc8dc.

PR #713, opened 2021-05-12
Relax update filenames as permitted for EAPI 8

mgorny:
Bug: https://bugs.gentoo.org/692774

### DIFF
--- a/lib/portage/update.py
+++ b/lib/portage/update.py
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 import errno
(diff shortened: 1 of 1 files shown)


lib/portage/update.py
```diff
@@ -175,14 +175,11 @@ def grab_updates(updpath, prev_mtimes=None):
 		raise
 	if prev_mtimes is None:
 		prev_mtimes = {}
-	# validate the file name (filter out CVS directory, etc...)
-	mylist = [myfile for myfile in mylist if len(myfile) == 7 and myfile[1:3] == "Q-"]
+	# skip dotfiles
+	mylist = [myfile for myfile in mylist if not myfile.startswith(".")]
```
ulm, line 179: Directories should also be skipped. The proposed spec says:
"All regular files in this directory will be processed, unless their name begins with a dot."

mgorny:
Made it skip entries that are not regular files. Also removed the empty list short-circuit because it makes little sense now — after all, `for` loop will just terminate immediately and the function will return empty list anyway.

lib/portage/update.py
```diff
@@ -175,18 +175,15 @@ def grab_updates(updpath, prev_mtimes=None):
...
 
 	update_data = []
 	for myfile in mylist:
+		if myfile.startswith("."):
+			continue
 		file_path = os.path.join(updpath, myfile)
 		mystat = os.stat(file_path)
+		if not stat.S_ISREG(mystat.st_mode):
+			continue
 		if update_data or \
```
ulm, line 187: I think without sorting, the logic of this following `if` statement makes even less sense than it did before. If `prev_mtimes` is given, then only the files with different mtimes should be considered.

Adding any other "later" files to the list doesn't make sense, because PMS doesn't specify any order for processing of these files, so there is no concept of "later", in the first place.

mgorny: Makes sense. I'm recombining the PRs now, I'll fix it after that.

PR #714, opened 2021-05-12
Disallow options in PATCHES in EAPI 8

mgorny:
Bug: https://bugs.gentoo.org/752486

### DIFF
--- a/bin/phase-helpers.sh
+++ b/bin/phase-helpers.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 if ___eapi_has_DESTTREE_INSDESTTREE; then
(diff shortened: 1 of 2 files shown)


bin/phase-helpers.sh
```diff
@@ -853,6 +853,23 @@ __eapi6_src_install() {
 	einstalldocs
 }
 
+__eapi8_src_prepare() {
+	local f
+	if ___is_indexed_array_var PATCHES ; then
+		for f in "${PATCHES[@]}"; do
+			[[ ${f} == -* ]] && die "Options disallowed in PATCHES array"
+		done
```
ulm, line 861: That's not as specified. `PATCHES` should just be passed to eapply without further filtering, i.e., all elements in `PATCHES` are supposed to be filenames, even if they start with `-`. (That's also how POSIX tools behave, `--` suppresses special treatment of `-` for all following arguments.)

(I think we don't want diverging behaviour between the default and explicit `eapply -- ${PATCHES}` in an ebuild.

bin/phase-helpers.sh
```diff
@@ -853,6 +853,23 @@ __eapi6_src_install() {
...
+		[[ ${#PATCHES[@]} -gt 0 ]] && eapply -- "${PATCHES[@]}"
+	elif [[ -n ${PATCHES} ]]; then
+		for f in ${PATCHES}; do
+			[[ ${f} == -* ]] && die "Options disallowed in PATCHES"
+		done
```
ulm, line 866: Same here.

ulm (changes requested):

mgorny:
Removed check for option-lookalikes.

ulm (approved):

PR #715, opened 2021-05-12
dosym: Implement -r option for EAPI 8.

ulm:
"dosym -r &lt;target&gt; &lt;link&gt;" will expand the (apparent) path of &lt;target&gt;
relative to the (apparent) path of the directory containing &lt;link&gt;.
The main aim of this is to allow for an absolute path to be specified
as the link target, and the function will count path components and
convert it into a relative path.

Since we're inside ED at this point but the image will finally be
installed in EROOT, we don't try to resolve any pre-existing symlinks
in &lt;target&gt; or &lt;link&gt;. In other words, path expansion only looks at
the specified apparent paths, without touching any actual files in ED
or EROOT.

Bug: https://bugs.gentoo.org/708360
Signed-off-by: Ulrich Müller &lt;ulm@gentoo.org&gt;

### DIFF
--- a/bin/eapi.sh
+++ b/bin/eapi.sh
@@ -238,6 +238,10 @@ ___eapi_has_DESTTREE_INSDESTTREE() {
 	[[ ${1-${EAPI-0}} =~ ^(0|1|2|3|4|4-python|4-slot-abi|5|5-progress|6)$ ]]
 }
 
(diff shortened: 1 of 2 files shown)


mgorny:
Merged into #718.

PR #719, opened 2021-05-18
Change _shell_quote to handle more special characters

fb913bf0de288ba84fe98f7a23d35edfdb22381:
This PR changes the function _shell_quote to ensure more
special characters are quoted correctly.

Without this commit, the following error message occurs when attempting
to merge an ebuild that inherits from an eclass contained in a
repository located in a filesystem path containing special characters:

.../ebuild.sh: eval: line 604: syntax error near unexpected token `('
.../ebuild.sh: eval: line 604: `PORTAGE_ECLASS_LOCATIONS=(/var/db/repos/gentoo /home/user/test(v2)/test-overlay)'

The paths that are handled in that line were sanitized with _shell_quote
beforehand, but the list of special characters handled by _shell_quote
did not include (). For this reason, the list has been extended to
handle the following special characters that were not handled before:

;&|(){}[]#!~?

Signed-off-by: Felix Bier <felix.bier@rohde-schwarz.com>

### DIFF
--- a/lib/portage/__init__.py
+++ b/lib/portage/__init__.py
@@ -403,7 +403,7 @@ def _get_stdin():
 		return sys.__stdin__
 	return sys.stdin
 
-_shell_quote_re = re.compile(r"[\s><=*\\\"'$`]")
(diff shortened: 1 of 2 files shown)


zmedico (approved):
Thank you!

PR #720, opened 2021-05-22
Bugfix 520378: Ensure EMERGE_LOG_DIR setting is respected

steeznson:
https://bugs.gentoo.org/520378

The path hardcoding issue appears to be resolved. EMERGE_LOG_DIR is parsed in `lib/_emerge/actions.py` if it exists and set to a variable, `emergelog._emerge_log_dir`. The output referring to logs are all using that variable currently.

That said, the EMERGE_LOG_DIR setting is not completely integrated with the code. I found two places where the setting was being ignored. One of them relates to the recent bugfix for https://bugs.gentoo.org/285614.

### DIFF
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -52,6 +52,7 @@ from portage._sets.base import InternalPackageSet
 from portage.util import cmp_sort_key, normalize_path, writemsg, varexpand, \
 	writemsg_level, writemsg_stdout
 from portage.util.digraph import digraph
+from portage.util.path import first_existing
 from portage.util.SlotObject import SlotObject
 from portage.util._async.run_main_scheduler import run_main_scheduler
 from portage.util._async.SchedulerInterface import SchedulerInterface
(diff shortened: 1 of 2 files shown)


lib/_emerge/actions.py
```diff
@@ -3165,8 +3165,12 @@ def run_action(emerge_config):
 	disable_emergelog = False
 	for x in ("--pretend", "--fetchonly", "--fetch-all-uri"):
 		if x in emerge_config.opts:
-			disable_emergelog = True
-			break
+			if x == "--fetchonly" and "--quiet" in emerge_config.opts:
+				# Log will be used to store fetch progress
```
zmedico, line 3169: You'll have to check for write access to emerge.log here, just as you did in e26a08896f098ff93530efcd2d77d60d2eff0257.

steeznson: Have added the access check now.

zmedico: Thanks!

PR #721, opened 2021-05-25
Use an explicit empty dir for pkg_* phases [WIP]

mgorny:
Create and use an explicit ${PORTAGE_BUILDDIR}/empty as working
directory for pkg_* phases, as proposed for EAPI 8.

Note that this patch doesn't work fully -- empty is not cleared between
pkg_preinst and pkg_postinst, and between pkg_prerm and pkg_postrm.

### DIFF
--- a/bin/ebuild.sh
+++ b/bin/ebuild.sh
@@ -186,9 +186,9 @@ export SANDBOX_ON=0
 # Ensure that $PWD is sane whenever possible, to protect against
 # exploitation of insecure search path for python -c in ebuilds.
 # See bug #239560, bug #469338, and bug #595028.
-if [[ -d ${HOME} ]]; then
-	# Use portage's temporary HOME directory if available.
-	cd "${HOME}" || die
+# EAPI 8 requires us to use an empty directory here.
+if [[ -d ${PORTAGE_BUILDDIR}/empty ]]; then
+	cd "${PORTAGE_BUILDDIR}/empty" || die
 else
 	cd "${PORTAGE_PYM_PATH}" || \
 		die "PORTAGE_PYM_PATH does not exist: '${PORTAGE_PYM_PATH}'"
--- a/bin/phase-functions.sh
+++ b/bin/phase-functions.sh
@@ -278,7 +278,8 @@ __dyn_clean() {
 	cd "${PORTAGE_PYM_PATH}" || \
 		die "PORTAGE_PYM_PATH does not exist: '${PORTAGE_PYM_PATH}'"
 
-	rm -rf "${PORTAGE_BUILDDIR}/image" "${PORTAGE_BUILDDIR}/homedir"
+	rm -rf "${PORTAGE_BUILDDIR}/image" "${PORTAGE_BUILDDIR}/homedir" \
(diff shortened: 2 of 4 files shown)


mgorny:
@zmedico, any suggestion where to put the empty dir cleanup to make it more reliable?

mgorny:
Hmm, I suppose an obvious solution might be to do it from bash.

mgorny:
> Hmm, I suppose an obvious solution might be to do it from bash.

…or not because it doesn't have permissions.

lib/portage/package/ebuild/prepare_build_dirs.py
```diff
@@ -71,6 +72,7 @@ def makedirs(dir_path):
 
 	mydirs = [os.path.dirname(mysettings["PORTAGE_BUILDDIR"])]
 	mydirs.append(os.path.dirname(mydirs[-1]))
+	mydirs.append(empty_dir)
```
zmedico, line 75: We also need to delete it in the `bin/phase-functions.sh` `__dyn_clean`function or else it will be preserved (this requirement stems from a desire to detect stray files or directories created by ebuilds).

ulm:
AFAICS this is the only missing EAPI 8 feature.

Any chance that we'll have a released version including this by the time of the next council meeting (2021-06-13)?

bin/ebuild.sh
```diff
@@ -186,12 +186,9 @@ export SANDBOX_ON=0
...
 # exploitation of insecure search path for python -c in ebuilds.
 # See bug #239560, bug #469338, and bug #595028.
-if [[ -d ${HOME} ]]; then
-	# Use portage's temporary HOME directory if available.
-	cd "${HOME}" || die
-else
-	cd "${PORTAGE_PYM_PATH}" || \
-		die "PORTAGE_PYM_PATH does not exist: '${PORTAGE_PYM_PATH}'"
+# EAPI 8 requires us to use an empty directory here.
+if [[ -d ${PORTAGE_BUILDDIR}/empty ]]; then
```
mgorny, line 190: Apparently we can't cd into it for some internal Portage phases (e.g. `clean`) but I'm not sure if we need to care about the starting directory for them anyway.

zmedico: We need to care for every phase, due to the persistent threat of issues like [bug 239560](https://bugs.gentoo.org/239560) and so on. So, if `${PORTAGE_BUILDDIR}/empty` doesn't exist then we should keep the existing behavior:
```
cd "${PORTAGE_PYM_PATH}" || die "PORTAGE_PYM_PATH does not exist: '${PORTAGE_PYM_PATH}'"
```

zmedico:
We have a _post_phase_userpriv_perms function for post-phase maintenance on `${T}`, so we could introduce a similar function for post-phase maintenance on `${PORTAGE_BUILDDIR}/empty`.  It will fit into EbuildPhase somewhere, and async is preferred but it doesn't necessarily have to be async because it shouldn't take long.

zmedico (approved):
Looks good!

mgorny:
689c79fc573879086aa809b62d9d4f4c8418f1fb

PR #722, opened 2021-05-29
Emit QA notices for setuptools warnings

mgorny:
Emit QA notices for setuptools warnings such as:

     * QA Notice: setuptools warnings detected:
     *
     *      Usage of dash-separated 'build-requires' will not be supported in future versions. Please use the underscore name 'build_requires' instead
     *      Usage of dash-separated 'description-file' will not be supported in future versions. Please use the underscore name 'description_file' instead
     *      Usage of dash-separated 'upload-dir' will not be supported in future versions. Please use the underscore name 'upload_dir' instead


mgorny:
Ping.

zmedico (approved):
Look good.

PR #723, opened 2021-05-31
Detect and report incorrect use of <stabilize-allarches/>

mgorny:
Report a QA warning when <stabilize-allarches/> is used on a package
installing ELF files.  We can easily check this through presence of
(non-empty) NEEDED build-info files.

### DIFF
--- a/bin/misc-functions.sh
+++ b/bin/misc-functions.sh
@@ -201,10 +201,15 @@ install_qa_check() {
 			echo "${QA_SONAME_NO_SYMLINK}" > \
 			"${PORTAGE_BUILDDIR}"/build-info/QA_SONAME_NO_SYMLINK
 
-		if has binchecks ${RESTRICT} && \
-			[ -s "${PORTAGE_BUILDDIR}/build-info/NEEDED.ELF.2" ] ; then
-			eqawarn "QA Notice: RESTRICT=binchecks prevented checks on these ELF files:"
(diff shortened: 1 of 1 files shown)


bin/misc-functions.sh
```diff
@@ -201,10 +201,15 @@ install_qa_check() {
 			echo "${QA_SONAME_NO_SYMLINK}" > \
 			"${PORTAGE_BUILDDIR}"/build-info/QA_SONAME_NO_SYMLINK
 
-		if has binchecks ${RESTRICT} && \
-			[ -s "${PORTAGE_BUILDDIR}/build-info/NEEDED.ELF.2" ] ; then
-			eqawarn "QA Notice: RESTRICT=binchecks prevented checks on these ELF files:"
-			eqawarn "$(while read -r x; do x=${x#*;} ; x=${x%%;*} ; echo "${x#${EPREFIX}}" ; done < "${PORTAGE_BUILDDIR}"/build-info/NEEDED.ELF.2)"
+		if [[ -s ${PORTAGE_BUILDDIR}/build-info/NEEDED.ELF.2 ]]; then
+			if grep -qs '<stabilize-allarches/>' "${EBUILD%/*}/metadata.xml"; then
```
zmedico, line 205: We also need to do `2>/dev/null` for this grep command in case this is a third-party repo and metadata.xml doesn't exist.

zmedico: Actually `grep -s` handles it that.

zmedico (approved):

bin/misc-functions.sh
```diff
@@ -201,10 +201,15 @@ install_qa_check() {
...
+				eqawarn "QA Notice: stabilize-allarches/> found on package installing ELF files"
```
Arfrever, line 206: Missing `<` before `stabilize-allarches/>` in this message?

PR #724, opened 2021-05-31
GLEP 75 layout fixes

doctaweeks:
It is not uncommon for PORTAGE_RO_DISTDIRS or GENTOO_MIRRORS to be used to point to a literal local mirror, not just a flat copy of files from DISTDIR. The GLEP 75 layout changes break these use cases when the upstream mirror begins using a non-flat layout.

These changes support parsing a layout.conf present in a local mirror, restoring the ability to symlink/copy distfiles and fixing the emerge output to indicate when distfiles do (not) actually need to be fetched.


zmedico (approved):
This looks great! Thank you!

In order to comply with [Gentoo's Certificate of Origin](https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin) policy, please if possible add a Signed-off-by line to each commit.

zmedico:
Merged in https://github.com/gentoo/portage/compare/95da84733ff80fec09601e677709197ff8b2a9a7...2fe8238b540240a62de3513839974e033f0bacd2.

PR #728, opened 2021-06-15
Consider autogroup scheduling (v2)

Flowdalic:


lib/_emerge/actions.py
```diff
@@ -2655,18 +2655,37 @@ def nice(settings):
...
+		return
 
-	autogroup_file = Path("/proc/self/autogroup")
-	try:
-		f = autogroup_file.open("r+")
-	except EnvironmentError:
+	sched_autogroup = Path("/proc/sys/kernel/sched_autogroup_enabled")
+	if not sched_autogroup.is_file():
 		# Autogroup scheduling is not enabled on this system.
 		return
```
zmedico, line 2663: The kind of TOCTOU race that I want it to handle would be where the `/proc/sys/kernel/sched_autogroup_enabled` state changed right during the critical section here.

lib/_emerge/actions.py
```diff
@@ -2655,18 +2655,37 @@ def nice(settings):
...
 
-	with f:
+	try:
+		# Will be "1" (enabled) or "0" disabled.
+		sched_autogroup_enabled = int(sched_autogroup.read_text())
```
zmedico, line 2667: The kind of TOCTOU race that I want to be sure we handle would be where the /proc/sys/kernel/sched_autogroup_enabled state changes before this function returns.

amadio: I think that the problem in my prefix system is that the kernel is just too old (2.6.32). Even when the feature is enabled and I just do `echo 10 > /proc/self/autogroup` as root, it fails.

Flowdalic: > The kind of TOCTOU race that I want to be sure we handle would be where the /proc/sys/kernel/sched_autogroup_enabled state changes before this function returns.

With this PR, If the value changes then between reading `sched_autogroup_enabled` and writing `/proc/self/autogroup` from 1 to 0, then the write will fail with an error message and the `atexit()` handler won't be installed. I think that is sensible, even though the TOCTOU still exists, but "only" results in one error message, and not two, as with the current master (one for the write exception and one from the atexit handler write throwing).

I am not sure how this TOCTOU can be prevented in this case, but maybe I am missing something.

Flowdalic: > This check can actually be removed,

I tend towards keeping the check, as it prevents an error message, due the exception thrown by `write(/proc/self/autogroup)`, in case `sched_autogroup_enabled` is 0.

amadio: On a different system with newer kernel (3.10), no errors occur, whether the feature is enabled or not. The value just doesn't change if the feature is disabled. Only with kernel 2.6.32 it fails no matter what I do.

Flowdalic:
@zmedico To conclude, I suggest the inclusion of the first two commits, the latest commit is not strictly necessary (but can't hurt, besides increasing the code size).

lib/_emerge/actions.py
```diff
@@ -2634,13 +2635,55 @@ def apply_priorities(settings):
...
 	except (OSError, ValueError) as e:
 		out = portage.output.EOutput()
-		out.eerror("Failed to change nice value to '%s'" % \
-			settings.get("PORTAGE_NICENESS", "0"))
+		out.eerror(f"Failed to change nice value to {nice_value}")
+		out.eerror("%s\n" % str(e))
+		return
+
+	sched_autogroup = Path("/proc/sys/kernel/sched_autogroup_enabled")
+	if not sched_autogroup.is_file():
```
zmedico, line 2649: Why can't we eliminate this stat call?

zmedico: I need a really really really solid argument for why this stat call is necessary.

zmedico: Please have amadio test a patch without the unecessary stat call, and get him to sign off on it.

zmedico:
@amadio I'm not going to approve this unless you sign off and say that it works.

zmedico:
> @amadio I'm not going to approve this unless you sign off and say that it works.

@juippis same to you please.

lib/_emerge/actions.py
```diff
@@ -2634,13 +2635,45 @@ def apply_priorities(settings):
 	nice(settings)
 
 def nice(settings):
+	nice_value: str = settings.get("PORTAGE_NICENESS", "0")
```
amadio, line 2638: There was a version of these changes that did nothing unless `PORTAGE_NICENESS` was explicitly set in the environment. We need that version to avoid failures on prefix with old kernels. This version will always try to write to `/proc/self/autogroup` and that will fail on old kernels (even as root, simply doing `echo 10 > /proc/self/autogroup` fails on 2.6.32, which is still unfortunately in use in very old systems).

amadio (changes requested):

lib/_emerge/actions.py
Flowdalic: Right, the somehow the improved version got lost: https://github.com/gentoo/portage/commit/055abe523c2c3f6c8f1dccfb53565209222f90c1#diff-dbd67931d37002a46abae37da609fef65a83cf1c90fe57325b2f73e0db831664R2638

Flowdalic:
@amadio restored the improved version

amadio:
Should be good to merge now. You can add my signoff. Thanks!

juippis (approved):
Can confirm, the current @Flowdalic's autogroup branch works for me.
AFAIK it's rebased correctly on top of portage's head right now.

amadio:
There's a typo in the third paragraph of the commit message of `eb2baad`: `orignal`. Not critical, just letting you know.

Flowdalic:
> There's a typo in the third paragraph of the commit message of `eb2baad`: `orignal`. Not critical, just letting you know.

Thanks, fixed.

Flowdalic:
@juippis @amadio I dramatically reduced the size and invasiveness, while increasing its robustness against failures, of the autogroup patch. May I ask you to test the current version if you have the cycles? That would be much appreciated. Thanks.

juippis (approved):
This latest revision works for me. Used portage for ~30 min for normal daily activities.

amadio:
It doesn't look like `pid-ns-init` is used on prefix, so I guess you don't have to worry about older kernels there. Maybe @zmedico knows better and can comment? This won't work as is if it's run on older kernels (i.e. if for some reason `pid-ns-init` starts being used, it may cause failures).

juippis (approved):
@amadio sorry I hadn't seen your ping, guess it doesn't notify when you edit your message.

Yes I have `FEATURES=pid-sandbox` set.

```
# emerge --info
Portage 3.0.21 (python 3.9.6-final-0, default/linux/amd64/17.1, gcc-11.1.0, glibc-2.33-r1, 5.12.0-pf6 x86_64)
=================================================================
System uname: Linux-5.12.0-pf6-x86_64-AMD_Ryzen_7_3700X_8-Core_Processor-with-glibc2.33
KiB Mem:    32862504 total,  32063080 free
KiB Swap:          0 total,         0 free
Timestamp of repository gentoo: Mon, 05 Jul 2021 03:38:09 +0000
Head commit of repository gentoo: 057273e7ada3c1d859f5bcde6bf193ccf06ea62e

sh bash 5.1_p8
ld GNU ld (Gentoo 2.36.1 p3) 2.36.1
app-shells/bash:          5.1_p8::gentoo
dev-lang/perl:            5.34.0::gentoo
dev-lang/python:          3.8.11::gentoo, 3.9.6::gentoo, 3.10.0_beta3::gentoo
dev-lang/rust-bin:        1.53.0::gentoo
dev-util/cmake:           3.20.5::gentoo
sys-apps/baselayout:      2.7-r3::gentoo
sys-apps/openrc:          0.43.3::gentoo
sys-apps/sandbox:         2.24::gentoo
sys-devel/autoconf:       2.13-r1::gentoo, 2.69-r5::gentoo
sys-devel/automake:       1.16.3-r1::gentoo
sys-devel/binutils:       2.36.1-r1::gentoo
sys-devel/gcc:            10.3.0-r1::gentoo, 11.1.0-r1::gentoo
sys-devel/gcc-config:     2.4::gentoo
sys-devel/libtool:        2.4.6-r6::gentoo
sys-devel/make:           4.3::gentoo
sys-kernel/linux-headers: 5.12::gentoo (virtual/os-headers)
sys-libs/glibc:           2.33-r1::gentoo
Repositories:

gentoo
    location: /var/db/repos/gentoo
    sync-type: git
    sync-uri: https://anongit.gentoo.org/git/repo/sync/gentoo.git
    priority: -1000

ACCEPT_KEYWORDS="amd64 ~amd64"
ACCEPT_LICENSE="*"
CBUILD="x86_64-pc-linux-gnu"
CFLAGS="-march=native -O2 -pipe"
CHOST="x86_64-pc-linux-gnu"
CONFIG_PROTECT="/etc /usr/share/gnupg/qualified.txt"
CONFIG_PROTECT_MASK="/etc/ca-certificates.conf /etc/dconf /etc/env.d /etc/fonts/fonts.conf /etc/gconf /etc/gentoo-release /etc/revdep-rebuild /etc/sandbox.d /etc/terminfo"
CXXFLAGS="-march=native -O2 -pipe"
DISTDIR="/var/cache/distfiles"
EMERGE_DEFAULT_OPTS="--jobs=8 --usepkg --autounmask=y --autounmask-write --autounmask-continue --autounmask-use=y --nospinner --buildpkg-exclude virtual --buildpkg-exclude acct-*"
ENV_UNSET="CARGO_HOME DBUS_SESSION_BUS_ADDRESS DISPLAY GOBIN GOPATH PERL5LIB PERL5OPT PERLPREFIX PERL_CORE PERL_MB_OPT PERL_MM_OPT XAUTHORITY XDG_CACHE_HOME XDG_CONFIG_HOME XDG_DATA_HOME XDG_RUNTIME_DIR"
FCFLAGS="-march=native -O2 -pipe"
FEATURES="assume-digests binpkg-docompress binpkg-dostrip binpkg-logs binpkg-multi-instance buildpkg config-protect-if-modified distlocks ebuild-locks fixlafiles ipc-sandbox merge-sync multilib-strict network-sandbox news parallel-fetch parallel-install pid-sandbox preserve-libs protect-owned qa-unresolved-soname-deps sandbox sfperms strict strict-keepdir unknown-features-warn unmerge-logs unmerge-orphans userfetch userpriv usersandbox usersync xattr"
FFLAGS="-march=native -O2 -pipe"
GENTOO_MIRRORS="https://ftp-stud.hs-esslingen.de/pub/Mirrors/gentoo/"
LANG="C.UTF-8"
LDFLAGS="-Wl,-O1 -Wl,--as-needed"
LINGUAS=" en en_US en-US fi sv "
MAKEOPTS="-j16 -l10"
PKGDIR="/var/cache/binpkgs"
PORTAGE_CONFIGROOT="/"
PORTAGE_RSYNC_OPTS="--recursive --links --safe-links --perms --times --omit-dir-times --compress --force --whole-file --delete --stats --human-readable --timeout=180 --exclude=/distfiles --exclude=/local --exclude=/packages --exclude=/.git"
PORTAGE_TMPDIR="/var/tmp"
USE="X acl amd64 bzip2 cli crypt dbus dri elogind fortran gdbm iconv ipv6 libglvnd libtirpc multilib ncurses nptl openmp openrc pam pcre readline seccomp split-usr ssl tcpd udev unicode xattr zlib" ABI_X86="64 32" ADA_TARGET="gnat_2018" ALSA_CARDS="ali5451 als4000 atiixp atiixp-modem bt87x ca0106 cmipci emu10k1x ens1370 ens1371 es1938 es1968 fm801 hda-intel intel8x0 intel8x0m maestro3 trident usb-audio via82xx via82xx-modem ymfpci" APACHE2_MODULES="authn_core authz_core socache_shmcb unixd actions alias auth_basic authn_alias authn_anon authn_dbm authn_default authn_file authz_dbm authz_default authz_groupfile authz_host authz_owner authz_user autoindex cache cgi cgid dav dav_fs dav_lock deflate dir disk_cache env expires ext_filter file_cache filter headers include info log_config logio mem_cache mime mime_magic negotiation rewrite setenvif speling status unique_id userdir usertrack vhost_alias" CALLIGRA_FEATURES="karbon sheets words" COLLECTD_PLUGINS="df interface irq load memory rrdtool swap syslog" CPU_FLAGS_X86="aes avx avx2 f16c fma3 mmx mmxext pclmul popcnt sha sse sse2 sse3 sse4_1 sse4_2 sse4a ssse3" ELIBC="glibc" GPSD_PROTOCOLS="ashtech aivdm earthmate evermore fv18 garmin garmintxt gpsclock greis isync itrax mtk3301 nmea ntrip navcom oceanserver oldstyle oncore rtcm104v2 rtcm104v3 sirf skytraq superstar2 timing tsip tripmate tnt ublox ubx" INPUT_DEVICES="evdev" KERNEL="linux" L10N="en en_US en-US fi sv" LCD_DEVICES="bayrad cfontz cfontz633 glk hd44780 lb216 lcdm001 mtxorb ncurses text" LIBREOFFICE_EXTENSIONS="presenter-console presenter-minimizer" LLVM_TARGETS="AMDGPU BPF" LUA_SINGLE_TARGET="lua5-1" LUA_TARGETS="lua5-1" OFFICE_IMPLEMENTATION="libreoffice" PHP_TARGETS="php7-3 php7-4" POSTGRES_TARGETS="postgres10 postgres11" PYTHON_SINGLE_TARGET="python3_9" PYTHON_TARGETS="pypy3 python3_8 python3_9 python3_10" RUBY_TARGETS="ruby26" USERLAND="GNU" VIDEO_CARDS="amdgpu radeonsi" XTABLES_ADDONS="quota2 psd pknock lscan length2 ipv4options ipset ipp2p iface geoip fuzzy condition tee tarpit sysrq proto steal rawnat logmark ipmark dhcpmac delude chaos account"
Unset:  CC, CPPFLAGS, CTARGET, CXX, INSTALL_MASK, LC_ALL, PORTAGE_BINHOST, PORTAGE_BUNZIP2_COMMAND, PORTAGE_COMPRESS, PORTAGE_COMPRESS_FLAGS, PORTAGE_RSYNC_EXTRA_OPTS, RUSTFLAGS
```

Once again, the current version works for me.

thesamesam:
@zmedico Would you mind applying this now given everybody seems happy?

mgorny:
Could you add the signoff to the commits?

mgorny:
> Could you add the signoff to the commits?

Ah, sorry, I've failed at pram ;-).

simonvanderveldt:
Was wondering how one can validate that this is working? The way I currently think it can be validated it doesn't seem to be working, but maybe I'm doing something wrong.

```
$ ps -C emerge                      
  PID TTY          TIME CMD
32017 pts/1    00:00:02 emerge
$ cat /proc/32017/autogroup
/autogroup-42 nice 0
```

Flowdalic:
Portage's autogroup nice'ing is based on the PID namespaces that portage creates per emerge job. Those jobs set their autogroup nice value, the main emerge process does not set an autogroup nice value.

PR #730, opened 2021-06-16
man/emerge.1: Fix typo in --with-bdeps description

thesamesam:


mjeveritt (commented):
For the non-man-fluent .. what does the change convert?

thesamesam:
Yeah, I'm definitely not fluent either yet..

It converts the output from:
>In dependency calculations, pull in build time dependencies that are not strictly required. This option is automatically enabled for installation actions, meaning they will be installed, and defaults to aqyaq for

to
>In dependency calculations, pull in build time dependencies that are not strictly required. This option is automatically enabled for installation actions, meaning they will be installed, and defaults to **y** for

PR #731, opened 2021-06-18
WIP: Add @unsatisfied-deps package set (bug 248026)

zmedico:
If emerge --depclean fails to resolve any dependencies, then it will
now suggest emerge `@unsatisfied-deps` as the simplest possible
solution, and will also suggest to unmerge `@unavailable` where
appropriate at the end:
```
$ emerge --depclean

 * Dependencies could not be completely resolved due to
 * the following required packages not being installed:
 *
 *   virtual/cdrtools pulled in by:
 *     app-cdr/cdrdao-1.2.4
 *
 * Have you forgotten to resolve unsatisfied dependencies prior to
 * depclean? The simplest possible command for this purpose is as
 * follows:
 *
 *   emerge @unsatisfied-deps
 *
 * The most comprehensive possible update command is this:
 *
 *   emerge --update --newuse --deep --with-bdeps=y @world
 *
 * Note that the --with-bdeps=y option is not required in many
 * situations. Refer to the emerge manual page (run `man emerge`)
 * for more information about --with-bdeps.
 *
 * Also, note that it may be necessary to manually uninstall
 * packages that no longer exist in the repository, since it may not
 * be possible to satisfy their dependencies. The simplest possible
 * command for this purpose is as follows, but be careful to examine
 * the resulting package list carefully:
 *
 *   emerge --ask --unmerge @unavailable
 *
```
Bug: https://bugs.gentoo.org/248026

### DIFF
--- a/doc/config/sets.docbook
+++ b/doc/config/sets.docbook
@@ -610,6 +610,13 @@
 			</itemizedlist>
 			</sect3>
 		</sect2>
+		<sect2 id='config-set-classes-UnsatisfiedDepsSet'>
+		<title>portage.sets.dbapi.UnsatisfiedDepsSet</title>
+		<para>
+		Package set which contains all installed packages
+		having one or more unsatisfied runtime dependencies.
+		</para>
+		</sect2>
 	</sect1>
 	
 	<sect1 id='config-set-defaults'>
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -1000,9 +1000,16 @@ def _calc_depclean(settings, trees, ldpath_mtimes,
 				msg.append("    %s" % (parent,))
 				msg.append("")
 			msg.extend(textwrap.wrap(
-				"Have you forgotten to do a complete update prior " + \
-				"to depclean? The most comprehensive command for this " + \
+				"Have you forgotten to resolve unsatisfied dependencies prior "
+				"to depclean? The simplest possible command for this "
 				"purpose is as follows:", 65
+			))
+			msg.append("")
+			msg.append("  " + \
+				good("emerge @unsatisfied-deps"))
+			msg.append("")
+			msg.extend(textwrap.wrap(
+				"The most comprehensive possible update command is this:", 65
 			))
 			msg.append("")
 			msg.append("  " + \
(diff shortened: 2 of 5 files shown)


doc/config/sets.docbook
```diff
@@ -610,6 +610,13 @@
 			</itemizedlist>
 			</sect3>
 		</sect2>
+		<sect2 id='config-set-classes-UnsatisfiedDepsSet'>
+		<title>portage.sets.dbapi.UnsatisfiedDepsSet</title>
+		<para>
+		Package set which contains all installed packages packages
```
mjeveritt, line 616: per email - copy/pasta on "packages packages "

lib/portage/_sets/dbapi.py
```diff
@@ -303,6 +311,69 @@ def singleBuilder(cls, options, settings, trees):
 
 	singleBuilder = classmethod(singleBuilder)
 
+class UnsatisfiedDepsSet(PackageSet):
+
+	_operations = ["merge", "unmerge"]
+
+	description = (
+		"Package set which contains all installed packages packages "
```
mjeveritt, line 319: more copy-pasta (as before...) :smiley:

zmedico: fixed thanks!

doc/config/sets.docbook
zmedico: fixed thanks!

lib/_emerge/actions.py
```diff
@@ -1000,11 +1000,18 @@ def unresolved_deps():
...
 			msg.extend(textwrap.wrap(
-				"Have you forgotten to do a complete update prior " + \
-				"to depclean? The most comprehensive command for this " + \
+				"Have you forgotten to resolve unsatisfied dependencies prior "
+				"to depclean? The simplest possible command for this "
 				"purpose is as follows:", 65
 			))
 			msg.append("")
+			msg.append("  " + \
+				good("emerge @unsatisfied-deps"))
```
zmedico, line 1009: Needs `--selective`.

zmedico, line 1009: It turns out that `@unsatisfied-deps` is often unsuitable here because it
pulls in a bunch of `@installed` packages, when you really want to use
`@world` as the source of truth.

zmedico: My plan is it introduce an `@unsatisfied-world` set which is equivalent to `@unsatisfied-deps` but filters out any non @world packages.

PR #734, opened 2021-06-20
Use PORTAGE_PROPERTIES and PORTAGE_RESTRICT internally

mgorny:
Use PORTAGE_PROPERTIES and PORTAGE_RESTRICT directly for internal
purposes rather than relying on PROPERTIES and RESTRICT having
the flattened values.


zmedico (approved):
Looks good, please merge.

PR #736, opened 2021-08-01
Bugfix 674212: Warn if portage home directory is missing

steeznson:
Signed-off-by: James Stevenson <james.al.stevenson@gmail.com>

### DIFF
--- a/lib/portage/package/ebuild/config.py
+++ b/lib/portage/package/ebuild/config.py
@@ -19,7 +19,10 @@ import warnings
 from _emerge.Package import Package
 import portage
 portage.proxy.lazyimport.lazyimport(globals(),
+	'portage.data:_portage_grpname',
+	'portage.data:_portage_username',
 	'portage.data:portage_gid',
+	'portage.data:portage_uid',
 	'portage.dep.soname.SonameAtom:SonameAtom',
 	'portage.dbapi.vartree:vartree',
 	'portage.package.ebuild.doebuild:_phase_func_map',
@@ -1086,6 +1089,24 @@ class config:
 					noiselevel=-1)
 				writemsg("!!! %s\n" % str(e),
 					noiselevel=-1)
+
+	def _check_portage_homedir(self):
+		"""
+		Warn if portage home directory is missing or has incorrect ownership
+		"""
+		try:
+			stat = os.stat(os.path.expanduser(f"~{_portage_username}"))
+			stat_gid = stat.st_gid
+			stat_uid = stat.st_uid
+			assert stat_gid == portage_gid
+			assert stat_uid == portage_uid
+		except (FileNotFoundError, AssertionError):
+			writemsg(
+				colorize("BAD",
+				_(f"!!! '{_portage_username}' user home directory either "
+				+ "does not exist, or is not owned by "
+				+ f"'{_portage_grpname}' group and '{_portage_username}' user."))
+				+ "\n", noiselevel=-1)
 
 	@property
 	def _keywords_manager(self):
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -2508,7 +2508,7 @@ def load_emerge_config(emerge_config=None, env=None, **kargs):
 	emerge_config.trees = portage.create_trees(trees=emerge_config.trees,
 				**kwargs)
 
-	for root_trees in emerge_config.trees.values():
+	for tree_key, root_trees in emerge_config.trees.items():
 		settings = root_trees["vartree"].settings
 		settings._init_dirs()
 		setconfig = load_default_config(settings, root_trees)
@@ -2519,6 +2519,10 @@ def load_emerge_config(emerge_config=None, env=None, **kargs):
 			root_trees["root_config"].update(root_config)
 		else:
 			root_trees["root_config"] = root_config
+
+		if tree_key == "/":
+			# warn once if portage home dir missing
+			settings._check_portage_homedir()
 
 	target_eroot = emerge_config.trees._target_eroot
 	emerge_config.target_config = \


lib/portage/data.py
```diff
@@ -178,6 +178,14 @@ def _get_global(k):
 				noiselevel=-1)
 			portage_group_warning()
 
+		portage_homedir_exists = os.access(
+				first_existing(os.path.expanduser("~portage")), portage.os.W_OK)
```
mgorny, line 182: I don't understand the purpose of the `first_existing()` call here.

steeznson: I was copying a pattern I'd seen elsewhere to check for existing directories and write access. 
Looking at it again, I suppose in this case `first_existing()` might lead to a false positive if portage was being run as root. I'll change it now.

lib/portage/data.py
```diff
@@ -178,6 +178,14 @@ def _get_global(k):
 				noiselevel=-1)
 			portage_group_warning()
 
+		portage_homedir_exists = os.access(
+				os.path.expanduser("~portage"), portage.os.W_OK)
```
mgorny, line 182: I think X_OK is worth checking too.

lib/portage/data.py
```diff
@@ -178,6 +178,14 @@ def _get_global(k):
...
+		if not portage_homedir_exists:
+			writemsg(
+				colorize("BAD",
+				  _("portage: home directory is missing for 'portage' user.")) \
```
mgorny, line 186: The error will be confusing if the directory exists but has wrong permissions.

mgorny (changes requested):

lib/portage/data.py
```diff
@@ -178,6 +178,17 @@ def _get_global(k):
 				noiselevel=-1)
 			portage_group_warning()
 
+		portage_homedir_writable = os.access(
+				os.path.expanduser("~portage"), portage.os.W_OK)
+		portage_homedir_executable = os.access(
+				os.path.expanduser("~portage"), portage.os.X_OK)
```
mgorny, line 184: You can just do a bitwise-OR of all requested flags (and I suppose check `R` then too).

steeznson: Took me a second to realise this meant changing the 2nd `os.access()` param to -

```
portage.os.R_OK | portage.os.W_OK | portage.os.X_OK
```

It's been fixed now though.

lib/portage/data.py
```diff
@@ -178,6 +178,16 @@ def _get_global(k):
 				noiselevel=-1)
 			portage_group_warning()
 
+		portage_homedir_ready = \
+				os.access(os.path.expanduser("~portage"),
+						  portage.os.R_OK | portage.os.W_OK | portage.os.X_OK)
```
mgorny, line 183:
```suggestion
		portage_homedir_ready = (
				os.access(os.path.expanduser("~portage"),
						  os.R_OK | os.W_OK | os.X_OK))
```

Please don't use backslash-wrapping, or `portage.os` wrapper in the new code.

mgorny (approved):
Two more nits but I'm not going to bother you more (sorry!) and just fix them while pushing.

Thanks for your work!

jospezial:
Why do I see the message and what can I do?
```
portage: unable to utilise 'portage' user home directory. Ensure that it exists with the correct permissions.

/etc/passwd: portage:/var/lib/portage/home:/sbin/nologin
drwxr-xr-x 2 portage portage 4096 28. Apr 14:11 /var/lib/portage/home
jlgentoo ~ # cd ~portage/
jlgentoo /var/lib/portage/home # 
```

thesamesam:
Could you please file a bug with the full build.log of it happening?

joecool1029:
Filed: https://bugs.gentoo.org/811930

mgorny:
I'm sorry but I've reverted this since the check seems to be overzealous. For example, the error keeps triggering when Portage API is used from other tools, and this certainly doesn't require Portage user homedir to be present.

steeznson:
Maybe this warning needs to be logged somewhere instead of printed. I hadn't realised how noisy it was. 
Can also change 'utilize' to the American spelling I guess.

mgorny:
I think that we either need to find a better place for the check, or add some condition to e.g. do the check once before emerge starts doing stuff.

steeznson:
No worries, that's fine with me. I'll take another look at it.

lib/portage/data.py
```diff
@@ -178,6 +178,16 @@ def _get_global(k):
...
+		if not portage_homedir_ready:
+			writemsg(
+				colorize("BAD",
+				  _("portage: unable to utilise 'portage' user home directory."
```
floppym, line 187: Please use the American English spelling of "utilize".

steeznson: No worries, I can certainly change it. I did not realise I'd used British English at the time but will keep an eye out for typos like this in the future.

floppym:
Something to note for future revisions: the default sandbox config does not permit writing to /var/lib/portage/home, so checking `os.W_OK` will fail for any code running within the sandbox environment.

See [bug 811930](https://bugs.gentoo.org/811930).

steeznson:
I've re-written the check so that it only occurs once, and have monitored the tox targets during the CI to double check the warning isn't being spammed. However, after reflecting on the original ticket again I'm beginning to think that this check on the home directory might be extraneous; potentially it's going to add more complexity to the code for marginal benefit. I'm happy to close the PR and look at a different bug ticket if that makes more sense in the long term.

lib/portage/package/ebuild/config.py
```diff
@@ -1087,6 +1088,24 @@ def _init_dirs(self):
 				writemsg("!!! %s\n" % str(e),
 					noiselevel=-1)
 
+	def _check_portage_homedir(self):
+		"""
+		Warn if portage home directory is missing or has incorrect ownership
+		"""
+		try:
+			stat = os.stat(os.path.expanduser("~portage"))
```
Arfrever, line 1096: You should not hardcode `portage`, but respect `portage.data._portage_username`.

`portage.data._portage_username` and `portage.data._portage_grpname` respect `PORTAGE_USERNAME` and `PORTAGE_GRPNAME` variables (which can be set in environment or `make.conf`).

lib/portage/package/ebuild/config.py
```diff
@@ -1087,6 +1088,24 @@ def _init_dirs(self):
...
+			stat_gid = stat.st_gid
+			stat_uid = stat.st_uid
+			assert stat_gid == portage_gid
+			assert stat_uid == portage_uid
+		except (FileNotFoundError, AssertionError):
+			writemsg(
+				colorize("BAD",
+				_("portage: 'portage' user home directory either "
+				+ "does not exist, or is not owned by "
+				+ "'portage' group and user."))
```
Arfrever, line 1106: And in this message.

lib/portage/package/ebuild/config.py
```diff
@@ -1087,6 +1090,24 @@ def _init_dirs(self):
...
+		try:
+			stat = os.stat(os.path.expanduser(f"~{_portage_username}"))
+			stat_gid = stat.st_gid
+			stat_uid = stat.st_uid
+			assert stat_gid == portage_gid
+			assert stat_uid == portage_uid
+		except (FileNotFoundError, AssertionError):
+			writemsg(
+				colorize("BAD",
+				_("portage: 'portage' user home directory either "
```
Arfrever, line 1106: Please also update this instance of `portage` to `{_portage_username}`.

Arfrever: Also I am not sure if this message should have `portage: ` prefix...

PR #741, opened 2021-08-20
Bugfix 343119: unclear error output when $PKGDIR is full 

steeznson:


mgorny (changes requested):
Thank you for your contribution but tbh I don't think that's the right way of handling that. Could you be more specific what output would you like to address?

steeznson:
Thanks for taking the time to look it over! The original bug report suggested this error:

```
IOError: [Errno 5] Input/output error
```

Was not a clear error in the event that the $PKGDIR mountpoint had no free space left. 
From what I could see looking at the code the error happens in the `_ensure_dir()` method in the `binarytree` class, so my thinking was to check the available space when a new binarytree instance was being initialised. If no space was found then I was hoping it would abort with a more descriptive error message.

I can close this pull request for now and try approaching the problem from a different angle.

mgorny:
Well, I think the best approach would be to wrap the relevant code in a `try`/`except` and amend the exception with some explanation.

PR #743, opened 2021-09-03
Modern metadata passing API & add INHERIT key to md5-cache

mgorny:
This patchset removes some leftover cruft after old cache support, switches the internal metadata passing between `ebuild.sh` and Python to use key-value format instead of old cache fixed line format and finally adds an `INHERIT` md5-cache key for better compatibility with PkgCore.


Arfrever:
There is already `INHERITED` key which contains names of eclasses inherited directly or indirectly.
Name `INHERIT` is not informative about content of this variable and will cause confusion for API consumers.
Could you rename this new key to e.g. `DIRECTLY_INHERITED`?

mgorny:
No.

PR #744, opened 2021-09-10
Detect corrupt ELF files in VDB

thesamesam:
A series of fixes for failure to preserve libraries in some situations.

Bug: https://bugs.gentoo.org/811462


lib/portage/package/ebuild/doebuild.py
```diff
@@ -2467,12 +2467,25 @@ def is_libdir(obj_parent):
...
+		# Let's check if we've got inconsistent results.
+		# If we're installing dynamic libraries (.so files), we should
+		# really have a PROVIDES.
+		# (This is a complementary check at the point of creation for the
+		# ingestion check in Binpkg.py)
+		# Note: we could check a non-empty PROVIDES against the list of .sos,
+		#		but this doesn't gain us anything. We're interested in failure
+		#		to properly parse the installed files at all, which should really
+		#		be a global problem (e.g. bug #811462)
+		installed_dynlibs = glob.glob(image_dir + '/**/*.so', recursive=True)
```
Arfrever, line 2486: This will cause false positives for almost all `.so` modules of Python, Perl, Ruby etc. Most of them do not have SONAMEs (and there is no reason for them to have SONAMEs).

Checking only `image_dir + '/lib*/lib*.so'` and `image_dir + '/usr/lib*/lib*.so'` would reduce number of false positives, but maybe some languages or applications put their modules directly in `/lib*` or `/usr/lib*` (instead of in some subdirectories).

lib/_emerge/Binpkg.py
```diff
@@ -344,6 +345,23 @@ def _unpack_contents_exit(self, unpack_contents):
...
+			# Let's check if we've got inconsistent results.
+			# If we're installing dynamic libraries (.so files), we should
+			# really have a PROVIDES.
+			# (This is a complementary check at the point of ingestion for the
+			# creation check in doebuild.py)
+			# Note: we could check a non-empty PROVIDES against the list of .sos,
+			#		but this doesn't gain us anything. We're interested in failure
+			#		to properly parse the installed files at all, which should really
+			#		be a global problem (e.g. bug #811462)
+			installed_dynlibs = glob.glob(self.settings["D"] + '/**/*.so', recursive=True)
```
Arfrever, line 360: The same problem here.

thesamesam: You don't say! ;)

lib/portage/package/ebuild/doebuild.py
thesamesam: Ah, I was worried about this (it is a draft -- but thank you for the comments).

Not checked yet but I'd expect just /lib* and /usr/lib* to be fine (although may need a way to obtain get_libdir if possible to avoid recursion usage). I agree on excluding subdirectories, as it's really unlikely they're going to be the only libraries installed (only one is needed for us to realise there's a problem) and it helps avoid false positives.

Ideas for better heuristics welcome too, but ideally something we could do from pure Python without having to invoke *elf again (which in theory we could do when ingesting binpkgs to compare, but I don't really want to do this - but is not possible in the normal install case).

hlein: Hm, should all registered libdirs (say by reading `/etc/ld.so.conf{,.d/*}`, or dumping `/etc/ld.so.cache`) be examined?

Because there are libs that are in PROVIDES and NEEDED* files that are not just in {,/usr}/lib*:

```
$ strings /etc/ld.so.cache | egrep -v '^[^/]|^(/usr)?/lib(64)?/[^/]+$' | wc -l
115
$ strings /etc/ld.so.cache | egrep -v '^[^/]|^(/usr)?/lib(64)?/[^/]+$' | sed 's%/[^/]*$%%' | sort | uniq -c
     52 /opt/cuda/lib64
      2 /opt/cuda/nvvm/lib64
     21 /usr/lib/gcc/x86_64-pc-linux-gnu/10.3.0
     15 /usr/lib/gcc/x86_64-pc-linux-gnu/10.3.0/32
      5 /usr/lib/llvm/12/lib
      9 /usr/lib/llvm/12/lib64
      3 /usr/lib/rust/lib
      8 /usr/lib64/fltk
$ strings /etc/ld.so.cache  | egrep /usr/lib/gcc/x86_64-pc-linux-gnu/10.3.0
...
/usr/lib/gcc/x86_64-pc-linux-gnu/10.3.0/libstdc++.so
...
$ equery b libstdc++.so
 * Searching for libstdc++.so ... 
sys-devel/gcc-10.3.0-r2 (/usr/lib/gcc/x86_64-pc-linux-gnu/10.3.0/32/libstdc++.so -> libstdc++.so.6.0.28)
sys-devel/gcc-10.3.0-r2 (/usr/lib/gcc/x86_64-pc-linux-gnu/10.3.0/libstdc++.so -> libstdc++.so.6.0.28)
$ egrep -l 'libstdc\+\+.so' /var/db/pkg/*/*/NEEDED* | sed 's%/[^/]*$%%' | sort -u | wc -l
208
```

Many .so's in non-main-paths are only NEEDED by the same package that PROVIDES them, so no big deal. But f'rinstance if someone had upgraded gcc during the time when scanelf was silently broken, their /var/db/pkg/sys-devel/gcc*/ may be missing a PROVIDES file, and they are now in danger of breaking many things if they upgrade to a gcc that changes the libstdc++.so soname, no?

Arfrever: > Ideas for better heuristics welcome too, but ideally something we could do from pure Python without having to invoke *elf again

You might consider using `dev-python/pyelftools`...
`dev-python/pyelftools` already happens to be dependency of `app-misc/pax-utils[python]`.
In Portage's source code, usage of `elftools` modules would have to be guarded by `try: ... except ImportError: pass`, but I would be fine with ebuilds of `sys-apps/portage` having almost unconditional dependency on `dev-python/pyelftools` (in `!build? ( ... )` branch of `RDEPEND`).

Arfrever: > if they upgrade to a gcc that changes the libstdc++.so soname

`SONAME` of `libstdc++.so` has not changed in IIRC about 20 years.
Upstreams of glibc and GCC prefer to use symbol versioning, and they sometimes even break ABI without `SONAME` bumps.

hlein: > `SONAME` of `libstdc++.so` has not changed in IIRC about 20 years.

Hah, true, I'm misled by still sometimes encountering `libstdc++` 4.x and 5.x on other distros...

Still, seems like this applies to llvm libs under `/usr/lib/llvm/${major_version}/lib{,64}/`, fltk libs under `/usr/${libdir}/fltk/`, nvidia-cuda-toolkit libs under `/opt/cuda/...`, etc.? And that's just what I happen to find on my desktop, no scientific or exhaustive check of packages that set `LDPATH=` env files (and then checking for other packages that actually depend on & have `NEEDED` entries for those, vs them being purely self-contained like rust libs appear to me to be).

thesamesam: In the end, I don't think this is actually an issue:
- We check for PROVIDES which is always around with NEEDED for the Perl case
- The '**' descends recursively into all subdirectories so we're fine there wrt paths

PR #750, opened 2021-09-21
bin/misc-function.sh: check scanelf return code

thesamesam:
This is part of a series of fixes for the linked bug (failure
to preserve libraries in some situations).

We need to check if scanelf failed when calling it in the
installed-files QA check as we later use it to populate the VDB.

Silently continuing results in either blank e.g. PROVIDES,
NEEDED{,.ELF.2} or those files may be missing entirely,
resulting in a corrupt state both on the system and in
any generated binpkgs.

Adds an escape variable (PORTAGE_NO_SCANELF_CHECK) to allow
re-emerging pax-utils if it's broken.

Bug: https://bugs.gentoo.org/811462


bin/misc-functions.sh
```diff
@@ -177,11 +177,45 @@ install_qa_check() {
...
+					eerror "Please report this bug at https://bugs.gentoo.org/!"
+					eerror "It may be possible to re-emerge pax-utils with:"
+					eerror " PORTAGE_NO_SCANELF_CHECK=1 emerge -v1 app-misc/pax-utils"
+					eerror "Aborting to avoid corrupting metadata"
+					die "${0##*/}: Failed to run scanelf!"
+				fi
+				;;
+		esac
+
+		[[ -n ${scanelf_output} ]] && echo "${scanelf_output[@]}" | { while IFS= read -r l; do
```
mgorny, line 218: Why the `[@]`? You don't declare it as an array, do you?

Also, use `<<<${scanelf_output}` at the end of the loop instead. You don't need the `{ ... }` then.

PR #752, opened 2021-09-21
[WIP] Add cache to Atom class to improve performance

ezzieyguywuf:
This is the first of (hopefully) a few PR's in which I will attempt to improve
the Portage performance a bit. While it's been said before that an entire
rewrite of the actual dependency-graph algorithm is needed (and in fact, after
working with the code base a bit I would definitely agree), I still think there
are a few "mid-hanging fruit" that we can take care of in the meantime.

My intent in doing this is twofold:

1. Make portage go faster
2. Learn more about the codebase and the relevant algorithms

ultimately, I'd like to contribute to a rewrite that more vastly improves the
performance, but until such a time as I am able to do so, I think these
incremental improvements should still be generally acceptable.

I am measuring performance using `cProfile` as follows. First, I wrote this
little script `my_profiler.py`

```py
#!/usr/bin/python -b
import portage

portage._internal_caller = True
portage._disable_legacy_globals()
from _emerge.main import emerge_main

import cProfile

STATFILE="/home/me/Program/portage/restats"
portage.process.sanitize_fds()

cProfile.run("emerge_main()", STATFILE)
```

Next, I run it like this: note that I run it as `su` - this is because I had
some issues with `cProfile` getting confused, I think because `emerge` runs as
the `portage` user. I'm sure there are clever ways to resolve this, but I
decided to just hit it with a bat:

```sh
su
cd ${PATH_TO_PORTAGE_DEV}
export PYTHONPATH=${PATH_TO_PORTAGE_DEV}/lib
python my_profiler.py -pvuND firefox # any valid portage arguments can go here
```

Here are the results I got on _my_ machine for `-pvuND firefox` before this
patch:

```sh
Tue Sep 21 14:10:35 2021    restats_baseline_firefox

         40253079 function calls (40116445 primitive calls) in 37.377 seconds

   Ordered by: internal time
   List reduced from 2576 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
245425/204640    2.831    0.000   10.177    0.000 /home/me/portage/lib/portage/dep/__init__.py:1505(__init__)
   668316    2.078    0.000    2.078    0.000 {method 'match' of 're.Pattern' objects}
   260864    1.794    0.000    2.650    0.000 /home/me/portage/lib/portage/versions.py:423(__init__)
    40785    1.478    0.000    1.819    0.000 /home/me/portage/lib/portage/dep/__init__.py:1093(__init__)
   287203    1.302    0.000    1.951    0.000 /home/me/portage/lib/portage/dep/__init__.py:2500(match_from_list)
     1208    0.929    0.001    1.916    0.002 /home/me/portage/lib/portage/package/ebuild/config.py:2651(regenerate)
  2900207    0.689    0.000    0.689    0.000 {method 'group' of 're.Match' objects}
  2602399    0.636    0.000    0.811    0.000 {method 'get' of 'dict' objects}
    16496    0.613    0.000    8.284    0.001 /home/me/portage/lib/portage/dep/__init__.py:505(_use_reduce_cached)
   322430    0.477    0.000    2.004    0.000 /home/me/portage/lib/portage/_sets/base.py:144(iterAtomsForPackage)
```

And here are the results after:

```sh

Tue Sep 21 14:12:21 2021    restats_cached_atom_clean

         34481019 function calls (34369236 primitive calls) in 29.068 seconds

   Ordered by: internal time
   List reduced from 2561 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   299326    1.317    0.000    1.999    0.000 /home/me/portage/lib/portage/dep/__init__.py:2564(match_from_list)
     1205    0.924    0.001    1.899    0.002 /home/me/portage/lib/portage/package/ebuild/config.py:2651(regenerate)
223043/208569    0.786    0.000    2.462    0.000 /home/me/portage/lib/portage/dep/__init__.py:1527(__init__)
   302808    0.671    0.000    0.671    0.000 {method 'match' of 're.Pattern' objects}
  2073569    0.542    0.000    0.724    0.000 {method 'get' of 'dict' objects}
    16482    0.510    0.000    2.392    0.000 /home/me/portage/lib/portage/dep/__init__.py:505(_use_reduce_cached)
   197740    0.481    0.000    3.709    0.000 /home/me/portage/lib/portage/_sets/base.py:115(findAtomForPackage)
   333336    0.480    0.000    2.023    0.000 /home/me/portage/lib/portage/_sets/base.py:144(iterAtomsForPackage)
   264785    0.478    0.000    0.632    0.000 /home/me/portage/lib/portage/dep/__init__.py:2094(__getitem__)
    71415    0.457    0.000    0.993    0.000 /home/me/portage/lib/portage/versions.py:423(__init__)
```

Notice that the total time dropped by like 8 seconds!

Signed-off-by: Wolfgang E. Sanyer <WolfgangESanyer@gmail.com>

### DIFF
--- a/lib/portage/dep/__init__.py
+++ b/lib/portage/dep/__init__.py
@@ -1464,6 +1464,9 @@ class _use_dep:
         )
 
 
+_atom_cache = {}
+
+
 class Atom(str):
 
     """
@@ -1500,6 +1503,25 @@ class Atom(str):
         is_valid_flag=None,
         allow_build_id=None,
     ):
+        if (
+            unevaluated_atom is None
+            and allow_wildcard == False
+            and (allow_repo is None)
+            or (allow_repo == False)
+            and _use is None
+            and eapi is None
+            and is_valid_flag is None
+            and allow_build_id is None
+        ):
+
+            if s in _atom_cache.keys():
+                cached = _atom_cache[s]
+
+                if isinstance(cached, Atom):
+                    return cached
+                else:
+                    raise cached
+
         return str.__new__(cls, s)
 
     def __init__(
@@ -1513,6 +1535,22 @@ class Atom(str):
         is_valid_flag=None,
         allow_build_id=None,
     ):
+        global _atom_cache
+
+        if (
+            unevaluated_atom is None
+            and allow_wildcard == False
+            and (allow_repo is None)
+            or (allow_repo == False)
+            and _use is None
+            and eapi is None
+            and is_valid_flag is None
+            and allow_build_id is None
+        ):
+
+            if s in _atom_cache.keys():
+                return
+
         if isinstance(s, Atom):
             # This is an efficiency assertion, to ensure that the Atom
             # constructor is not called redundantly.
@@ -1562,6 +1600,7 @@ class Atom(str):
                 atom_re = _get_atom_wildcard_re(eapi_attrs)
                 m = atom_re.match(s)
                 if m is None:
+                    _atom_cache[s] = InvalidAtom(self)
                     raise InvalidAtom(self)
                 m_group = m.group
                 if m_group("star") is not None:
@@ -1574,14 +1613,17 @@ class Atom(str):
                     op = None
                     cpv = cp = m_group("simple")
                     if m_group(atom_re.groupindex["simple"] + 3) is not None:
+                        _atom_cache[s] = InvalidAtom(self)
                         raise InvalidAtom(self)
                 if cpv.find("**") != -1:
+                    _atom_cache[s] = InvalidAtom(self)
                     raise InvalidAtom(self)
                 slot = m_group("slot")
                 repo = m_group("repo")
                 use_str = None
                 extended_syntax = True
             else:
+                _atom_cache[s] = InvalidAtom(self)
                 raise InvalidAtom(self)
         elif m.group("op") is not None:
             m_group = m.group
@@ -1601,12 +1643,15 @@ class Atom(str):
                     build_id = cpv_build_id[len(cpv) + 1 :]
                     if len(build_id) > 1 and build_id[:1] == "0":
                         # Leading zeros are not allowed.
+                        _atom_cache[s] = InvalidAtom(self)
                         raise InvalidAtom(self)
                     try:
                         build_id = int(build_id)
                     except ValueError:
+                        _atom_cache[s] = InvalidAtom(self)
                         raise InvalidAtom(self)
                 else:
+                    _atom_cache[s] = InvalidAtom(self)
                     raise InvalidAtom(self)
         elif m.group("star") is not None:
             base = atom_re.groupindex["star"]
@@ -1618,6 +1663,7 @@ class Atom(str):
             repo = m_group(atom_re.groups - 1)
             use_str = m_group(atom_re.groups)
             if m_group(base + 3) is not None:
+                _atom_cache[s] = InvalidAtom(self)
                 raise InvalidAtom(self)
         elif m.group("simple") is not None:
             op = None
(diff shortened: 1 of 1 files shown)


lib/portage/dep/__init__.py
```diff
@@ -1562,6 +1600,7 @@ def __init__(
                 atom_re = _get_atom_wildcard_re(eapi_attrs)
                 m = atom_re.match(s)
                 if m is None:
+                    _atom_cache[s] = InvalidAtom(self)
```
mgorny, line 1603: Is there a real gain from caching invalid atoms? I mean, they shouldn't occur during normal operation, should they?

lib/portage/dep/__init__.py
```diff
@@ -1718,32 +1771,39 @@ def __init__(
...
+                    _atom_cache[s] = err
+                    raise err
                 if is_valid_flag is not None and self.use.conditional:
                     invalid_flag = None
                     try:
                         for conditional_type, flags in self.use.conditional.items():
                             for flag in flags:
                                 if not is_valid_flag(flag):
                                     invalid_flag = (conditional_type, flag)
+                                    _atom_cache[s] = StopIteration()
```
mgorny, line 1806: `StopIteration` is caught below, so reraising it via cache is wrong.

thesamesam:
On IRC, I mentioned ideally using functools where possible.

mgorny:
Also, have you compared memory usage? Since you're doing unlimited cache without expiration, I suspect a lot of atoms may end up being cached after major depgraph processing, and they will stay in the memory for the whole emerge process even if they're never going to be used again.

lib/portage/dep/__init__.py
ezzieyguywuf: There's lots of different places that instantiate an `Atom`, and while I didn't check them all super thoroughly, I think some of them use logic around whether or not they were able to succesfully create one.

Similarly, my understanding is that in python, it is idiomatic to use exceptions for control flow.

That's why I chose to cache the exception, so that I could throw it sooner rather than having to plough through the huge `__init__` just to get to it again.

lib/portage/dep/__init__.py
```diff
@@ -1513,6 +1535,22 @@ def __init__(
...
+
+        if (
+            unevaluated_atom is None
+            and allow_wildcard == False
+            and (allow_repo is None)
+            or (allow_repo == False)
+            and _use is None
+            and eapi is None
+            and is_valid_flag is None
+            and allow_build_id is None
```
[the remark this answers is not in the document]
ezzieyguywuf: you're right, I should have provided some explanation here.

I did some testing, and it seems like that vast majority of the time, them `Atom` constructor is called with all of these parameters defaulted. Therefore, I decided to build the cache only for these default values, in order to improve the readability of the cache implementation.

That being said, per @thesamesam  's comment above, I'll be looking to use `functools`s cache mechanism, which will get rid of this anyway.

ezzieyguywuf:
> 
> 
> Also, have you compared memory usage? Since you're doing unlimited cache without expiration, I suspect a lot of atoms may end up being cached after major depgraph processing, and they will stay in the memory for the whole emerge process even if they're never going to be used again.

I have not, though I did mention in IRC that I need to do so. I'll include the memory profiling as part of the `functools` conversion.

lib/portage/dep/__init__.py
```diff
@@ -1464,6 +1464,9 @@ def _eval_qa_conditionals(self, use_mask, use_force):
         )
 
 
+_atom_cache = {}
```
zmedico, line 1467: This needs to be a lru_cache in order to prevent unbounded growth.

lib/portage/dep/__init__.py
```diff
@@ -1500,6 +1503,25 @@ def __new__(
...
+            and allow_wildcard == False
+            and (allow_repo is None)
+            or (allow_repo == False)
+            and _use is None
+            and eapi is None
+            and is_valid_flag is None
+            and allow_build_id is None
+        ):
+
+            if s in _atom_cache.keys():
```
zmedico, line 1517: This should be `s in atom_cache` because there is no need to call `keys()`.

zmedico (changes requested):

PR #754, opened 2021-09-24
Some minor code refactors

ezzieyguywuf:
- portage.dep.Atom: Clean up __new__ parameters
- portage.eapi: use list instead of str for namedtuple definition
- portage.eapi: move None check to helper functions
- portage.eapi: use functools @cache decorator instead of custom implementation

### DIFF
--- a/lib/portage/eapi.py
+++ b/lib/portage/eapi.py
@@ -2,288 +2,167 @@
 # Distributed under the terms of the GNU General Public License v2
 
 import collections
-import operator
-import types
+import logging
+from functools import lru_cache
+from typing import Optional
 
 from portage import eapi_is_supported
 
 
-def eapi_has_iuse_defaults(eapi):
-    return eapi != "0"
+def eapi_has_iuse_defaults(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).iuse_defaults
 
 
-def eapi_has_iuse_effective(eapi):
-    return eapi not in ("0", "1", "2", "3", "4", "4-python", "4-slot-abi")
+def eapi_has_iuse_effective(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).iuse_effective
 
 
-def eapi_has_slot_deps(eapi):
-    return eapi != "0"
+def eapi_has_slot_deps(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).slot_deps
 
 
-def eapi_has_slot_operator(eapi):
-    return eapi not in ("0", "1", "2", "3", "4", "4-python")
+def eapi_has_slot_operator(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).slot_operator
 
 
-def eapi_has_src_uri_arrows(eapi):
+def eapi_has_src_uri_arrows(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).src_uri_arrows
+
+
+def eapi_has_selective_src_uri_restriction(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).selective_src_uri_restriction
+
+
+def eapi_has_use_deps(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).use_deps
+
+
+def eapi_has_strong_blocks(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).strong_blocks
+
+
+def eapi_has_src_prepare_and_src_configure(eapi: str) -> bool:
     return eapi not in ("0", "1")
 
 
-def eapi_has_selective_src_uri_restriction(eapi):
-    return eapi not in (
-        "0",
-        "1",
-        "2",
-        "3",
-        "4",
-        "4-python",
-        "4-slot-abi",
-        "5",
-        "5-progress",
-        "6",
-        "7",
-    )
+def eapi_supports_prefix(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).prefix
 
 
-def eapi_has_use_deps(eapi):
-    return eapi not in ("0", "1")
+def eapi_exports_AA(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).exports_AA
 
 
-def eapi_has_strong_blocks(eapi):
-    return eapi not in ("0", "1")
+def eapi_exports_KV(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).exports_KV
 
 
-def eapi_has_src_prepare_and_src_configure(eapi):
-    return eapi not in ("0", "1")
+def eapi_exports_merge_type(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).exports_merge_type
 
 
-def eapi_supports_prefix(eapi):
-    return eapi not in ("0", "1", "2")
+def eapi_exports_replace_vars(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).exports_replace_vars
 
 
-def eapi_exports_AA(eapi):
-    return eapi in ("0", "1", "2", "3")
+def eapi_exports_EBUILD_PHASE_FUNC(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).exports_EBUILD_PHASE_FUNC
 
 
-def eapi_exports_KV(eapi):
-    return eapi in ("0", "1", "2", "3")
+def eapi_exports_PORTDIR(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).exports_PORTDIR
 
 
-def eapi_exports_merge_type(eapi):
+def eapi_exports_ECLASSDIR(eapi: str) -> bool:
+    return _get_eapi_attrs(eapi).exports_ECLASSDIR
+
+
+def eapi_exports_REPOSITORY(eapi: str) -> bool:
+    return eapi in ("4-python", "5-progress")
+
+
+def eapi_has_pkg_pretend(eapi: str) -> bool:
     return eapi not in ("0", "1", "2", "3")
 
 
-def eapi_exports_replace_vars(eapi):
-    return eapi not in ("0", "1", "2", "3")
+def eapi_has_implicit_rdepend(eapi: str) -> bool:
+    return eapi in ("0", "1", "2", "3")
 
 
-def eapi_exports_EBUILD_PHASE_FUNC(eapi):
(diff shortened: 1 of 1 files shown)


lib/portage/eapi.py
```diff
@@ -353,9 +368,7 @@ def _decorate_eapi_funcs():
 _decorate_eapi_funcs()
 
 
-_eapi_attrs_cache = {}
-
-
+@cache
```
mgorny, line 466: You've seemed to mix changes from the other commit here.

ezzieyguywuf: Great catch! I'll fix this.

ezzieyguywuf: Ok this should be fixed now.

lib/portage/eapi.py
```diff
@@ -2,33 +2,54 @@
...
 
 import collections
-import operator
-import types
-
-from portage import eapi_is_supported
+from functools import lru_cache
 
 
+@lru_cache(None)
```
mattst88, line 8: I don't understand why we expect an lru_cache to help here. The function is trivial. Same goes for the rest of the eapi_has_* functions.

lib/portage/dep/__init__.py
```diff
@@ -1489,17 +1489,7 @@ def __init__(self, forbid=False):
...
-        s,
-        unevaluated_atom=None,
-        allow_wildcard=False,
-        allow_repo=None,
-        _use=None,
-        eapi=None,
-        is_valid_flag=None,
-        allow_build_id=None,
-    ):
+    def __new__(cls, s, *args, **kwargs):
```
mattst88, line 1492: I'm not a Python expert. Could you explain this change?

lib/portage/eapi.py
```diff
@@ -288,25 +414,39 @@ def eapi_has_sysroot(eapi):
...
+        "selective_src_uri_restriction",
+        "slot_operator",
+        "slot_deps",
+        "src_uri_arrows",
+        "strong_blocks",
+        "use_deps",
+        "use_dep_defaults",
+        "empty_groups_always_true",
+        "sysroot",
+    ],
```
mattst88, line 449: I don't understand the intention of this change. Is there reason to believe that a list of field names is somehow better than a space-separated list of field names?

lib/portage/eapi.py
```diff
@@ -387,49 +477,40 @@ def _get_eapi_attrs(eapi):
...
+        dots_in_use_flags=eapi_allows_dots_in_use_flags.func(eapi),
+        empty_groups_always_true=eapi_empty_groups_always_true.func(eapi),
+        exports_AA=eapi_exports_AA.func(eapi),
+        exports_EBUILD_PHASE_FUNC=eapi_exports_EBUILD_PHASE_FUNC.func(eapi),
+        exports_ECLASSDIR=eapi_exports_ECLASSDIR.func(eapi),
+        exports_KV=eapi_exports_KV.func(eapi),
+        exports_merge_type=eapi_exports_merge_type.func(eapi),
+        exports_PORTDIR=eapi_exports_PORTDIR.func(eapi),
+        exports_replace_vars=eapi_exports_replace_vars.func(eapi),
+>>>>>>> 50fd2caa2 (portage.eapi: move None check to helper functions)
```
mattst88: Merge conflict marker.

lib/portage/eapi.py
```diff
@@ -9,26 +9,44 @@
 
 
 def eapi_has_iuse_defaults(eapi):
+    if eapi is None:
+        return True
+
```
mattst88, line 12: Moving these blocks into the functions seems like a good idea to me.

lib/portage/eapi.py
Arfrever: A real tuple or list is certainly more Pythonic than whitespace-delimited-string fake list.
(If tuple instead of list is working, then I suggest to use tuple, i.e. change `[` `]` to `(` `)` around those elements.)

lib/portage/eapi.py
Arfrever: I have not measured anything, but some quite trivial functions are called thousands or millions times by dependency resolver...

lib/portage/dep/__init__.py
ezzieyguywuf: I changed this simply to clean up the code - these parameters are not used at all, and thus it doesn't seem to make sense to name them.

In fact, I'd prefer to get rid of them altogether, but that would be a _slightly_ bigger refacter, as e.g. [this test](https://github.com/gentoo/portage/blob/master/lib/portage/tests/dep/testAtom.py#L216) seems to think that these parameters actually do something.

lib/portage/eapi.py
ezzieyguywuf: The main intent of this commit is to replace [this custom cache implementation](https://github.com/gentoo/portage/blob/master/lib/portage/eapi.py#L322-L356) with the builtin one.

probably I should make this clearer in the commit message if it's still confusing.

Anyway, since the original implementation cached _every_ function in here, I did the same in order to provide a 1-to-1 mapping.

That being said, these types of conversations are _exactly_ the reasons why I wanted to do this sort of cleanup. By making it more clear what's going on, it makes it easier to question the utility of such things.

lib/portage/eapi.py
ezzieyguywuf: `tuple` works fine, I'll make that change.

@mattst88 from a technical standpoint, either the list or the space-separated string work equally fine.

However, it's my opinion that the space-separated list is more difficult to reason about, as well as more error prone.

For example, in the original code, [line 306](https://github.com/gentoo/portage/blob/master/lib/portage/eapi.py#L306) identifies three named fields. This is easy to gloss over, though, as the majority of the fields are defined one-per-line.

Additionally, using this style lends itself to errors such as:

```py
  "my_new_field",
  "some_other_field ",
```

where a missing space results in only _one_ additional field being added versus the expected two.

lib/portage/eapi.py
ezzieyguywuf: fixed

thesamesam (approved):
Good work!

lib/portage/eapi.py
```diff
@@ -2,33 +2,54 @@
...
-from portage import eapi_is_supported
+from functools import lru_cache
 
 
+@lru_cache(None)
 def eapi_has_iuse_defaults(eapi):
+    if eapi is None:
+        return True
+
     return eapi != "0"
```
mgorny, line 13: To be honest, I don't like the added verbosity. How about covering the `None` in the actual  check? e.g.:
```suggestion
    return eapi not in ("0",)
```

lib/portage/eapi.py
```diff
@@ -2,33 +2,54 @@
...
 
 
+@lru_cache(None)
 def eapi_has_iuse_effective(eapi):
+    if eapi is None:
+        return False
+
     return eapi not in ("0", "1", "2", "3", "4", "4-python", "4-slot-abi")
```
mgorny, line 21:
```suggestion
    return eapi not in (None, "0", "1", "2", "3", "4", "4-python", "4-slot-abi")
```

and so on…

lib/portage/eapi.py
mattst88: My only worry is that readers will be confused by the presence of `None` in some of these functions but not others. The fact that `eapi` can be `None` is strange to begin with.

zmedico:
The patch "portage.eapi: use functools @cache decorator instead of custom implementation" misses the point of the original caching mechanism. It doesn't make sense to cache the eapi_* functions if they no longer contribute to the _eap_attrs cache.

ezzieyguywuf:
@zmedico I reworked the commit that you mentioned, do you mind taking a second look?

lib/portage/eapi.py
ezzieyguywuf: This has been reworked, @mgorny @mattst88 please take a second look and let me know your thoughts.

lib/portage/eapi.py
ezzieyguywuf: resolving this comment, leaving the other one open.

lib/portage/eapi.py
```diff
@@ -324,113 +203,124 @@ def eapi_has_sysroot(eapi):
...
+        "4",
+        "4-python",
+        "4-slot-abi",
+        "5",
+        "5-progress",
+        "6",
+        "7",
+        "8",
+    )
+    EAPI_RANGE_MAP = {
```
zmedico, line 221: This EAPI_RANGE_MAP variable seems to be unused.

lib/portage/eapi.py
```diff
@@ -324,113 +203,124 @@ def eapi_has_sysroot(eapi):
...
+        7: (10, 11),
+        8: (11, 12),
+    }
+
+    _eapi_val: int = -1
+
+    def __init__(self, eapi_string: str):
+        if not eapi_string in self.ALL_EAPIS:
+            raise ValueError(f"'{eapi_string}' not recognized as a valid EAPI")
+        self._eapi_val = int(eapi_string[0])
```
zmedico, line 238: This `int(eapi_string[0])` usage seems fragile, since we will eventually have EAPI 10 with two digits instead of one. Maybe use `int(eapi_string.partition("-")[0])`.

zmedico (commented):
Thanks, this is looking much better than before.

lib/portage/eapi.py
ezzieyguywuf: You're right, this is an artifact of an older implementation. I'll take it out.

lib/portage/eapi.py
ezzieyguywuf: Great point, I'll make the suggested change.

ezzieyguywuf: ...actually, I think I'll go with a regex, i.e. `^(\d+)`, otherwise the singular strings, i.e. `"8"` would not match.

another option is to first try converting the string as-is, i.e. `try: int(eapi_string)`, but I'm leaning more towards the regex solution. LMK if you have an opinion.

zmedico: Note that `int(eapi_string.partition("-")[0])` actually does work with a singular string like `"8"`, but I suppose a regular expression is also fine.

ezzieyguywuf: Ah, great point!

I think my unfamiliarity with python is showing 😅

I'll abandon the regex approach, it's a pretty big hammer to use when clearly it is not needed. thanks for the help!

PR #756, opened 2021-09-27
install-qa-check.d: Add a QA check for installing xattrs

mgorny:
Warn the developers if ebuilds install files with xattrs to ${ED}.
The xattrs may or may not be preserved when installing the package,
making them unreliable on one hand, and somewhat suprising in other
cases (e.g. when they unintentionally leak from developer's system).

This is the first step towards restoring PMS compliance and *not*
preserving extended metadata.

### DIFF
--- a/bin/install-qa-check.d/95xattr
+++ b/bin/install-qa-check.d/95xattr
@@ -1,0 +1,54 @@
+# Check for xattrs.  See https://bugs.gentoo.org/815220.
+
+xattr_check() {
+	type -P getfattr >/dev/null || return
+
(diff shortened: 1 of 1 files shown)


bin/install-qa-check.d/95xattr
```diff
@@ -0,0 +1,54 @@
+# Check for xattrs.
```
thesamesam, line 1: Can you include a bit more information/background here? (Similar to floppym's [query](https://archives.gentoo.org/gentoo-portage-dev/message/67a537a1e97692db1f9903fb003b0bbd) on the ML).

PR #757, opened 2021-09-28
prepare_build_dirs: apply permissions to filesdir

zmedico:
Bug: https://bugs.gentoo.org/815196

### DIFF
--- a/lib/portage/package/ebuild/prepare_build_dirs.py
+++ b/lib/portage/package/ebuild/prepare_build_dirs.py
@@ -484,8 +484,15 @@ def _prepare_fake_filesdir(settings):
 
     # Copy files from real directory to ebuild directory (without metadata).
     if os.path.isdir(real_filesdir) and not os.path.isdir(filesdir):
-        shutil.copytree(
-            real_filesdir, filesdir, copy_function=copyfile
+        shutil.copytree(real_filesdir, filesdir, copy_function=copyfile)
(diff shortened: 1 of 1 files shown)


thesamesam (approved):

lib/portage/package/ebuild/prepare_build_dirs.py
```diff
@@ -484,8 +484,15 @@ def _prepare_fake_filesdir(settings):
...
     # Copy files from real directory to ebuild directory (without metadata).
     if os.path.isdir(real_filesdir) and not os.path.isdir(filesdir):
-        shutil.copytree(
-            real_filesdir, filesdir, copy_function=copyfile
+        shutil.copytree(real_filesdir, filesdir, copy_function=copyfile)
+        apply_recursive_permissions(
+            filesdir,
+            uid=portage_uid,
+            gid=portage_gid,
+            dirmode=0o700,
```
mgorny, line 492: Maybe we could open these files to the Portage group at least? I suppose it would ease debugging.

mgorny (approved):
Thanks for looking into it.

lib/portage/package/ebuild/prepare_build_dirs.py
zmedico: Sure, updated to use `dirmode=0o750` and `filemode=0o640`.

PR #758, opened 2021-09-28
Use a distinct color for QA warnings

mgorny:
As a first step towards making QA warnings distinct from user-facing
warnings, introduce a separate color style (brown instead of yellow)
and explicit eqawarn printing API.


mgorny:
Will refactor this more.

PR #759, opened 2021-09-28
Output rewrite for better clarify and greppability

mgorny:
Ok, so it's more major than I originally intended but I think it's a good direction (once you get used to it). Roughly:

1. All bash color vars are now prefixed with `PORTAGE_COLOR_` to avoid accidental collisions with ebuild vars (e.g. ebuild setting `GOOD=foo` broke `elog` before).
2. There are specific color vars for all kinds of output functions, and now `einfo` messages use distinct color (dark green) from `elog`, and `eqawarn` (brown) from `ewarn`.
3. Messages are now prefixed by their kind, making it possible to distinguish them without colors and grep for specific kind of logs:
   - `[--]` for einfo & ebegin
   - `[II]` for elog
   - `[WW]` for ewarn
   - `[QA]` for eqawarn
   - `[EE]` for eerror
4. Finally, I've replaced most of `>>>` and `!!!` in Portage output with four `>>>>` and `!!!!` to align the output again.


mgorny:
elog

mgorny:
_newform

RinCat:
Can we add `[SS]` esec for security issues?

mgorny:
> Can we add `[SS]` esec for security issues?

Please file a 'future EAPI' bug for this. That said, I'm not sure if 'SS" is good label, it has pretty negative meaning in Europe.

petteyg:
Feels backwards to have einfo use `--` and elog use `II`. If ewarn and eerorr are respectively `WW` and `EE`, it would follow that `II` is einfo, not something else.

mgorny:
Retracting because of gentoo-dev discussion.

PR #761, opened 2021-10-03
prepare_build_dirs: handle copytree FileExistsError

zmedico:
Bug: https://bugs.gentoo.org/815871

### DIFF
--- a/lib/portage/package/ebuild/prepare_build_dirs.py
+++ b/lib/portage/package/ebuild/prepare_build_dirs.py
@@ -477,13 +477,21 @@ def _ensure_log_subdirs(logdir, subdir):
         ensure_dirs(current, uid=uid, gid=gid, mode=grp_mode, mask=0)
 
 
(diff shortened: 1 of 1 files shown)


lib/portage/package/ebuild/prepare_build_dirs.py
```diff
@@ -477,13 +477,21 @@ def _ensure_log_subdirs(logdir, subdir):
         ensure_dirs(current, uid=uid, gid=gid, mode=grp_mode, mask=0)
 
 
+def _safe_copytree(src, dst, **kwargs):
```
mgorny, line 480: That's a weird name for a function that removes `dst` without warning.

zmedico: Good point. Renamed to just `_copytree`.

mgorny (approved):
Makes sense. Thanks!

PR #764, opened 2021-10-28
install-qa-check.d/60pkgconfig: enhance QA checks

thesamesam:
- Run pkg-config --validate
- Check if EPREFIX is respected
- Check if libdir is respected (weak check)

### DIFF
--- a/bin/install-qa-check.d/60pkgconfig
+++ b/bin/install-qa-check.d/60pkgconfig
@@ -107,7 +107,7 @@ pkgconfig_check() {
 	# Skip result reporting if *_p* because for both _pN and _preN, we
 	# don't generally expect the versions to be exactly accurate, and
 	# we want to avoid false positives.
-	if [[ ${all_bad} == "yes" && ${PV} != *_p* ]] ; then
+	if [[ ${all_bad} == "yes" && ${PV} != *_p* ]] && ! has live ${PROPERTIES} ; then
 		eqawarn "QA Notice: pkg-config files with mismatched Version found!"
 		eqawarn "At least ${bad_file}'s Version field does not match ${PV}"
 		eqawarn "Please check all .pc files installed by this package."


bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,8 +1,14 @@
 # Check for pkg-config file issues
 
 pkgconfig_check() {
+	# Make a list of .pc files and bail out if there aren't any
+	local files=$(find "${ED}"/usr/lib*/pkgconfig -name '*.pc' -print 2>/dev/null)
```
mgorny, line 5: Why u no `-print0` and `read`, and bash array? Also, do we really expect files in subdirs there? If pkg-config doesn't recur into subdirs, neither should we.

bin/install-qa-check.d/60pkgconfig
```diff
@@ -13,6 +13,12 @@ pkgconfig_check() {
 		eqawarn "QA Notice: pkg-config files with wrong LDFLAGS detected:"
 		eqawarn "${f//${D}}"
 	fi
+
+	# Validate using pkgconfig
+	if ! pkg-config --validate "${files[@]}" ; then
```
mgorny, line 18: Is the `--validate` option portable across different pkg-config implementations? Hmm, I suppose we only support pkgconf these days, so feel free to ignore the problem. OTOH, you could check for pkgconf being available and just call it directly.

bin/install-qa-check.d/60pkgconfig
```diff
@@ -19,6 +19,21 @@ pkgconfig_check() {
 		eqawarn "QA Notice: pkg-config files which fail validation found!"
 		eqawarn "Run 'pkg-config --validate ...' for more information"
 	fi
+
+	# Check for unexpected paths
+	# e.g. https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=c90ab38e3577aae61fac2341b34ad593948de1cd
+	for f in "${files[@]}" ; do
+		local key
+		for key in prefix exec_prefix libdir includedir ; do
+			local value=$(pkg-config --variable=prefix ${f})
```
mgorny, line 28: Don't you wan to use `${key}` here?

bin/install-qa-check.d/60pkgconfig
```diff
@@ -19,6 +19,21 @@ pkgconfig_check() {
...
+
+			if [[ ${value#${EPREFIX}} == ${value} ]] ; then
```
mgorny: Isn't this going to be triggered by empty EPREFIX?

bin/install-qa-check.d/60pkgconfig
```diff
@@ -27,13 +27,40 @@ pkgconfig_check() {
...
+				# (Don't bother if /usr/lib is a symlink to /usr/lib64)
+				continue
+			fi
+
+			if egrep -q "/lib64" ${f} ; then
+				eqawarn "QA Notice: pkg-config files not respecting libdir found"
+				eqawarn "${f} contains reference to 'lib64'"
+			fi
+		elif [[ ${f} == *lib64/pkgconfig* ]] ; then
+			if egrep -q "/lib$" ${f} ; then
```
mgorny, line 58: I think you should cover `lib/` too.

bin/install-qa-check.d/60pkgconfig
gyakovlev: this one should be safer indeed, `-d` arg is double singlequote.
mapfile -d '' files < <(find "${ED}"/usr/lib*/pkgconfig -maxdepth 1 -type f -name '*.pc' -print0 2>/dev/null)

we need to search `/usr/share/pkgconfig` too, it contains pc files for arch-independent things (headers, icons, etc)

gyakovlev:
I think it'd be helpful to tag and output offending files too, `eqatag` should do it.

bin/install-qa-check.d/60pkgconfig
thesamesam: Yes!

bin/install-qa-check.d/60pkgconfig
thesamesam: OpenBSD has it and so does fd.o, so let's go with this for now, but I'll add a comment that it's possible some less-common impls (like some pure Perl ones, etc) may not implement it

bin/install-qa-check.d/60pkgconfig
thesamesam: Good point.

bin/install-qa-check.d/60pkgconfig
thesamesam: Oh, nice.

thesamesam:
Should have addressed everything apart from the listing files / QA tag thing. Will do that soon.

bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,12 +1,70 @@
 # Check for pkg-config file issues
 
 pkgconfig_check() {
+	# Make a list of .pc files and bail out if there aren't any
+	mapfile -d '' files < <(find "${ED}"/usr/lib*/pkgconfig "${ED}"/usr/share/pkgconfig -maxdepth 1 -type f -name '*.pc' -print0 2>/dev/null)
```
mgorny, line 5:
```suggestion
	mapfile -d '' files < <(find "${ED}"/usr/{lib*,share}/pkgconfig -maxdepth 1 -type f -name '*.pc' -print0 2>/dev/null)
```

bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,12 +1,70 @@
...
+	[[ -n "${files[@]}" ]] || return
```
mgorny, line 6:
```suggestion
	[[ -z "${files[@]}" ]] && return
```

Generally people find it easier to read the logic without negations, i.e. 'if X then Y'.

bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,12 +1,70 @@
...
+
+	local f
+
 	# Look for leaking LDFLAGS into pkg-config files
-	local f=$(egrep -sH '^Libs.*-Wl,(-O[012]|--hash-style)' "${ED}"/usr/*/pkgconfig/*.pc)
+	f=$(egrep -sH '^Libs.*-Wl,(-O[012]|--hash-style)' "${files[@]}")
```
mgorny, line 11: `grep` has `-z` option for NUL-termination.

bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,12 +1,70 @@
...
+	# Note: some less common implementations may not support this?
+	# Seems like f.d.o, OpenBSD, and of course pkgconf do though.
+	if ! pkg-config --validate "${files[@]}" ; then
+		eqawarn "QA Notice: pkg-config files which fail validation found!"
+		eqawarn "Run 'pkg-config --validate ...' for more information"
+	fi
+
+	# Check for unexpected paths
+	# e.g. https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=c90ab38e3577aae61fac2341b34ad593948de1cd
+	if [[ -n "${EPREFIX}" ]] ; then
```
mgorny, line 27:
```suggestion
	if [[ -n ${EPREFIX} ]] ; then
```

Quoting inside `[[ ... ]]` is only needed for RHS of `==` (to disable wildcards).

bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,12 +1,70 @@
...
+		for f in "${files[@]}" ; do
+			local key
+			for key in prefix exec_prefix libdir includedir ; do
+				local value=$(pkg-config --variable=${key} ${f})
```
mgorny, line 31: You're losing quotes.

bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,12 +1,70 @@
...
+
+				if [[ ${value#${EPREFIX}} == ${value} ]] ; then
```
mgorny, line 33:
```suggestion
				if [[ ${value} != "${EPREFIX}"* ]] ; then
```

right?

bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,12 +1,70 @@
...
+	#
+	# If we're installing to ${ED}/usr/lib/pkgconfig, let's make sure
+	# we're not referencing lib64.
+	#
+	# e.g. https://bugs.gentoo.org/729642
+	for f in "${files[@]}" ; do
+		# In ${ED}/usr/lib, we shouldn't reference lib64
+		if [[ ${f} == *lib/pkgconfig* ]] ; then
+			if [[ -d "${ED}"/usr/lib && ! -L "${ED}"/usr/lib ]] ; then
+				# (Don't bother if /usr/lib is a symlink to /usr/lib64)
```
mgorny, line 53: Isn't the condition the other way around?

bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,12 +1,70 @@
...
+				continue
+			fi
+
+			if egrep -q "/lib64" ${f} ; then
+				eqawarn "QA Notice: pkg-config files not respecting libdir found"
+				eqawarn "${f} contains reference to 'lib64'"
+			fi
+		elif [[ ${f} == *lib64/pkgconfig* ]] ; then
+			if egrep -q "/lib[/]?$" ${f} ; then
```
mgorny, line 62: That certainly won't match inline `/lib/foo`.

mgorny (changes requested):

bin/install-qa-check.d/60pkgconfig
thesamesam: Ah, thanks.

bin/install-qa-check.d/60pkgconfig
SoapGentoo: `grep -P '/lib(?![[:digit:]])'` maybe? yes sue me for using PCRE

bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,12 +1,80 @@
...
+				# (Don't bother if /usr/lib is a symlink to /usr/lib64)
+				continue
+			fi
+
+			if egrep -q "/lib64" ${f} ; then
+				bad_libdir+=( ${f} )
+			fi
+		elif [[ ${f} == *lib64/pkgconfig* ]] ; then
+			# We want to match /lib/, /lib/foo/, but not e.g. /lib64 or /lib64/
+			if grep -qP '/lib(?![[:digit:]])' ${f} ; then
```
mgorny, line 66:
```suggestion
			if grep -qP '/lib\b' ${f} ; then
```

bin/install-qa-check.d/60pkgconfig
```diff
@@ -75,6 +75,29 @@ pkgconfig_check() {
...
+	# To be safe, let's make sure _all_ installed .pcs have a bad Version
+	# before warning, as this should catch the general cases we're worried
+	# about, while avoiding any pathological cases e.g. multiple libraries
+	# with different versioning within one package.
+	# Example bugs: bug #833895, bug #833887.
+	local all_bad=yes
+	# Record the last bad file matched
+	local bad_file
+	for f in "${files[@]}" ; do
+		if [[ ${PV} == $(sed -n -e 's/^Version: //p' "${f}" || die) ]] ; then
```
thesamesam, line 88: There's a problem here where ${PV} has e.g. _p1 for say, a Debian patchset, and it obviously then won't match the .pc file. Not sure how we should handle that best.

thesamesam: Meh. For now, I might just skip it if the version contains _p* (including _pre) given that I don't think we necessarily expect correct versions there anyway and it complicates the logic. QA checks are intended to be overall wins rather than perfect, and so avoiding the false positives is more important than trying to catch every case IMO.

bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,11 +1,105 @@
 # Check for pkg-config file issues
 
 pkgconfig_check() {
+	# Make a list of .pc files and bail out if there aren't any
+	mapfile -d '' files < <(find "${ED}"/usr/{lib*,share}/pkgconfig -maxdepth 1 -type f -name '*.pc' -print0 2>/dev/null)
```
mgorny, line 5:
```suggestion
	mapfile -d '' files < <(
		find "${ED}"/usr/{lib*,share}/pkgconfig -maxdepth 1 -type f -name '*.pc' -print0 2>/dev/null
	)
```

Let's try to keep the lines shorter when we can.

bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,11 +1,105 @@
 # Check for pkg-config file issues
 
 pkgconfig_check() {
+	# Make a list of .pc files and bail out if there aren't any
```
mgorny, line 4:
```suggestion
	local files=()
	# Make a list of .pc files and bail out if there aren't any
```

bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,11 +1,105 @@
...
+		for f in "${files[@]}" ; do
+			local key
+			for key in prefix exec_prefix libdir includedir ; do
+				local value=$(pkg-config --variable="${key}" "${f}")
+
+				if [[ ${value} != "${EPREFIX}"* ]] ; then
+					eqawarn "QA Notice: pkg-config files not respecting EPREFIX found"
+					eqawarn "${f}'s key=${key} does not respect EPREFIX:"
+					eqawarn "${key}=${value}"
+					eqatag -v pkgconfig.bad-paths "${f}"
```
mgorny, line 39: Sounds like you're being very verbose here. It's probably quite likely that if one variable is wrong, then others also will be wrong, and you're repeating the whole message for every single instance. Plus you're listing the file path twice (`eqatag -v` is going to do that), and the key twice. Also, `eqatag` accepts key-value pairs, so you could include the variable name there as well.

bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,11 +1,105 @@
...
-		eqawarn "${f//${D}}"
+		eqatag -v pkgconfig.bad-ldlags "${f//${D}}"
+	fi
+
+	# Validate using pkgconfig
+	# Some less common implementations may not support this?
+	# seems like f.d.o, OpenBSD, and of course pkgconf do though.
+	# Need --maximum-traverse-depth=1 to avoid checking deps and giving
+	# unrelated warnings/errors.
+	if ! pkg-config --maximum-traverse-depth=1 --validate "${files[@]}" ; then
```
mgorny, line 22: Am I missing something or aren't you accounting for the possibility that pkg-config is not installed?

bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,11 +1,105 @@
...
+	# To be safe, let's make sure _all_ installed .pcs have a bad Version
+	# before warning, as this should catch the general cases we're worried
+	# about, while avoiding any pathological cases e.g. multiple libraries
+	# with different versioning within one package.
+	# Example bugs: bug #833895, bug #833887.
+	local all_bad=yes
+	# Record the last bad file matched
+	local bad_file
+	for f in "${files[@]}" ; do
+		if [[ ${PV} == $(sed -n -e 's/^Version: //p' "${f}" || die) ]] ; then
```
mgorny, line 88: Any reason you're calling pkg-config in the above checks but working on the file directly here?

bin/install-qa-check.d/60pkgconfig
```diff
@@ -1,11 +1,105 @@
...
+			all_bad=no
+			break
+		fi
+
+		bad_file=f
```
mgorny, line 93: Did you test this? :-P Not that you're using it.

mgorny (changes requested):

mattst88 (approved):
Looks like a valuable improvement to me.

bin/install-qa-check.d/60pkgconfig
thesamesam: In some cases, the version has been a ridiculous string ("This version was generated by a git tar ball") or something and I wasn't convinced pkg-config would handle it correctly/notice. But I don't mind changing and experimenting.

bin/install-qa-check.d/60pkgconfig
thesamesam: Yeah, I've been running with this series for several months, although this part is new.

thesamesam:
TODO: strip ${D} from printed paths, eqatags fixes

bin/install-qa-check.d/60pkgconfig
mgorny: Well, it seems to truncate it to the first word but I suppose that may be good enough. Your choice.

```
# pkg-config --modversion libical
yo
```

thesamesam: I'm going to stick to my original so that I can notice something like "1.2.0 (git)" or similar.

bin/install-qa-check.d/60pkgconfig
thesamesam: Thanks!

thesamesam:
Merged.

PR #765, opened 2021-11-03
EbuildIpc.communicate: lockfile PermissionDenied retry

zmedico:
The lockfile function is expected to raise PermissionDenied if
the (root) parent process holds the lock, so retry in this case.

Bug: https://bugs.gentoo.org/468990

### DIFF
--- a/bin/ebuild-ipc.py
+++ b/bin/ebuild-ipc.py
@@ -158,7 +158,16 @@ class EbuildIpc:
         # Make locks quiet since unintended locking messages displayed on
         # stdout could corrupt the intended output of this program.
         portage.locks._quiet = True
-        lock_obj = portage.locks.lockfile(self.ipc_lock_file, unlinkfile=True)
+        # Acquire lock with PermissionDenied retry for bug #468990.
+        for _ in range(1000):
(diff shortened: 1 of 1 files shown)


vapier:
didn't know if you wanted to do it at this high level, or should the retry loop be inside the existing `lib/portage/locks.py:_lockfile_iteration` which has checks for different errno failure modes

zmedico:
> didn't know if you wanted to do it at this high level, or should the retry loop be inside the existing `lib/portage/locks.py:_lockfile_iteration` which has checks for different errno failure modes

I think it's probably good enough to retry inside EbuildIpc.communicate for now, and we can evaluate adding support to _lockfile_iteration later if we find that this is useful in other cases. I'm not yet aware of any other cases where we would need this kind of retry.

vapier (approved):

zmedico:
We could solve a permission race on the lockfile creation side, by doing a `chmod g+s` on the parent directory, and `umask 0002` during creation of the lockfile.

zmedico:
Closing in favor of https://github.com/gentoo/portage/pull/768.

PR #766, opened 2021-11-04
Exclude binary cache build for live ebuilds

MadhuPriyaMurugan:
This commit introduces a new value "buildpkg-live" for FEATURES, which
is enabled by default (so the default behavior of building binary cache
for all packages is retained). When it is disabled by calling emerge
with FEATURES="-buildpkg-live", binary caches will not be built live
ebuilds even if we specify --buildpkg. So that it is no longer necessary
to pass a list of packages with live ebuilds to --buildpkg-exclude.

Before this commit, when an emerge is called with the option
'--buildpkg', a binary cache for the package is created under
/var/cache/binpkgs. For example, when we do a, 'emerge --ask
--verbose --buildpkg some-gitpkg/abc', a binary cache abc-1.1.1.tbz2
is created under /var/cache/binpkgs/some-gitpkg.

With this commit, even if we explicitly use the options, '--buildpkg'
for the packages with live ebuilds, no binary cache will be created
(given we disable it calling emerge with FEATURES="-buildpkg-live").

Motivation: Since binary caches are created for all packages, including
packages with live ebuilds, a separate list of (for eg.,) git packages
needs to be maintained. And this is then passed to the options
'--buildpkg-exclude' via, EMERGE_DEFAULT_OPTS. So the motivation behind
this patch was to reduce redundancy, while we can simply disable binary
cache for live ebuilds with this option.

Signed-off-by: Madhu Priya Murugan <madhu.murugan@rohde-schwarz.com>

### DIFF
--- a/lib/_emerge/EbuildBuild.py
+++ b/lib/_emerge/EbuildBuild.py
@@ -343,9 +343,32 @@ class EbuildBuild(CompositeTask):
             and opts.buildpkg != "n"
         )
 
+        # Do not build binary cache for packages from volatile sources.
+        # For volatile sources (eg., git), the PROPERTIES parameter in
+        # the ebuild is set to 'live'.
+
+        # The default behavior is to build binary cache for all pkgs.
+        # "buildpkg-live" is a FEATURE that is enabled by default.
+        # To not build binary cache for live pkgs, we disable it by
+        # specifying FEATURES="-buildpkg-live"
+
+        buildpkg_live = True
+
+        if not "buildpkg-live" in features:
+            buildpkg_live = False
+
+        live_ebuild = False
+
+        if "live" in self.settings.get("PROPERTIES", "").split():
+            live_ebuild = True
+
+        binpkg_live_disabled = live_ebuild and not buildpkg_live
+
         if (
-            "buildpkg" in features or self._issyspkg
-        ) and not self.opts.buildpkg_exclude.findAtomForPackage(pkg):
+            ("buildpkg" in features or self._issyspkg)
+            and not binpkg_live_disabled
+            and not self.opts.buildpkg_exclude.findAtomForPackage(pkg)
+        ):
 
             self._buildpkg = True
 
--- a/cnf/make.globals
+++ b/cnf/make.globals
@@ -54,7 +54,7 @@ FETCHCOMMAND_SFTP="bash -c \"x=\\\${2#sftp://} ; host=\\\${x%%/*} ; port=\\\${ho
 
 # Default user options
 FEATURES="assume-digests binpkg-docompress binpkg-dostrip binpkg-logs
-          binpkg-multi-instance
+          binpkg-multi-instance buildpkg-live
           config-protect-if-modified distlocks ebuild-locks
           fixlafiles ipc-sandbox merge-sync multilib-strict
           network-sandbox news parallel-fetch pid-sandbox
--- a/lib/portage/const.py
+++ b/lib/portage/const.py
@@ -131,6 +131,7 @@ SUPPORTED_FEATURES = frozenset(
         "binpkg-logs",
         "binpkg-multi-instance",
         "buildpkg",
+        "buildpkg-live",
         "buildsyspkg",
         "candy",
         "case-insensitive-fs",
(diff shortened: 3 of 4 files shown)


floppym (commented):
This change assumes that nobody wants to build binpkgs for live packages. I don't believe this is a safe assumption.

MadhuPriyaMurugan:
@floppym Thank you for the review. Do you think explicitly requesting to build binpkgs for live packages will make more sense, may be something like --buildpkg_live = y? So, the build binpkgs for live packages is by default off, but can be requested if the user wants to?

floppym:
Hmm. I think it might make more sense to maintain the current behavior by default, and add an option to enable the behavior you want to see.

MadhuPriyaMurugan:
@floppym I have taken your suggestion and adapted accordingly. Now the behavior we want can be achieved only by calling emerge with FEATURES="binpkg-nolive".

MadhuPriyaMurugan:
@floppym I have made some changes to variable names

zmedico:
Our convention it to use "positive" names for FEATURES, for instance `binpkg-live` instead of `binpkg-nolive`. If we call it `binpkg-live` then it should be enabled be default in order to preserve backward compatibility.

MadhuPriyaMurugan:
@zmedico Thank you for the review. I have adapted the code based on your suggestions.

zmedico:
If you run `black` on the touched files then that should correct these issues:
https://github.com/gentoo/portage/runs/4206002634?check_suite_focus=true
https://github.com/gentoo/portage/runs/4206002755?check_suite_focus=true
```
  ************* Module _emerge.EbuildBuild
  lib/_emerge/EbuildBuild.py:348:38: C0303: Trailing whitespace (trailing-whitespace)
  lib/_emerge/EbuildBuild.py:351:66: C0303: Trailing whitespace (trailing-whitespace)
  lib/_emerge/EbuildBuild.py:352:67: C0303: Trailing whitespace (trailing-whitespace)
  ************* Module _emerge.depgraph
  lib/_emerge/depgraph.py:7373:83: C0303: Trailing whitespace (trailing-whitespace)
  lib/_emerge/depgraph.py:7374:82: C0303: Trailing whitespace (trailing-whitespace)
```

MadhuPriyaMurugan:
@zmedico I ran `black` on the files, and I have pushed the changes.

zmedico:
The depgraph behavior change is really a separate thing, so I'd prefer that you submit that as a separate PR from the  FEATURES="buildpkg-live" PR, and implement it as a `--usepkg-exclude-live` option or something like that.

MadhuPriyaMurugan:
@zmedico For clarification, in your first suggestion, you are recommending adding an option "--usepkg-exclude-live". Will it be okay if we add a FEATURE="usepkg-live", like we did for buildpkg?

zmedico:
> @zmedico For clarification, in your first suggestion, you are recommending adding an option "--usepkg-exclude-live". Will it be okay if we add a FEATURE="usepkg-live", like we did for buildpkg?

I think it's better if this is just an emerge option without a corresponding FEATURES flag. You can enable it in EMERGE_DEFAULT_OPTS if you want it to be a persistent setting.

MadhuPriyaMurugan:
@zmedico As you suggested, I will make a separate PR for depgraph. I have updated this PR to address only EbuildBuild.

MadhuPriyaMurugan:
The second PR for depgraph, https://github.com/gentoo/portage/pull/769

lib/_emerge/EbuildBuild.py
```diff
@@ -343,9 +343,34 @@ def _fetch_exit(self, fetcher):
...
+        if not "buildpkg-live" in features:
+            buildpkg_live = False
+
+        live_ebuild = False
+
+        if "live" in self.settings.get("PROPERTIES", "").split():
+            live_ebuild = True
+
+        binpkg_live_disabled = (
+            "buildpkg" in features and live_ebuild and not buildpkg_live
```
zmedico, line 366: I think it's better to omit `"buildpkg" in features`, since we already have `"buildpkg" in features or self._issyspkg` in the boolean expression that follows, and checking `"buildpkg" in features` in both expressions would be redundant.

MadhuPriyaMurugan:
@zmedico Yeap makes sense. I removed it.

zmedico (approved):
LGTM. Thanks!

PR #769, opened 2021-11-15
Exclude binary pkg installation for live ebuilds

MadhuPriyaMurugan:
This commit introduces a new option "--usepkg-exclude-live" for emerge.
Passing this option with the emerge call, will disable binary pkgs from
being installed for live ebuilds. So it is no longer necessary to pass
a list of live packages to --usepkg-exclude.

Before this commit, when an emerge is called with the option
'--usepkg', the corresponding package is installed from the binary cache
under /var/cache/binpkgs.

With this commit, even if we explicitly use the options, '--usepkg'
for the packages with live ebuilds, no binary package will not be
installed for live ebuilds (given we disable it calling emerge with
"--usepkg-exclude-live" in the emerge).

Motivation: We no longer need to maintain a separate list for VCS based
pkgs (eg., git pkgs), from being installed from the binary cache, and
then pass it to --usepkg-exclude. This reduces some redundancy.

Signed-off-by: Madhu Priya Murugan <madhu.murugan@rohde-schwarz.com>

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -7303,6 +7303,7 @@ class depgraph:
         rebuilt_binaries = "rebuilt_binaries" in self._dynamic_config.myparams
         usepkg = "--usepkg" in self._frozen_config.myopts
         usepkgonly = "--usepkgonly" in self._frozen_config.myopts
+        usepkg_exclude_live = "--usepkg-exclude-live" in self._frozen_config.myopts
         empty = "empty" in self._dynamic_config.myparams
         selective = "selective" in self._dynamic_config.myparams
         reinstall = False
@@ -7370,11 +7371,23 @@ class depgraph:
                     ):
                         continue
 
+                    # We can choose not to install a live package from using binary
+                    # cache by disabling it with option --usepkg-exclude-live in the
+                    # emerge call.
+
+                    live_ebuild = False
+
+                    if "live" in pkg._metadata.get("PROPERTIES", "").split():
+                        live_ebuild = True
+
                     if (
                         built
                         and not installed
-                        and usepkg_exclude.findAtomForPackage(
-                            pkg, modified_use=self._pkg_use_enabled(pkg)
+                        and (
+                            usepkg_exclude.findAtomForPackage(
+                                pkg, modified_use=self._pkg_use_enabled(pkg)
+                            )
+                            or (usepkg_exclude_live and live_ebuild)
                         )
                     ):
                         break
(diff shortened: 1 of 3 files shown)


lib/_emerge/main.py
```diff
@@ -51,6 +51,7 @@
     "--skipfirst",
     "--tree",
     "--unordered-display",
+    "--usepkg-exclude-live",
```
zmedico, line 54: Do you anticipate that a user would want to set this in EMERGE_DEFAULT_OPTS and override it via the command line? If so, then it would be useful to make this a `--usepkg-exclude-live=<y|n>` option.

MadhuPriyaMurugan:
@zmedico Yes I would want the possibility of setting it in EMERGE_DEFAULT_OPTS. I have update the PR.

zmedico:
Could you also add this option to the emerge man page? Thanks!

MadhuPriyaMurugan:
@zmedico Sure, I have added the option to the emerge man page.

lib/_emerge/depgraph.py
```diff
@@ -7370,11 +7370,28 @@ def _wrapped_select_pkg_highest_available_imp(
...
                         continue
 
+                    # We can choose not to install a live package from using binary
+                    # cache by disabling it with option --usepkg-exclude-live in the
+                    # emerge call.
+
+                    usepkg_exclude_live = False
+
+                    if "--usepkg-exclude-live" in self._frozen_config.myopts:
+                        usepkg_exclude_live = True
```
zmedico, line 7380: We can define this variable before the loop, like this:
```
usepkg_exclude_live = "--usepkg-exclude-live" in self._frozen_config.myopts
```

MadhuPriyaMurugan: Yea, that makes sense.

MadhuPriyaMurugan: I pushed the changes.

lib/_emerge/depgraph.py
```diff
@@ -7370,11 +7370,25 @@ def _wrapped_select_pkg_highest_available_imp(
                     ):
                         continue
 
+                    # We can choose not to install a live package from using binary
+                    # cache by disabling it with option --usepkg-exclude-live in the
+                    # emerge call.
+
+                    usepkg_exclude_live = "--usepkg-exclude-live" in self._frozen_config.myopts
```
zmedico, line 7377: Since this variable doesn't change inside the loop, we should define it before the loop, next to the other definitions like these:
```
        usepkg = "--usepkg" in self._frozen_config.myopts
        usepkgonly = "--usepkgonly" in self._frozen_config.myopts
```

zmedico (approved):
LGTM. Thanks!

PR #770, opened 2021-11-17
estrip: fix race condition in save_elf_debug

floppym:
https://bugs.gentoo.org/823798


floppym:
Added a retry limit for the lockfile.

mgorny (commented):
Cool, thanks. I still don't like keeping the ugly vapiercode around but meh.

PR #771, opened 2021-11-17
estrip: drop multijob code

floppym:
https://bugs.gentoo.org/823798


bin/estrip
```diff
@@ -280,12 +280,9 @@ process_elf() {
 	# dev-vcs/git which creates ~111 hardlinks to one file in
 	# /usr/libexec/git-core).
 	lockfile=${inode_link}_lockfile
-	if ! ln "${inode_link}" "${lockfile}" 2>/dev/null ; then
-		while [[ -f ${lockfile} ]] ; do
-			sleep 1
-		done
-		unset lockfile
-	fi
+	while ! ln "${inode_link}" "${lockfile}" 2>/dev/null; do
```
mgorny: Hmm, doesn't that imply that when some 'other' error happens, Portage will hang forever? Not saying that's a regression but still sucks. Maybe you could add some retry limit and fail after N retries?

mgorny: Oh, I see you remove this anyay, so I guess not worth changing.

mgorny (approved):

bin/estrip
floppym: > Oh, I see you remove this anyay, so I guess not worth changing.

That's still a maybe.

> Hmm, doesn't that imply that when some 'other' error happens, Portage will hang forever? Not saying that's a regression but still sucks. Maybe you could add some retry limit and fail after N retries?

I'm not sure what a "safe" retry limit would be. It really depends on a couple factors:

1. The time it takes to strip a single object.
2. The number of hard links that exist for said object.

I guess I could pick an arbitrary number and hope for the best? Maybe start with 30 seconds or so?

floppym: I think I would rather leave the locking code as is, and we can revisit the issue if/when somebody actually runs into a hung process due to some unknown error when calling `ln`.

mgorny: I would not go that route. We know that there's a hang potential here and that it's easy to fix. Hangs on unattended system upgrades (or tinderboxes) aren't nice, and especially hangs that are hard to reproduce are unlikely to actually be reported.

floppym: Ok, I have added a retry limit.

floppym:
Closing this since it does not fix any known bugs and apparently there is a greater benefit for systems with poor single code performance.

PR #772, opened 2021-11-23
portage/util/_dyn_libs/dyn_libs.py: fix symlink recursion issue

floppym:
Python's glob.glob() function follows symlinks recursively. Use
os.walk() instead.

Also re-order the operands to short-circuit the directory walk when
PROVIDES is not empty.

Bug: https://bugs.gentoo.org/826982
Fixes: cba2156dba89a22f2858238013469b4d80208854


mgorny (approved):
I like it!

mgorny:
(that said, I suppose you could make it inline with some `any(...)` magic but the verbose way is nice to read too)

thesamesam:
Thanks.

PR #774, opened 2021-11-29
emerge: Default enable soname dependencies (bug 687956)

zmedico:
Default emerge --ignore-soname-deps=n, in order to enable
soname dependencies by default. As always, soname dependencies
remain inapplicable in the absence of the --usepkgonly option
(or --getbinpkgonly). Therefore, this change only affects
commands that specify --usepkgonly or --getbinpkgonly.

Bug: https://bugs.gentoo.org/687956


floppym (approved):
Sounds reasoable.

PR #776, opened 2021-12-04
portage.output: Replace darkblue color with teal

mgorny:
The "darkblue" color is often barely visible on dark terminals which
makes reading emerge output really hard (I basically have to copy-paste
it a lot in order to be able to read it at all).  Replace it with teal
that does not seem to have any significant use in the output.

Here's a before-after:
portage-teal


thesamesam (approved):

floppym:
I like it.

PR #780, opened 2021-12-22
Add -X shortopt for --exclude

idl0r:
Adding a shortopt, "-X", for --exclude


zmedico (approved):
LGTM, thanks!

I have a man page update that I'll merge with this.

PR #782, opened 2022-01-18
cnf/make.globals: use https by default for fetching distfiles

juippis:


floppym:
SSL on distfiles.gentoo.org is made possible by using a CDN.

@robbat2 expressed concern about what would happen if we turn the CDN off. He suggested adding both https and http to GENTOO_MIRRORS as a fallback.

I don't see much point in having both https and http since that makes it very easy for an attacker to force users from fetching over a secure channel to an insecure one.

robbat2:
To be clearer:
- if we have to turn off the CDN, we lose https.
- https DOES NOT add any integrity in our case (we have manifests and other mechanisms that provide integrity)
- https DOES provide privacy/secrecy: the trivial attackers can't see what you're downloading (but they could block HTTPS & force your downloads via HTTP)

If Portage uses the entries in order, there are no downsides to listing both. Multi-file fetch with HTTPS esp may be completed w/ HTTP2 and be faster than plain HTTP.

Please do add a comment that http is included since the non-CDN rotation does not provide SSL.

floppym:
> Multi-file fetch with HTTPS esp may be completed w/ HTTP2 and be faster than plain HTTP.

In theory, yes. However, Portage currently uses wget for fetching by default, which doesn't support HTTP2.

robbat2:
this is probably somewhere that the concepts from the GSOC2010 project "idfetch" are worthwhile. They didn't get significantly merged at the time, but it moved to a fetch daemon that could provide significant speedups when fetching many files.

cnf/make.globals
```diff
@@ -21,7 +21,7 @@ FCFLAGS=""
 
 # Default distfiles mirrors. This rotation has multiple hosts and is reliable.
 # Approved by the mirror-admin team.
-GENTOO_MIRRORS="http://distfiles.gentoo.org"
+GENTOO_MIRRORS="https://distfiles.gentoo.org"
```
juippis, line 24:
```suggestion
# Default distfiles mirrors. This rotation has multiple hosts and is reliable.
# https and http enabled; http as backup, since the non-CDN rotation
# does not provide SSL.
# Approved by the mirror-admin team.
GENTOO_MIRRORS="https://distfiles.gentoo.org http://distfiles.gentoo.org"
```
Like this?

ajakk:
> To be clearer:
> 
>     * if we have to turn off the CDN, we lose https.
> 
>     * https DOES NOT add any integrity in our case (we have manifests and other mechanisms that provide integrity)
> 
>     * https DOES provide privacy/secrecy: the trivial attackers can't see what you're downloading (but they could block HTTPS & force your downloads via HTTP)
> 
> 
> If Portage uses the entries in order, there are no downsides to listing both. Multi-file fetch with HTTPS esp may be completed w/ HTTP2 and be faster than plain HTTP.

If we're at all concerned about the privacy/secrecy part above, can't an attacker in between the user and distfiles.gentoo.org still MITM the connection and block HTTPS connections to force a downgrade?

> Please do add a comment that http is included since the non-CDN rotation does not provide SSL.

thesamesam:
The only benefit is privacy against a passive attacker/MITM. It's minor but cheap if we can opportunistically provide it.

floppym:
I would rather keep the current setting (http-only), and let privacy-conscious users override it to the https variant via make.conf. That way we are not pretending to be secure by default, and if the CDN goes down the user is responsible for updating their config.

floppym:
I'm closing this since I feel pretty strongly about my stance. Somebody else can merge it if I get out voted.

PR #785, opened 2022-01-28
Ignore all XPAK when "binpkg-request-signature" enabled.

syu-nya:

### DIFF
--- a/lib/portage/dbapi/bintree.py
+++ b/lib/portage/dbapi/bintree.py
@@ -42,8 +42,10 @@ from portage.exception import (
     ParseError,
     PortageException,
     PortagePackageException,
+    SignatureException,
 )
 from portage.localization import _
+from portage.output import colorize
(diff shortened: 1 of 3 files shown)


thesamesam:
In the commit message, I would note why (because signatures don't exist for those).

thesamesam:
Thank you!

lib/portage/dbapi/bintree.py
```diff
@@ -887,6 +889,14 @@ def _populate_local(self, reindex=True):
         # the Packages file will not be needlessly re-written due to
         # missing digests.
         minimum_keys = self._pkgindex_keys.difference(self._pkgindex_hashes)
+
+        if "binpkg-request-signature" in self.settings.features:
```
thesamesam, line 893: Merged but we should be able to do something like:
`gpkg_only = "binpkg-request-signature" in self.settings.features`
instead of if with true/false, right?

PR #787, opened 2022-02-09
Blacken previously missed Python files

ajakk:
These were found with this rough pipeline:

```sh
find . | xargs file | grep 'Python script' | cut -d':' -f 1 | xargs black
```


thesamesam:
Can you update CI to cover these too?

PR #789, opened 2022-02-21
Remove Personal EAPIs `4-python` and `5-progress`

mattst88:
Deprecated a year ago in commit 67cf9c2b05042de37f36f5b6840c450128a065bd ("Mark EAPIs "4-python" and "5-progress" as deprecated")

See https://archives.gentoo.org/gentoo-portage-dev/message/c4f7d25ab420123e51c9ec1befb7c24d —

    I've asked Arfrever multiple times if these are still used anywhere, and
    he seemingly has not responded intentionally.
    
    According to https://bugs.gentoo.org/174536#c27 these EAPIs were only
    used in Arfrever's personal overlay, and even in 2012 there were
    questions about why they were supported in portage.
    
    The "Progress Overlay" does contain ebuilds using these EAPIs but it has
    not been updated since 2018 and doesn't look like it is useful at this
    point.


Arfrever:
Could there firstly be an effectively working mechanism of EAPI plug-ins, which would allow ebuild repositories to inject support for custom EAPIs in Portage?

Also much of this code is useful and could be added to EAPI="9" or EAPI="10"...
(Would there be objection to moving this code to EAPI="8-progress" or EAPI="9-progress" and only removing older EAPIs?)

Maybe some of these features could be made to be repository-specific-configurable (i.e. they would be enabled e.g. by setting some attribute in `metadata/layout.conf` or in another file).

Due to frequent refactorings and reformattings in Portage source code, it is hard to externally maintain patchset for Portage. (I already locally maintain about 15 patches for Portage.)

mattst88:
> Could there firstly be an effectively working mechanism of EAPI plug-ins, which would allow ebuild repositories to inject support for custom EAPIs in Portage?

That's not something I'm personally interested in pursuing.

> Also much of this code is useful and could be added to EAPI="9" or EAPI="10"...
(Would there be objection to moving this code to EAPI="8-progress" or EAPI="9-progress" and only removing older EAPIs?)

Yes, I would object. This code has existed for ~10 years. If it was going to be enabled by new EAPIs it would have happened by now. At the same time, it's trivial to revive bits of it if we want to in the future.

> Maybe some of these features could be made to be repository-specific-configurable (i.e. they would be enabled e.g. by setting some attribute in metadata/layout.conf or in another file).

I'm not interested in that either.

> Due to frequent refactorings and reformattings in Portage source code, it is hard to externally maintain patchset for Portage. (I already locally maintain about 15 patches for Portage.)

Now that the codebase has been reformatted with black and we have a GitHub CI check for black, I don't think it will be difficult to manage patches.

thesamesam:
> > Due to frequent refactorings and reformattings in Portage source code, it is hard to externally maintain patchset for Portage. (I already locally maintain about 15 patches for Portage.)
> 
> Now that the codebase has been reformatted with black and we have a GitHub CI check for black, I don't think it will be difficult to manage patches.

Yes, agreed. That shouldn't be an issue anymore. But also, I really wish we had such extensive refactorings. Not sure they happen though right now.

Arfrever:
Even if you want to continue with this, please firstly apply #754 to avoid creating another major conflicts for my patches.

> …or you could just use regular EAPIs instead of putting a lot of work on others because you have to be special and use your own special EAPI.

Sometimes other users have asked for features which happen to be implemented in these EAPIs.
Examples:
https://bugs.gentoo.org/461948
https://bugs.gentoo.org/541398

A use case was mentioned in https://bugs.gentoo.org/414815#c10 but was silently ignored by @ulm in https://bugs.gentoo.org/414815#c11

> > (Would there be objection to moving this code to EAPI="8-progress" or EAPI="9-progress" and only removing older EAPIs?)
> 
> Yes, I would object. This code has existed for ~10 years. If it was going to be enabled by new EAPIs it would have happened by now.

Unfortunately the PMS team rejects many useful ideas.
It would be sad if Portage was not allowed to contain outside-of-PMS features.
(gentoo.git may have requirement to closer follow PMS, but maintainers of other repositories should have free will to make decisions.)

mattst88:
> Even if you want to continue with this, please firstly apply #754 to avoid creating another major conflicts for my patches.

I would be happy to. Could you help review the patches? I left some questions yesterday but I'm not expecting to hear from the author who appears to have disappeared from Gentoo recently.

mattst88:
I plan to push this in a day or two unless there are further comments.

PR #790, opened 2022-03-05
Allow post_emerge to skip some expensive operations.

antarus12345:
This is the start of something for https://bugs.gentoo.org/792195

I suspect a second patchseries will need some setting to enable this; I presume most user do not want this behavior (and in fact want these things to run every time.)

Signed-off-by: Alec Warner <antarus@gentoo.org>


lib/_emerge/post_emerge.py
```diff
@@ -88,6 +89,16 @@ def post_emerge(myaction, myopts, myfiles, target_root, trees, mtimedb, retval):
     settings = vardbapi.settings
     info_mtimes = mtimedb["info"]
 
+    # Do not update the mtimedb if we updated it recently.
+    if time.time() - mtimedb.lastUpdated() > 60 * 5:
```
Arfrever, line 94: Does it affect updates of `resume` / `resume_backup` lists in mtimedb (which are quite important)?

PR #792, opened 2022-03-06
emerge --keep-going: deduplicate atoms in 'dropped' message

ajakk:
The --keep-going 'dropped' message sometimes contains duplicate atoms:

 * emerge --keep-going: app-containers/podman-4.0.2 dropped because it
 * requires app-containers/catatonit, app-containers/catatonit

This patch deduplicates the atoms printed in that message:

 * emerge --keep-going: app-containers/podman-4.0.2 dropped because it
 * requires app-containers/catatonit


zmedico:
Thanks!

PR #793, opened 2022-03-09
Drop -b from shebangs

ajakk:
The switch to using '/usr/bin/env python' shebangs doesn't naively
work with arguments after 'python', so the shebangs were broken. '-b'
is of dubious necessity anyway, so let's just drop it.

Fixes: 41f4f6d25019c4f944b8d9576d96482f92231fda


floppym (approved):

floppym:
Pushed as 781aefd8be33c3bd699e5a665225013c5f76b531.

PR #794, opened 2022-03-15
estrip: avoid calling scanelf twice

thesamesam:
We can use the previous scanelf data to not call it again to find
all of the dynamically linked executables/libraries which need stripping.

Bug: https://bugs.gentoo.org/749624

### DIFF
--- a/bin/estrip
+++ b/bin/estrip
@@ -410,8 +410,9 @@ while read -r x ; do
 done < <(
 	# Use sort -u to eliminate duplicates for bug #445336.
 	(
-		scanelf -yqRBF '#k%F' -k '.symtab' "$@"
 		find "$@" -type f ! -type l -name '*.a'
+		cut -d ' ' -f1 < "${PORTAGE_BUILDDIR}"/build-info/NEEDED \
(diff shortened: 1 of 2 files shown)


bin/estrip
```diff
@@ -410,8 +410,8 @@ while read -r x ; do
 done < <(
 	# Use sort -u to eliminate duplicates for bug #445336.
 	(
-		scanelf -yqRBF '#k%F' -k '.symtab' "$@"
-		find "$@" -type f ! -type l -name '*.a'
+		cut -d ' ' -f1 < "${PORTAGE_BUILDDIR}"/build-info/NEEDED \
+			| sed -e "s:^:${ED}:"
```
zmedico, line 414: It looks like it's not going to include `*.a` here, so maybe add that find call back. For reference, I have ncurses-6.2_p20210619 installed, and CONTENTS contains `*.a` files which do not appear in NEEDED.

thesamesam: Oh, right! Good spot.

zmedico (approved):
LGTM. Thanks!

thesamesam:
Thank you!

PR #795, opened 2022-03-15
autoclean: only enable for global updates and depcleans

thesamesam:
Closes: https://bugs.gentoo.org/792195

### DIFF
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -672,12 +672,6 @@ def action_build(
                         ldpath_mtimes,
                         autoclean=1,
                     )
-                else:
-                    portage.writemsg_stdout(
-                        colorize("WARN", "WARNING:")
-                        + " AUTOCLEAN is disabled.  This can cause serious"
-                        + " problems due to overlapping packages.\n"
-                    )
 
         return retval
 
@@ -775,6 +769,15 @@ def action_depclean(
     # that should have been pulled into the graph. On the other hand, it's
     # relatively safe to ignore missing deps when only asked to remove
     # specific packages.
+
+    # Force autoclean for depcleans (but not purges), as it was changed
+    # to default off to not run it on every unmerge.
+    # bug #792195
+    if action == "depclean":
+        settings.unlock()
+        settings["AUTOCLEAN"] = "yes"
+        settings.backup_changes("AUTOCLEAN")
+        settings.lock()
 
     msg = []
     if (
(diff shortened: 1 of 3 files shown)


lib/_emerge/actions.py
```diff
@@ -2903,6 +2910,7 @@ def load_emerge_config(emerge_config=None, env=None, **kargs):
         ("target_root", "ROOT"),
         ("sysroot", "SYSROOT"),
         ("eprefix", "EPREFIX"),
+        ("autoclean", "AUTOCLEAN"),
```
zmedico, line 2913: Looks like this is a stray change.

thesamesam: Thanks!

thesamesam:
@zmedico How should we handle the "AUTOCLEAN is disabled.  This can cause serious ..." message? Just drop it now?

lib/_emerge/actions.py
```diff
@@ -776,6 +776,13 @@ def action_depclean(
     # relatively safe to ignore missing deps when only asked to remove
     # specific packages.
 
+    # Let's autoclean if we applied updates, rather than always doing it
+    # bug #792195
```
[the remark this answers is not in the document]
thesamesam: Oh right, thanks! Done on both counts.

zmedico:
> @zmedico How should we handle the "AUTOCLEAN is disabled. This can cause serious ..." message? Just drop it now?

I don't think that the autoclean code in vartree.py should be disabled by default, since that really does have dangerous implications (old version is not uninstalled during upgrade). For the purposes of this PR, I think it would make sense to introduce a separate GLOBAL_AUTOCLEAN variable which only affects the global autoclean action that you are trying to optimize.

thesamesam:
I'm wondering if instead we should just unconditionally autoclean for that case?

zmedico:
Sure, unconditional autoclean in vartree.py seems reasonable. The AUTOCLEAN=n case there really doesn't seem worth supporting.

lib/portage/dbapi/vartree.py
```diff
@@ -5086,21 +5086,18 @@ def onerror(e):
...
-
-        if autoclean:
-            emerge_log(_(" >>> AUTOCLEAN: %s") % (slot_atom,))
+        emerge_log(_(" >>> AUTOCLEAN: %s") % (slot_atom,))
 
         others_in_slot.append(self)  # self has just been merged
         for dblnk in list(others_in_slot):
             if dblnk is self:
                 continue
-            if not (autoclean or dblnk.mycpv == self.mycpv or reinstall_self):
```
thesamesam, line 5102: When changing to just:
```
	if not (dblnk.mycpv == self.mycpv or reinstall_self):
```

this broke the `lib/portage/tests/emerge/test_emerge_slot_abi.py` test, presumably because something overlapping was left, and normally we were testing with autoclean on.

@zmedico let me know if I should somehow adapt the test (and how?) if I haven't made the right fix here. It seems right to me to just _always_ autoclean here though like we discussed but I might be missing something (it feels wrong the test broke in the AUTOCLEAN=no case?)

zmedico: Right, we always want to cleanup after ourselves. The rest of this condition was for the "autoclean is False" case, which no longer exists, so the whole condition is unnecessary now.

zmedico (approved):
LGTM

thesamesam:
Thanks for all the help Zac, really appreciated!

PR #800, opened 2022-03-31
GitSync.update: default to "--depth 1" to avoid fetching whole Git history

duxsco:
This commit is explained in the corresponding bug report: https://bugs.gentoo.org/824782#c17

PR #801, opened 2022-04-02
GitSync.update: default to "--depth 1" (bug 824782 comment 17)

duxsco:
Enforce the use of "--depth" in both GitSync.new and GitSync.update following the same logic.

Each function gets its own designated portage option of "clone-depth" and "sync-depth". This means that portage option "sync-depth" is not taken into consideration anymore in GitSync.new.

Portage option "clone-depth" and "sync-depth" lead to following behaviour with a value of:
0: The use of "--depth <value>" is disabled causing the retrieval of the whole Git history.
ℕ (positive integer): "--depth <value>" is used.
Unset portage option: "--depth 1" is used.

### DIFF
--- a/lib/portage/sync/modules/git/git.py
+++ b/lib/portage/sync/modules/git/git.py
@@ -84,9 +84,6 @@ class GitSync(NewBase):
         if self.repo.clone_depth is not None:
             if self.repo.clone_depth != 0:
                 git_cmd_opts += " --depth %d" % self.repo.clone_depth
-        elif self.repo.sync_depth is not None:
-            if self.repo.sync_depth != 0:
-                git_cmd_opts += " --depth %d" % self.repo.sync_depth
         else:
             # default
             git_cmd_opts += " --depth 1"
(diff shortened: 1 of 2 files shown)


duxsco:
I accidentally deleted the wrong repo and https://github.com/gentoo/portage/pull/800 got closed. sry

duxsco:
Sry for the inconvenience.

duxsco:
I added now the "Signed-off-by" line to the commit message. Beware that the current state of this PR doesn't solve: https://github.com/gentoo/portage/pull/800#issuecomment-1086965767

duxsco:
The issue linked in my previous comment occured again. But, it vanished after executing `emerge --sync` several times.

ajakk:
> The issue linked in my previous comment occured again. But, it vanished after executing `emerge --sync` several times.

I'm really not sure how that issue happened. I've been running with `::gentoo`'s sync depth set to 1, and I've never seen such a problem, though I have seen some issues where the `::gentoo` on disk somehow diverges from the remote at https://github.com/gentoo-mirror/gentoo. To fix that we'll need to clobber the repo like rsync does.

ajakk:
I've also applied this patch locally, synced, and got no breakage.

zmedico (approved):
LGTM. Thanks!

thesamesam:
Thanks for reviewing Zac!

PR #805, opened 2022-04-10
Fix dispatch-conf

floppym:


floppym:
CC; @kenrap

kenrap:
Fair enough. I will avoid the dispatch related files. There is a `diff_mixed_wrapper` class that's makes the code confusing for me to understand and fix.

Thanks for taking the time to resolve this.

thesamesam (approved):

PR #807, opened 2022-04-11
eend: Output QA warning if ebegin has not been called

laumann:
The assumption here is that ebegin-eend aren't nested.

Closes: https://bugs.gentoo.org/835824


laumann:
Also, I intend to add a similar check to pkgcheck.

zmedico (approved):

thesamesam:
I get a false positive with openvpn in pkg_setup I think?

laumann:
> I get a false positive with openvpn in pkg_setup I think?

I think the problem is in linux-info.eclass. The `check_extra_config()` function has an `eend 0` that looks pretty lonely: https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/linux-info.eclass#n870

PR #811, opened 2022-04-12
ebegin: Output QA warning if call to eend is missing

laumann:
The idea here is to check if EBEGIN_EEND is set after the phase has been
executed. If so, then a call to eend is probably missing. This is under
the assumption that ebegin-eend invocations should be properly paired
within the same phase.

In ebegin, the EBEGIN_EEND variable is also checked, and, if set,
indicates that ebegin has been called and is missing the closing call to
eend.

I doubt that this check is perfect, but it seems to work in what little
testing I've done.

Closes: https://bugs.gentoo.org/835823

### DIFF
--- a/bin/isolated-functions.sh
+++ b/bin/isolated-functions.sh
@@ -339,6 +339,9 @@ ebegin() {
 	[[ ${RC_ENDCOL} == "yes" ]] && echo >&2
 	LAST_E_LEN=$(( 3 + ${#RC_INDENTATION} + ${#msg} ))
 	LAST_E_CMD="ebegin"
+	if [[ -v EBEGIN_EEND ]] ; then
+		eqawarn "QA Notice: ebegin called, but missing call to eend (phase: ${EBUILD_PHASE})"
+	fi
 	EBEGIN_EEND=1
 	return 0
 }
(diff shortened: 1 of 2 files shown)


laumann:
For the sake of completeness, here's the very simple ebuild I've tested with:

```bash
# Copyright 2022 Gentoo Authors

EAPI=8

DESCRIPTION="Testing package"
HOMEPAGE="http://nowhere.com"

LICENSE="GPL-2"
SLOT="0"

pkg_pretend() {
	ebegin "well hello there"
	ebegin "well hello again"
}

pkg_setup() {
	eend 0
}
```
and the output:
```
$ ebuild dev-libs/foobar/foobar-0.0.1.ebuild clean setup
Appending /home/tj/sources/gentoo/portage-tester to PORTDIR_OVERLAY...
 * well hello there ...
 * well hello again ...
 * QA Notice: ebegin called, but missing call to eend (phase: pretend)
 * QA Notice: ebegin called, but missing call to eend (phase: pretend)
 * QA Notice: eend called without preceding ebegin (phase: setup)
 [ ok ]
```

thesamesam (approved):

zmedico (approved):
LGTM

PR #812, opened 2022-04-12
cleanup/style: Replace '== None' with 'is None' and '!= None' with 'is not None'

laumann:
**cleanup: replace '== None' with 'is None'**

Essentially the following script:

    find -name '*.py' |xargs sed -i -e 's/== None/is None/g'

Reading through DEVELOPING it says to prefer 'foo is not None' over 'foo
!= None' and grepping for '== None' got a few hits.

**cleanup: replace '!= None' with 'is not None'**

Result from running this:

    find -name '*.py' |xargs sed -i -e 's/!= None/is not None/g'

Same as the previous commit, just figured I'd keep the results in
separate commits.

### DIFF
--- a/bin/ebuild
+++ b/bin/ebuild
@@ -435,7 +435,7 @@ for arg in pargs:
     except PermissionDenied as e:
         portage.writemsg("!!! Permission Denied: %s\n" % (e,), noiselevel=-1)
         a = 1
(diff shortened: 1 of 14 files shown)


thesamesam:
Thanks!

PR #815, opened 2022-04-13
ebuild.sh: add missing IDEPEND re-sourcing

thesamesam:


thesamesam:
I'm not sure what this is about but I figured I'd throw this up either as a reminder to look into it or for somebody else to confirm it looks odd to not be there.

zmedico:
It looks like we're also missing a B_IDEPEND setting where B_DEPEND etc are set.

zmedico (approved):
LGTM

thesamesam:
thanks!

PR #816, opened 2022-04-14
movefile: merge symlinks atomically

thesamesam:
Since POSIX allows renaming symlinks with symlinks, use that to get
atomic updates.  This is faster and less flaky to parallel processes:
removing a live symlink and then recreating it leaves a small window
where other things might try to use it but fail.

[sam: cherry-picked from chromiumos' third_party/portage_tool repo]
(cherry picked from commit cherry-pick 80ed2dac95db81acac8043e6685d0a853a08d268)
Bug: https://bugs.gentoo.org/836400

### DIFF
--- a/lib/portage/util/movefile.py
+++ b/lib/portage/util/movefile.py
@@ -171,7 +171,7 @@ def movefile(
             bsd_chflags.chflags(os.path.dirname(dest), 0)
 
     if destexists:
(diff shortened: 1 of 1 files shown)


zmedico (approved):
LGTM

thesamesam:
Thanks!

PR #817, opened 2022-04-14
Fix an instance of git-rev-parse called as root when FEATURES=usersync

ajakk:


ajakk:
Closing in favor of mike's

PR #818, opened 2022-04-14
Add portage.process.check_output() function and use it in GitSync class

floppym:
This allows us to utilize existing logic for changing the user id in the
called process.

### DIFF
--- a/lib/portage/sync/modules/git/git.py
+++ b/lib/portage/sync/modules/git/git.py
@@ -104,7 +104,7 @@ class GitSync(NewBase):
 
         exitcode = portage.process.spawn_bash(
             "cd %s ; exec %s" % (portage._shell_quote(self.repo.location), git_cmd),
-            **self.spawn_kwargs
+            **self.spawn_kwargs,
         )
         if exitcode != os.EX_OK:
             msg = "!!! git clone error in %s" % self.repo.location
@@ -154,7 +154,7 @@ class GitSync(NewBase):
 
         try:
             remote_branch = portage._unicode_decode(
-                subprocess.check_output(
+                portage.process.check_output(
                     [
                         self.bin_command,
                         "rev-parse",
@@ -163,6 +163,7 @@ class GitSync(NewBase):
                         "@{upstream}",
                     ],
                     cwd=portage._unicode_encode(self.repo.location),
+                    **self.spawn_kwargs,
                 )
             ).rstrip("\n")
         except subprocess.CalledProcessError as e:
@@ -184,7 +185,7 @@ class GitSync(NewBase):
             exitcode = portage.process.spawn(
                 gc_cmd,
                 cwd=portage._unicode_encode(self.repo.location),
-                **self.spawn_kwargs
+                **self.spawn_kwargs,
             )
             if exitcode != os.EX_OK:
                 msg = "!!! git gc error in %s" % self.repo.location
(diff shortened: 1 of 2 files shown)


floppym:
In testing, I found that syncing with `sync-git-verify-commit-signature` fails when we run git as the "portage" user, probably due to a permissions issue on the temporary directory that we use for WKD key refresh. I'll have to look into that.

lib/portage/sync/modules/git/git.py
```diff
@@ -285,11 +290,13 @@ def verify_head(self, revision="-1"):
 
             rev_cmd = [self.bin_command, "log", "-n1", "--pretty=format:%G?", revision]
             try:
+                kwargs = self.spawn_kwargs.copy()
+                kwargs["env"] = env
                 status = portage._unicode_decode(
-                    subprocess.check_output(
+                    portage.process.check_output(
```
floppym, line 296: This call is problematic. The files gnupg files generated by gemato will be owned by root, but we need to run git as non-root to avoid the "'/var/db/repos/gentoo' is owned by someone else" error.

I think we would need to invoke the gemato setup code (`self._get_openpgp_env()`) as the appropriate non-root user.

thesamesam:
@ajakk

floppym:
This PR breaks emerge --info when a git repo is configured.
```
% emerge --info
  File "/home/floppym/src/portage/bin/emerge", line 60, in <module>
    retval = emerge_main()
  File "/home/floppym/src/portage/lib/_emerge/main.py", line 1294, in emerge_main
    return run_action(emerge_config)
  File "/home/floppym/src/portage/lib/_emerge/actions.py", line 3942, in run_action
    return action_info(
  File "/home/floppym/src/portage/lib/_emerge/actions.py", line 1956, in action_info
    head_commit = sync.retrieve_head(options=options)
  File "/home/floppym/src/portage/lib/portage/sync/modules/git/git.py", line 367, in retrieve_head
    portage.process.check_output(
TypeError: portage.process.check_output() argument after ** must be a mapping, not NoneType
```

ajakk:
Lifting the uid/gid/etc grabbing functionality from controller.py, we can get something like:

```
diff --git a/lib/portage/sync/modules/git/git.py b/lib/portage/sync/modules/git/git.py
index 10dabb5a4..4b4e8cf45 100644
--- a/lib/portage/sync/modules/git/git.py
+++ b/lib/portage/sync/modules/git/git.py
@@ -3,6 +3,7 @@

 import io
 import logging
+import pwd
 import subprocess

 import portage
@@ -361,6 +362,27 @@ class GitSync(NewBase):
             return (1, False)
         rev_cmd = [self.bin_command, "rev-list", "--max-count=1", "HEAD"]
         try:
+            repo = kwargs['options']['repo']
+            st = os.stat(repo.location)
+            try:
+                pw = pwd.getpwuid(st.st_uid)
+            except KeyError:
+                pass
+            else:
+                # Drop privileges when syncing, in order to match
+                # existing uid/gid settings.
+                self.spawn_kwargs = {}
+                self.usersync_uid = st.st_uid
+                self.spawn_kwargs["uid"] = st.st_uid
+                self.spawn_kwargs["gid"] = st.st_gid
+                self.spawn_kwargs["groups"] = [st.st_gid]
+                self.spawn_kwargs["env"] = {}
+                self.spawn_kwargs["env"]["HOME"] = pw.pw_dir
+                self.spawn_kwargs["env"]["LOGNAME"] = pw.pw_name
+                umask = 0o002
+                if not st.st_mode & 0o020:
+                    umask = umask | 0o020
+                self.spawn_kwargs["umask"] = umask
             ret = (
                 os.EX_OK,
                 portage._unicode_decode(
```

Though that duplicates a bit of code and I haven't dug into trying to make it more generic to fit into its own function.

floppym:
I'm really not sure where to go with this one. I feel like some code needs to be reworked (possibly in gemato), or we need to approach this problem from a different angle.

I wonder if we could just disable git's new behavior without re-introducing whatever security issue it "solves".

thesamesam:
We might be able to do something like `git config --add safe.directory /path/to/sync/dir` if `sync-type = git`.

PR #819, opened 2022-04-14
Delete repoman

mattst88:


mattst88:
I think this is ready. Please review/ack.

mgorny (approved):
WFM, I guess. I also guess we're going to need to require prior Portage version in our existing repoman ebuilds, until they're gone.

mattst88:
> WFM, I guess. I also guess we're going to need to require prior Portage version in our existing repoman ebuilds, until they're gone.

I could split the changes to `lib/` in the first patch out, if that would allow us to retain compatibility. We could commit those small clean ups after `app-portage/repoman` is gone from the tree.

@kenrap: would that allow your clean ups to go ahead? Would your clean ups break `app-portage/repoman`'s ability to run against portage?

kenrap:
@mattst88 

> @kenrap: would that allow your clean ups to go ahead? Would your clean ups break `app-portage/repoman`'s ability to run against portage?

With my clean ups as they are now, they would break repoman's ability to run against portage. I would have to go through that traumatizing and intense labor again to fix repoman.

However, as an experiment by testing your changes with the work I did, I would like to share about those experiences.

I added a remote to your portage fork and fetched your branches to checkout your `delete-repoman` branch. And there were two merge conflicts to fix. After integrating them, I ran the tests and got one error from `lib/portage/util/install_mask.py` saying that `parent.decode()` on line 178 was a string and can't be decoded. Seeing that made me sweat with fear. From running a few tests again after an adjustment each time, it became clear that it was a circular type error. I didn't give up and I wanted to see if I can solve it. With great luck, I literally overcame one!! And both Python 3.9 and 3.10 tests passed for me. I don't know if it was me becoming more familiar with the codebase or my pattern recognition skills improved. I might have a chance at solving those other functions with their local coercion closure hacks.

However this experiment won't solve repoman's ebuild compatibility issue as mentioned earlier.

thesamesam (approved):

PR #823, opened 2022-04-20
phase-functions.sh: do not "addpredict /" for src_test

floppym:
Silencing the sandbox by default makes it more difficult to track down
file access problems that may effect tests.

Removing this will likely cause an increased number of test failures
until affected ebuilds can be adjusted.

Bug: https://bugs.gentoo.org/836671


zmedico (approved):

floppym:
Before this gets merged, we should ensure that most of the bugs blocking [bug 839747](https://bugs.gentoo.org/839747) have been addressed.

floppym:
The fallout from this is too much; it's disturbingly common for test suites to try things like `rmdir /` with failure as the expected result.

Closing this.

PR #824, opened 2022-04-21
ebegin/eend: accept properly nested calls in different functions

laumann:
Turn the EBEGIN_EEND variable into a stack that tracks which function
called ebegin. Calls to ebegin may be stacked and do not generate a QA
warning if the callers are different functions. Calls to eend then check
that the function name at the top of the stack matches the caller's
function name.

Bug: https://bugs.gentoo.org/839585
Bug: https://bugs.gentoo.org/839588

### DIFF
--- a/bin/isolated-functions.sh
+++ b/bin/isolated-functions.sh
@@ -340,9 +340,15 @@ ebegin() {
 	LAST_E_LEN=$(( 3 + ${#RC_INDENTATION} + ${#msg} ))
 	LAST_E_CMD="ebegin"
 	if [[ -v EBEGIN_EEND ]] ; then
-		eqawarn "QA Notice: ebegin called, but missing call to eend (phase: ${EBUILD_PHASE})"
+		# Already a call to ebegin
+		local prev="${EBEGIN_EEND[-1]}"
+		if [[ "${prev}" == "${FUNCNAME[1]}" ]] ; then
+			eqawarn "QA Notice: ebegin called in ${prev}, but missing call to eend (${FUNCNAME[1]})"
+		fi
+		EBEGIN_EEND+=( "${FUNCNAME[1]}" )
+	else
+		EBEGIN_EEND=( "${FUNCNAME[1]}" )
 	fi
-	EBEGIN_EEND=1
 	return 0
 }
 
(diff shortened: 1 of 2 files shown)


laumann:
Small demo:

```bash
EAPI=8

DESCRIPTION="Testing package"
HOMEPAGE="http://nowhere.com"

LICENSE="GPL-2"
SLOT="0"

somefunc() {
	ebegin "i'm somefunc"
	#eend 0
}

pkg_pretend() {
	ebegin "well hello there"
	somefunc
	eend 0
}
```

and we get:
```
 * well hello there ...
 * i'm somefunc ...
 * QA Notice: eend (in pkg_pretend) improperly matched with ebegin (called in somefunc)
 [ ok ]
 * QA Notice: ebegin called in pkg_pretend but missing call to eend
```

Uncommenting the `eend 0` line, produces no warnings (but the output does get messed up a bit):
```
Appending /home/tj/sources/gentoo/portage-tester to PORTDIR_OVERLAY...
 * well hello there ...
 * i'm somefunc ...
 [ ok ]
 [ ok ]
```

bin/isolated-functions.sh
```diff
@@ -340,9 +340,15 @@ ebegin() {
 	LAST_E_LEN=$(( 3 + ${#RC_INDENTATION} + ${#msg} ))
 	LAST_E_CMD="ebegin"
 	if [[ -v EBEGIN_EEND ]] ; then
-		eqawarn "QA Notice: ebegin called, but missing call to eend (phase: ${EBUILD_PHASE})"
+		# Already a call to ebegin
+		local prev="${EBEGIN_EEND[-1]}"
+		if [[ "${prev}" == "${FUNCNAME[1]}" ]] ; then
+			eqawarn "QA Notice: ebegin called in ${prev}, but missing call to eend (${FUNCNAME[1]})"
```
laumann, line 346: err, maybe not the best message, as ${prev} and ${FUNCNAME[1]} will be the same string :-)

How about:
```
QA Notice: ebegin already called in ${prev}, but missing call to eend
```
?

bin/isolated-functions.sh
```diff
@@ -340,9 +340,15 @@ ebegin() {
 	LAST_E_LEN=$(( 3 + ${#RC_INDENTATION} + ${#msg} ))
 	LAST_E_CMD="ebegin"
 	if [[ -v EBEGIN_EEND ]] ; then
-		eqawarn "QA Notice: ebegin called, but missing call to eend (phase: ${EBUILD_PHASE})"
+		# Already a call to ebegin
+		local prev="${EBEGIN_EEND[-1]}"
+		if [[ "${prev}" == "${FUNCNAME[1]}" ]] ; then
```
laumann, line 345: It was pointed out to me that properly recursive functions with properly nested ebegin-eend will still generate a QA notice here.

floppym:
Maybe it would be better to ban nested ebegin calls and adjust the affected eclasses to not do that.

laumann:
> Maybe it would be better to ban nested ebegin calls and adjust the affected eclasses to not do that.

The more I've thought about it, the more I would agree.

PR #825, opened 2022-04-22
MiscFunctionsProcess: use a separate SANDBOX_LOG

floppym:
This avoids clobbering the main sandbox.log file when
"misc-functions.sh die_hooks" is invoked after an ebuild phase.

Signed-off-by: Mike Gilbert <floppym@gentoo.org>


thesamesam:
Oh, is this why sandbox.log is blank so often? 👀

floppym:
> Oh, is this why sandbox.log is blank so often?

Yes, it's why we are seeing `head: cannot open '/var/tmp/portage/dev-vcs/rcs-5.10.1/temp/sandbox.log' for reading: No such file or directory` in the latest tinderbox bug reports.

zmedico (approved):

PR #827, opened 2022-04-26
install-qa-check.d/60pkgconfig: add QA_PKGCONFIG_VERSION

floppym:
This allows ebuild maintainers to override the expected version in the .pc file when it differs from ${PV}.


thesamesam (approved):
For cases where several libraries are being installed, maybe we should allow setting it to blank and just skip the check?

anyway, lgtm

floppym:
I pushed an update that skips the check if QA_PKGCONFIG_VERSION is set to empty string.

floppym:
Pushed as 2c25a3a6e88f32d94c63ab38baa34f2d79a2699e.

PR #828, opened 2022-04-29
Use writemsg util to fix duplicate log issue

ezzieyguywuf:
Closes: https://bugs.gentoo.org/838406

Signed-off-by: Wolfgang E. Sanyer <WolfgangESanyer@gmail.com>


lib/portage/eapi.py
```diff
@@ -3,6 +3,7 @@
 
 import collections
 import logging
```
thesamesam, line 5: Should we maybe do `from logging import logging.warning` or something instead?

ezzieyguywuf: I tend to prefer just the module, since it provides a form of documentation in the code itself regarding where functions/constants come from.

Are you thinking it's more lightweight to import just the one constant?

I dug into it a bit, and [this SO](https://softwareengineering.stackexchange.com/a/187471) seems to suggest that the same import occurs, just with different semantics.

kenrap: > Are you thinking it's more lightweight to import just the one constant?

Objectively speaking, using `logging.WARNING` in a few places is not going to have any real performance hit. It only matters when the dot operator does multiple lookups during a loop, especially very large loops. For single use statements, what @ezzieyguywuf did is fine and is more explicit too.

thesamesam: I'd got confused -- the original issue wasn't an import-time thing, it was to do with logging and implicit logger name. So yeah, all fine!

kenrap:
Question: Is there any reason to delay merging this PR? Given how small the diff is, shouldn't merging this be a no-brainer?

Personally speaking, I'm getting tired of seeing this output all of the time:
```bash
# revdep-rebuild
WARNING:root: * This is the new python coded version
 * This is the new python coded version
WARNING:root: * Please report any bugs found using it.
 * Please report any bugs found using it.
WARNING:root: * The original revdep-rebuild script is installed as revdep-rebuild.sh
 * The original revdep-rebuild script is installed as revdep-rebuild.sh
WARNING:root: * Please file bugs at: https://bugs.gentoo.org/
 * Please file bugs at: https://bugs.gentoo.org/
WARNING:root: * Collecting system binaries and libraries
 * Collecting system binaries and libraries
WARNING:root: * Checking dynamic linking consistency
 * Checking dynamic linking consistency
WARNING:root:
Your system is consistent

Your system is consistent
```

ezzieyguywuf:
> Question: Is there any reason to delay merging this PR? Given how small the diff is, shouldn't merging this be a no-brainer?
> 
> Personally speaking, I'm getting tired of seeing this output all of the time:
> 
> ```shell
> # revdep-rebuild
> WARNING:root: * This is the new python coded version
>  * This is the new python coded version
> WARNING:root: * Please report any bugs found using it.
>  * Please report any bugs found using it.
> WARNING:root: * The original revdep-rebuild script is installed as revdep-rebuild.sh
>  * The original revdep-rebuild script is installed as revdep-rebuild.sh
> WARNING:root: * Please file bugs at: https://bugs.gentoo.org/
>  * Please file bugs at: https://bugs.gentoo.org/
> WARNING:root: * Collecting system binaries and libraries
>  * Collecting system binaries and libraries
> WARNING:root: * Checking dynamic linking consistency
>  * Checking dynamic linking consistency
> WARNING:root:
> Your system is consistent
> 
> Your system is consistent
> ```

Ugh, just realized I accidentally changed the log level to WARNING, I think INFO makes more sense.

I can probs fix it tomorrow fwiw

thesamesam:
> Question: Is there any reason to delay merging this PR? Given how small the diff is, shouldn't merging this be a no-brainer?
> 

There's only one sam, although other people have privileges to merge - I haven't had a chance to test it yet. It doesn't help that I was unwell this week either 😞.

> Personally speaking, I'm getting tired of seeing this output all of the time:
> 

You could put the patch (https://github.com/gentoo/portage/pull/828.patch) in `/etc/portage/patches/sys-apps/portage/fix-log-spam.patch` to kill two birds with one stone: stop the annoying message and then also tell me it works fine for you in daily usage.

kenrap:
I'm going to swap out the `WARNING` with `INFO` in the patch. Might as well at least.

ezzieyguywuf:
> > Question: Is there any reason to delay merging this PR? Given how small the diff is, shouldn't merging this be a no-brainer?
> 
> There's only one sam, although other people have privileges to merge - I haven't had a chance to test it yet. It doesn't help that I was unwell this week either 😞.
> 
> > Personally speaking, I'm getting tired of seeing this output all of the time:
> 
> You could put the patch (https://github.com/gentoo/portage/pull/828.patch) in `/etc/portage/patches/sys-apps/portage/fix-log-spam.patch` to kill two birds with one stone: stop the annoying message and then also tell me it works fine for you in daily usage.

I updated the log-level to `DEBUG`. I got some anecdotal reports that logging the cache information is mostly spam, e.g. not useful.

I'm not quite sure how to reproduce, but I'm fine with removing this log altogether if it's too much trouble. Or perhaps I can comment it out and add a note to future devs to uncomment if they need/want the information for debugging/profiling

floppym (changes requested):
This change causes the cache info output to show up in `portageq envvar -v`, which breaks etc-update.
```
% ./portageq envvar -v CONFIG_PROTECT
cache info: CacheInfo(hits=0, misses=1, maxsize=32, currsize=0)
cache info: CacheInfo(hits=1, misses=2, maxsize=32, currsize=1)
cache info: CacheInfo(hits=146, misses=3, maxsize=32, currsize=2)
CONFIG_PROTECT="/etc /usr/share/config /usr/share/gnupg/qualified.txt"
```

thesamesam:
This is the last issue blocking the next release (https://github.com/gentoo/portage/pull/848).

floppym:
I think this should be fixed in gentoolkit instead, since it add a custom log handler.

See https://github.com/gentoo/gentoolkit/pull/22.

ezzieyguywuf:
> I think this should be fixed in gentoolkit instead, since it add a custom log handler.
> 
> See https://github.com/gentoo/gentoolkit/pull/22.

The gentoolkit fix is similar to what I was planning to do here. Not sure if it's appropriate to do the same thing in both places (probably is)

floppym:
We don't want `writemsg_level` here. That's going to send output to stdout, which is not desired.

ezzieyguywuf:
> We don't want `writemsg_level` here. That's going to send output to stdout, which is not desired.

Even though the level is DEBUG?

floppym:
> Even though the level is DEBUG?

writemsg sends any level above WARNING to stderr. Any level below WARNING (including DEBUG) gets sent to stdout.

The noiselevel determines if the message gets output at all. It defaults to 0. This gets compared to the noiselimit global variable, which also defaults to 0. If the noiselevel is greater than or equal to the noiselimit, the message gets output. The noiselimit may be decreased to -1 by passing `--quiet` to emerge.

We could maybe set noiselevel=1 for this debug message if we really want to use writemsg.

However, I don't see any obvious reason to switch from using the `logging` module to using `writemsg` here.

floppym:
Also, in https://github.com/gentoo/portage/pull/849, I change the logging call from `logging.info` to `logger.debug`.

ezzieyguywuf:
> However, I don't see any obvious reason to switch from using the `logging` module to using `writemsg` here.

I think the original motivation for using `writemsg` was in order to avoid duplicating log output. I believe based on my research, by reusing the existing configured logger i could avoid the duped messages

Maybe i got it wrong though? Or (more likely?) maybe #849 obviates the need to do this?

I'll need to find some time to take a closer look.

I don't want to keep blocking the release though, so I'll just comment out the log for now

floppym:
> I believe based on my research, by reusing the existing configured logger i could avoid the duped messages

writemsg does not an "existing configured logger".

writemsg does not actually use the logging module for anything but the level constants (WARNING, DEBUG, etc). It always writes messages to stdout or stderr directly rather than using any outputs configured via the logging module.

I really think this PR is unnecessary since gentoolkit has been updated to avoid adding a second log handling function to the root-level logger. It should not block the Portage release.

floppym:
After thinking about it, it's probably safer to drop the logging statement. I submitted a PR with a proper commit message. https://github.com/gentoo/portage/pull/850

ezzieyguywuf:
Obsoleted by #850

PR #829, opened 2022-05-06
bin/*.sh: use bash extended-test builtin ('[[' and ']]')

Flowdalic:

### DIFF
--- a/bin/ebuild.sh
+++ b/bin/ebuild.sh
@@ -224,7 +224,7 @@ inherit() {
 	local B_IDEPEND
 	local B_PROPERTIES
 	local B_RESTRICT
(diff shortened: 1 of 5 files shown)


floppym:
This seems ok to me.

Flowdalic:
> This seems ok to me.

Thanks for the feedback. I wasn't actually sure if this is a safe transformation, but I assumed so, since bash's extended-test builtin seems to be available since bash 2.0. I am  currently in the process of testing this patch on my laptop and desktop system, and so far, did not run into any trouble.

floppym:
Pushed as 110f999894f2dfef4b5d9aff9cc999c1cf0d1283.

PR #832, opened 2022-05-19
lib/portage/util: fix bundled whirlpool on empty bytestring input

laumann:
The WhirlpoolAdd function did not consider zero-length input, so calls
to update(b'') would produce out-of-bounds errors. This was not covered
by any tests, because the constructor implicitly skipped the call to
update on zero-length input.

Add check for zero-length input to WhirlpoolAdd, and have the Whirlpool
constructor skip calling update() only if arg is None.

Closes: https://bugs.gentoo.org/846389

### DIFF
--- a/lib/portage/util/whirlpool.py
+++ b/lib/portage/util/whirlpool.py
@@ -37,11 +37,9 @@ class Whirlpool:
     may be provided; if present, this string will be automatically
     hashed."""
 
-    def __init__(self, arg=None):
+    def __init__(self, arg=b""):
         self.ctx = WhirlpoolStruct()
-        if arg:
-            self.update(arg)
-        self.digest_status = 0
+        self.update(arg)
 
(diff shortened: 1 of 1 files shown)


laumann:
@mgorny @thesamesam @dwfreed

lib/portage/util/whirlpool.py
```diff
@@ -39,7 +39,7 @@ class Whirlpool:
 
     def __init__(self, arg=None):
```
mgorny, line 40: How about:

```suggestion
    def __init__(self, arg=b""):
```

and then you wouldn't have to check `arg` at all, the other condition will catch it.

laumann: Works for me.

mgorny:
I'm rerunning the failed jobs.

PR #835, opened 2022-06-04
[PREFIX] Prepend EPREFIX to bash (and other programs)

CyberTailor:
Fixes issue where "/bin/bash" does not exist.


thesamesam:
Sorry, I'd missed this was to the prefix branch. FWIW, we do want to get as much as possible in master instead, but that's a bigger project.

CyberTailor:
Decided not to prefix "mv" since it lives in $ROOT/tmp/bin during
stage1, not in $ROOT/bin.

Although bootstrap-prefix.sh can make a symlink..

PR #836, opened 2022-06-04
[PREFIX] bin/*: fix hardcoded shebangs

CyberTailor:
Relace with @PREFIX_PORTAGE_PYTHON@ and @PORTAGE_BASH@ (where
appropriate).

### DIFF
--- a/bin/doins.py
+++ b/bin/doins.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python -b
+#!@PREFIX_PORTAGE_PYTHON@ -b
 # Copyright 2017-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 #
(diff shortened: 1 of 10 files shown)


floppym:
It would be nice to have a solution that would allow usage from a git checkout without installation.

CyberTailor:
> It would be nice to have a solution that would allow usage from a git checkout without installation.

Portage from the `master` branch doesn't use autotools. These changes only affect the `prefix` branch.

grobian:
this is not entirely true, we use the python based install from the ebuild, the autoconf-based one from bootstrap-prefix

As to be able to run from a development snapshot, with Prefix tools this is really difficult if you don't want/can't use /usr/bin/env.  In the past this wasn't possible due to using python -b and so on.  Not sure how that is today.

floppym:
> As to be able to run from a development snapshot, with Prefix tools this is really difficult if you don't want/can't use /usr/bin/env. In the past this wasn't possible due to using python -b and so on. Not sure how that is today.

We already switched to using /usr/bin/env on the master branch See 41f4f6d25019c4f944b8d9576d96482f92231fda and 781aefd8be33c3bd699e5a665225013c5f76b531.

grobian:
I basicall did this

PR #840, opened 2022-06-15
depclean: return failure when requested packages aren't depcleaned

ajakk:
Closes: https://bugs.gentoo.org/851750


lib/_emerge/actions.py
```diff
@@ -1697,6 +1697,11 @@ def cmp_reference_count(node1, node2):
                     cleanlist.append(node.cpv)
 
         return _depclean_result(0, cleanlist, ordered, required_pkgs_total, resolver)
+    if bool(args_set) and "--pretend" not in myopts:
```
floppym, line 1700: The explicit conversion to `bool` is unnecessary.

floppym (approved):

PR #841, opened 2022-06-15
bin/*: make executable

ajakk:
Tests already make these executable, and since they're in bin/ they
should probably be executable anyway.


floppym:
This doesn't seem right. Many of these files are partial scripts that cannot be executed by themselves. I don't think it makes sense to set the execute bit on those.

floppym:
It looks like setup.py passes the contents of bin/ in the `scripts` parameter to the setup() function, which results in them being installed with the execute bit set.

As well, there's some logic to change the file mode in venv_data_files() in setup.py. This is probably what causes tox to set the execute bit whenever it runs.

ajakk:
I'll close this for now as it's an incorrect fix.

PR #842, opened 2022-06-21
git: mark repository as safe for newer gits

thesamesam:
While this doesn't solve the odd permissions issue for communication
b/t gemato & git & portage, it does stop it manifesting.

This fixes compatibility with >=dev-vcs/git-2.35.2.

Bug: https://bugs.gentoo.org/838271

### DIFF
--- a/lib/portage/sync/modules/git/git.py
+++ b/lib/portage/sync/modules/git/git.py
@@ -113,6 +113,9 @@ class GitSync(NewBase):
             return (exitcode, False)
         if not self.verify_head():
             return (1, False)
+
+        self.add_safe_directory(self.repo.location)
+
         return (os.EX_OK, True)
 
     def update(self):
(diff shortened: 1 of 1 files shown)


lib/portage/sync/modules/git/git.py
```diff
@@ -341,3 +346,29 @@ def retrieve_head(self, **kwargs):
...
         return ret
+
+    def add_safe_directory(self, directory):
+        # https://github.com/gentoo/portage/pull/818
+        # Workaround for bug #838271 and bug #838223
+        # We unset first to avoid duplicate entries and also to handle
+        # relocations.
+        cmds = [
+            [self.bin_command, "config", "--unset-all", "safe.directory"],
+            [self.bin_command, "config", "--add", "safe.directory", directory],
```
floppym, line 357: This doesn't work because git ignores the setting in .git/config within a repo. We would need to pass `--global` to add it to root's .gitconfig, or `--system` to add it to /etc/gitconfig.

floppym: I'll work on a patch to check the status of `git config --system --get` and call `git config --system --add`.

thesamesam: I've got no idea why it seemed to work for me. Maybe my permissions were wonky. Thanks.

PR #844, opened 2022-06-26
eend: drop QA warning when ebegin was called from another function

floppym:
check-reqs.eclass does this:

```
_check_reqs_start_phase() {
    ...
    ebegin
    ...
}

_check_reqs_foo() {
    _check_reqs_start_phase
    ...
    eend
}
```

This seems like a valid pattern.

Fixes: 605ad0d675a64eb39144122cf284100192cdfea0

### DIFF
--- a/bin/isolated-functions.sh
+++ b/bin/isolated-functions.sh
@@ -380,9 +380,6 @@ eend() {
 	if [[ -v EBEGIN_EEND ]] ; then
 		local caller="${FUNCNAME[1]}"
 		local tos="${EBEGIN_EEND[-1]}"
-		if [[ "${caller}" != "${tos}" ]] ; then
-			eqawarn "QA Notice: eend (in ${caller}) improperly matched with ebegin (called in ${tos})"
-		fi
 		unset EBEGIN_EEND[-1]
 		if [[ ${#EBEGIN_EEND[@]} -eq 0 ]] ; then
 			unset EBEGIN_EEND
--- a/bin/isolated-functions.sh
+++ b/bin/isolated-functions.sh
@@ -339,16 +339,10 @@ ebegin() {
 	[[ ${RC_ENDCOL} == "yes" ]] && echo >&2
 	LAST_E_LEN=$(( 3 + ${#RC_INDENTATION} + ${#msg} ))
 	LAST_E_CMD="ebegin"
-	if [[ -v EBEGIN_EEND ]] ; then
-		# Already a call to ebegin
-		local prev="${EBEGIN_EEND[-1]}"
-		if [[ "${prev}" == "${FUNCNAME[1]}" ]] ; then
-			eqawarn "QA Notice: ebegin called in ${prev}, but missing call to eend (${FUNCNAME[1]})"
-		fi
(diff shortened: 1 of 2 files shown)


floppym:
cc @laumann

laumann:
On June 26, 2022 8:27:44 PM GMT+02:00, Mike Gilbert ***@***.***> wrote:
>cc @laumann 
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/gentoo/portage/pull/844#issuecomment-1166609320
>You are receiving this because you were mentioned.
>
>Message ID: ***@***.***>
Thanks @floppym, im travelling right now, but am back next week to give a proper response.

At a glance, i think the stack handling can be simplified even more, afaict it basically just needs to pop the stack.

bin/isolated-functions.sh
```diff
@@ -380,9 +380,6 @@ eend() {
 	if [[ -v EBEGIN_EEND ]] ; then
 		local caller="${FUNCNAME[1]}"
 		local tos="${EBEGIN_EEND[-1]}"
```
laumann, line 382: These could also just be dropped then.

You could also consider changing the EBEGIN_EEND stack to a simple counter and increment in ebegin, then decrement here in eend.

laumann: ebegin also uses the stack to see if there's an unclosed call to ebegin from another function. If so, a warning is emitted that it's missing a call to eend.

floppym:
I pushed an update to replace the stack with a counter.

bin/isolated-functions.sh
```diff
@@ -339,15 +339,8 @@ ebegin() {
...
-		# Already a call to ebegin
-		local prev="${EBEGIN_EEND[-1]}"
-		if [[ "${prev}" == "${FUNCNAME[1]}" ]] ; then
-			eqawarn "QA Notice: ebegin called in ${prev}, but missing call to eend (${FUNCNAME[1]})"
-		fi
-		EBEGIN_EEND+=( "${FUNCNAME[1]}" )
-	else
-		EBEGIN_EEND=( "${FUNCNAME[1]}" )
+	if (( __EBEGIN_EEND_COUNT++ > 0 )); then
+		eqawarn "QA Notice: nested ebegin call in ${FUNCNAME[1]}"
```
laumann, line 343: In #824 I recall being told on #gentoo-qa that properly nested ebegin-eend pairs should be allowed (it's definitely used in some places). But you also [commented](https://github.com/gentoo/portage/pull/824#issuecomment-1105638609) that maybe it would be better to ban them and fix eclasses that do this.

This change would definitely start generating QA warnings, but I don't know that there's consensus to ban nested ebegin-eend.

floppym: I opened #846 to simply revert your changes.

floppym: I also updated this PR to only emit the nesting warning when ebegin is called twice in the same function.

floppym:
Merged #846 instead.

bin/isolated-functions.sh
laumann: Alright.

PR #847, opened 2022-07-03
Fix _check_pre_quickpkg_files image prefix

syu-nya:

### DIFF
--- a/lib/portage/gpkg.py
+++ b/lib/portage/gpkg.py
@@ -1847,11 +1847,12 @@ class gpkg:
             image_total_size,
         )
 
-    def _check_pre_quickpkg_files(self, contents, root):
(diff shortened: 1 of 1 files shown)


lib/portage/gpkg.py
```diff
@@ -1847,11 +1847,12 @@ def _check_pre_image_files(self, root_dir, image_prefix="image"):
             image_total_size,
         )
 
-    def _check_pre_quickpkg_files(self, contents, root):
+    def _check_pre_quickpkg_files(self, contents, root, image_prefix="image"):
```
thesamesam, line 1850: When would the prefix not be "image"?

syu-nya: None at the time. I just do that so easier to do extension.

thesamesam: Alrighty.

PR #849, opened 2022-07-09
Avoid calling root-level logging functions

floppym:
Instead, establish a named logger in each module. This will allow for easier filtering of log messages if we ever expand the use of the logging module.

### DIFF
--- a/bin/doins.py
+++ b/bin/doins.py
@@ -26,6 +26,8 @@ import sys
 from portage.util import movefile
 from portage.util.file_copy import copyfile
 
+logger = logging.getLogger("portage.bin.doins")
+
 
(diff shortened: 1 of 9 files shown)


thesamesam (approved):
This might be useful to allow us to refine debugging output, as we currently struggle with that. The options we have when debugging something with a user are "emerge --debug" or bust.

thesamesam:
FWIW, I'd probably wait until after we tag 3.0.31 for this, not that it's really risky though, more that it's big enough, but it's only a very weak preference.

floppym:
> FWIW, I'd probably wait until after we tag 3.0.31 for this, not that it's really risky though, more that it's big enough.

I agree.

PR #851, opened 2022-07-10
env-update: handle BROOT vs ROOT distinction for e.g. PATH

thesamesam:
When Portage builds packages with a non-trivial ROOT setting, say,
ROOT=/tmp/stage1root, the build dependencies (compilers, tools, etc.)
should be taken from the actual build environment BROOT, say BROOT=/ ,
and that must also be reflected in the environment.

An example situation that leads to problems otherwise is the following:
* We build a clang-based stage1 with CC=clang
* The clang binary exists only in a custom directory pulled into PATH
  via env.d.
* ROOT=/tmp/stage1root is initially empty, meaning portage tries to
  build with settings from an empty env.d, and suddently cannot find
  the clang that perfectly exists in BROOT=/ anymore.
* The build fails.

The patch takes a more versatile approach and separates environment
variables into three categories:
* constructed from ROOT
* constructed from BROOT
* combined from ROOT and BROOT (which order??)

Thanks-to: Andreas K. Hüttel <dilfridge@gentoo.org>
Thanks-to: James Le Cuirot <chewi@gentoo.org>

### DIFF
--- a/lib/portage/package/ebuild/config.py
+++ b/lib/portage/package/ebuild/config.py
@@ -485,14 +485,7 @@ class config:
             # interaction with the calling environment that might
             # lead to unexpected results.
 
-            env_d = (
-                getconfig(
-                    os.path.join(eroot, "etc", "profile.env"),
-                    tolerant=tolerant,
-                    expand=False,
-                )
-                or {}
-            )
+            env_d = self._get_env_d(broot=broot, eroot=eroot, tolerant=tolerant)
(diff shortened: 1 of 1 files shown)


thesamesam:
@akhuettel please test the latest revision when you get a chance

thesamesam:
cc @chewi

chewi (approved):

zmedico (approved):

PR #853, opened 2022-07-11
news: fix isRelevant check

thesamesam:
Manifested as all news items being shown, even if
(very) irrelevant to the running system (e.g.
different arch, packages not installed, ...).

I think the distinction here is that with the previous state,
we'd end up with _only_ Trues, or nothing (an element
would be omitted), whereas this commit means we end
up considering a possible mixed sequence.

Fixes: 9e24d0143450628f334cdb62e579efafd1bfd2ba
Reported-by: kurly

### DIFF
--- a/lib/portage/news.py
+++ b/lib/portage/news.py
@@ -281,10 +281,9 @@ class NewsItem:
         kwargs = {"vardb": vardb, "config": config, "profile": profile}
 
         all_match = all(
-            True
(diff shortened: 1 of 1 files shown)


mgorny (approved):
Seems to make sense.

floppym (approved):

thesamesam:
Merged.

PR #857, opened 2022-07-17
Fix incorrect BINPKG_COMPRESS_FLAGS split.

syu-nya:


thesamesam:
Relevant bug is https://bugs.gentoo.org/857933

PR #861, opened 2022-07-19
Scheduling policy switching

slashbeast:
Adds ability to control the scheduler policy that is used for emerge and
all child processes. Mainly to interface the ability to switch to
SCHED_IDLE as the solution to keep interactive tasks unaffected by
building process happening in the background.

On a test sample N=1 with AMD Ryzen 5950x and 64 GB of ram building
sys-devel/gcc with lto enabled significantly reduces responsiveness of
the system, even with CONFIG_SCHED_AUTOGROUP and PREEMPT enabled. Using
a web browser result in visible lags, video playback in web browser,
when using CPU decoding, also suffers greatly.

Switching Portage to SCHED_IDLE (PORTAGE_SCHEDULING_POLICY="idle")
results in no visible slowdowns and responsiveness is as if nothing in
the background was happening.

This is especially worthy feature when running on powerful CPUs, where
users often opt in to build not only with parallel build jobs, but also
with multiple packages at once. Anyone running with PORTAGE_NICENESS="19"
will undoubtedly want to use this feature to force SCHED_IDLE policy.

Signed-off-by: KARBOWSKI Piotr <slashbeast@gentoo.org>

### DIFF
--- a/cnf/make.conf.example
+++ b/cnf/make.conf.example
@@ -291,6 +291,23 @@
 #     unset.
 #PORTAGE_IONICE_COMMAND="ionice -c 3 -p \${PID}"
 #
+# PORTAGE_SCHEDULING_POLICY allows changing the current scheduling policy. The
+# supported options are 'other', 'batch', 'idle', 'fifo' and 'round-robin'. When
+# unset, the scheduling policy remains unchanged, by default Linux uses 'other'
+# policy. Users that wish to minimize the Portage's impact on system
+# responsiveness should set scheduling policy to 'idle' which significantly
+# reduces the disruption to the rest of the system by scheduling Portage as
+# extremely low priority processes.
+#
+#PORTAGE_SCHEDULING_POLICY="idle"
+#
+# PORTAGE_SCHEDULING_PRIORITY allows changing the priority (1-99) of the current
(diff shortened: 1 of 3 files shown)


cnf/make.conf.example
```diff
@@ -291,6 +291,20 @@
 #     unset.
 #PORTAGE_IONICE_COMMAND="ionice -c 3 -p \${PID}"
 #
+# PORTAGE_SCHEDULING_POLICY allows changing the current scheduling policy 
+# (defaults to 'other'). The supported options are 'other', 'batch', 'idle',
```
thesamesam, line 295: Both here and in the man page, could you include a suggestion for what people should use for minimal Portage disruption to the rest of the system?

(What has the effect PORTAGE_NICENESS=19 is supposed to have?)

slashbeast: I will add this information, modify a bit tthe variable names and force-push the branch.

floppym (approved):

thesamesam (approved):
Looks good to me .

thesamesam:
Once new Portage is released, please update https://wiki.gentoo.org/wiki/Portage_niceness, thanks!

PR #862, opened 2022-07-19
Use verbose prefixes for output messages

thesamesam:
Rebased version of #759. Just throwing this up so I don't forget about it and for visibility.

Need to read over the ML feedback from before and see whether we should make it optional (not sure I love this given we already have a lot of debt in that sense), change something else, or keep it as in this PR.

----

Use verbose prefixes for output functions einfo, elog... in order to
make the message types distinguishable on non-color terminals,
and to enable the possibility of grepping them.  The prefixes are
roughly inspired by Xorg logs, and are:

- `[..]` for einfo (and ebegin)
- `[II]` for elog
- `[WW]` for ewarn
- `[QA]` for eqawarn
- `[EE]` for eerror

Bug: https://bugs.gentoo.org/728046
Bug: https://bugs.gentoo.org/778860


mgorny:
We probably don't want to go with the full refactoring and instead do something like `QA>`, `EE>`, etc.

mgorny:
Also making them customizable would be fun.

lib/portage/output.py
```diff
@@ -631,7 +631,7 @@ def __eend(self, caller, errno, msg):
             self._write(
                 out,
                 "%*s%s\n"
-                % ((self.term_columns - self.__last_e_len - 7), "", status_brackets),
+                % ((self.term_columns - self.__last_e_len - 8), "", status_brackets),
```
Arfrever, line 634: Here and ...

lib/portage/output.py
```diff
@@ -645,7 +645,7 @@ def ebegin(self, msg):
         msg += " ..."
         if not self.quiet:
             self.einfon(msg)
-        self.__last_e_len = len(msg) + 3
+        self.__last_e_len = len(msg) + 4
```
Arfrever, line 648: ... here. Why these exact values?
`" * "` has lenght 3, while `"[EE] "` etc. have length 5.

To avoid excessive length, maybe single-letter identifiers would be better (`"[E] "`, `"[W] "`, `"[Q] "` etc.)?

mgorny:
I don't think on/off switch is useful. Better make the prefixes configurable.

PR #865, opened 2022-07-25
ResolverPlayground: unconditionally setup essential_binaries

zmedico:
Closes: https://bugs.gentoo.org/861077

### DIFF
--- a/lib/portage/tests/resolver/ResolverPlayground.py
+++ b/lib/portage/tests/resolver/ResolverPlayground.py
@@ -129,8 +129,11 @@ class ResolverPlayground:
         """
 
         self.debug = debug
-        if eprefix is None:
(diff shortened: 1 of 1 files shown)


lib/portage/tests/resolver/ResolverPlayground.py
```diff
@@ -129,8 +129,11 @@ def __init__(
         """
 
         self.debug = debug
-        if eprefix is None:
-            self.eprefix = normalize_path(tempfile.mkdtemp())
+        if True:
```
floppym, line 132: Is this just to avoid a change in indentation?

zmedico: Yeah, we can go ahead and dedent the whole thing.

PR #866, opened 2022-07-26
bin/ebuild-helpers/doins: fix D check, add EPREFIX check

grobian:
ED = D/EPREFIX, so checking for ED includes EPREFIX, which when this is
absent fails to check for D.  Simply check for D instead, which catches
both the case for D and ED.

Add check for usage of EPREFIX, like for using D with helpers.

### DIFF
--- a/bin/ebuild-helpers/doins
+++ b/bin/ebuild-helpers/doins
@@ -42,12 +42,22 @@ if ! ___eapi_has_prefix_variables; then
 	export ED="${D}"
 fi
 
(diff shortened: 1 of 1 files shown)


floppym:
The change makes sense to me.

However. it might make sense to move the whole thing to `insinto` (phase-functions.sh), since that's where `_E_INSDESTTREE_` gets set.

floppym (approved):

grobian:
pushed, thanks

PR #867, opened 2022-07-26
bin/install-qa-check.d/80libraries: support Darwin/Mach-O objects

grobian:
Check for dylib on Darwin, so on everything else.


thesamesam (approved):
The whole function could do with some better annotation but that's nothing to do with this. Thanks!

mgorny (approved):

grobian:
pushed, thanks

PR #868, opened 2022-07-26
bin/install-qa-check.d/80multilib-strict: use file/find from Prefix

grobian:

### DIFF
--- a/bin/install-qa-check.d/80multilib-strict
+++ b/bin/install-qa-check.d/80multilib-strict
@@ -1,7 +1,7 @@
 # Strict multilib directory checks
 multilib_strict_check() {
 	if has multilib-strict ${FEATURES} && \
-	   [[ -x /usr/bin/file && -x /usr/bin/find ]] && \
+	   type find &>/dev/null && type file &>/dev/null && \
 	   [[ -n ${MULTILIB_STRICT_DIRS} && -n ${MULTILIB_STRICT_DENY} ]]
 	then
 		rm -f "${T}/multilib-strict.log"


thesamesam (commented):
BROOT for these I think

bin/install-qa-check.d/80multilib-strict
```diff
@@ -1,7 +1,7 @@
 # Strict multilib directory checks
 multilib_strict_check() {
 	if has multilib-strict ${FEATURES} && \
-	   [[ -x /usr/bin/file && -x /usr/bin/find ]] && \
+	   [[ -x ${EPREFIX}/usr/bin/file && -x ${EPREFIX}/usr/bin/find ]] && \
```
mgorny, line 4: To be honest, this looks to be some historical inconsistency. I think it'd be better to actually verify `type -P` or alike, to be consistent with calling it via PATH.

grobian: you mean something like:
```
local FIND=$(type -P find)
local FILE=$(type -P file)

if [[ -x ${FIND} ... ]]
...
  <(${FIND} "${ED}/${dir}" ...)
```
?

floppym: I was going to suggest the same, though I wonder if we want to avoid calling `find` or `file` if they are installed outside of `EPREFIX`. Presumably these could be provided by the host OS, and might not work as we expect them to.

floppym: Thinking about this more, I do think we should change it to check PATH via `type`, regardless of the possible compatibility issue on prefix.

The alternative would be to change the calls to `${BROOT}/usr/bin/find` and `${BROOT}/usr/bin/file` if we really want to avoid calling host-provided binaries on prefix.

mgorny (approved):
LGTM. If this turns out to cause any problems in the future, we can always change to using full paths.

thesamesam (approved):

grobian:
thanks, pushed

PR #870, opened 2022-07-26
misc-functions and phase-functions prefix fixes

grobian:


bin/misc-functions.sh
```diff
@@ -83,7 +85,7 @@ install_qa_check() {
 		local EPREFIX= ED=${D}
 	fi
 
-	cd "${ED}" || die "cd failed"
+	cd "${D}" || die "cd failed"
```
floppym, line 88: I wonder if some install QA checks expect to be invoked from `${ED}` rather than `${D}`.

Anyway, this should be safe for the typically (non-prefix) case, so I guess we can roll with it.

floppym (approved):

grobian:
thanks pushed

PR #871, opened 2022-07-28
bin: style tweaks

thesamesam:

### DIFF
--- a/bin/install-qa-check.d/90gcc-warnings
+++ b/bin/install-qa-check.d/90gcc-warnings
@@ -113,25 +113,26 @@ gcc_warn_check() {
 		fi
 
 		local cat_cmd=cat
-		[[ $PORTAGE_LOG_FILE = *.gz ]] && cat_cmd=zcat
-		[[ $reset_debug = 1 ]] && set -x
-		# Use safe cwd, avoiding unsafe import for bug #469338.
-		f=$(cd "${PORTAGE_PYM_PATH}" ; $cat_cmd "${PORTAGE_LOG_FILE}" | \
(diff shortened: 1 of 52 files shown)


floppym (approved):
I wonder if it makes sense to add this to .git-blame-ignore-revs?

thesamesam:
> I wonder if it makes sense to add this to .git-blame-ignore-revs?

Ah, good idea.

bin/install-qa-check.d/90gcc-warnings
```diff
@@ -113,25 +113,26 @@ gcc_warn_check() {
...
 			# In the future this will be a forced "die". In preparation,
 			# increase the log level from "qa" to "eerror" so that people
-			# are aware this is a problem that must be fixed asap.
+			# are aware this is a problem that must be fixed ASAP.
 
-			# just warn on 32bit hosts but bail on 64bit hosts
+			# Just warn on 32bit hosts but bail on 64bit hosts
 			case ${CHOST} in
-				alpha*|hppa64*|ia64*|powerpc64*|mips64*|sparc64*|sparcv9*|x86_64*) gentoo_bug=yes ;;
+				alpha*|aarch64*|arm64*|hppa64*|ia64*|powerpc64*|loongarch64*|mips64*|riscv64*|sparc64*|sparcv9*|s390x*|x86_64*) gentoo_bug=yes ;;
```
Arfrever, line 130: Instead of hardcoding these, maybe better to call `${CC:-${CHOST}-gcc}` and check `__SIZEOF_POINTER__`?
```
if ${CC:-${CHOST}-gcc} -E -P -dM - < /dev/null 2> /dev/null | grep -q "^#define __SIZEOF_POINTER__ 8$" ; then
```

PR #872, opened 2022-07-28
bin/install-qa-check.d/05prefix: optimise, fix and deal with init scr…

grobian:
…ipts

- canonicalise EPREFIX only once (pull out of the loop)
- do not process binary files (grep -I)
- ensure we match files, avoid double / (bug #861305)
- match /etc/init.d scripts

### DIFF
--- a/bin/install-qa-check.d/05prefix
+++ b/bin/install-qa-check.d/05prefix
@@ -34,9 +34,12 @@ install_qa_check_prefix() {
 	# check shebangs, bug #282539
 	rm -f "${T}"/non-prefix-shebangs-errs
 	local WHITELIST=" /usr/bin/env "
+	# shebang can be an absolutised path, bug #342929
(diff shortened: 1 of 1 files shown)


bin/install-qa-check.d/05prefix
```diff
@@ -49,9 +52,8 @@ install_qa_check_prefix() {
 		line=( ${line#"#!"} )
 		IFS=${oldIFS}
 		[[ ${WHITELIST} == *" ${line[0]} "* ]] && continue
-		local fp=${fn#${D}} ; fp=/${fp%/*}
-		# line[0] can be an absolutised path, bug #342929
-		local eprefix=$(canonicalize ${EPREFIX})
+		# bug 861305: fp really needs to start with just a single /
+		local fp=${fn#${D}/} ; fp=/${fp%/*}
```
floppym, line 56: This change may fail if `${D}` contains a trailing slash (before EAPI 6).

PR #874, opened 2022-07-29
portage.5: document restrict-allowed, properties-allowed

thesamesam:
Note that Portage doesn't really do anything with these
right now AFAICT but it may in future (not planned though) decide
to ignore non-acceptable tokens, etc.

Bug: https://bugs.gentoo.org/861659

### DIFF
--- a/man/portage.5
+++ b/man/portage.5
@@ -1,4 +1,4 @@
-.TH "PORTAGE" "5" "Apr 2021" "Portage VERSION" "Portage"
+.TH "PORTAGE" "5" "July 2022" "Portage VERSION" "Portage"
 .SH NAME
 portage \- the heart of Gentoo
 .SH "DESCRIPTION"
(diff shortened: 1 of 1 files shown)


zmedico (approved):

man/portage.5
```diff
@@ -1455,6 +1455,14 @@ listed toward the left of the list.
 .BR repo\-name " = <value of profiles/repo_name>"
 The name of this repository (overrides profiles/repo_name if it exists).
 .TP
+.BR properties\-allowed
+List of PROPERTIES tokens which are allowed for ebuilds within
+the repo.  If unset, all tokens are allowed.
```
mgorny, line 1460:
```suggestion
the repo.  If unset, any token is allowed.
```

I think this way we make it more clear that we allow arbitrary token, not just some well-defined set.

PR #875, opened 2022-07-30
install-qa-check.d/60pkgconfig: improve pkg-config version check

Flowdalic:
Make the pkg-config .pc-file version check display the actual found
version and hint towards the QA_PKGCONFIG_VERSION variable.

### DIFF
--- a/bin/install-qa-check.d/60pkgconfig
+++ b/bin/install-qa-check.d/60pkgconfig
@@ -101,15 +101,19 @@ pkgconfig_check() {
 	# Skip the check if QA_PKGCONFIG_VERSION is set to empty string.
 	if [[ -n ${QA_PKGCONFIG_VERSION} ]]; then
 		local all_bad=yes
-		# Record the last bad file matched
-		local bad_file
+		local -A bad_files
 		for f in "${files[@]}" ; do
-			if [[ ${QA_PKGCONFIG_VERSION} == $(pkg-config --modversion "${f}") ]] ; then
+			local file_version=$(pkg-config --modversion "${f}")
+			if [[ ${QA_PKGCONFIG_VERSION} == ${file_version} ]] ; then
 				all_bad=no
 				break
 			fi
 
-			bad_file="${f//${D}}"
+			# Record a special value if the .pc file has no version set at all.
+			if [[ -z ${file_version} ]] ; then
+				file_version="<no-set>"
+			fi
+			bad_files["${f//${D}}"]="${file_version}"
 		done
 
 		# Skip result reporting if *_p* because for both _pN and _preN, we
@@ -117,8 +121,15 @@ pkgconfig_check() {
 		# we want to avoid false positives.
 		if [[ ${all_bad} == "yes" && ${PV} != *_p* ]] && ! has live ${PROPERTIES} ; then
 			eqawarn "QA Notice: pkg-config files with mismatched Version found!"
-			eqawarn "At least ${bad_file}'s Version field does not match ${PV}"
+			eqawarn "The Version field of the following files does not match ${PV}"
+			local bad_file
+			for bad_file in "${!bad_files[@]}"; do
+				local bad_file_version="${bad_files[${bad_file}]}"
+				eqawarn "- ${bad_file}: ${bad_file_version}"
+			done
 			eqawarn "Please check all .pc files installed by this package."
+			eqawarn "You can use QA_PKGCONFIG_VERSION to set the expected version,"
+			eqawarn "or set to the empty string to disable this QA check."
 		fi
 	fi
 }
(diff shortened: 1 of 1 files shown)


bin/install-qa-check.d/60pkgconfig
```diff
@@ -102,23 +102,35 @@ pkgconfig_check() {
...
 		# Skip result reporting if *_p* because for both _pN and _preN, we
 		# don't generally expect the versions to be exactly accurate, and
 		# we want to avoid false positives.
 		if [[ ${all_bad} == "yes" && ${PV} != *_p* ]] && ! has live ${PROPERTIES} ; then
 			eqawarn "QA Notice: pkg-config files with mismatched Version found!"
-			eqawarn "At least ${bad_file}'s Version field does not match ${PV}"
+			eqawarn "The following file's Version field does not match ${PV}"
+			local bad_file
+			for bad_file in "${!bad_files[@]}"; do
+				local bad_file_version="${bad_files[${bad_file}]}"
```
thesamesam, line 128: See if there's a way to use eqatag? (https://www.gentoo.org/glep/glep-0065.html#eqatag)

(I'm guessing no.)

bin/install-qa-check.d/60pkgconfig
```diff
@@ -102,23 +102,35 @@ pkgconfig_check() {
 	if [[ -n ${QA_PKGCONFIG_VERSION} ]]; then
 		local all_bad=yes
 		# Record the last bad file matched
-		local bad_file
+		local -A bad_files
 		for f in "${files[@]}" ; do
-			if [[ ${QA_PKGCONFIG_VERSION} == $(pkg-config --modversion "${f}") ]] ; then
+			local file_version=$(pkg-config --modversion "${f}")
+			if [[ ${QA_PKGCONFIG_VERSION} == ${file_version} ]] ; then
```
MatthewGentoo, line 108: Could this comparison be aware of version components, so that e.g. it treats `25.0` and `25.0.0` as equal?

Flowdalic: Indeed, it believe using `ver_test ${QA_PKGCONFIG_VERSION} -eq ${file_version}` would achieve this. I am not sure if `ver_test` is available in `install-qa-check.d`, will have a look.

bin/install-qa-check.d/60pkgconfig
Flowdalic: Why are you guessing 'no'. My naive assumption would be that, given that `eqawarn` is available, `eqatag` would also be available here. and we can simply

```
eqatag pkgconfig.unexpected-version ${bad_files[@]}
local bad_file
for bad_file in "${!bad_files[@]}"; do
    …
```

thesamesam: I think I'd assumed the output wouldn't fit and I think I assumed wrong :)

bin/install-qa-check.d/60pkgconfig
```diff
@@ -104,7 +104,7 @@ pkgconfig_check() {
 		local -A bad_files
 		for f in "${files[@]}" ; do
 			local file_version=$(pkg-config --modversion "${f}")
-			if [[ ${QA_PKGCONFIG_VERSION} == ${file_version} ]] ; then
+			if ver_test ${QA_PKGCONFIG_VERSION} -eq ${file_version}; then
```
Flowdalic, line 107: I am not entirely sure if we can use `ver_test` here. A quick test showed that it works, but it may be fragile. That is, it only worked because the ebuild had the right EAPI.

Furthermore, `ver_test` requires all arguments to conform to PMS version strings. This may not be the case for `$file_version`. Not sure if this is then undefined behavior that could cause havoc?

Flowdalic:
I've moved the commit that makes the pkgconfig version check use `ver_test()` to compare the versions into https://github.com/gentoo/portage/pull/882.

thesamesam (approved):

bin/install-qa-check.d/60pkgconfig
```diff
@@ -101,24 +101,36 @@ pkgconfig_check() {
...
-			if [[ ${QA_PKGCONFIG_VERSION} == $(pkg-config --modversion "${f}") ]] ; then
+			local file_version=$(pkg-config --modversion "${f}")
+			if [[ ${QA_PKGCONFIG_VERSION} == ${file_version} ]] ; then
 				all_bad=no
 				break
 			fi
 
-			bad_file="${f//${D}}"
+			# Record a special value if the .pc file has no version set at all.
+			if [[ -z ${file_version} ]]; then
```
thesamesam, line 113: Match ; style

bin/install-qa-check.d/60pkgconfig
```diff
@@ -101,24 +101,36 @@ pkgconfig_check() {
...
+			bad_files["${f//${D}}"]="${file_version}"
 		done
 
 		# Skip result reporting if *_p* because for both _pN and _preN, we
 		# don't generally expect the versions to be exactly accurate, and
 		# we want to avoid false positives.
 		if [[ ${all_bad} == "yes" && ${PV} != *_p* ]] && ! has live ${PROPERTIES} ; then
 			eqawarn "QA Notice: pkg-config files with mismatched Version found!"
-			eqawarn "At least ${bad_file}'s Version field does not match ${PV}"
+			eqawarn "The following file's Version field does not match ${PV}"
```
thesamesam, line 124: file(s), I guess. Happy with it then I reckon.

thesamesam:
Thanks!

PR #876, opened 2022-07-30
lib/_emerge/actions.py: support Clang in getgccversion

grobian:
- call gcc-config with full path for Prefix
- try Clang when GCC is absent

### DIFF
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -2950,6 +2950,18 @@ def getgccversion(chost=None):
     gcc_ver_command = ["gcc", "-dumpversion"]
     gcc_ver_prefix = "gcc-"
 
+    # accept clang as system compiler too
+    clang_ver_command = ["clang", "--version"]
+    clang_ver_prefix = "clang-"
+
+    ubinpath = os.path.join("/", portage.const.EPREFIX, "usr", "bin")
+
+    def getclangversion(output):
+        version = re.search("clang version ([0-9.]+) ", output)
+        if version:
+            return version.group(1)
+        return "unknown"
+
     gcc_not_found_error = red(
         "!!! No gcc found. You probably need to 'source /etc/profile'\n"
         + "!!! to update the environment of this terminal and possibly\n"
@@ -2959,7 +2971,9 @@ def getgccversion(chost=None):
     if chost:
         try:
             proc = subprocess.Popen(
-                ["gcc-config", "-c"], stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+                [ubinpath + "/" + "gcc-config", "-c"],
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
             )
         except OSError:
             myoutput = None
@@ -2972,7 +2986,8 @@ def getgccversion(chost=None):
 
         try:
             proc = subprocess.Popen(
-                [chost + "-" + gcc_ver_command[0]] + gcc_ver_command[1:],
+                [ubinpath + "/" + chost + "-" + gcc_ver_command[0]]
+                + gcc_ver_command[1:],
                 stdout=subprocess.PIPE,
                 stderr=subprocess.STDOUT,
             )
@@ -2984,6 +2999,23 @@ def getgccversion(chost=None):
             mystatus = proc.wait()
         if mystatus == os.EX_OK:
             return gcc_ver_prefix + myoutput
+
+        # no GCC? try Clang
+        try:
+            proc = subprocess.Popen(
+                [ubinpath + "/" + chost + "-" + clang_ver_command[0]]
+                + clang_ver_command[1:],
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+            )
+        except OSError:
+            myoutput = None
+            mystatus = 1
+        else:
+            myoutput = _unicode_decode(proc.communicate()[0]).rstrip("\n")
+            mystatus = proc.wait()
+        if mystatus == os.EX_OK:
+            return clang_ver_prefix + getclangversion(myoutput)
 
     try:
         proc = subprocess.Popen(


lib/_emerge/actions.py
```diff
@@ -2950,6 +2950,18 @@ def getgccversion(chost=None):
     gcc_ver_command = ["gcc", "-dumpversion"]
     gcc_ver_prefix = "gcc-"
 
+    # accept clang as system compiler too
+    clang_ver_command = ["clang", "--version"]
+    clang_ver_prefix = "clang-"
+
+    ubinpath = os.path.join("/", portage.const.EPREFIX, "usr", "bin")
```
thesamesam, line 2957:
```suggestion
    usr_bin_path = os.path.join("/", portage.const.EPREFIX, "usr", "bin")
```

lib/_emerge/actions.py
```diff
@@ -2959,7 +2971,9 @@ def getgccversion(chost=None):
     if chost:
         try:
             proc = subprocess.Popen(
-                ["gcc-config", "-c"], stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+                [ubinpath + "/" + "gcc-config", "-c"],
```
thesamesam, line 2974:
```suggestion
                [usr_bin_path + "/" + "gcc-config", "-c"],
```

lib/_emerge/actions.py
```diff
@@ -2950,6 +2950,18 @@ def getgccversion(chost=None):
     gcc_ver_command = ["gcc", "-dumpversion"]
     gcc_ver_prefix = "gcc-"
 
+    # accept clang as system compiler too
+    clang_ver_command = ["clang", "--version"]
+    clang_ver_prefix = "clang-"
+
+    ubinpath = os.path.join("/", portage.const.EPREFIX, "usr", "bin")
+
+    def getclangversion(output):
```
thesamesam, line 2959:
```suggestion
    def get_clang_version(output):
```

lib/_emerge/actions.py
```diff
@@ -2985,6 +3000,23 @@ def getgccversion(chost=None):
...
+                stderr=subprocess.STDOUT,
+            )
+        except OSError:
+            myoutput = None
+            mystatus = 1
+        else:
+            myoutput = _unicode_decode(proc.communicate()[0]).rstrip("\n")
+            mystatus = proc.wait()
+        if mystatus == os.EX_OK:
+            return clang_ver_prefix + getclangversion(myoutput)
```
thesamesam, line 3018:
```suggestion
            return clang_ver_prefix + get_clang_version(myoutput)
```

lib/_emerge/actions.py
```diff
@@ -2985,6 +3000,23 @@ def getgccversion(chost=None):
         if mystatus == os.EX_OK:
             return gcc_ver_prefix + myoutput
 
+        # no GCC? try Clang
+        try:
+            proc = subprocess.Popen(
+                [ubinpath + "/" + chost + "-" + clang_ver_command[0]]
```
thesamesam, line 3006:
```suggestion
                [usr_bin_path + "/" + chost + "-" + clang_ver_command[0]]
```

lib/_emerge/actions.py
thesamesam: Eh, nevermind, let's not change GCC as well I guess.

Arfrever:
Relying on `PATH` is better than hardcoding `${EPREFIX}/usr/bin`. Some users may have intentional overrides (e.g. using `/usr/local/...`).
`EPREFIX`-ed components of `PATH` should already be present in Prefix environment.
Also `${CHOST}-` prefix apparently does not matter for Clang.

grobian:
In Prefix we often don't want to "see" host-provided stuff, which is in PATH.  Hence, the explicit EPREFIX target.  It also seems questionable whether we want to support stuff coming from random locations, this is about the integrity of the package manager.  In this case, it's actually about some info display, about which compiler is in use.  It's completely "as-is", for it doesn't use the same logic to find the compiler while building, gcc-config can be changed in the meanwhile (even during a series of emerge) and the only thing I tried to achieve here, is to show Clang + version on macOS with Prefix, at the time we still used llvm/clang as system compiler.  Today we don't because it is impossible, but perhaps someday in the future it will be again.  In any case, there was/is a CHOST with Clang in that setup.

Nevertheless, since this change is purely cosmetic, and doesn't do anything useful, I don't think it's worth it if it raises too many questions.  It could be used as inspiration for those that want to fix this informational bit when they have a LLVM system compiler.

Arfrever:
Also see:
* [Gentoo Bug #730032 ("emerge --info prints about gcc and GNU ld while CC and LD point to something else")](https://bugs.gentoo.org/730032)
* [Gentoo Bug #730152 ("emerge --info: Show information about LLVM toolchain (sys-devel/clang, sys-devel/lld, sys-devel/llvm), dev-util/meson, sys-apps/systemd")](https://bugs.gentoo.org/730152)

Fix for Gentoo Bug #730152 was applied on [2021-12-14](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=009ba7c96b78c0b2b98dffcc3dac717d8d733e50), so installed versions of `sys-devel/clang` are shown in `emerge --info`.
If there are no overrides by `CC`/`CXX`, `PATH` or `llvm.eclass:llvm_pkg_setup()`, then highest version of `sys-devel/clang` would be used.

`CC`/`CXX` were shown since [2015-08-23](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=baac630cd83ed1e2bba176a4a5703df48f9787c7).

`/usr/local/bin/gcc`, `/usr/local/bin/clang` could be e.g. wrappers which call system compilers with some enforced options.

Preparation of GCC for crosscompilation is different than preparation of Clang for crosscompilation.

Arfrever in Gentoo Bug #730032 said:
> After (...) change is implemented, gcc version in first line could be dropped.

These farther-reaching suggestions would be still good to implement.

Maybe minimal fix for now would be to drop `gcc_not_found_error` text and make `getgccversion()` return `None` instead of `"[unavailable]"` and make `getportageversion` skip part of output when `gccver` is `None`.

grobian:
I agree that we should just drop the compiler from the first line.  I think other actions (like what Arfrever suggests) should be pursued, but this patch is not the right way forward.  I'll drop it from Prefix too.

PR #877, opened 2022-07-30
lib/_emerge/depgraph.py: check EPREFIX compatibility for binpkgs

grobian:
one cannot just install any binpkg on a Prefix system, it needs to
contain a compatible EPREFIX, that is, the same (preferable), or one
that can be adjusted to using chpathtool.

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -11871,6 +11871,23 @@ def _get_masking_status(pkg, pkgsettings, root_config, myrepo=None, use=None):
         if not pkgsettings._accept_chost(pkg.cpv, pkg._metadata):
             mreasons.append(_MaskReason("CHOST", "CHOST: %s" % pkg._metadata["CHOST"]))
 
+    # check EPREFIX is compatible to be adjusted (chpathtool)
+    eprefix = pkgsettings["EPREFIX"]
+    if (
+        eprefix is not None
+        and len(eprefix.rstrip("/")) > 0
+        and pkg.built
+        and not pkg.installed
+    ):
+        if not "EPREFIX" in pkg._metadata:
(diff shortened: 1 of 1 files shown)


floppym:
This triggers a test failure.
```
======================================================================
FAIL: testSonameAutoUnmask (portage.tests.resolver.soname.test_autounmask.SonameAutoUnmaskTestCase)
----------------------------------------------------------------------
  File "/home/runner/work/portage/portage/build/lib/portage/tests/__init__.py", line 273, in run
    testMethod()
  File "/home/runner/work/portage/portage/build/lib/portage/tests/resolver/soname/test_autounmask.py", line 110, in testSonameAutoUnmask
    self.assertEqual(
AssertionError: False != True : atoms: (dev-libs/icu), key: success, expected: False, got: True
atoms: (dev-libs/icu), key: mergelist, expected: ['[binary]dev-libs/icu-49', '[binary]dev-libs/libxml2-2.7.8'], got: ['[binary]dev-libs/icu-4.8']
atoms: (dev-libs/icu), key: unstable_keywords, expected: {'dev-libs/libxml2-2.7.8'}, got: None
```
I would guess this is because we do not specify EPREFIX in the fake binpkgs specified in test_autounmask.py. This causes the fake binpkgs to always be masked by this new code.

grobian:
> I would guess this is because we do not specify EPREFIX in the fake binpkgs in test_autounmask.py. This causes the fake binpkgs to always be masked by this new code.

This is what I was trying to figure out, the check shouldn't run when EPREFIX isn't set...

Arfrever:
Does it work if you add `"EPREFIX": portage.const.EPREFIX` line in each member of `binpkgs` in `lib/portage/tests/resolver/soname/test_autounmask.py`?

PR #879, opened 2022-08-01
estrip: avoid spurious NEEDED warning when no ELF installed; apply optimisation to EAPI 7+ too

thesamesam:
```
commit 26c292b0794c72bf4a52499f6bf3366f61181780
Author: Sam James <sam@gentoo.org>
Date:   Mon Aug 1 01:56:51 2022 +0100

    estrip: apply scanelf optimisation to EAPI 7+ / dostrip

    See: bb88e766897f5e7e0b0a10c48cf99a04edb73a40
    Bug: https://bugs.gentoo.org/749624
    Bug: https://bugs.gentoo.org/862606
```

```
commit 2980e0325e826f5cff7d12fb1fc9fea3c543f08b
Author: Sam James <sam@gentoo.org>
Date:   Mon Aug 1 01:05:47 2022 +0100

    estrip: avoid spurious NEEDED warning when no ELF installed

    If no ELF files are installed, then we're obviously
    not going to have a NEEDED file.

    The reason this didn't get hit sooner is because
    apparently very few of us are actually testing
    EAPI 6 actively - which figures.

    We only take this path for < EAPI 7 (non-dostrip)
    which means we're actually missing the estrip
    optimisation in bug #749624 for newer EAPIs!

    (Will be fixed in a followup).

    Bug: https://bugs.gentoo.org/749624
    Bug: https://bugs.gentoo.org/862606
    Fixes: bb88e766897f5e7e0b0a10c48cf99a04edb73a40
```

### DIFF
--- a/bin/estrip
+++ b/bin/estrip
@@ -419,9 +419,10 @@ while read -r x ; do
 done < <(
 	# NEEDED may not exist for some packages (bug #862606)
 	if [[ -f "${PORTAGE_BUILDDIR}"/build-info/NEEDED ]] ; then
-		needed_contents=$(cat "${PORTAGE_BUILDDIR}"/build-info/NEEDED | cut -d ' ' -f1 | sed -e "s:^:${D%/}:")
-	else
-		needed_contents=""
+		while IFS= read -r needed_entry ; do
+			needed_entry="${needed_entry% *}"
+			needed_contents+=( "${D%/}${needed_entry}" )
+		done < "${PORTAGE_BUILDDIR}"/build-info/NEEDED
 	fi
 
 	# Use sort -u to eliminate duplicates (bug #445336).
@@ -429,8 +430,6 @@ done < <(
 		[[ -n ${needed_contents[@]} ]] && printf "%s\n" "${needed_contents[@]}"
 		find "$@" -type f ! -type l -name '*.a'
 	) | LC_ALL=C sort -u
-
-	unset needed_contents
 )
 else
 while IFS= read -d '' -r x ; do
--- a/bin/estrip
+++ b/bin/estrip
@@ -64,12 +64,51 @@ while [[ $# -gt 0 ]] ; do
 		done
 
 		if [[ ${find_paths[@]} ]]; then
+			# We can avoid scanelf calls for binaries we already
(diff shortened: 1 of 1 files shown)


thesamesam:
Sorry, not ready yet, just noticed an issue again..

bin/estrip
```diff
@@ -64,12 +64,51 @@ while [[ $# -gt 0 ]] ; do
...
+						#
+						# find_path entries may be exact paths (like /usr/lib64/libfoo.so)
+						# or instead /usr/lib64, or ${ED}/usr, etc.
+						#
+						# We check if the beginning (i.e. first entry) of the NEEDED line
+						# matches the path given
+						# e.g. find_path="/usr/lib64" will match needed_entry="/usr/lib64/libfoo.so libc.so".
+						needed_entry_file="${needed_entry% *}"
+						if [[ "${needed_entry_file}" =~ ^${find_path##${D}} ]] ; then
+							scanelf_results+=( $(echo "${D}${needed_entry_file}") )
```
floppym, line 96: I don't understand the point of calling `echo` in a command substitution here. I think this would be equivalent.
```suggestion
							scanelf_results+=( "${D}${needed_entry_file}" )
```

bin/estrip
```diff
@@ -408,12 +447,21 @@ while read -r x ; do
...
+
+	# Use sort -u to eliminate duplicates (bug #445336).
 	(
+		[[ -n ${needed_contents[@]} ]] && printf "%s\n" "${needed_contents[@]}"
 		find "$@" -type f ! -type l -name '*.a'
-		cut -d ' ' -f1 < "${PORTAGE_BUILDDIR}"/build-info/NEEDED \
-			| sed -e "s:^:${D%/}:"
 	) | LC_ALL=C sort -u
+
+	unset needed_entry needed_contents
```
floppym, line 464: This `unset` is a bit pointless at the end of a subshell, but I guess it doesn't hurt.

floppym (changes requested):

bin/estrip
thesamesam: Sorry, that's left over from earlier. Good spot

floppym (approved):

PR #881, opened 2022-08-04
dispatch-conf: respect (E)ROOT

alfredfo:
Signed-off-by: Alfred Persson Forsberg <cat@catcream.org>

### DIFF
--- a/bin/dispatch-conf
+++ b/bin/dispatch-conf
@@ -115,8 +115,8 @@ class dispatch:
         count = 0
 
         perform_conf_update_session_hooks("pre-session")
-
-        config_root = portage.settings["EPREFIX"] or os.sep
+        
+        config_root = portage.settings["EROOT"]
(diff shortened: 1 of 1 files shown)


thesamesam (approved):

bin/dispatch-conf
```diff
@@ -115,8 +115,13 @@ class dispatch:
...
         perform_conf_update_session_hooks("pre-session")
-
-        config_root = portage.settings["EPREFIX"] or os.sep
+        
+        config_root = (
+            portage.settings["EROOT"]
+            or portage.settings["ROOT"]
+            or portage.settings["EPREFIX"]
+            or os.sep
+        )
```
zmedico, line 124: Since portage.settings["EROOT"] is always non-empty, we may as well just do this:
```
config_root = portage.settings["EROOT"] 
```

alfredfo: updated!

zmedico (approved):

PR #883, opened 2022-08-06
bin: ebuild.sh: drop QA interceptors

thesamesam:
Long-obsolete since fb2459330cf226ee34d3875a1143531bd109aaf2
and 40da7ee19c4c195da35083bf2d2fcbd852ad3846.

Indeed, pkgcore dropped QA interceptors for the same
reason after its analogue of those commits.

Arfrever notably pointed out that for the binaries
listed, it'd actually be a downgrade in severity
from fatal -> warning, as it overrides the
command not found abort.

### DIFF
--- a/bin/ebuild.sh
+++ b/bin/ebuild.sh
@@ -480,9 +480,6 @@ __try_source() {
 export SANDBOX_ON="1"
 export S=${WORKDIR}/${P}
 
-# Turn off extended glob matching so that g++ doesn't get incorrectly matched.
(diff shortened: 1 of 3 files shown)


thesamesam:
@Arfrever points out in #gentoo-portage that we may want to implement the autotools check (for non-global scope) in pkgcheck or similar.

I wonder if we should go further and have a @REPLACES eclassdoc tag for e.g. pytest -> epytest as well.

PR #885, opened 2022-08-08
EbuildPhase: print SLOT along other build metadata

laumann:
Requested as a useful part in the output from tinderbox runs.

Closes: https://bugs.gentoo.org/864382

### DIFF
--- a/lib/_emerge/EbuildPhase.py
+++ b/lib/_emerge/EbuildPhase.py
@@ -135,7 +135,9 @@ class EbuildPhase(CompositeTask):
                     maint_str = "<invalid metadata.xml>"
 
             msg = []
-            msg.append("Package:    %s" % self.settings.mycpv)
+            msg.append(
+                "Package:    %s:%s" % (self.settings.mycpv, self.settings["SLOT"])
+            )
             if self.settings.get("PORTAGE_REPO_NAME"):
                 msg.append("Repository: %s" % self.settings["PORTAGE_REPO_NAME"])
             if maint_str:


laumann:
cc @asarubbo @mgorny @thesamesam

laumann:
Quick test build with `dev-libs/libdaemon`:

```
 * Package:    dev-libs/libdaemon-0.14-r3
 * Slot:       0/5
 * Repository: gentoo
 * USE:        abi_x86_64 amd64 elibc_glibc kernel_linux userland_GNU
 * FEATURES:   network-sandbox preserve-libs sandbox userpriv usersandbox
>>> Unpacking source...
>>> Unpacking libdaemon-0.14.tar.gz to /var/tmp/portage/dev-libs/libdaemon-0.14-r3/work
>>> Source unpacked in /var/tmp/portage/dev-libs/libdaemon-0.14-r3/work
```

lib/_emerge/EbuildPhase.py
```diff
@@ -136,6 +136,7 @@ async def _async_start(self):
 
             msg = []
             msg.append("Package:    %s" % self.settings.mycpv)
+            msg.append("Slot:       %s" % self.settings["SLOT"])
```
mgorny, line 139: Perhaps we could print it as part of 'package' instead?

mgorny: I'd personally prefer atom-style `%s:%s`.

laumann: done

mgorny (approved):
Thanks!

PR #888, opened 2022-08-15
pyupgrade everything

ajakk:

Anthony Sottile wrote a tool that replaces python2-idiomatic code with python3. This is the result of me running it over the whole of the portage codebase. Most of it is trivial replacements of percent-formatted strings with str().format(), and replacing set and dict instantiations with their literal alternative (ie `set([1,2,3]) -> {1,2,3}`).


Flowdalic:
Is the tool available somewhere? I am asking because I hoped that it may supports also transformation to f-strings. Or is there a particular reason this did not directly transform to f-strings (which I find more appealing and would prefer to `format()`)?

thesamesam:
> Is the tool available somewhere? I am asking because I hoped that it may supports also transformation to f-strings. Or is there a particular reason this did not directly transform to f-strings (which I find more appealing and would prefer to `format()`)?

It's on pypi, but we can't use f strings as our base version is Python 3.7 for now. We can discuss changing it though.

floppym:
> It's on pypi, but we can't use f strings as our base version is Python 3.7 for now.

f-strings were introduced in Python 3.6.

thesamesam:
> > It's on pypi, but we can't use f strings as our base version is Python 3.7 for now.
> 
> f-strings were introduced in Python 3.6.

Duh. Sorry, I was thinking of f"{...=}".

thesamesam:
dev-python/pyupgrade is now in tree by the way.

lib/_emerge/actions.py
```diff
@@ -1648,7 +1651,8 @@ def create_cleanlist():
 
                 if debug:
                     writemsg_level(
-                        "Candidates: [%s]\n" % ", ".join("'%s'" % (x,) for x in atoms),
+                        "Candidates: [%s]\n"
+                        % ", ".join("'{}'".format(x) for x in atoms),
```
zmedico, line 1655: Looks like it missed a `.format` opportunity here.

lib/_emerge/depgraph.py
```diff
@@ -2040,7 +2043,9 @@ def _slot_confict_backtrack(self, root, slot_atom, all_parents, conflict_pkgs):
                 "  package(s) to mask: %s" % str(to_be_masked),
                 "      slot: %s" % slot_atom,
                 "   parents: %s"
-                % ", ".join("(%s, '%s')" % (ppkg, atom) for ppkg, atom in all_parents),
+                % ", ".join(
```
zmedico, line 2046: Here's another missed `.format` opportunity.

lib/portage/dbapi/bintree.py
```diff
@@ -575,7 +570,7 @@ def __init__(
             "SLOT": "0",
             "USE": "",
         }
-        self._pkgindex_inherited_keys = ["CHOST", "repository"]
+        self._pkgindex_inherited_keys = ["BINPKG_FORMAT", "CHOST", "repository"]
```
zmedico, line 573: This looks like an unintentional revert.

lib/portage/update.py
```diff
@@ -429,11 +425,13 @@ def match_callback(repo_name, atoma, atomb):
                             # add a comment with the update command, so
                             # the user can clearly see what happened
                             contents[pos] = "# %s\n" % " ".join(
```
zmedico, line 427: Here's another missed `.format` opportunity.

lib/_emerge/actions.py
ajakk: Hm, yeah, seems like there's some moderately odd constructs that it doesn't handle:

```
 ~ $ cat pyupgrade-test.py
msg = [
    "%s" % "foo"
]

msg = ["{} {}".format("foo", "%s" % "bar")]
 ~ $ pyupgrade pyupgrade-test.py
 ~ $ cat pyupgrade-test.py
msg = [
    "%s" % "foo"
]

msg = ["{} {}".format("foo", "%s" % "bar")]
```

lib/portage/dbapi/bintree.py
ajakk: Indeed, thanks! Fixed.

lib/portage/update.py
ajakk: These 3 should be fixed in a new commit

zmedico (approved):

lib/_emerge/AbstractPollTask.py
```diff
@@ -93,7 +93,7 @@ def _read_buf(self, fd):
 
     def _async_wait(self):
         self._unregister()
-        super(AbstractPollTask, self)._async_wait()
+        super()._async_wait()
```
thesamesam, line 96: How did super() work before? (What changed?)

ajakk: I am not sure I understand the details, but this pattern is what's recommended in documentation (which says it's the same as the old pattern):

https://docs.python.org/3.11/library/functions.html#super

thesamesam: Oh okay, that's all I needed - thanks!

thesamesam:
I think this is good to go but I figure we'll wait until after https://github.com/gentoo/portage/pull/955 is in given the fix for that may require a quick release.

PR #893, opened 2022-08-19
Add gpkg-sign tool

syu-nya:

### DIFF
--- a/bin/misc-functions.sh
+++ b/bin/misc-functions.sh
@@ -512,6 +512,10 @@ __dyn_package() {
 		die "PORTAGE_BINPKG_TMPFILE is unset"
 	mkdir -p "${PORTAGE_BINPKG_TMPFILE%/*}" || die "mkdir failed"
 
+	if [[ ! -z "${BUILD_ID}" ]]; then
+		echo -n "${BUILD_ID}" > "${PORTAGE_BUILDDIR}"/build-info/BUILD_ID
+	fi
+
 	if [[ "${BINPKG_FORMAT}" == "xpak" ]]; then
 		local tar_options=""
 
@@ -547,7 +551,7 @@ __dyn_package() {
 	elif [[ "${BINPKG_FORMAT}" == "gpkg" ]]; then
 		PYTHONPATH=${PORTAGE_PYTHONPATH:-${PORTAGE_PYM_PATH}} \
 			"${PORTAGE_PYTHON:-/usr/bin/python}" "${PORTAGE_BIN_PATH}"/gpkg-helper.py compress \
-			"${CATEGORY}/${PF}" "${PORTAGE_BINPKG_TMPFILE}" "${PORTAGE_BUILDDIR}/build-info" "${D}"
+			"${PF}${BUILD_ID:+-${BUILD_ID}}" "${PORTAGE_BINPKG_TMPFILE}" "${PORTAGE_BUILDDIR}/build-info" "${D}"
 		if [[ $? -ne 0 ]]; then
 			rm -f "${PORTAGE_BINPKG_TMPFILE}"
 			die "Failed to create binpkg file"
--- a/lib/portage/dbapi/bintree.py
+++ b/lib/portage/dbapi/bintree.py
@@ -1039,6 +1039,12 @@ class binarytree:
                     if match:
                         binpkg_format = match.get("BINPKG_FORMAT", None)
 
+                    if not binpkg_format:
+                        if myfile.endswith(SUPPORTED_XPAK_EXTENSIONS):
+                            binpkg_format = "xpak"
+                        elif myfile.endswith(SUPPORTED_GPKG_EXTENSIONS):
+                            binpkg_format = "gpkg"
+
                     if gpkg_only:
                         if binpkg_format != "gpkg":
                             if not gpkg_only_warned:
--- a/lib/portage/dbapi/bintree.py
+++ b/lib/portage/dbapi/bintree.py
@@ -47,6 +47,7 @@ from portage.exception import (
 from portage.localization import _
 from portage.output import colorize
 from portage.package.ebuild.profile_iuse import iter_iuse_vars
+from portage.util import ensure_dirs
 from portage.util.file_copy import copyfile
 from portage.util.futures import asyncio
 from portage.util.futures.executor.fork import ForkExecutor
@@ -733,7 +734,8 @@ class binarytree:
             # assuming that it will be deleted by eclean-pkg when its
             # time comes.
             mynewcpv = _pkg_str(mynewcpv, metadata=metadata, db=self.dbapi)
-            update_path = self.getname(mynewcpv, allocate_new=True) + ".partial"
+            allocated_pkg_path = self.getname(mynewcpv, allocate_new=True)
+            update_path = allocated_pkg_path + ".partial"
             self._ensure_dir(os.path.dirname(update_path))
             update_path_lock = None
             try:
@@ -747,7 +749,11 @@ class binarytree:
                     mybinpkg.update_metadata(mydata, new_basename=mynewcpv)
                 else:
                     raise InvalidBinaryPackageFormat(binpkg_format)
-                self.inject(mynewcpv, filename=update_path)
+                self.inject(
+                    mynewcpv,
+                    current_pkg_path=update_path,
+                    allocated_pkg_path=allocated_pkg_path,
+                )
             finally:
                 if update_path_lock is not None:
                     try:
@@ -1590,13 +1596,17 @@ class binarytree:
                 self._additional_pkgs[instance_key] = pkg
                 self.dbapi.cpv_inject(pkg)
 
-    def inject(self, cpv, filename=None):
+    def inject(self, cpv, current_pkg_path=None, allocated_pkg_path=None):
         """Add a freshly built package to the database.  This updates
         $PKGDIR/Packages with the new package metadata (including MD5).
         @param cpv: The cpv of the new package to inject
         @type cpv: string
-        @param filename: File path of the package to inject, or None if it's
-                already in the location returned by getname()
+        @param current_pkg_path: File path of the package to inject,
+               or None if it's already in the location returned by getname()
+        @type filename: string
+        @rtype: _pkg_str or None
+        @param allocated_pkg_path: File path of the package that was newly
+               allocated or None if it's not allocated.
         @type filename: string
         @rtype: _pkg_str or None
         @return: A _pkg_str instance on success, or None on failure.
@@ -1604,10 +1614,10 @@ class binarytree:
         mycat, mypkg = catsplit(cpv)
         if not self.populated:
             self.populate()
-        if filename is None:
+        if current_pkg_path is None:
             full_path = self.getname(cpv)
         else:
-            full_path = filename
+            full_path = current_pkg_path
         try:
             s = os.stat(full_path)
         except OSError as e:
@@ -1615,7 +1625,7 @@ class binarytree:
                 raise
             del e
             writemsg(
-                _("!!! Binary package does not exist: '%s'\n") % full_path,
+                f"!!! Binary package does not exist: '{full_path}'\n",
                 noiselevel=-1,
             )
             return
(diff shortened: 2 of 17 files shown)


mgorny:
Can't look at it right now but big thanks! This is exactly a tool I was looking for, in order to sign kernel binpkgs built on containers without having to figure out some reasonably safe agent setup there!

lib/portage/gpkg.py
```diff
@@ -764,10 +764,10 @@ class gpkg:
     https://www.gentoo.org/glep/glep-0078.html
     """
 
-    def __init__(self, settings, base_name=None, gpkg_file=None):
+    def __init__(self, settings, basename=None, gpkg_file=None):
         """
         gpkg class handle all gpkg operations for one package.
-        base_name is the package basename.
+        basename is the package basename.
```
mgorny, line 770: It's a bit weird to call it "basename" when it contains category name as well.

lib/portage/gpkg.py
```diff
@@ -778,10 +778,16 @@ def __init__(self, settings, base_name=None, gpkg_file=None):
             self.gpkg_file = _unicode_decode(
                 gpkg_file, encoding=_encodings["fs"], errors="strict"
             )
-        self.base_name = base_name
+
+        if basename:
```
mgorny, line 782: We really prefer explicit `is None` if you don't expect any other non-true values.

lib/portage/gpkg.py
```diff
@@ -1081,35 +1097,38 @@ def update_metadata(self, metadata, newcpv=None):
...
+                    old_data_tarinfo = container_old.getmember(
+                        os.path.join(self.prefix, file_name_old)
+                    )
                     new_data_tarinfo = copy(old_data_tarinfo)
-                    if newcpv:
-                        m[1] = m[1].replace(oldcpv, newcpv, 1)
-                        new_data_tarinfo.name = new_data_tarinfo.name.replace(
-                            oldcpv, newcpv, 1
-                        )
+                    m[1] = m[1].replace(old_basename, new_basename, 1)
```
mgorny, line 1128: Is this still relevant now that Manifest doesn't contain full paths?

lib/portage/gpkg.py
```diff
@@ -1451,7 +1494,7 @@ def _load_manifest(self, manifest_string):
         manifest_filenames = []
 
         for manifest_record in manifest_string.splitlines():
-            if manifest_record == "":
+            if manifest_record == "" or manifest_record.startswith("[GNUPG:]"):
```
mgorny, line 1497: Why would it contain `[GNUPG:]`?

lib/portage/gpkg.py
```diff
@@ -1523,10 +1566,30 @@ def _verify_binpkg(self, metadata_only=False):
                     f"Cannot read tar file: {self.gpkg_file}"
                 )
 
-            # Check gpkg header
-            if self.gpkg_version not in container_files:
+            # Check if gpkg version file exists in any place
+            if self.gpkg_version not in [os.path.basename(f) for f in container_files]:
```
mgorny, line 1570: Use a generator rather than list comprehension here, as there should be no need to inline the result into a list first (i.e. change `[...]` to `(...)`).

mgorny (commented):
Thanks for doing this.

lib/portage/gpkg.py
syu-nya: It suppose to be basename, but also support pass cpv to it, or the caller need to split it themselves.

lib/portage/gpkg.py
syu-nya: I thought the basename could be empty "" if it does not know what the package is, but the caller could set it to None. So going to use `is None`.

lib/portage/gpkg.py
syu-nya: No, it is no longer needed.

lib/portage/gpkg.py
syu-nya: I don't know if this is a bug or expected behavior, but if GPG verification passes but the signature is untrusted or unknown, gpg will also send `[GNUPG:] XXX` to stdout instead of stderr.

mgorny: Weird. I've never seen it do that, and I see that you're using `--status-fd`, so it shouldn't happen really.

syu-nya: Decrypt untrusted Manifest I got:
stderr:
```
 gpg: Signature made Thu Sep  1 10:06:15 2022 EDT
gpg:                using RSA key 8DEDA2CDED49C8809287B89D8812797DDF1DD192
gpg: Good signature from "Gentoo Portage Test Untrusted Key (Test Only, Do NOT Trust!!!) (Gentoo Test Key) <test@example.org>" [unknown]
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: 8DED A2CD ED49 C880 9287  B89D 8812 797D DF1D D192
```

stdout:
```
[GNUPG:] PLAINTEXT 74 0 
DATA gpkg-1 0 SHA512 cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e BLAKE2B 786a02f742015903c6c6fd852552d272912f4740e15847618a86e217f71f5419d25e1031afee585313896444934eb04b903a685b1448b755d56f701afe9be2ce
DATA metadata.tar.zst 22 SHA512 489abee14a5c50af166e8f639fb9caf89cb871560c8d156bf782eede2bc8d2ab08515121015e67b061d7fb754892d6879cf71a622cc1c4e4eb3c4b0d97d841a6 BLAKE2B 7a3dfa9a6496afc90543e556d949bd77a7532716192b23711988104959e4d9f115b17f6ad8c8d5b817aac0e8d9b08b0024c62da3c08bd0b581aacacd40227686
DATA metadata.tar.zst.sig 833 SHA512 26d51d9fbca64e23394ac8940dbcb131581da63c2fe96d45d63df7fb81691cd9b2a592a28710fd14b6efcc450b93356e94b3384bb5527a3155cb562cb61ef2b6 BLAKE2B 2d2f3876abcb59b432367d4aa55012c7e028c8add40f9b3dffb8c2b87e11bc9da558e7e3b38da5eb68bc15841cc84d0ac9f76f72cfa1390546c83c751eca343e
DATA image.tar.zst 1049649 SHA512 97d6b3b4f38422c5ff048347e516199dab2238d50e2da83d31c46cd7df822870261e9d2c7f2485e0b9f88e3f4209c167670409f54adee20f081a7074f8ac105f BLAKE2B 6519051adbfe4d0077242a2313f5d70a975d20d95ceb306c2ebb496124471b5abe7a69f89f61411b878b9862d7a6fc6db15182e1db2f80575f322964671986ce
DATA image.tar.zst.sig 833 SHA512 543a9430b42a309f43767877f297ca74144442edaea4695e7d1c8f2822108a1c1665340c43a4b8b002b3498bf06dd8aa928c27c780fe3d73010d97eaeecfbf87 BLAKE2B 618474560052d74935e66c7db97efcb79aa9788c71e3aba9449299a81fd52ecbee8ab3863fb775a7e2c9464830a34c91c591367fa0be5b69c284b2f019601594
[GNUPG:] NEWSIG
[GNUPG:] KEY_CONSIDERED 8DEDA2CDED49C8809287B89D8812797DDF1DD192 0
[GNUPG:] SIG_ID uEoFdrEAEF2ZHJMgYTRU0qmBUKs 2022-09-01 1662041175
[GNUPG:] KEY_CONSIDERED 8DEDA2CDED49C8809287B89D8812797DDF1DD192 0
[GNUPG:] GOODSIG 8812797DDF1DD192 Gentoo Portage Test Untrusted Key (Test Only, Do NOT Trust!!!) (Gentoo Test Key) <test@example.org>
[GNUPG:] VALIDSIG 8DEDA2CDED49C8809287B89D8812797DDF1DD192 2022-09-01 1662041175 0 4 0 1 10 01 8DEDA2CDED49C8809287B89D8812797DDF1DD192
[GNUPG:] KEY_CONSIDERED 8DEDA2CDED49C8809287B89D8812797DDF1DD192 0
[GNUPG:] TRUST_UNDEFINED 0 pgp
```

This seems only happened when using cleartext signatures.

syu-nya: Ok, I found what happened, `status-fd` wrong number in test case.  😂

bin/misc-functions.sh
```diff
@@ -545,9 +549,15 @@ __dyn_package() {
...
-			"${CATEGORY}/${PF}" "${PORTAGE_BINPKG_TMPFILE}" "${PORTAGE_BUILDDIR}/build-info" "${D}"
+		if [[ -z "${BUILD_ID}" ]]; then
+			PYTHONPATH=${PORTAGE_PYTHONPATH:-${PORTAGE_PYM_PATH}} \
+				"${PORTAGE_PYTHON:-/usr/bin/python}" "${PORTAGE_BIN_PATH}"/gpkg-helper.py compress \
+				"${PF}" "${PORTAGE_BINPKG_TMPFILE}" "${PORTAGE_BUILDDIR}/build-info" "${D}"
+		else
+			PYTHONPATH=${PORTAGE_PYTHONPATH:-${PORTAGE_PYM_PATH}} \
+				"${PORTAGE_PYTHON:-/usr/bin/python}" "${PORTAGE_BIN_PATH}"/gpkg-helper.py compress \
+				"${PF}-${BUILD_ID}" "${PORTAGE_BINPKG_TMPFILE}" "${PORTAGE_BUILDDIR}/build-info" "${D}"
+		fi
```
mgorny, line 560:
```suggestion
		PYTHONPATH=${PORTAGE_PYTHONPATH:-${PORTAGE_PYM_PATH}} \
			"${PORTAGE_PYTHON:-/usr/bin/python}" "${PORTAGE_BIN_PATH}"/gpkg-helper.py compress \
			"${PF}${BUILD_ID:+-${BUILD_ID}}" "${PORTAGE_BINPKG_TMPFILE}" "${PORTAGE_BUILDDIR}/build-info" "${D}"
```

(wrapping if necessary)

lib/portage/dbapi/bintree.py
```diff
@@ -2199,6 +2173,41 @@ def _allocate_filename_multi(self, cpv):
...
+
+        if binpkg_format == "xpak":
+            filename_format = "%s-%s.xpak"
+        elif binpkg_format == "gpkg":
+            filename_format = "%s-%s.gpkg.tar"
+        else:
+            raise InvalidBinaryPackageFormat(binpkg_format)
+
+        while True:
+            filename = filename_format % (
```
mgorny, line 2202: These days we really prefer `.format()` for new code.

lib/portage/dbapi/bintree.py
```diff
@@ -2199,6 +2173,41 @@ def _allocate_filename_multi(self, cpv):
...
+                os.path.join(self.pkgdir, cpv.cp, pf),
+                build_id,
+            )
+            if os.path.exists(filename):
```
mgorny, line 2206: Can you perhaps create the file here (e.g. as an empty file) to actually avoid races? Using `x` as mode should do the trick.

mgorny (commented):
Oh, and please leave a comment if you force push. GitHub doesn't send mail when that happens, so we don't notice that you've made changes.

bin/misc-functions.sh
syu-nya: Ok, didn't thought this.

lib/portage/dbapi/bintree.py
syu-nya: Was reused old code. Do you prefer new `f"{var}"` or `.format()`.

lib/portage/dbapi/bintree.py
syu-nya: Sure.

syu-nya: This will leave an empty file if binpkg creation failed, and increase the build id. Is this ok?

syu-nya: This one actually created a lot of invalid binary package warning. I guess I need silence it if the file is empty.

lib/portage/dbapi/bintree.py
mgorny: `f`, I suppose. Make the `if` above set `suffix` rather than the full format string, and use that.

lib/portage/dbapi/bintree.py
mgorny: Does binpkg creation fail that lot, or are you talking about tests?

lib/portage/dbapi/bintree.py
```diff
@@ -2183,22 +2166,73 @@ def _allocate_filename_multi(self, cpv):
...
+                    # Avoid races
+                    ensure_dirs(os.path.dirname(filename))
+                    with open(filename, "x") as f:
+                        pass
+                except FileExistsError:
+                    build_id += 1
+                    continue
                 return filename
 
+    def _allocate_build_id(self, cpv):
```
mgorny, line 2193: Actually, why are there two almost-identical functions? How about using one function that returns a tuple of `(filename, build_id)`?

lib/portage/dbapi/bintree.py
syu-nya: It is actually because `getname` does not know what status the file is, and try to read the empty placeholder.

* Exists package.
* New package.
* New allocated empty placeholder.
* Remote package.

The placeholder file is implemented, but the logic in a lot of other places has been modified.

lib/portage/dbapi/bintree.py
syu-nya: Put the 2 functions together, but create 2 more warp functions that return `filename` and `(filename, build_id)`. Since `getname`used in many place that will broke other tools.

lib/portage/dbapi/bintree.py
```diff
@@ -2049,7 +2036,17 @@ def dep_bestmatch(self, mydep):
...
+        return self._getname(
+            cpv, allocate_new=allocate_new, remote_binpkg_format=remote_binpkg_format
+        )[0]
+
+    def getname_build_id(self, cpv, allocate_new=None, remote_binpkg_format=None):
+        return self._getname(
+            cpv, allocate_new=allocate_new, remote_binpkg_format=remote_binpkg_format
+        )
+
+    def _getname(self, cpv, allocate_new=None, remote_binpkg_format=None):
```
mgorny, line 2049: Any reason for the indirection though? It seems that `_getname` and `getname_build_id` are pretty much the same thing.

```suggestion
    def getname(self, cpv, allocate_new=None, remote_binpkg_format=None):
        return self.getname_build_id(
            cpv, allocate_new=allocate_new, remote_binpkg_format=remote_binpkg_format
        )[0]

    def getname_build_id(self, cpv, allocate_new=None, remote_binpkg_format=None):
```

lib/portage/dbapi/bintree.py
```diff
@@ -2175,29 +2177,40 @@ def _allocate_filename_multi(self, cpv):
...
+            except AttributeError:
+                binpkg_format = self.settings.get(
+                    "BINPKG_FORMAT", SUPPORTED_GENTOO_BINPKG_FORMATS[0]
+                )
+        else:
+            binpkg_format = remote_binpkg_format
 
         if binpkg_format == "xpak":
-            filename_format = "%s-%s.xpak"
+            filename_format = "{}-{}.xpak"
```
mgorny, line 2191:
```suggestion
            binpkg_suffix = "xpak"
```

lib/portage/dbapi/bintree.py
```diff
@@ -2175,29 +2177,40 @@ def _allocate_filename_multi(self, cpv):
...
         elif binpkg_format == "gpkg":
-            filename_format = "%s-%s.gpkg.tar"
+            filename_format = "{}-{}.gpkg.tar"
```
mgorny, line 2193:
```suggestion
            binpkg_suffix = "gpkg.tar"
```

lib/portage/dbapi/bintree.py
```diff
@@ -2175,29 +2177,40 @@ def _allocate_filename_multi(self, cpv):
...
         else:
             raise InvalidBinaryPackageFormat(binpkg_format)
 
         while True:
-            filename = filename_format % (
+            filename = filename_format.format(
                 os.path.join(self.pkgdir, cpv.cp, pf),
                 build_id,
             )
```
mgorny, line 2201:
```suggestion
            filename = f"{os.path.join(self.pkgdir, cpv.cp, pf)}-{build_id}.{binpkg_suffix}"
```

mgorny (commented):
Thanks for doing this. A few more suggestions inline.

lib/portage/dbapi/bintree.py
syu-nya: Oh, that's right.

mgorny (approved):
No more comments from me. Thank you for all the good work!

mgorny:
Ok, I've merged it. Thanks again!

thesamesam:
Thank you! @syu-nya, remember to use Bug:/Closes: tags in future if you can.

PR #894, opened 2022-08-21
Add emerge-hooks to execute user code during the emerge process

Schievel1:
Right now Portage has already bashrc to hook into the ebuild process. However, initially this was thought to manipulate environment variables (I think so, hence the name?) for a single package build. 
Bashrc is sourced several times DURING the ebuild process. However it is not sourced BEFORE and AFTER the ebuild process. 

My proposed feature would be to source emergerc, a script located in /etclportage as well, more times and change a environment variable (e.g. EMERGE_STATE) to let the users control with code is called.
A purpose for that for example would be to make a btrfs snapshot every time before emerging an update.

There are several situations that come to mind, where a user could benefit from a source of emergerc:
- every time emerge starts
- after dependencies are successfully calculated
- when there is a dependency conflict
- after the prompt "Would you like to merge these packages?" (if --ask) right before emerge starts working with the first package
- before --sync, after --sync (although there is postsync.d for that)
- when emerge installed all packages successfully
- when emerge failed with one package

See also https://bugs.gentoo.org/839777


Schievel1:
With this this tell the different phases of emerge apart, one would do the same like in bashrc with envirenment variables 
https://wiki.gentoo.org/wiki//etc/portage/bashrc#Environment_variable

```
if [ "${EMERGE_PHASE}" == "setup" ]
then
  echo "your code here"
fi
```

Schievel1:
Maybe dep clean and unmerge should also get their own hook

PR #895, opened 2022-08-22
Improve "An update to portage" message

amarz45:
I changed "\_highly\_" to "highly". I am not sure if that was a typo or intentional but it looks weird. Also, the name "Portage" should be capitalized in my opinion. And replace apostrophes with backticks to indicate that it is a command.


scardracs:
It's not a typo. It's like that because you can notice better. Also portage is written without capital letters, that's why portage and not Portage

PR #896, opened 2022-09-02
Follow PKGDIR when downloading remote binpkgs

syu-nya:
https://bugs.gentoo.org/864160


AndrewAmmerlaan:
Great, this does indeed fix the problem for me!

syu-nya:
This bug also fixed in https://github.com/gentoo/portage/pull/893.
Because both PR changed same functions and have conflict, this one should not be merged.

PR #897, opened 2022-09-09
Bugfix/bug 864259

palao:
This should close https://bugs.gentoo.org/864259

### DIFF
--- a/lib/portage/tests/dbapi/test_bintree.py
+++ b/lib/portage/tests/dbapi/test_bintree.py
@@ -1,0 +1,54 @@
+# Copyright 2020-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+from unittest.mock import MagicMock
+
+from portage.tests import TestCase
+
+from portage.dbapi.bintree import binarytree
+
+
+class BinarytreeTestCase(TestCase):
+    def test_required_init_params(self):
+        with self.assertRaises(TypeError) as cm:
+            binarytree()
+        self.assertEqual(str(cm.exception), "pkgdir parameter is required")
+        with self.assertRaises(TypeError) as cm:
+            binarytree(pkgdir="/tmp")
+        self.assertEqual(str(cm.exception), "settings parameter is required")
+
+    def test_init_with_legacy_params_warns(self):
+        with self.assertWarns(DeprecationWarning):
+            binarytree(_unused=None, pkgdir="/tmp", settings=MagicMock())
+        with self.assertWarns(DeprecationWarning):
+            binarytree(virtual=None, pkgdir="/tmp", settings=MagicMock())
+
+    def test_instance_has_required_attrs(self):
+        # Quite smoky test. What would it be a better testing strategy?
+        # Not sure yet...
+        required_attrs_no_multi_instance = {
+            "pkgdir", "_multi_instance", "dbapi", "update_ents",
+            "move_slot_ent", "populated", "tree", "_binrepos_conf",
+            "_remote_has_index", "_remotepkgs", "_additional_pkgs",
+            "invalids", "settings", "_pkg_paths", "_populating",
+            "_all_directory", "_pkgindex_version", "_pkgindex_hashes",
+            "_pkgindex_file", "_pkgindex_keys", "_pkgindex_aux_keys",
+            "_pkgindex_use_evaluated_keys", "_pkgindex_header",
+            "_pkgindex_header_keys", "_pkgindex_default_pkg_data",
+            "_pkgindex_inherited_keys", "_pkgindex_default_header_data",
+            "_pkgindex_translated_keys", "_pkgindex_allowed_pkg_keys",
+        }
+        no_multi_instance_settings = MagicMock()
+        no_multi_instance_settings.features = ""
+        no_multi_instance_bt = binarytree(
+            pkgdir="/tmp", settings=no_multi_instance_settings)
+        multi_instance_settings = MagicMock()
+        multi_instance_settings.features = "binpkg-multi-instance"
+        multi_instance_bt = binarytree(
+            pkgdir="/tmp", settings=multi_instance_settings)
+        for attr in required_attrs_no_multi_instance:
+            getattr(no_multi_instance_bt, attr)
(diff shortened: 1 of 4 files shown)


palao:
AFAICS, the failures have nothing to do with this PR, am I wrong?

floppym:
> AFAICS, the failures have nothing to do with this PR, am I wrong?

I'm not sure what gives you that impression; it looks like this PR does cause the failure. It will need to be investigated before this can be merged.

palao:
It's true. My bad, sorry. I'll have a look at it.
Thank you for the comment.

palao:
I fixed the problem following a different approach (since the other was broken).
As I'm quite new to portage I would appreciate a critical view.

Among all possible solutions that I saw, I tried the one that is more testable and less invasive. I hope I did it well...

floppym:
I plan to take a look at this tomorrow.

floppym:
Sorry, but I have to ask you to clean up the commit history a bit.

- Rebase your changes on current master (no merge commits!).
- Some of the commits undo changes made in prior commits. Please squash these into a single commit.
- In cases where you are making large amounts of whitespace changes (de-indenting blocks), please move the whitespace changes into a separate commit. We can then add that commit to .git-blame-ignore-revs.

palao:
Thank you for the feedback.
I'll hope to do the git surgery by Friday. But I cannot promise: these days I'm quite busy.

palao:
> Sorry, but I have to ask you to clean up the commit history a bit.
> 
>     * Rebase your changes on current master (no merge commits!).
> 
>     * Some of the commits undo changes made in prior commits. Please squash these into a single commit.
> 
>     * In cases where you are making large amounts of whitespace changes (de-indenting blocks), please move the whitespace changes into a separate commit. We can then add that commit to .git-blame-ignore-revs.

Done it. Please let me know if I missed something, or there is another problem.

palao:
Hmmm. Before cleaning up the history that was passing. During the git surgery I didn't touch anything related to the (what I see is) the fail.
Any thoughts?

floppym:
> Hmmm. Before cleaning up the history that was passing. During the git surgery I didn't touch anything related to the (what I see is) the fail.
> Any thoughts?

Don't worry about it: there's a flakey test that passed after I restarted it. See bug [850127](https://bugs.gentoo.org/850127).

thesamesam:
@syu-nya

syu-nya:
I would say set `getbinpkg_refresh=False` at https://github.com/gentoo/portage/blob/06e43403a8d198cbf67c4dab81c799351f44f02f/lib/_emerge/actions.py#L171-L174
should be enough.

There are only 3 places in portage will call `populate()` with binpkg index.

The db init at 
https://github.com/gentoo/portage/blob/06e43403a8d198cbf67c4dab81c799351f44f02f/lib/_emerge/actions.py#L3512-L3514

this issue, which will always called after db init 
https://github.com/gentoo/portage/blob/06e43403a8d198cbf67c4dab81c799351f44f02f/lib/_emerge/actions.py#L172-L174

and followed
https://github.com/gentoo/portage/blob/06e43403a8d198cbf67c4dab81c799351f44f02f/lib/_emerge/actions.py#L429-L431
This one already disabled refresh binpkg index.

I was thought `action_build()` could be called by others but did not found one.

If external program call `action_build()`, maybe 
`getbinpkg_refresh=(not emerge_config.target_config.trees["bintree"].populated)`

lib/portage/tests/dbapi/test_bintree.py
```diff
@@ -0,0 +1,155 @@
...
+        bt.populate(add_repos=repos)
+        ppopulate_additional.assert_called_once_with(repos)
+
+    @patch("portage.dbapi.bintree.BinRepoConfigLoader")
+    @patch("portage.dbapi.bintree.binarytree._populate_remote")
+    @patch("portage.dbapi.bintree.binarytree._populate_local")
+    def test_populate_with_getbinpkgs(
+        self, ppopulate_local, ppopulate_remote, pBinRepoConfigLoader
+    ):
+        refresh = "something"
```
floppym, line 109: I'm confused by this. Why are you passing a string in the getbinpkg_refresh kwarg?

floppym:
I don't have a strong opinion on changing the default versus changing the single call site as @syu-nya suggests.

palao:
@syu-nya, the advantage of your proposal is obviously that it requires to change less code.

However, the idea behind my proposal is as follows. The bug fix must ensure that the remote package indices are transferred only once. If that must be the case, I thought it would be better to be explicit about it: only one call to `binarytree.populate` is explicitly done with `getbinpkg_refresh=True` in the PR, as it is now.

Of course I'm more inclined towards my proposal, but yours is also good. I can change to yours, if necessary.

In any case, all this is telling me that, in my opinion, a deeper change is needed in the interface of the `binarytree` class, particularly the way the `binarytree.populate` method works and must be used. Somehow is not clean. That is clear to me. But, honestly, I don't have a clear proposal to do yet... Still I thought that changing the default behavior to make more explicit what is going on is an improvement.

syu-nya:
@palao I don't have strong prefer, but explicitly to refresh index is a good reason. Indeed a lot of old logic could be improved, but it would be very time consuming.

lib/portage/tests/dbapi/test_bintree.py
palao: The idea of this test is to make sure that if `populate` is called with `getbinpkgs=True`, then `_populate_remote` is called, and it is *blindly* given whatever is passed in the `getbinpkg_refresh` argument of `populate`.
This test is also telling us that there is no validation done by `populate` and the semantics of the `getbinpkg_refresh` parameter (and all the responsibility to handle it) is translated to `_populate_remote`.

I chose a meaningless string to make clear the point that for `populate` it is irrelevant what is the value of that parameter.

Do you think that this is wrong or unnecessary? Do you have a better idea?

floppym: Ah, thanks for explaining.

floppym (approved):
This PR looks ok to me now. I'll merge tomorrow if there is no negative feedback before then.

PR #898, opened 2022-09-10
portage: sets: VariableSet: parse *DEPEND in includes/excludes

thesamesam:
VariableSet takes a metadata variable and checks
its contents based on 'includes' or 'excludes'
from the set definition.

Unfortunately, until now, it didn't parse/expand the chosen variable
and would only match it literally (it'd also not check effective
metadata, just what's listed in the ebuild -- no *DEPEND
from an eclass, for example).

If variable is *DEPEND, actually parse includes/excludes
so we can effecitvely match say, includes="dev-lang/go"
against ">=dev-lang/go-1.18" in an ebuild.

Bug: https://bugs.gentoo.org/827974
Bug: https://bugs.gentoo.org/865115

### DIFF
--- a/lib/portage/_sets/dbapi.py
+++ b/lib/portage/_sets/dbapi.py
@@ -168,14 +168,32 @@ class VariableSet(EverythingSet):
             return False
         (values,) = self._metadatadb.aux_get(ebuild, [self._variable])
         values = values.split()
+
+        if "DEPEND" in self._variable:
+            include_atoms = []
+            for include in self._includes:
+                include_atoms.append(Atom(include))
+
+            for x in use_reduce(values, token_class=Atom):
+                if not isinstance(x, Atom):
+                    continue
+
+                for include_atom in include_atoms:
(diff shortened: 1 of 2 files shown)


lib/portage/_sets/dbapi.py
```diff
@@ -168,14 +168,32 @@ def _filter(self, atom):
...
+            include_atoms = []
+            for include in self._includes:
+                include_atoms.append(Atom(include))
+
+            for x in use_reduce(values, token_class=Atom):
+                if not isinstance(x, Atom):
+                    continue
+
+                for include_atom in include_atoms:
+                    print(f" Checking if {x=} intersects with {include_atom=}")
```
floppym, line 182: Support for `{x=}` was added in [Python 3.8](https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging), but we claim to support Python 3.7.

floppym: Oh, I suppose this is just a leftover debug statement anyway.

PR #899, opened 2022-09-10
gpkg-sign add gpg configuration check.

syu-nya:
Closes: https://bugs.gentoo.org/869470

### DIFF
--- a/bin/gpkg-sign
+++ b/bin/gpkg-sign
@@ -16,6 +16,15 @@ def main(
     gpkg_file, keep_current_signature=False, allow_unsigned=False, skip_signed=False
 ):
     eout = portage.output.EOutput()
+
+    if not portage.settings.get("BINPKG_GPG_SIGNING_GPG_HOME"):
+        eout.eerror("BINPKG_GPG_SIGNING_GPG_HOME is not set")
+        exit(1)
+
+    if not portage.settings.get("BINPKG_GPG_SIGNING_KEY"):
+        eout.eerror("BINPKG_GPG_SIGNING_KEY is not set")
+        exit(1)
+
     try:
         package = gpkg.gpkg(settings=portage.settings, gpkg_file=gpkg_file)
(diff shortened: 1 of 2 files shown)


bin/gpkg-sign
```diff
@@ -16,6 +16,25 @@ def main(
     gpkg_file, keep_current_signature=False, allow_unsigned=False, skip_signed=False
 ):
     eout = portage.output.EOutput()
+
+    if (
+        "BINPKG_GPG_SIGNING_GPG_HOME" in portage.settings
+        and portage.settings["BINPKG_GPG_SIGNING_GPG_HOME"]
```
mgorny, line 22:
```suggestion
        not portage.settings.get("BINPKG_GPG_SIGNING_GPG_HOME")
```

perhaps? And so on.

mgorny (approved):
LGTM

mgorny:
(note that I haven't actually tested it, just eyeball-review)

RinCat:
I will push a new update soon.

syu-nya:
How did I trigger a github bug again. 😂

lib/_emerge/actions.py
```diff
@@ -614,6 +614,28 @@ def action_build(
                         in trees[eroot]["root_config"].settings.features
                     )
                 ):
+                    if not trees[eroot]["root_config"].settings.get(
```
mgorny, line 617: Well, the one above was rather short but this really asks for a:

```
for x in ("BINPKG_GPG_SIGNING_GPG_HOME", "BINPKG_GPG_SIGNING_KEY"):
    ...
```

syu-nya: OK, changed.

mgorny (approved):
Again, without testing, but eyeball LGTM.

thesamesam:
> Again, without testing, but eyeball LGTM.

Could you test it please, given you reported the bug?

PR #901, opened 2022-09-16
Fix wrong creation of empty binpkg from binhost

syu-nya:
Closes: https://bugs.gentoo.org/870283


mgorny (approved):
I don't understand it but it looks as if it makes sense, so I trust it does.

PR #903, opened 2022-09-18
Revert changes to portage.checksum._apply_hash_filter

floppym:
The modified code may discard the file size from the output, which is never desired.

Also, it calls remove on the object being iterated over, which is problematic.

Reverts: 18e5a8170c69aecd10f162918de571d85055ae81
Reported-by: Marek Behún

### DIFF
--- a/lib/portage/checksum.py
+++ b/lib/portage/checksum.py
@@ -442,10 +442,10 @@ def _apply_hash_filter(digests, hash_filter):
     """
 
     verifiable_hash_types = set(digests).intersection(hashfunc_keys)
+    verifiable_hash_types.discard("size")
     modified = False
     if len(verifiable_hash_types) > 1:
-        verifiable_hash_types.discard("size")
-        for k in verifiable_hash_types:
+        for k in list(verifiable_hash_types):
             if not hash_filter(k):
                 modified = True
                 verifiable_hash_types.remove(k)
@@ -453,7 +453,11 @@ def _apply_hash_filter(digests, hash_filter):
                     break
 
     if modified:
-        digests = {k: v for k, v in digests.items() if k in verifiable_hash_types}
+        digests = {
+            k: v
+            for k, v in digests.items()
+            if k == "size" or k in verifiable_hash_types
+        }
 
     return digests
 
--- a/lib/portage/tests/util/test_checksum.py
+++ b/lib/portage/tests/util/test_checksum.py
@@ -1,9 +1,9 @@
-# Copyright 2011-2017 Gentoo Foundation
+# Copyright 2011-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 from portage.tests import TestCase
 
-from portage.checksum import checksum_str
+from portage.checksum import checksum_str, _apply_hash_filter
 from portage.exception import DigestException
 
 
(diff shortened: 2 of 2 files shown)


mgorny:
Could you add a minimal regression test, or at least link the reported bug?

floppym:
> Could you add a minimal regression test, or at least link the reported bug?

There is no bug report to link; the issue was reported via IRC.
I would need to spend some time to familiarize myself with how to add a regression test, so I would prefer to leave that for another PR.

mgorny:
Could you at least summarize how to reproduce it, so I could look into adding a regression test?

floppym:
The issue was reported in `#gentoo-portage`. See the comments by kabel.

He submitted 2 patches via this link: https://blackhole.sk/~kabel/src/portage/

The second patch contains this description:
```
From b46d78b815ffc146efeed74656b592a9f0e90f36 Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Marek=20Beh=C3=BAn?= <kabel@kernel.org>
Date: Sun, 18 Sep 2022 13:42:55 +0200
Subject: [PATCH 2/2] BinpkgVerifier: check for existence of "size" key after
 applying hash filter
MIME-Version: 1.0
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit

Currently we check for "size" key in the digests dict before we call
_apply_hash_filter(), resulting in the "size" key being lost, which then
causes:

  File "/usr/lib/python3.10/site-packages/_emerge/BinpkgVerifier.py", line 54, in _start
    if size != digests["size"]:
  KeyError: 'size'

This can be reproduced by for example setting
  PORTAGE_CHECKSUM_FILTER="-* sha512"
in make.conf and then trying to fetch a binkpkg.
```
The patch itself seems incorrect/unnecessary after the _apply_hash_filter code is reverted.

floppym:
I think a regression test could simply check to ensure that  "size"  is not discarded by _apply_hash_filter. I'm just not certain where to put said test.

lib/portage/checksum.py
```diff
@@ -442,18 +442,22 @@ def _apply_hash_filter(digests, hash_filter):
     """
 
     verifiable_hash_types = set(digests).intersection(hashfunc_keys)
+    verifiable_hash_types.discard(size)
```
mgorny, line 445: I'm looking at this and trying to figure out how could it work, given that `size` is not defined here.

mgorny:
Here's a test for you:

```diff
diff --git a/lib/portage/tests/util/test_checksum.py b/lib/portage/tests/util/test_checksum.py
index da864ba22..141b0ff7f 100644
--- a/lib/portage/tests/util/test_checksum.py
+++ b/lib/portage/tests/util/test_checksum.py
@@ -1,9 +1,9 @@
-# Copyright 2011-2017 Gentoo Foundation
+# Copyright 2011-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 from portage.tests import TestCase
 
-from portage.checksum import checksum_str
+from portage.checksum import checksum_str, _apply_hash_filter
 from portage.exception import DigestException
 
 
@@ -146,3 +146,24 @@ class ChecksumTestCase(TestCase):
             )
         except DigestException:
             self.skipTest("STREEBOG512 implementation not available")
+
+
+class ApplyHashFilterTestCase(TestCase):
+    def test_apply_hash_filter(self):
+        indict = {"MD5": "", "SHA1": "", "SHA256": "", "size": ""}
+
+        self.assertEqual(
+            sorted(_apply_hash_filter(indict, lambda x: True)),
+            ["MD5", "SHA1", "SHA256", "size"])
+        self.assertEqual(
+            sorted(_apply_hash_filter(indict, lambda x: x == "MD5")),
+            ["MD5", "size"])
+        self.assertEqual(
+            sorted(_apply_hash_filter(indict, lambda x: x != "MD5")),
+            ["SHA1", "SHA256", "size"])
+        self.assertEqual(
+            sorted(_apply_hash_filter(indict, lambda x: x == "SHA256")),
+            ["SHA256", "size"])
+        # this should return size + one of the hashes
+        self.assertEqual(
+            len(list(_apply_hash_filter(indict, lambda x: False))), 2)
```

lib/portage/checksum.py
floppym: My mistake (forgot quotes).

floppym:
> Here's a test for you:

Thanks! I have added it to the PR.

thesamesam (approved):
LGTM and I'm sorry for not catching it. I remember looking at this and being confused but I think I satisfied myself somehow it was fine.

mgorny (approved):
LGTM

PR #906, opened 2022-09-23
Use pkg_allocated_path in Scheduler

syu-nya:
Bug: https://bugs.gentoo.org/872392

### DIFF
--- a/lib/_emerge/Scheduler.py
+++ b/lib/_emerge/Scheduler.py
@@ -975,7 +975,11 @@ class Scheduler(PollScheduler):
                         continue
 
                     if fetched:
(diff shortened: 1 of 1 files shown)


thesamesam:
Thanks! Would you mind putting some explanation into the commit message?

syu-nya:
Well, this is a mistake that I should added `pkg_allocated_path` to keep the filename and not let Scheduler to create new one and will not be used. Kind feel do not have much to say to fix this.

thesamesam:
Right, I see now. I'd still put it in but I'll chuck that in myself. Thank you!

PR #907, opened 2022-09-24
bin: pass -j from MAKEOPTS to xz for parallel decompression

thesamesam:
>= xz 5.3.3_alpha supports parallel decompression.

### DIFF
--- a/bin/phase-helpers.sh
+++ b/bin/phase-helpers.sh
@@ -544,7 +544,7 @@ unpack() {
 						"with EAPI '${EAPI}'. Instead use 'xz'."
 				fi
 				if ___eapi_unpack_supports_xz; then
-					__unpack_tar "xz -d"
+					__unpack_tar "xz -T$(___makeopts_jobs) -d"
 				else
 					__vecho "unpack ${x}: file format not recognized. Ignoring."
 				fi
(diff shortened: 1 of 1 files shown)


bin/phase-helpers.sh
```diff
@@ -544,7 +544,7 @@ unpack() {
 						"with EAPI '${EAPI}'. Instead use 'xz'."
 				fi
 				if ___eapi_unpack_supports_xz; then
-					__unpack_tar "xz -d"
+					__unpack_tar "xz -T$(___makeopts_jobs) -d"
```
floppym, line 547: If MAKEOPTS does not contain a `-j` flag, `___makeopts_jobs` outputs `-1`. Does xz handle this appropriately?

thesamesam: Is it not 1?

thesamesam: With MAKEOPTS="", I get:
````
portage   560660  1.0  0.0  14628  5796 pts/27   S+   22:53   0:00 xz -T1 -d -c -- /var/tmp/portage/dev-lang/rust-1.64.0/distdir/rustc-1.64.0-src.tar.xz
````

This seems to be working correctly.

floppym: My mistake, nevermind.

thesamesam:
Merged.

PR #909, opened 2022-09-25
Add per format compression flags for binpkg

syu-nya:
This added new config BINPKG_COMPRESS_FLAGS_[format] to be used
for different compression method. General BINPKG_COMPRESS_FLAGS
will be ignored if current method had specified BINPKG_COMPRESS_FLAGS_[format]

e.g. BINPKG_COMPRESS_FLAGS_ZSTD="-9"

Closes: https://bugs.gentoo.org/871573

### DIFF
--- a/bin/quickpkg
+++ b/bin/quickpkg
@@ -153,6 +153,18 @@ def quickpkg_atom(options, infos, arg, eout):
                         "compress": "cat",
                         "package": "sys-apps/coreutils",
                     }
+
+                if (
+                    settings.get(
+                        f"BINPKG_COMPRESS_FLAGS_{binpkg_compression.upper()}", None
+                    )
+                    is not None
+                ):
+                    compression["compress"] = compression["compress"].replace(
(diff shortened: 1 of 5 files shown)


mgorny:
> BINPKG_COMPRESS_FLAGS_[foramt]

Is that german? ;-)

mgorny (commented):
At a first glance, it looks correct to me. However, I wish we wouldn't repeat almost the same code in 4 different places.

syu-nya:
> At a first glance, it looks correct to me. However, I wish we wouldn't repeat almost the same code in 4 different places.

They all in different code paths, and quickpkg and gpkg.py can be used independently. So I don't think it is easy to put them together.

thesamesam:
Can you rebase then I think we can get this in?

syu-nya:
> Can you rebase then I think we can get this in?

Done.

thesamesam:
Thank you!

PR #910, opened 2022-09-25
Add more error handles

syu-nya:
And catch process Exception spawned by portage itself. e.g. bad decompress command.

Closes: https://bugs.gentoo.org/871570


mgorny (approved):
Visually, LGTM. Haven't tested it though.

PR #911, opened 2022-09-25
Print a QA warning if nonfatal is used without dying

mgorny:

### DIFF
--- a/bin/isolated-functions.sh
+++ b/bin/isolated-functions.sh
@@ -108,8 +108,13 @@ __bashpid() {
 }
 
 __helpers_die() {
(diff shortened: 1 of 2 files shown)


ulm:
I guess technically the code is fine, but I wonder if there aren't legitimate uses of `nonfatal` without `die`.

Especially, ebuilds seem to make use of the feature in `src_test`. See for example toolchain.eclass which ignores return status but compares with reference output instead.

floppym:
I agree with @ulm on this; it's not terribly difficult to come up with a use case where using nonfatal without calling die makes sense.

PR #912, opened 2022-09-26
data.py: fix PORTAGE_USERNAME default (bug 873088)

zmedico:
Fix the PORTAGE_USERNAME default for the case where the portage.settings attribute exists and PORTAGE_USERNAME is unset.

Fixes: 18e5a8170c69aecd10f162918de571d85055ae81
Bug: https://bugs.gentoo.org/873088


thesamesam:
Thank you!

PR #913, opened 2022-09-27
Add experimental job server

syu-nya:
To use the job server, enable "jobserver" feature in make.conf. It currently needs to turn off the "network-sandbox" feature.

Added new execute:
bin/jobeserver            standalone job server
bin/jobwrapper.sh         program wrapper

Added new variables:
PORTAGE_JOBSERVER_ADDRESS
PORTAGE_JOBSERVER_PORT
PORTAGE_JOBSERVER_MAX_JOBS
PORTAGE_JOBSERVER_MIN_FREE_MEMORY
PORTAGE_JOBSERVER_MAX_SYSTEM_LOAD
PORTAGE_JOBSERVER_DELAY
PORTAGE_JOBSERVER_REMOTE

For details check man make.conf


syu-nya:
Quick start:

Enable feature `jobserver` and disable `network-sandbox` in `make.conf`
`FEATURES="jobserver -network-sandbox"`

Create wrapper links like:
```
mkdir -p /usr/local/lib/portage/jobserver

ln -s /usr/bin/jobwrapper.sh /usr/local/lib/portage/jobserver/gcc
ln -s /usr/bin/jobwrapper.sh /usr/local/lib/portage/jobserver/g++
ln -s /usr/bin/jobwrapper.sh /usr/local/lib/portage/jobserver/clang
ln -s /usr/bin/jobwrapper.sh /usr/local/lib/portage/jobserver/clang++
ln -s /usr/bin/jobwrapper.sh /usr/local/lib/portage/jobserver/x86_64-pc-linux-gnu-gcc
ln -s /usr/bin/jobwrapper.sh /usr/local/lib/portage/jobserver/x86_64-pc-linux-gnu-g++
ln -s /usr/bin/jobwrapper.sh /usr/local/lib/portage/jobserver/x86_64-pc-linux-gnu-clang
ln -s /usr/bin/jobwrapper.sh /usr/local/lib/portage/jobserver/x86_64-pc-linux-gnu-clang++
ln -s /usr/bin/jobwrapper.sh /usr/local/lib/portage/jobserver/rustc
```
Add more if you needed. But do not add tools like `make`, `cmake`, `cargo`, `python`, as they could spawn others and could run into deadlock.

System installed should go to `/usr/lib/portage/jobserver`.

Adjust `PORTAGE_JOBSERVER_MAX_JOBS` in `make.conf` if you want.

If you want to run the primary job server yourself and see the stats, run `jobserver -s` before you start portage. No root needed.
More configurations check `jobserver --help`

Set portage run parallel jobs use `--jobs X` if you not already. Disable all limiters like `--load-average`. Make sure you had enough memory.

And its done.

If you have multiple build containers. Run `jobserver -a [IP]` at you host system, and set `PORTAGE_JOBSERVER_ADDRESS` and `PORTAGE_JOBSERVER_REMOTE="true"` in your containers.

bin/jobwrapper.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
zmedico, line 1: Please change this to `/usr/bin/env bash` like we use in other files now.  BTW, where is jobwrapper.sh called from?

syu-nya: `jobwrapper.sh` called from ebuild modified PATH. When the build system try to call gcc / clang, they actually called this fake script to handle the corresponding command spawn.

bin/jobserver
```diff
@@ -0,0 +1,386 @@
...
+                if wait_memory:
+                    sleep(1)
+                    continue
+
+            if __max_load__ is not None:
+                if os.getloadavg()[0] > __max_load__:
+                    __jobs_status__ = "System Load Too High"
+                    sleep(1)
+                    continue
+            break
```
xxc3nsoredxx, line 105: Instead of an explicit `while True` loop a cleaner approach would be something like
```python
wait = True
while wait:
    wait = False
    if not_enough_memory:
        wait = True
    elif load_too_high:
        wait = True
```

xxc3nsoredxx (changes requested):

bin/jobserver
```diff
@@ -0,0 +1,389 @@
...
+                with open("/proc/meminfo") as f:
+                    for line in f.readlines():
+                        if line.startswith("MemAvailable:"):
+                            if int(line.split()[1]) < __min_free_mem__:
+                                __jobs_status__ = "Insufficient Free Memory"
+                            else:
+                                wait_memory = False
+                            break
+                if wait_memory:
+                    wait = True
```
xxc3nsoredxx, line 100: `wait_memory` seems useless since you could directly set `wait = True` in the `if` block where you set `__job_status__`.

xxc3nsoredxx (changes requested):

bin/jobserver
```diff
@@ -0,0 +1,386 @@
...
+                break
+
+            if __min_free_mem__ is not None:
+                with open("/proc/meminfo") as f:
+                    for line in f.readlines():
+                        if line.startswith("MemAvailable:"):
+                            if int(line.split()[1]) < __min_free_mem__:
+                                __jobs_status__ = "Insufficient Free Memory"
+                            else:
+                                wait = True
```
xxc3nsoredxx, line 96: This is incorrect since you'd want `wait` to be set to `True` if there isn't enough memory available, not if there _is_ enough.

xxc3nsoredxx (changes requested):

xxc3nsoredxx:
Question for the future: what would install the `jobwrapper.sh` symlinks? Portage install script? Portage ebuild? Individual tool ebuilds? The user manually into `/usr/local`?

I'm leaning towards the individual tool ebuilds, so for example, GCC installs `/usr/lib/.../gcc`, `/usr/lib/.../$target-gcc`, etc. That way only symlinks for tools that exist on the machine are created.

syu-nya:
I think the ebuild should install the symlinks, as it should knows what it going to install.

thesamesam:
Note that we can possibly integrate with `shadowman` here with already tries to handle ccache and distcc (cc @mgorny)?

bin/jobserver
```diff
@@ -0,0 +1,386 @@
...
+    status_thread = threading.Thread(target=job_status, daemon=True)
+    status_thread.start()
+
+    if __primary_server__:
+        ThreadingTCPServer.allow_reuse_address = True
+        __server__ = ThreadingTCPServer((str(__address__), options.port), JobServer)
+        try:
+            __server__.serve_forever()
+        except KeyboardInterrupt:
+            __server__.shutdown()
```
xxc3nsoredxx, line 377: It would be nice to print a newline (or do something similar) after `Ctrl-C` so that the shell prompt doesn't get printed on the same line as the jobserver status info when running `jobserver -s`

xxc3nsoredxx (changes requested):

syu-nya:
New options for `jobserver`, `-S` enable sandbox. Must be run as root.
The job server no longer need to disable `network-sandbox` for local building. This is done by created a shared network namespace between `jobserver` and `ebuild`, that is still isolated and loopback only.
To verify the sandbox, run as root:
```
jobserver -s -S
```
open another root shell:
```
nsenter --net=/proc/PID_TO_JOBSERVER/ns/net
ip addr
```
You can also try to install something and see if portage will using it.

The `jobserver` will now drop root privileges after set up PID file and namespace.

bin/jobserver
```diff
@@ -0,0 +1,453 @@
...
+        __job_running_lock__.release()
+        __job_counter_lock__.release()
+        __job_free_slot_lock__.release()
+
+    def _allow_job(self):
+        global __job_running_lock__
+        global __min_free_mem__
+        global __max_load__
+        global __jobs_status__
+        global __parent_status__
```
Flowdalic, line 95: This global variable does not seem to be used by this function?

bin/jobserver
```diff
@@ -0,0 +1,453 @@
...
+        global __job_counter_lock__
+        global __job_free_slot_lock__
+        global __job_running_lock__
+        global __job_delay__
+
+        __job_running_lock__.release()
+        __job_counter_lock__.release()
+        __job_free_slot_lock__.release()
+
+    def _allow_job(self):
```
Flowdalic, line 90: Please let me note that this function is probably the heart and core of the jobserver feature. I am not saying that it is wrong how it is done currently, but I suspect that we will have a few more iterations over this function once this got merged. :)

I think the hardest part is to get the interaction with `os.getloadavg()` right, because the load average is a dampened value. This means an action, like spawning a new job, is not immediately reflected in the load average. (The same could be said to be true for `MemAvailable`.)

Therefore, I wonder if we may end up with different sleep values depending on wehter or not a job has been allowed.  Sleep longer if a new job has been granted, sleep shorter if we perform just another round of constraint checks. But of course, if that is a sensible idea has been to shown in real-world usage, and it is probably sensible to start with a simple approach.

syu-nya: I did consider sleep longer to get the new job occupied resources. However, this waiting time is highly uncertain. Some tasks may take at least a few seconds to initialize, while other simple tasks may finished in less than 1 second. In this case it can lead to severe performance degradation because it cannot keep up with the running speed. 

On the other hand I do not consider available memory and system avg load as safe guards. An unreasonable amount of parallelism can totally lead to OOM. And the memory and CPU usage of a particular task is far above the average, which is out of the control of the jobserver.

bin/jobserver
```diff
@@ -0,0 +1,453 @@
...
+                        if line.startswith("MemAvailable:"):
+                            if int(line.split()[1]) < __min_free_mem__:
+                                __jobs_status__ = "Insufficient Free Memory"
+                                wait = True
+                            break
+
+            if not wait and __max_load__ is not None:
+                if os.getloadavg()[0] > __max_load__:
+                    __jobs_status__ = "System Load Too High"
+                    wait = True
```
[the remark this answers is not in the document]
xxc3nsoredxx: As it is right now, the `while wait` loop should terminate by default since it's basically a do-while loop. So if `os.getloadavg()[0] <= __max_load__` then `wait` continues to be `False`.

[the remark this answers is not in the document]
Flowdalic: Ahh, I see that now. Thanks for the explanation.

eli-schwartz (commented):
It seems a bit unfortunate to not share an implementation with the GNU make jobserver protocol for the purpose of interoperating with other programs that expect a GNU make jobserver.

For example, gcc's `-flto=jobserver` will check for a GNU make jobserver in order to run the link-time `lto1` processes in parallel. It is unlikely GCC will implement support for "the gentoo jobserver".

It would also mean that all Make-based packages would get seamless support for all build edges without verifying every program you want to wrap and making a shellscript symlink. A wrapper could still be useful for packages that don't natively support the GNU make jobserver.

Note that this PR was submitted roughly two months after GNU make implemented -- and a month before it tagged a release containing -- a new jobserver style using named pipes: https://git.savannah.gnu.org/cgit/make.git/commit/?id=7ad2593b2d2bb5b9332f4444d8bf93ac6f958bc6

This gets rid of the primary flaw with the jobserver, namely that it required passing around file descriptors to exactly the processes you want to participate in the jobserver. It's now possible to have recursive programs handle this correctly out of the box, which means that if a build system doesn't support the jobserver then you *can* still make wrapper scripts that listen to `--jobserver-auth=fifo:PATH` and then use that as the central coordinating manager.

lib/portage/const.py
```diff
@@ -98,6 +98,11 @@
 LOCALE_DATA_PATH = f"{PORTAGE_BASE_PATH}/locale"  # FIXME: not used
 EBUILD_SH_BINARY = f"{PORTAGE_BIN_PATH}/ebuild.sh"
 MISC_SH_BINARY = f"{PORTAGE_BIN_PATH}/misc-functions.sh"
+JOBSERVER_BINARY = f"{PORTAGE_BIN_PATH}/jobserver"
```
xxc3nsoredxx, line 101: I get an error: "Could not find job server binary" when trying to install anything.

This is being installed as a system-wide binary, not an "internal Portage executable" (as the `meson_options.txt` describes the `portage-bindir` option, which gets propagated to `PORTAGE_BIN_PATH` via meson), so it should be `{BINARY_PREFIX}/usr/bin/jobserver` I think. At least it fixes the error for me.

xxc3nsoredxx (changes requested):

lib/portage/const.py
syu-nya: Yes, it should use the full the full path. fixed

syu-nya:
I move the 2 executable
`jobserver` -> `ejobserver`
`jobwrapper.sh` -> `ejobwrapper.sh`
that maybe conflicting with others.
Symlinks need to be updated.

thesamesam:
Closing per https://bugs.gentoo.org/937165#c1, as we have an alternative now. Thank you for this though

PR #916, opened 2022-09-29
ecompress bugfixes

thesamesam:
…ssion

- Pass -9 like we do for bzip2 & gzip.
- Pass -j from MAKEOPTS for parallel compression.

### DIFF
--- a/bin/ecompress
+++ b/bin/ecompress
@@ -71,9 +71,9 @@ while [[ $# -gt 0 ]] ; do
 				# note: to save time, we need to do this only if there's
 				# at least one compressed file
 				case ${path} in
-					*.Z|*.gz|*.bz2|*.lzma|.lz4|*.xz|*.zst)
+					*.Z|*.gz|*.bz2|*.lzma|.lz|.lzo|.lz4|*.xz|*.zst)
 						vpath=${path%.*}
-						for comp in '' .Z .gz .bz2 .lzma .lz4 .xz .zst; do
+						for comp in '' .Z .gz .bz2 .lzma .lz .lzo .lz4 .xz .zst; do
 							if [[ ${vpath}${comp} != ${path} && \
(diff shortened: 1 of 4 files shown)


bin/ecompress
```diff
@@ -128,6 +128,7 @@ if [[ ${PORTAGE_COMPRESS_FLAGS+set} != "set" ]] ; then
 		# See: https://bugs.gentoo.org/672916
 		# Setting '--rm' will remove the source files after a successful compression.
 		lz4)  PORTAGE_COMPRESS_FLAGS="-m --rm";;
+		xz)   PORTAGE_COMPRESS_FLAGS="-9 -T$(__makeopts_jobs)";;
```
mgorny, line 131: I have somewhat mixed feelings about this. On one hand, we're passing `-9` to maximize compression. On the other, we're now defaulting to an arbitrary number of jobs that actually worsens compression. But I guess most people won't really care, and it has probably better gain-to-cost ratio than going for lower compression level.

bin/ecompress
```diff
@@ -71,9 +71,9 @@ while [[ $# -gt 0 ]] ; do
 				# note: to save time, we need to do this only if there's
 				# at least one compressed file
 				case ${path} in
-					*.Z|*.gz|*.bz2|*.lzma|*.xz)
+					*.Z|*.gz|*.bz2|*.lzma|*.xz|*.zst)
```
mgorny: Don't we want `.lz4`, `.lz`, `lzo` here as well?

bin/ecompress
thesamesam: Right, people can still override it if they want.

bin/ecompress
thesamesam: Good point, I'll add that now.

thesamesam:
Merged.

PR #918, opened 2022-09-29
binpkg: compress/decompress xz in parallel; compress zstd in parallel

thesamesam:
- As before, needs >= xz-5.3.3_alpha for parallel decompression.
- zstd does not support parallel decompression.

### DIFF
--- a/lib/portage/package/ebuild/doebuild.py
+++ b/lib/portage/package/ebuild/doebuild.py
@@ -111,7 +111,7 @@ from portage.util import (
     writemsg_stdout,
     write_atomic,
 )
-from portage.util.cpuinfo import get_cpu_count
+from portage.util.cpuinfo import get_cpu_count, makeopts_to_job_count
 from portage.util.lafilefixer import rewrite_lafile
 from portage.util.compression_probe import _compressors
 from portage.util.futures import asyncio
@@ -667,8 +667,12 @@ def doebuild_environment(
                 mysettings["PORTAGE_COMPRESSION_COMMAND"] = "cat"
         else:
             try:
+                compression_binary = compression["compress"].replace(
+                    "{JOBS}",
+                    str(makeopts_to_job_count(mysettings.get("MAKEOPTS", "1"))),
+                )
                 compression_binary = shlex_split(
-                    varexpand(compression["compress"], mydict=settings)
+                    varexpand(compression_binary, mydict=settings)
                 )[0]
             except IndexError as e:
                 writemsg(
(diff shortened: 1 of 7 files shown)


thesamesam:
@syu-nya

lib/portage/util/cpuinfo.py
```diff
@@ -48,3 +50,22 @@ def get_cpu_count():
...
+    """
+    Parse the job count (-jN) from MAKEOPTS. Python version of
+    bin/isolated-functions.sh's ___makeopts_jobs().
+
+    @return: Number of jobs to run or number of CPUs if none set.
+    """
+    if not makeopts:
+        return get_cpu_count()
+
+    # TODO: Fix
```
thesamesam, line 65: This TODO is just because I think the regex could be better like the one at __makeopts_jobs. Suggestions welcome!

https://github.com/gentoo/portage/blob/master/bin/isolated-functions.sh#L472

lib/portage/package/ebuild/doebuild.py
```diff
@@ -667,8 +667,12 @@ def doebuild_environment(
                 mysettings["PORTAGE_COMPRESSION_COMMAND"] = "cat"
         else:
             try:
+                compression_binary = compression["compress"].replace(
```
thesamesam, line 670: I need to test this path (only tested gpkg path so far).

syu-nya:
I think N should be limited in compression, IIRC multiple threads will reduce compression efficiency.

syu-nya:
Seems ok https://yeah.nah.nz/misc/xz-thread/

thesamesam:
> I think N should be limited in compression, IIRC multiple threads will reduce compression efficiency.

It usually does, but only by a small amount. It's hard because we're balancing it with decompression speed (which gets better when compressing xz in parallel at least).

zstd considers the difference between parallel compression and serial compression to be minimal: https://github.com/facebook/zstd/issues/517#issuecomment-275545877 (note that we're talking about zstd here, not pzstd).

Newer xz will default to the threaded compressor even with -T1 / default, so it's not a concern there either I think?

But it is an important question and it's one mgorny raised on another PR. My gut is most of the time, people really want faster installs of both binpkgs and also creating them.

syu-nya:
Yeah, faster is better, I don't think most people do care a few percent size increase.

lib/portage/util/cpuinfo.py
```diff
@@ -48,3 +50,22 @@ def get_cpu_count():
...
+    jobs = re.match(".*(j|--jobs=\s)\s*([0-9]+)", makeopts)
```
mgorny, line 66:
```suggestion
    jobs = re.match(r".*(j|--jobs=\s)\s*([0-9]+)", makeopts)
```

lib/portage/util/cpuinfo.py
```diff
@@ -48,3 +50,22 @@ def get_cpu_count():
...
+
+    if not jobs or not jobs.groups()[1]:
```
mgorny, line 68: I don't think the group can be empty, can it?

thesamesam: Let me check again, I think I had it somehow before, but it may have been with an inferior regex I was using at first.

PR #920, opened 2022-10-04
man: emerge.1: mention gpkg

thesamesam:


thesamesam:
@syu-nya - anything else we should add?

syu-nya:
Looks good.

PR #921, opened 2022-10-07
portage: manifest: fix Manifest.addFile() call to renamed updateFileH…

thesamesam:
…ashes

Bug: https://bugs.gentoo.org/875860
Fixes: 5ac2a8f08101501d0416ad9d79e277c363cea85e

### DIFF
--- a/lib/portage/manifest.py
+++ b/lib/portage/manifest.py
@@ -464,21 +464,21 @@ class Manifest:
 
     def addFile(self, ftype, fname, hashdict=None, ignoreMissing=False):
         """Add entry to Manifest optionally using hashdict to avoid recalculation of hashes"""
-        if ftype == "AUX":
-            if not fname.startswith("files/"):
-                fname = os.path.join("files", fname)
-            if fname.startswith("files"):
-                fname = fname[6:]
+        if ftype == "AUX" and not fname.startswith("files/"):
+            fname = os.path.join("files", fname)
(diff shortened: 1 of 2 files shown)


thesamesam:
Draft because I feel like we could do a test here.

lib/portage/manifest.py
```diff
@@ -477,7 +477,7 @@ def addFile(self, ftype, fname, hashdict=None, ignoreMissing=False):
         if hashdict is not None:
             self.fhashdict[ftype][fname].update(hashdict)
         if self.required_hashes.difference(set(self.fhashdict[ftype][fname])):
-            self.updateFileHashes(
+            self.updateAllFileHashes(
                 ftype, fname, checkExisting=False, ignoreMissing=ignoreMissing
```
hemmond, line 481: There is one more thing in [gentoo bug #875860](https://bugs.gentoo.org/875860) - fname should be in square brackets. 

At least that's what our internal testing with Petr Bahula seems like.

thesamesam: Thanks - and thanks for giving me motivation to actually write a test for it, we clearly need it!

thesamesam:
Added a test. Confirmed it would've caught each of the issues here too.

hemmond:
Looks good to me. Both the test and the fix seems to be OK.

hemmond (approved):
Looks good to me. Both the test and the fix seems to be OK.

thesamesam:
Thanks!

PR #926, opened 2022-10-21
wip: ensure_dirs: check ownership of existing directory

floppym:
Bug: https://bugs.gentoo.org/853283

This is work in progress. I haven't tested all places `ensure_dirs` is used.


floppym:
This seems like it is going to be messy to get working right, and I don't really feel motivated to work on it further.

PR #927, opened 2022-10-26
 Do not redirect stderr from gpkg subprocess

syu-nya:
Closes: https://bugs.gentoo.org/877215

### DIFF
--- a/lib/portage/gpkg.py
+++ b/lib/portage/gpkg.py
@@ -91,6 +91,7 @@ class tar_stream_writer:
         gid              # drop root group to gid
         """
         self.checksum_helper = checksum_helper
+        self.cmd = cmd
         self.closed = False
         self.container = container
         self.killed = False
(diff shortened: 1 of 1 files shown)


mgorny:
Is there a reason you're redirecting stderr in the first place? Do we expect compressors to print unexpected noise? I think most of the tools are quiet by default, and if user passes `-v` via compress flags explicitly, then I suppose it makes sense to have it seen.

syu-nya:
> Is there a reason you're redirecting stderr in the first place? Do we expect compressors to print unexpected noise? I think most of the tools are quiet by default, and if user passes `-v` via compress flags explicitly, then I suppose it makes sense to have it seen.

I was worrying that they will print some noise (like progress and speed), especially in parallel. But it seems to be quiet by default.

lib/portage/gpkg.py
```diff
@@ -123,7 +124,7 @@ def __init__(
                     cmd,
                     stdin=subprocess.PIPE,
                     stdout=subprocess.PIPE,
-                    stderr=subprocess.PIPE,
+                    stderr=None,
```
mgorny, line 127:
```suggestion
```

I think it'd be cleaner not to pass it at all.

PR #928, opened 2022-10-26
 Use binpkg extensions and header to get format

syu-nya:
This deprecates the BINPKG_FORMAT in the metadata and Package index.

Also fixed a bug that obsolete packages that have been moved may be
attempted to be read during package searching.

Closes: https://bugs.gentoo.org/877357
Closes: https://bugs.gentoo.org/877419

### DIFF
--- a/lib/portage/binpkg.py
+++ b/lib/portage/binpkg.py
@@ -2,18 +2,31 @@
 # Distributed under the terms of the GNU General Public License v2
 
 import tarfile
+from portage import os
 from portage.const import SUPPORTED_XPAK_EXTENSIONS, SUPPORTED_GPKG_EXTENSIONS
+from portage.exception import InvalidBinaryPackageFormat
 from portage.output import colorize
 from portage.util import writemsg
 
 
-def get_binpkg_format(binpkg_path):
+def get_binpkg_format(binpkg_path, check_file=False, remote=False):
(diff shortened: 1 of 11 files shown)


mgorny (approved):
I don't understand the code but if it does what it says it does, I think it makes sense in the end. Even though right now we can determine the format based on suffix, we may have e.g. `gpkg-2` in the future that differs only in content.

syu-nya:
Well, I actually changed this to detect the format using extensions and header. So you need to check again.

lib/portage/binpkg.py
```diff
@@ -2,19 +2,32 @@
...
+    if remote:
+        if file_ext_format is not None:
+            return file_ext_format
+        else:
+            raise InvalidBinaryPackageFormat(
+                f"Unsupported binary package format from '{binpkg_path}'"
+            )
+
+    if file_ext_format is not None:
+        return file_ext_format
```
mgorny, line 29: Now I'm confused. When called on a local binpkg, the function would previously check both suffix and content, and report a mismatch. Now it will always use the suffix, so the content check would only be used if someone misnamed the package (would that even work?), and the part about suffix/format mismatch would never be used.

syu-nya: This can lead to a lot of extra IO. especially if you have many versions, but only one of them will be used in the end.
I will add a read check only if the format is invalid.

syu-nya:
File check added.

PR #929, opened 2022-10-27
Unconditional update first level of dir name

syu-nya:
Enable / disable binpkg-multi-instance while doing metadata move / update may cause binary package structure mismatch. This will unconditional replace the first level name (cpv) in the package.

Closes: https://bugs.gentoo.org/877271

### DIFF
--- a/lib/portage/gpkg.py
+++ b/lib/portage/gpkg.py
@@ -1047,6 +1047,14 @@ class gpkg:
             self._add_signature(checksum_info, image_tarinfo, container)
 
         self._add_manifest(container)
+
+        # Check if all directories are the same in the container
+        prefix = os.path.commonpath(container.getnames())
+        if not prefix:
+            raise InvalidBinaryPackageFormat(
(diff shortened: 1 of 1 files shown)


mgorny (changes requested):
I don't think this really solves the problem I've been facing because the files not being updated have the correct directory name, the directory of new files is wrong.

syu-nya:
This does not solves the problem, because I don't know where caused it and I cannot reproduce. But this can eliminate a factor of errors caused by package updates. 
I just added a sanity check, if the directory is wrong it should throw an error when it created.

thesamesam:
> This does not solves the problem, because I don't know where caused it and I cannot reproduce. But this can eliminate a factor of errors caused by package updates. I just added a sanity check, if the directory is wrong it should throw an error when it created.

Thanks, this is what I particularly wanted to see -- we want to assert as soon as we see a mismatch to help narrow it down.

thesamesam:
Could you refresh the commit message then I think this is good to go in? (it's essentially sanity checks)

PR #932, opened 2022-10-30
Add support for a distinct SYSROOT location and fix definition of ESYSROOT

chewi:
This new `SYSROOT` location cannot be prefix-aware as we don't currently have any variable to represent that. Given that this is intended to be used with crossdev toolchains under `/usr/${CHOST}`, that should not matter. Even if that location is nested within a prefixed system, it does not need to be aware of that as it is not intended for execution.

For a given location, we were basing `ESYSROOT` on the prefix of that location instead of the prefix used by the `SYSROOT` location. For example, if `BROOT=/foo`, `ROOT=/bar`, `EPREFIX=/baz`, `EROOT=/bar/baz`, and `SYSROOT=/` then `ESYSROOT` should be `/foo`, not `/baz`.

As discussed, this should have been done years ago, but the drawn out debate we had about changing the definition of `ESYSROOT` dragged on so long that I forgot to do the follow up. Sorry!

I saw that ChromeOS were patching out the overly restrictive `SYSROOT` check years ago, and I hear they still are. They should not need to following this change.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,6 +10,10 @@ Features:
 
 * cleanups: Drop long-obsolete Jython compatibility code.
 
+* Add support for a distinct SYSROOT location that is not equal to / or ROOT.
+  This is only expected to be used for cross-bootstrapping a new system from
+  scratch using a crossdev environment under /usr/${CHOST}.
+
 Bug fixes:
 * sync: Clobber repositories using sync-type=git to match rsync behavior. This
   helps with issues where git-synced repositories can become confused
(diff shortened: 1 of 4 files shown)


ulm (commented):
I cannot say much here; @chewi is really the expert on this.

AFAICS this agrees with PMS for the case when `SYSROOT` is empty and `ROOT` is non-empty.

floppym (approved):

zmedico (approved):
The logic seems to agree with table 8.3 in PMS.

chewi:
Done!

chewi:
Ah, thought I did!

PR #933, opened 2022-11-02
Add emerge hooks

Schievel1:
Right now Portage has already bashrc to hook into the ebuild process. However, initially this was thought to manipulate environment variables (I think so, hence the name?) for a single package build.
Bashrc is sourced several times DURING the ebuild process. However it is not sourced BEFORE and AFTER the ebuild process.

My proposed feature would be to source emergerc, a script located in /etc/portage as well, several times and change a environment variable (e.g. EMERGE_PHASE) to let the users control which code is called.
A purpose for that for example would be to make a btrfs snapshot every time before emerging an update.

There are several situations that come to mind, where a user could benefit from a source of emergerc. The value that the environment variable $EMERGE_PHASE will have it that situation is written behind those points:
- every time emerge starts (emerge_startup)
- right before the dependencies are calculated (pre_calc_deps)
- after dependencies are successfully calculated (post_calc_deps_success)
- after dependencies and they were not successfully calculated (post_calc_deps_fail)
- after the prompt "Would you like to merge these packages?" (if --ask) right before emerge starts working with the first package (pre_first_emerge)
- before --sync (pre_sync)
- after --sync (although there is postsync.d for that) (post_sync_success) / (post_sync_fail)
- when emerge installed all packages successfully  (post_emerge_success)
- when emerge failed with one package (post_emerge_fail)
- when emerge exits (emerge_exit)
- before and after depclean (pre_depclean) / (post_depclean)
- before and after unmerge (pre_unmerge) / (post_unmerge)

To tell the different phases of emerge apart, one would do the same like in bashrc with environment variables
https://wiki.gentoo.org/wiki//etc/portage/bashrc#Environment_variable
```
if [ "${EMERGE_PHASE}" == "pre_first_emerge" ]
then
  echo "your code here"
fi
```

See also:
Bug: https://bugs.gentoo.org/839777

### DIFF
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -104,7 +104,7 @@ def callemergerc(phase='', packages='', opts=''):
     if os.path.isfile(emergerc_script):
         subprocess.run([os.path.join(". /", emergerc_script)],
                        shell=True,
-                       env={"EMERGE_PHASE": phase, "PACKAGES": packages, "OPTS": opts})
+                       env={"EMERGE_PHASE": phase, "PACKAGES": packages, "OPTS": opts, "EMERGE_PID": str(os.getpid())})
 
 def action_build(
     emerge_config,
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -100,13 +100,11 @@ from _emerge.UseFlagDisplay import pkg_use_display
 from _emerge.UserQuery import UserQuery
 
 emergerc_script = os.path.join("/", portage.const.USER_CONFIG_PATH, "emergerc")
-def callemergerc(phase):
+def callemergerc(phase='', packages='', opts=''):
     if os.path.isfile(emergerc_script):
         subprocess.run([os.path.join(". /", emergerc_script)],
                        shell=True,
-                       env={"EMERGE_PHASE": phase})
-
-callemergerc('emerge_startup')
+                       env={"EMERGE_PHASE": phase, "PACKAGES": packages, "OPTS": opts})
 
 def action_build(
     emerge_config,
@@ -269,7 +267,7 @@ def action_build(
     quiet = "--quiet" in myopts
     myparams = create_depgraph_params(myopts, myaction)
     mergelist_shown = False
-    callemergerc('pre_calc_deps')
+    callemergerc(phase='pre_calc_deps', opts=str(myopts))
 
     if pretend or fetchonly:
         mtimedb.make_readonly()
@@ -454,9 +452,9 @@ def action_build(
             return 1
 
     if success:
-        callemergerc('post_calc_deps_success')
+        callemergerc(phase='post_calc_deps_success', packages=str(mydepgraph.altlist()), opts=str(myopts))
     else:
-        callemergerc('post_calc_deps_fail')
+        callemergerc(phase='post_calc_deps_fail', packages=str(mydepgraph.altlist()), opts=str(myopts))
 
     mergecount = None
     if (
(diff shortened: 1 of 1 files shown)


Schievel1:
This works, I have no idea whatsoever about the security implications when calling a shell script like that in python. 
Maybe someone has some better suggestions and I would be happy to put hat in.

Schievel1:
To check out when which phases are run, use 
```
#!/bin/bash

echo ${EMERGE_PHASE}
```
in `/etc/portage/emergerc`

lib/_emerge/actions.py
```diff
@@ -99,6 +99,10 @@
 from _emerge.UseFlagDisplay import pkg_use_display
 from _emerge.UserQuery import UserQuery
 
+emergerc_script = os.path.join("/", portage.const.USER_CONFIG_PATH, "emergerc")
+def callemergerc(phase):
+    os.system(" [[ -f " + emergerc_script + " ]] && export EMERGE_PHASE=" + phase + " && " + os.path.join(". /", emergerc_script))
```
floppym, line 104: Don't use `os.system` for this, especially if `emergerc` is expected to be a bash script (not POSIX-compatible sh).

The `portage.process.spawn_bash` function might be more appropriate if `emergerc` is to be sourced as a bash script.

Alternatively, we could require that `emergerc` be executable and call `portage.process.spawn` or `subprocess.run` on it directly.

The `EMERGE_PHASE` environment var could be passed in via the `env` keyword argument.

Instead of testing for the existence of `emergerc_script` in bash, you could do it in python directly using `os.path.isfile`.

floppym (changes requested):

lib/_emerge/actions.py
Schievel1: Hi, thanks for you suggestions.

That's a lot of options, and I don't really understand the pros and cons for them. Which one would you prefer the most? 

I will do the isFile check in python, no problemo 👌

floppym: I think I would opt for `subprocess.run`, since we don't really have a need for the more advanced features of `portage.process.spawn`.

Schievel1:
Do you have any ideas what else could be passed as an environment variable?

Schievel1:
I added some functionality to send the list of packages and the command line options of emerge to the hook, so users can write scripts that react to that accordingly. I converted this back to draft for the time being, because I still need to test that thoroughly.

MrByte:
Noob question. Can you make an ebuild for those of us that want to use it, ala local overlay?

Schievel1:
> Noob question. Can you make an ebuild for those of us that want to use it, ala local overlay?

Not really. You can make a user patch, but the script is too much of a moving target. I tried that for a while but gave up. Every time the script is changed from upstream, you would need to update the patch.

MrByte:
> > Noob question. Can you make an ebuild for those of us that want to use it, ala local overlay?
> 
> Not really. You can make a user patch, but the script is too much of a moving target. I tried that for a while but gave up. Every time the script is changed from upstream, you would need to update the patch.

Yeah, sadly true. Right now I've got a script "zmerge" that looks for "world" in the parameters and does a ZFS snap for me if that's there.

I may take a swing at a patch at some point, or some other form of hackery.

Thanks for the reply.

floppym:
Hey, sorry for ignoring this PR. I'm not really on board with the idea in concept, so I'm not very motivated to re-review it. I think someone else will need to run with this one.

MrByte:
It gives options, options are always a good thing.

nagelp:
Just a user's opinion here: I'd really like to have the ability to go back in time after emerge has done changes to the system. This feature would enable a btrfs snapshot hook before emerge makes changes. It can make life much more convenient, and it's standard practice in other distributions already - so why not allow the user to do this on Gentoo as well?

PR #934, opened 2022-11-08
Do not drop default PATH in the package index

syu-nya:


thesamesam:
Thanks so much for the quick fix!

PR #935, opened 2022-11-13
Replace repoman with pkgdev in man

yuri1969:
Repoman is deprecated by bug [#835013](https://bugs.gentoo.org/835013).

This one just caught my eye during `man emerge`...

### DIFF
--- a/man/emerge.1
+++ b/man/emerge.1
@@ -570,9 +570,9 @@ ignored because any libraries that have consumers will simply be
 preserved.
 .TP
 .BR \-\-digest
-Prevent corruption from being noticed. The `repoman manifest` command is the
+Prevent corruption from being noticed. The \fBpkgdev manifest\fR command is the
 preferred way to generate manifests and it is capable of doing an entire
-repository or category at once (see \fBrepoman\fR(1)).
+repository or category at once (see \fBpkgdev\fR(1)).
 .TP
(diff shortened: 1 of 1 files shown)


man/emerge.1
```diff
@@ -570,9 +570,9 @@ ignored because any libraries that have consumers will simply be
 preserved.
 .TP
 .BR \-\-digest
-Prevent corruption from being noticed. The `repoman manifest` command is the
+Prevent corruption from being noticed. The \fBpkgdev manifest\fR command is the
 preferred way to generate manifests and it is capable of doing an entire
```
mattst88, line 574: Does `pkgdev manifest` support handling an entire repository? (I have no idea)

We should definitely replace references to `repoman`. My only question is whether the surrounding text is true of `pkgdev` as well.

yuri1969: Removing manifest files from packages in my local repository followed by calling `pkgdev manifest` within the root directory of the repository resulted in generating fresh manifest files for the packages with previously removed manifests.

My conclusion is, it works in the same way `repoman` did.

floppym:
It feels a bit weird to refer to a tool that relies on a competing package manager in the emerge man page. But it does reflect reality.

We might want to enhance Portage to better support Manifest generation.

PR #936, opened 2022-11-19
Make EAPI 8 `--disable-static` logic libtool-specific

SoapGentoo:
* The intention has always been to only target `configure` scripts that use libtool, not just any script with a `--disable-static*` option.

* libtool has been using the same `configure` format for at least the past 15 years (going back to libtool 1.5.22):

  1. shared and static libraries enabled (the main use case): --enable-shared[=PKGS]  build shared libraries [default=yes] --enable-static[=PKGS]  build static libraries [default=yes]

  2. shared libraries enabled and static libraries disabled: --enable-static[=PKGS]  build static libraries [default=no] --enable-shared[=PKGS]  build shared libraries [default=yes]

  3. shared libraries disabled and static libraries enabled: --enable-shared[=PKGS]  build shared libraries [default=no] --enable-static[=PKGS]  build static libraries [default=yes]

Bug: https://bugs.gentoo.org/814380

### DIFF
--- a/bin/phase-helpers.sh
+++ b/bin/phase-helpers.sh
@@ -655,8 +655,8 @@ econf() {
 			fi
 
 			if ___eapi_econf_passes_--disable-static; then
-				if [[ ${conf_help} == *--disable-static* || \
-						${conf_help} == *--enable-static* ]]; then
+				if [[ ${conf_help} == *--enable-shared[^A-Za-z0-9+_.-]* &&
+						${conf_help} == *--enable-static[^A-Za-z0-9+_.-]* ]]; then
 					conf_args+=( --disable-static )
 				fi
 			fi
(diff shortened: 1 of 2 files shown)


bin/phase-helpers.sh
```diff
@@ -655,8 +655,8 @@ econf() {
 			fi
 
 			if ___eapi_econf_passes_--disable-static; then
-				if [[ ${conf_help} == *--disable-static* || \
-						${conf_help} == *--enable-static* ]]; then
+				if [[ ${conf_help} == *--enable-shared\[=PKGS\]* &&
+						${conf_help} == *--enable-static\[=PKGS\]* ]]; then
```
ulm, line 659: If we are to follow what has been proposed for the spec, then it should be like this:
```suggestion
				if [[ ${conf_help} == *--enable-shared[^A-Za-z0-9+_.-]* &&
						${conf_help} == *--enable-static[^A-Za-z0-9+_.-]* ]]; then
```

mgorny (approved):
LGTM, though let's wait for the PMS change to land first.

thesamesam:
Please update NEWS.

floppym (approved):

PR #937, opened 2022-11-19
ebuild/doebuild.py: add -Oline to default MAKEOPTS

xxc3nsoredxx:
Add `-Oline` to ensure proper synchronization of output when doing parallel builds.

### DIFF
--- a/lib/portage/package/ebuild/doebuild.py
+++ b/lib/portage/package/ebuild/doebuild.py
@@ -592,6 +592,8 @@ def doebuild_environment(
             nproc = get_cpu_count()
             if nproc:
                 mysettings["MAKEOPTS"] = "-j%d" % (nproc)
+            if "GNUMAKEFLAGS" not in mysettings:
+                mysettings["GNUMAKEFLAGS"] = "--output-sync=line"
 
         if not eapi_exports_KV(eapi):
             # Discard KV for EAPIs that don't support it. Cached KV is restored
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,10 @@ portage-3.0.45 (UNRELEASED)
 Features:
 * Support new ELF machine code: AMDGPU
 
+* ebuild: Set GNUMAKEFLAGS="--output-sync=line" to ensure build logs are written
+  to synchronously when running GNU make in parallel. This option is only set if
+  MAKEOPTS and GNUMAKEFLAGS are left unset by the user.
+
 Bug fixes:
 * gpkg: Handle out-of-space errors (bug #891391).
 


thesamesam:
I like this personally but bit worried about a revolt as some folks complained about GCC and Binutils output "looking slow". But given this is the default rather than forcefully appended, there's no harm done I think.

Flowdalic:
My 2cents: I wonder how portable this option is. And I personally would prefer the long version "--output-sync=line", if possible, as it reduced the mental load when reading the options.

xxc3nsoredxx:
I think that's a good idea. `-O` is probably associated with compiler optimizations first and foremost.

As for portability? GNU Make (obviously) has that option, with the following comment [indirectly tagged on to it](https://www.gnu.org/software/make/manual/html_node/Features.html#Features):

> If you are concerned with writing portable makefiles, you should not use the features of make listed here, nor the ones in [Missing](https://www.gnu.org/software/make/manual/html_node/Missing.html).

I didn't see anything in [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=make&apropos=0&sektion=0&manpath=FreeBSD+13.1-RELEASE+and+Ports&arch=default&format=html), [NetBSD](https://man.netbsd.org/NetBSD-9.3-STABLE/make.1), or [OpenBSD](https://man.openbsd.org/OpenBSD-7.2/make.1) for output synchronization at a quick glance. Which I would guess to be the next most popular ones.

thesamesam:
We can use GNUMAKEFLAGS to avoid the portability issue.

xxc3nsoredxx:
>  Flags such as ‘--no-print-directory’ or ‘--output-sync’ may be appropriate for GNUMAKEFLAGS.

From [here](https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html) (at the bottom of the page).

Good catch.

lib/portage/package/ebuild/doebuild.py
```diff
@@ -609,6 +609,7 @@ def doebuild_environment(
             nproc = get_cpu_count()
             if nproc:
                 mysettings["MAKEOPTS"] = "-j%d" % (nproc)
+            mysettings["GNUMAKEFLAGS"] = "--output-sync=line"
```
floppym, line 612: You should check to see if `GNUMAKEFLAGS` is already in mysettings before setting it. Otherwise, you will clobber the user's settings if they have GNUMAKEFLAGS in the environment or in make.conf.

thesamesam:
1. Could you update NEWS?
2. By setting `GNUMAKEFLAGS`, we're doing something interesting here: we're no longer just changing it for when people don't set `MAKEOPTS` (I imagine a lot of users do set it), but instead changing it for _almost all_ users (I imagine absolutely nobody sets `GNUMAKEFLAGS`). I think this is acceptable and I think this makes logs a lot more legiible, but it's something to keep in mind.

xxc3nsoredxx:
1. Sure.
2. Not true. The relevant line is _just_ outside of the diff, but here's more context:
```python
if "MAKEOPTS" not in mysettings:
    nproc = get_cpu_count()
    if nproc:
        mysettings["MAKEOPTS"] = "-j%d" % (nproc)
    if "GNUMAKEFLAGS" not in mysettings:
        mysettings["GNUMAKEFLAGS"] = "--output-sync=line"
```
It currently only gets set if users don't have `MAKEOPTS` set _and_ don't have `GNUMAKEFLAGS` set. My thought was to not touch any `make` settings if the user has specified their own, but it could be promoted outside of the `MAKEOPTS` check if you think that's better.

thesamesam:
Ah, wonderful! I'm sorry for missing that. Let's maybe go with it as-is for now, then possibly revisit if nobody complained down the line in a bit.

PR #938, opened 2022-11-19
Check MD5 hashes of all packages installed on your system.

ddwek:
Although it's primarily designed for Gentoo Authors and maintainers, this short and simple script can be used for developers or any users too who want to verify existence and integrity of each file on the filesystem.

P.S.: I really love Gentoo, it's my favorite Linux distro, ever! I think on Portage is the most perfect package management, but very often I had to download over an over again entire packs with missing files such as headers and libraries. That was the reason why I wrote this shell script. I believe that it's useful for you and many people, but if you have to modify it in somehow just do so. Last but not least, if you think this script is not suitable to your needs and decide drop the branch, do that, but don't go breaking my heart :-(

### DIFF
--- a/bin/verify_all_hashes.sh
+++ b/bin/verify_all_hashes.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+#
+# Copyright 2022 Daniel Dwek
+#
+# This file is part of Portage.
+#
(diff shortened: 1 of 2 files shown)


ajakk:
How does this differ from `qcheck` from `app-portage/portage-utils`?

PR #939, opened 2022-11-20
GitSync.update: `git clean` the repo before updating

ajakk:

### DIFF
--- a/lib/portage/sync/modules/git/git.py
+++ b/lib/portage/sync/modules/git/git.py
@@ -228,6 +228,24 @@ class GitSync(NewBase):
         if not self.verify_head(revision="refs/remotes/%s" % remote_branch):
             return (1, False)
 
+        # Clean up the repo before trying to sync to upstream's
+        clean_cmd = [self.bin_command, "clean", "--force", "-d", "-x"]
+
+        if quiet:
+            clean_cmd.append("--quiet")
+
+        portage.process.spawn(
+            clean_cmd,
+            cwd=portage._unicode_encode(self.repo.location),
+            **self.spawn_kwargs,
+        )
+
+        if exitcode != os.EX_OK:
+            msg = "!!! git clean error in %s" % self.repo.location
+            self.logger(self.xterm_titles, msg)
+            writemsg_level(msg + "\n", level=logging.ERROR, noiselevel=-1)
+            return (exitcode, False)
+
         # `git diff --quiet` returns 0 on a clean tree and 1 otherwise
         is_clean = (
             portage.process.spawn(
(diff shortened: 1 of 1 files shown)


lib/portage/sync/modules/git/git.py
```diff
@@ -228,6 +228,24 @@ def update(self):
         if not self.verify_head(revision="refs/remotes/%s" % remote_branch):
             return (1, False)
 
+        # Clean up the repo before trying to sync to upstream's
+        clean_cmd = [self.bin_command, "clean", "--force", "-d", "-x"]
```
thesamesam, line 232: Let's add a ceiling on this so that it doesn't go above the repo dir as a sanity check in case e.g. /var is a git repo too or something.

ajakk: I've just pushed a commit that should do this, not tested it sufficiently yet. Will git really traverse to parent directories of a directory with a `.git` though? We should already only be operating on the directory if `.git` exists within it, via `GitSync.exists()`, called by `NewBase.sync()`.

lib/portage/sync/modules/git/git.py
```diff
@@ -117,6 +117,22 @@ def new(self, **kwargs):
 
         return (os.EX_OK, True)
 
+    def _gen_ceiling_string(self, path):
```
thesamesam, line 120: Why can't we set GIT_CEILING_DIRECTORIES="${PARENT_OF_REPO}", e.g. /var/db/repos if my repo is /var/db/repos/gentoo?

If /var/db/repos/gentoo doesn't have a git repository, it might move up into /var/db/repos, but the environment variable forbids it. git doesn't search laterally as far as I know (imagine /var/db/repos/local with a git repository). Is that what you're aiming to prevent here?

After writing this out, I guess this might be needed with symlinks?

ajakk: The git manpage says, about the environment variable, "If set, it is a list of directories that Git should not chdir up into while looking for a repository". Does that imply it definitely won't chdir into `/var/db/` if `/var/db/repos/` is in the CEILING_DIRECTORIES? I can read it both ways. Maybe I'm just being overly cautious or misunderstanding the manpage, though.

thesamesam: Ah! I was obsessing over how to interpret it to allow lateral movement, but allowing non-linear jumps makes more sense as something which an obnoxious implementation could do. Could you put that in a comment?

lib/portage/sync/modules/git/git.py
```diff
@@ -228,6 +228,24 @@ def update(self):
         if not self.verify_head(revision="refs/remotes/%s" % remote_branch):
             return (1, False)
 
+        # Clean up the repo before trying to sync to upstream's
```
Flowdalic, line 254: What is the motivation for doing so? It appears this could cause data loss for ebuild developers with unadded files in the repo.

PR #940, opened 2022-11-20
src: use designated initialiser syntax for PyModuleDef

thesamesam:
It's far easier to keep track of the arguments this way.


mattst88 (approved):
Yes, please.

PR #941, opened 2022-11-20
depgraph: Resolve atoms using the correct root

chewi:
This bug may have played out in different ways, but one example was a rebuild in / causing the same package to be added to ROOT, even when it had no other reason to be.

Admittedly not entirely sure about this, but it has the right affect.


lib/_emerge/depgraph.py
```diff
@@ -5169,6 +5170,11 @@ def _resolve(self, myfavorites):
                     )
                     raise
 
+        try:
+            del myroot, pkgsettings, pprovideddict, virtuals
+        except NameError:
+            pass
+
```
zmedico, line 5177: Maybe do this in the scope of the loop, so that we don't have to ignore the NameError.

zmedico (approved):
LGTM. Thanks!

lib/_emerge/depgraph.py
chewi: I was going to, but there's a couple of `break`s in there. Someone else suggested adding `del` for safety, but they hadn't considered what would happen if it didn't enter the loop, so then I added the `try`.

chewi:
Done!

PR #942, opened 2022-11-21
Remove Jython compat

xxc3nsoredxx:
Jython is stuck in Python 2-land. Remove Jython compatibility code and references to Jython.

### DIFF
--- a/lib/portage/tests/process/test_PopenProcessBlockingIO.py
+++ b/lib/portage/tests/process/test_PopenProcessBlockingIO.py
@@ -1,4 +1,4 @@
-# Copyright 2012 Gentoo Foundation
+# Copyright 2012-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 import subprocess
@@ -18,10 +18,13 @@ from portage.util._async.PipeReaderBlockingIO import PipeReaderBlockingIO
 
 class PopenPipeBlockingIOTestCase(TestCase):
     """
-    Test PopenProcess, which can be useful for Jython support:
+    Test PopenProcess, which is historically useful for Jython support:
             * use subprocess.Popen since Jython does not support os.fork()
             * use blocking IO with threads, since Jython does not support
               fcntl non-blocking IO)
+
+    Portage does not currently support Jython, but re-introducing support
+    in The Future (TM) may be possible.
     """
 
     _echo_cmd = "echo -n '%s'"
--- a/lib/portage/tests/runTests.py
+++ b/lib/portage/tests/runTests.py
@@ -1,12 +1,11 @@
 #!/usr/bin/env python
 # runTests.py -- Portage Unit Test Functionality
-# Copyright 2006-2020 Gentoo Authors
+# Copyright 2006-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 import grp
 import os
 import os.path as osp
-import platform
 import pwd
 import signal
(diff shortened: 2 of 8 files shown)


lib/portage/tests/process/test_PopenProcessBlockingIO.py
```diff
@@ -18,10 +18,7 @@
 
 class PopenPipeBlockingIOTestCase(TestCase):
     """
-    Test PopenProcess, which can be useful for Jython support:
```
thesamesam, line 21: This is a bit misleading -- we should note why we're doing it, the tests are still the way they are, right?

lib/portage/tests/runTests.py
```diff
@@ -21,12 +21,7 @@ def debug_signal(signum, frame):
     pdb.set_trace()
 
 
-if platform.python_implementation() == "Jython":
```
thesamesam, line 24: Tag the bug in your commit just for completeness

lib/portage/util/_async/PipeReaderBlockingIO.py
```diff
@@ -15,10 +15,7 @@ class PipeReaderBlockingIO(AbstractPollTask):
     """
     Reads output from one or more files and saves it in memory, for
     retrieval via the getvalue() method. This is driven by a thread
-    for each input file, in order to support blocking IO.  This may
-    be useful for using threads to handle blocking IO with Jython,
-    since Jython lacks the fcntl module which is needed for
-    non-blocking IO (see http://bugs.jython.org/issue1074).
+    for each input file, in order to support blocking IO.
```
thesamesam, line 18: I would again keep a historical comment so we can re-evaluate if the whole thing is needed in future?

lib/portage/tests/runTests.py
xxc3nsoredxx: Sure.

lib/portage/util/_async/PipeReaderBlockingIO.py
xxc3nsoredxx: Alright, I can leave the historical comments in.

xxc3nsoredxx: Actually, I'll just remove these files from the commit since it'd be just a comment change.

thesamesam: I think it might be worth leaving the historical comments in but rephrasing to make clear that Jython is longer a concern.

xxc3nsoredxx: Ah, I can do that.

xxc3nsoredxx:
I only had part of the changes staged in the previous push smh my head :facepalm:

thesamesam (approved):

thesamesam:
pylint isn't happy

xxc3nsoredxx:
I see that. Is it the unused imports it's complaining about? This makes it almost sound like it's not:
```
Error: ERROR: InvocationError for command /home/runner/work/portage/portage/run-pylint (exited with code 4)
```
https://github.com/gentoo/portage/actions/runs/3512355979/jobs/5899540455#step:6:123
(Similar for 3.8-3.10)

thesamesam:
Yeah, it's the imports

xxc3nsoredxx:
Should be an easy fix then. I don't like how ambiguous the error message is, especially since it rated my code "10/10" lol.

PR #943, opened 2022-11-23
WIP: install-qa-check.d/20perl-directories: add new QA check

Flowdalic:
@akhuettel: @sam [asked me for such a check](https://github.com/gentoo/gentoo/pull/28222#issuecomment-1311528464), does this go into the right direction? As of now, this is untested WIP.

Check for perl modules installed into invalid directories.

Bug: https://bugs.gentoo.org/876757

### DIFF
--- a/bin/install-qa-check.d/20perl-directories
+++ b/bin/install-qa-check.d/20perl-directories
@@ -1,0 +1,28 @@
+# Check for perl modules installed in invalid directories.
+
+perl_dir_check() {
+	local invalid_dirs=(
+		usr/share/perl5
+	)
+	if [[ ${PN} != dev-lang/perl ]]; then
+		invalid_dirs+=( usr/lib64/perl5/5.* )
+	fi
+
+	# See, for example, bug #876757
+	local dir detected_invalid_dirs=()
+	for dir in "${invalid_dirs[@]}" ; do
+		if [[ -d ${ED}/${dir} ]]; then
+		   detected_invalid_dirs+=( "${x}" )
+		fi
+	done
+	if [[ ${#detected_invalid_dirs[@]} -gt 0 ]] ; then
+		eqawarn "QA Notice: This ebuild installs perl modules into invalid directories."
+		eqawarn " Maybe PERLDIR is not set accordingly in the ebuild?"
+		eqatag -erl.invalid-module-directory -v "${detected_invalid_dirs[@]}"
+	fi
+}
+
+perl_dir_check
+: # guarantee successful exit
+
+# vim:ft=sh


bin/install-qa-check.d/20perl-directories
```diff
@@ -0,0 +1,22 @@
...
+	for x in usr/share/perl5 ; do
+		if [[ -d "${ED}/${x}" ]]; then
+		   f+=( ${x} )
+		fi
+	done
+	if [[ ${#f[@]} -gt 0 ]] ; then
+		eqawarn "QA Notice: This ebuild installs perl modules into invalid directories."
+		eqawarn " Maybe PERLDIR is not set accordingly in the ebuild?"
+		eqawarn
+		eqawarn $(printf '%s\n' "${f[@]}")
```
mgorny, line 15: There's `eqatag` for that kind of thing, that also aids automated processing (not that anything that part).

bin/install-qa-check.d/20perl-directories
```diff
@@ -0,0 +1,22 @@
+# Check for perl modules installed in invalid directories.
+
+perl_dir_check() {
+	# See, for example, bug #876757
+	local dir f=()
+	for x in usr/share/perl5 ; do
```
mgorny, line 6: I suppose you want to check for `/usr/lib64/perl5/5.*` directory (i.e. stuff being installed outside `vendor_perl`) by packages other than perl itself?

bin/install-qa-check.d/20perl-directories
```diff
@@ -0,0 +1,22 @@
...
+		if [[ -d "${ED}/${x}" ]]; then
```
mgorny, line 7: Unnecessary quotes.

bin/install-qa-check.d/20perl-directories
```diff
@@ -0,0 +1,22 @@
...
+		   f+=( ${x} )
```
mgorny, line 8: Missing quotes.

akhuettel:
This is in principle a good idea.

My only comment is, there are packages that deliberately install their own module tree (for use only within the package). autoconf, fvwm, texlive, texinfo, ... This is OK as long as the modules have no binary part (well, the code still needs to be working under new perl, but it doesn't have to be forcefully reinstalled). 

So what we need is some variable we can set in the ebuild to disable this warning.

akhuettel:
Looking at the code, you're only checking for the presence of "invalid directories" (given in some list) by checking for the existence of these directories. Anything could be in there, and they could even be empty.

Wouldn't it make more sense to check for the presence of Perl modules (*.pm files) and then collect the directories where such files exist, and check that list against the "invalid directories"?

Also, the correct place is  `/usr/$(get_libdir)/perl5/vendor_perl/`

PR #946, opened 2022-11-25
Add caching to _slot_operator_check_reverse_dependencies

zmedico:
Add lru_cache to speed up the running time of "Calculating
dependencies".

In a ChromeOS use case, this patch decreases the running time from
311s to 197s with almost no memory usage increase.

Bug: https://bugs.gentoo.org/883071
Signed-off-by: Pin-yen Lin <treapking@chromium.org>

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -2240,6 +2240,7 @@ class depgraph:
 
         return None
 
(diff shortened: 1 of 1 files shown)


floppym (approved):
Sounds like a nice improvement.

PR #947, opened 2022-11-26
cnf: sets: add @nim-rebuild set

CyberTailor:
Nim is statically linked (like Go and Rust) and so this set could be helpful.

*(Although it would also rebuild all packages that just install source code)*


thesamesam:
I think the Nim package should probably install this (and we're likely to move the same for Rust & Go).

PR #948, opened 2022-11-27
Fix some type errors

xxc3nsoredxx:
Knock out some low-hanging fruit given by running

    mypy --pretty --show-error-context .

inside lib/ (191 total errors found).
```
portage/package/ebuild/_config/special_env_vars.py:248: error: Incompatible types in assignment (expression has type "FrozenSet[str]", variable has type "List[str]")  [assignment]
    environ_whitelist = frozenset(environ_whitelist)
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
portage/package/ebuild/_config/special_env_vars.py:353: error: Incompatible types in assignment (expression has type "FrozenSet[str]", variable has type "List[str]")  [assignment]
    environ_filter = frozenset(environ_filter)
                     ^~~~~~~~~~~~~~~~~~~~~~~~~
portage/__init__.py: note: At top level:
portage/__init__.py:537: error: Need type annotation for "_testing_eapis" [var-annotated]
    _testing_eapis = frozenset([])
                     ^~~~~~~~~~~~~
portage/checksum.py:342: error: Incompatible types in assignment (expression has type "int", variable has type "bool")  [assignment]
            prelink_capable = 1
                              ^
portage/process.py:64: error: Incompatible types in assignment (expression has type "None", variable has type "str")  [assignment]
            _fd_dir = None
                      ^~~~
portage/process.py:68: error: Incompatible types in assignment (expression has type "None", variable has type "str")  [assignment]
        _fd_dir = None
                  ^~~~
_emerge/Package.py: note: At top level:
_emerge/Package.py:865: error: Incompatible types in assignment (expression has type "FrozenSet[str]", variable has type "Set[str]")  [assignment]
    _all_metadata_keys = frozenset(_all_metadata_keys)
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
portage/getbinpkg.py:46: error: Incompatible types in assignment (expression has type "Tuple[Type[Exception], ...]", variable has type "List[Type[Exception]]")  [assignment]
    _all_errors = tuple(_all_errors)
                  ^~~~~~~~~~~~~~~~~~
```


xxc3nsoredxx:
Setting in `lib/portage/__init__.py`
```python
_testing_eapis = frozenset([""])
```
was incorrect since that would result in `eapi_is_supported()` returning `True` for empty string. Reverted.

lib/_emerge/Package.py
```diff
@@ -860,9 +860,10 @@ def with_use(self, use):
         return pkg
 
 
-_all_metadata_keys = set(x for x in portage.auxdbkeys)
-_all_metadata_keys.update(Package.metadata_keys)
-_all_metadata_keys = frozenset(_all_metadata_keys)
+_all_metadata_keys_temp = set(x for x in portage.auxdbkeys)
+_all_metadata_keys_temp.update(Package.metadata_keys)
+_all_metadata_keys = frozenset(_all_metadata_keys_temp)
+del _all_metadata_keys_temp
```
floppym, line 866: This change is pretty ugly. Can you come up with a way to avoid using a temporary variable?

lib/portage/package/ebuild/_config/special_env_vars.py
```diff
@@ -348,9 +349,10 @@
 ]
 
 # No longer supported variables
-environ_filter += ["SYNC"]
+environ_filter_temp += ["SYNC"]
 
-environ_filter = frozenset(environ_filter)
+environ_filter = frozenset(environ_filter_temp)
+del environ_filter_temp
```
floppym, line 355: Again, this temporary variable just clutters the code. I don't like this.

lib/portage/process.py
```diff
@@ -61,13 +61,13 @@
     if os.path.isdir(_fd_dir):
         break
     else:
-        _fd_dir = None
+        _fd_dir = ""
```
floppym, line 64: I think this change makes the code worse.

floppym (changes requested):

lib/portage/process.py
xxc3nsoredxx: Can you elaborate on why you think using `""` makes it worse?

lib/_emerge/Package.py
xxc3nsoredxx: Yeah, I think I can.

xxc3nsoredxx: I've removed the temp var.

lib/portage/package/ebuild/_config/special_env_vars.py
xxc3nsoredxx: Removed this one too.

lib/portage/process.py
floppym: Setting it to `None` conveys that we are don't have a valid value to use. If we were to accidentally pass None to a function that expects a valid path, we would likely get a helpful error at runtime.

Setting it to `""` looks like it might just be a mistake, and the behavior of functions that accept a path equal to empty string is less predictable.

lib/_emerge/Package.py
```diff
@@ -860,9 +860,7 @@ def with_use(self, use):
         return pkg
 
 
-_all_metadata_keys = set(x for x in portage.auxdbkeys)
-_all_metadata_keys.update(Package.metadata_keys)
-_all_metadata_keys = frozenset(_all_metadata_keys)
+_all_metadata_keys = frozenset(set(portage.auxdbkeys).union(Package.metadata_keys))
```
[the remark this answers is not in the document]
xxc3nsoredxx: It seems to come from when elements of `auxdbkeys` were conditionally included, up until 1dbe0f2

```patch
--- a/lib/_emerge/Package.py
+++ b/lib/_emerge/Package.py
@@ -791,8 +791,7 @@ def with_use(self, use):
 			pkg = self
 		return pkg
 
-_all_metadata_keys = set(x for x in portage.auxdbkeys \
-	if not x.startswith("UNUSED_"))
+_all_metadata_keys = set(x for x in portage.auxdbkeys)
 _all_metadata_keys.update(Package.metadata_keys)
 _all_metadata_keys = frozenset(_all_metadata_keys)
```
Some of the `x for x` did get dropped, so it was probably oversight.

lib/portage/process.py
xxc3nsoredxx: Makes sense, I'll drop that change.

xxc3nsoredxx: Reverted.

lib/_emerge/Package.py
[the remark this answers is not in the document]
floppym: Ah, that explanation makes me feel better. Thanks!

floppym (approved):

zmedico (approved):

PR #950, opened 2022-11-30
Properly assign filenames for remote packages

syu-nya:


mgorny (approved):
At a first glance, makes sense.

PR #952, opened 2022-12-01
vartree: use ewarn, not eerror, if merge is going ahead despite colli…

thesamesam:
…sions

Collisions aren't always fatal - say if FEATURES="protect-owned unmerge-orphans" and the collision is over an orphaned file.

This changes the colors used once we discover that all relevant files are orphaned to be warning rather than an error, to make it less scary and distinguish from other situations.


thesamesam:
Old output:
Screenshot 2022-12-01 at 06 36 27

New output:
image

mgorny (approved):
LGTM. Thanks!

PR #954, opened 2022-12-01
Add a pre-commit config

ajakk:
Using pre-commit instead of a manually written pre-commit hook will run the linter only over files that have changes, resulting in a much faster linting hook.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,10 @@
+---
+repos:
+  - repo: https://github.com/psf/black
+    rev: 22.10.0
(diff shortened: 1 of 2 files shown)


thesamesam:
Should this handle pylint too?

ajakk:
Sure!

mgorny:
Please don't. pre-commit is very bad design (look at the size of your home directory), and upstream is extremely toxic.

ajakk:
Regardless of anyone's personal feelings about it, it's still an opt-in and there are obvious benefits as stated in the patch and commit message.

PR #955, opened 2022-12-02
bintree: Fix breaking GPKG structure on updates

mgorny:
Fix passing the wrong basename to the gpkg initializer when performing package updates.  This caused newly written files to be missing build_id and therefore causing a structure mismatch.

Closes: https://bugs.gentoo.org/877271


mgorny:
CC @syu-nya

PR #958, opened 2022-12-05
glsa: abort on incorrect arch delimiter

thesamesam:
Bug: https://bugs.gentoo.org/882797

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,8 +5,10 @@ Features:
 * TODO
 
 Bug fixes:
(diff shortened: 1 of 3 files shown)


thesamesam:
Going to rework this.

ajakk (approved):
Looks great, thanks! In tinkering I found that `glsa-check --dump` doesn't catch this (it just naively converts the xml to plaintext), but I'm not sure I object to that behavior.

Looks good to me!

PR #959, opened 2022-12-13
logging: Write logs under BROOT, not EPREFIX

chewi:
Logs should be written to the build system, not the host system. When building for a prefixed system under `ROOT`, logs were being written under `/${EPREFIX}` even though that directory did not exist at all.


zmedico (approved):
LGTM. Thanks!

thesamesam:
Thanks Chewi and also thanks @zmedico for always being on-hand for these reviews, it genuinely is invaluable and a big help.

PR #931, opened 2022-10-30
GitSync.update: clobber changes to git repo as rsync does

ajakk:
We can also simplify the logic slightly by using `git reset --hard` to set the new HEAD for both shallow and deep clones.

Signed-off-by: John Helmert III <ajak@gentoo.org>

This seems to work at runtime for both shallow and deep clones of the repo.

PR #961, opened 2022-12-17
Add 'volatile' configuration option

thesamesam:
config: Add 'volatile' configuration option

This introduces a 'volatile' configuration option for repos.conf.

* If a repository is marked as volatile (volatile=true), then Portage
  will assume the repository may be changed at any time, and that
  it may contain e.g. local changes.

  Portage will not perform optimizations based on assuming its integrity.

  It will also not perform destructive operations if it is a git repository
  (it will not prioritise being able to continue to sync).

* If a repository is marked as non-volatile (volatile=false), then Portage
  will assume that only Portage changes the repository and that it will not
  be modified otherwise.

  Portage will perform optimizations based on assuming this, like not
  needing to verify the metadata cache for ebuilds & eclasses.

  Portage will prioritise the repository continuing to work and may
  e.g. 'git reset --hard' or 'git clean -fdx' in order to keep git repositories
  working for sync purposes. (If you do not like this, please set 'volatile=true').

See: https://github.com/gentoo/portage/pull/801
See: https://github.com/gentoo/portage/pull/931
See: https://github.com/gentoo/portage/pull/939
See: https://github.com/gentoo/portage/pull/960

### DIFF
--- a/lib/_emerge/Package.py
+++ b/lib/_emerge/Package.py
@@ -210,7 +210,7 @@ class Package(Task):
     @property
     def invalid(self):
         if self._invalid is None:
-            self._validate_deps()
+            self._validate_deps(True)
             if self._invalid is None:
                 self._invalid = False
         return self._invalid
@@ -235,7 +235,7 @@ class Package(Task):
         atoms left unevaluated.
         """
         if self._validated_atoms is None:
-            self._validate_deps()
+            self._validate_deps(True)
         return self._validated_atoms
 
     @property
@@ -305,7 +305,7 @@ class Package(Task):
 
         return tuple(elements)
 
-    def _validate_deps(self):
+    def _validate_deps(self, force=False):
         """
         Validate deps. This does not trigger USE calculation since that
         is expensive for ebuilds and therefore we want to avoid doing
@@ -314,6 +314,10 @@ class Package(Task):
         eapi = self.eapi
         dep_eapi = eapi
         dep_valid_flag = self.iuse.is_valid_flag
+
+        if not force and not self.root_config.settings.repositories[self.repo].volatile:
+            return
+
         if self.installed:
             # Ignore EAPI.incompatible and conditionals missing
             # from IUSE for installed packages since these issues


thesamesam:
If we pursue https://github.com/gentoo/portage/pull/939, it'll be gated behind `volatile = false` too.

thesamesam:
@zmedico Would you mind taking a look at the optimisation commits in particular?

lib/portage/sync/modules/git/git.py
```diff
@@ -227,6 +273,29 @@ def update(self):
...
+            # - Only done for volatile=false repositories to avoid losing
+            # data.
+            # - This is needed to avoid orphaned files preventing further syncs
+            # on shallow clones.
+            clean_cmd = [self.bin_command, "clean", "--force", "-d", "-x"]
+
+            if quiet:
+                clean_cmd.append("--quiet")
+
+            portage.process.spawn(
```
Flowdalic, line 287:
```suggestion
            exitcode = portage.process.spawn(
```

Flowdalic (changes requested):
Please add some rationale for the "sync: git: git clean the repo before updating" clean commit. I would prefer the information in form of a code comment block, but having it in the git commit message is also fine. But since it is not immediately obvious why a `git clean` is necessary for repositories that are fully controller by portage, some words on the motivation for doing so would be very helpful. If I understand correctly, it boils down to
1. users creating files in the repo when they should not 
2. the "CDN issue" where a subsequent fetch returns a later commit, causing git to leave untracked files in the repo

As for the logic to determine the state of `volatile` per repo. My [suggestion](https://archives.gentoo.org/gentoo-dev/message/816172d515c0bb85dc84dd306a6420dc) is

```
if volatile_explicitly_configured:
    volatile = explicitly_configured_volatile_value
else if prefix
    volatile = false
else if Path(repo_dir).user != (root|portage) # Or, if uid >= 1000
    volatile = true
else
    volatile = false
```

lib/portage/dbapi/porttree.py
```diff
@@ -637,7 +638,9 @@ def _pull_valid_cache(self, cpv, ebuild_path, repo_path):
                 # EAPI from _parse_eapi_ebuild_head, we disregard cache entries
                 # for unsupported EAPIs.
                 continue
-            if auxdb.validate_entry(metadata, ebuild_hash, eclass_db):
+            if not repo.volatile or auxdb.validate_entry(
+                metadata, ebuild_hash, eclass_db
+            ):
```
zmedico, line 643: I feel like we probably need a _pull_valid_cache to indicate cases like egencache where the entries are expected to be invalid.

thesamesam:
> Please add some rationale for the "sync: git: git clean the repo before updating" clean commit. 

The rationale is in the subsequent commits (last commit but also NEWS), but I'll put it in there as well wrt the orphaned files bit at least (to not end up copy/pasting the same explanation repeatedly).

>I would prefer the information in form of a code comment block, but having it in the git commit message is also fine.

Sure, I can do that as well.

> But since it is not immediately obvious why a `git clean` is necessary for repositories that are fully controller by portage, some words on the motivation for doing so would be very helpful. If I understand correctly, it boils down to
> 
>     1. users creating files in the repo when they should not
> 
>     2. the "CDN issue" where a subsequent fetch returns a later commit, causing git to leave untracked files in the repo
> 

Right, although the motivation was really 2., not 1. We're not really worried about self-inflicted breakage.

I think your compromise sounds good and I'll implement it. Thanks for the comments.

lib/portage/sync/modules/git/git.py
thesamesam: Good spot, thanks!

lib/portage/dbapi/porttree.py
thesamesam: Ah, good point. But this is enough to make me think we should revisit this separately from the git changes in order to get those in. Thank you!

man/portage.5
```diff
@@ -1259,6 +1259,14 @@ Keep snapshots in \fBDISTDIR\fR (do not delete). Defaults to no, false.
...
 signature. This uses the OpenPGP key(ring) specified by the
 sync\-openpgp\-key\-path setting. Defaults to no, false.
+.B volatile
+Specifies whether a repository is volatile.  Volatile repositories
+are assumed to contain changes made outside of Portage.  This prohibits
+optimizations from occurring by assuming the integrity of the repository
+and its caches may be violated.  Disabling this option allows local changes to
+be made to the repository for e.g. development safely.  For git-synced
+repositories, Portage will not perform destructive operations.
+.TP
```
Flowdalic, line 1269: Maybe a sentence about the default value. For example:
> Per default, repositories are considered non-volatile, unless their directory is not owned by 'root' or 'portage', or if they are not under `/var/db/repos/`.

Flowdalic (approved):
LGTM

man/portage.5
Flowdalic: Or maybe the inverse
> Per default, repositories are considered volatile if their directory is owned by either 'root' or 'portage' and if it is under `/var/db/repos`. Otherwise, the repository is considered non volatile.

ajakk:
Is the added complexity here justified by the tiny minority of users that would find use in it? Wouldn't this be solved by `auto-sync=false`? If someone's manually fiddling with their Gentoo repository, I'd think it would be minimal overhead for them to then sync manually, or maybe even sync via a `postsync.d` hook.

thesamesam:
> Is the added complexity here justified by the tiny minority of users that would find use in it? Wouldn't this be solved by `auto-sync=false`? If someone's manually fiddling with their Gentoo repository, I'd think it would be minimal overhead for them to then sync manually, or maybe even sync via a `postsync.d` hook.

I still think it's a usecase I wouldn't like to support, but enough people do it that it's not worth arguing over, and it also means we can implement the optimisation I wanted to add (which requires all of this anyway).

thesamesam:
Merged modulo the validate_deps bit which will be done separately.

PR #962, opened 2022-12-21
Volatile repository optimisations

thesamesam:
Split out from https://github.com/gentoo/portage/pull/961.

### DIFF
--- a/lib/_emerge/Package.py
+++ b/lib/_emerge/Package.py
@@ -208,7 +208,7 @@ class Package(Task):
     @property
     def invalid(self):
         if self._invalid is None:
-            self._validate_deps()
+            self._validate_deps(force=True)
             if self._invalid is None:
                 self._invalid = False
         return self._invalid
@@ -233,7 +233,7 @@ class Package(Task):
         atoms left unevaluated.
         """
         if self._validated_atoms is None:
-            self._validate_deps()
+            self._validate_deps(force=True)
         return self._validated_atoms
 
     @property
@@ -302,7 +302,21 @@ class Package(Task):
 
         return tuple(elements)
 
-    def _validate_deps(self):
+    def _populate_soname_metadata(self):
+        if self.built:
+            k = "PROVIDES"
+            try:
+                self._provides = frozenset(parse_soname_deps(self._metadata[k]))
+            except InvalidData as e:
+                self._invalid_metadata(k + ".syntax", f"{k}: {e}")
+
+            k = "REQUIRES"
+            try:
+                self._requires = frozenset(parse_soname_deps(self._metadata[k]))
+            except InvalidData as e:
+                self._invalid_metadata(k + ".syntax", f"{k}: {e}")
+
+    def _validate_deps(self, force=False):
         """
         Validate deps. This does not trigger USE calculation since that
         is expensive for ebuilds and therefore we want to avoid doing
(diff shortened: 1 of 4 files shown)


lib/_emerge/Package.py
```diff
@@ -314,6 +314,10 @@ def _validate_deps(self):
         eapi = self.eapi
         dep_eapi = eapi
         dep_valid_flag = self.iuse.is_valid_flag
+
+        if not force and not self.root_config.settings.repositories[self.repo].volatile:
```
thesamesam, line 318: Need to address @zmedico's comment at https://github.com/gentoo/portage/pull/961#discussion_r1051689888.

lib/portage/dbapi/porttree.py
```diff
@@ -637,7 +638,9 @@ def _pull_valid_cache(self, cpv, ebuild_path, repo_path):
                 # EAPI from _parse_eapi_ebuild_head, we disregard cache entries
                 # for unsupported EAPIs.
                 continue
-            if auxdb.validate_entry(metadata, ebuild_hash, eclass_db):
+            if not repo.volatile or auxdb.validate_entry(
```
thesamesam, line 641: Need to address @zmedico's comment at #961 (comment).

lib/_emerge/Package.py
```diff
@@ -314,6 +314,10 @@ def _validate_deps(self):
...
+            return
+
```
zmedico, line 320: The _validate_deps method has parse_soname_deps calls for PROVIDES and REQUIRES that we don't want to skip for binary packages.

thesamesam:
I remember a while ago wanting to check if PORTDIR_OVERLAY being set (like `ebuild` does too) correctly triggers a force check. Need to look...

thesamesam:
Note that we could drop bbe658f7e8ad8d9457a5d00fe6b64f369bb64d30 as `_validate_deps` doesn't have any non-Force callers. I felt it made it a bit more explicit at the time, but looking back, I could go either way now.

lib/portage/dbapi/porttree.py
```diff
@@ -741,7 +748,11 @@ def async_aux_get(self, mycpv, mylist, mytree=None, myrepo=None, loop=None):
             future.set_exception(PortageKeyError(mycpv))
             return future
 
-        mydata, ebuild_hash = self._pull_valid_cache(mycpv, myebuild, mylocation)
+        # If PORTDIR_OVERLAY is set, we allow checking for modifications because
+        # this happens when ebuild(1) is pointed at another location for a repository.
+        mydata, ebuild_hash = self._pull_valid_cache(
+            mycpv, myebuild, mylocation, force=lambda: "PORTDIR_OVERLAY" in os.environ
```
zmedico, line 754: The force parameter should be a plain bool rather than a lambda. Maybe we don't need the force parameter here though, since RepoConfigLoader already sets volatile to True here:
```python
                            # If PORTDIR_OVERLAY is set, we have to require volatile,
                            # because it'll break changes e.g. with ebuild(1) or
                            # development in a local repository with the same repo_name.
                            if k == "volatile" and portdir_overlay:
                                v = True
```

lib/portage/dbapi/porttree.py
```diff
@@ -645,6 +646,12 @@ def _pull_valid_cache(self, cpv, ebuild_path, repo_path):
                 # EAPI from _parse_eapi_ebuild_head, we disregard cache entries
                 # for unsupported EAPIs.
                 continue
+
+            # egencache for example needs to be able to invalidate and force
+            # fresh entries.
+            if not force and not repo.volatile:
+                break
```
zmedico, line 653: It's not really clear which auxdb we've trusted here. Maybe we should only trust pregen_auxdb since that's the one that's distributed with the repository. However, it's expected for pregen_auxdb to contain stale metadata if the repository inherits eclasses from another repository that have changed.

PR #963, opened 2022-12-23
lib/portage/process.py: warn on process spawn with large environment

Flowdalic:
We have discussed that such a warning may be sensible to make people aware of the "large environment" issue. This is a first shot at it. I wonder if emitting the warning unconditionally, as opposed to only if an ebuild bash is spawned, is sensible, or if it may cause to much noise. Then again, first tests shows that most environments are far smaller than 96 KiB, and if you above the 96 KiB warning limit, then some proactive action may be necessary anway.

### DIFF
--- a/lib/portage/process.py
+++ b/lib/portage/process.py
@@ -6,6 +6,7 @@
 import atexit
 import errno
 import fcntl
+import logging
 import multiprocessing
 import platform
 import signal
@@ -15,6 +16,9 @@ import sys
 import traceback
 import os as _os
 
+from dataclasses import dataclass
+from functools import lru_cache
+
 from portage import os
 from portage import _encodings
 from portage import _unicode_encode
@@ -22,7 +26,7 @@ import portage
 
 portage.proxy.lazyimport.lazyimport(
     globals(),
-    "portage.util:dump_traceback,writemsg",
+    "portage.util:dump_traceback,writemsg,writemsg_level",
 )
 
 from portage.const import BASH_BINARY, SANDBOX_BINARY, FAKEROOT_BINARY
@@ -241,6 +245,37 @@ def cleanup():
     pass
 
 
+@dataclass(frozen=True)
+class EnvStats:
+    env_size: int
+    env_largest_name: str
+    env_largest_size: int
+
+
+def calc_env_stats(env) -> EnvStats:
+    @lru_cache(1024)
+    def encoded_length(s):
+        return len(os.fsencode(s))
+
+    env_size = 0
+    env_largest_name = None
+    env_largest_size = 0
+    for env_name, env_value in env.items():
+        env_name_size = encoded_length(env_name)
+        env_value_size = encoded_length(env_value)
+        # Add two for '=' and the terminating null byte.
+        total_size = env_name_size + env_value_size + 2
+        if total_size > env_largest_size:
+            env_largest_name = env_name
+            env_largest_size = total_size
+        env_size += total_size
+
+    return EnvStats(env_size, env_largest_name, env_largest_size)
+
+
+env_too_large_warnings = 0
+
+
 def spawn(
     mycommand,
     env=None,
@@ -261,6 +296,7 @@ def spawn(
     unshare_mount=False,
     unshare_pid=False,
     cgroup=None,
+    warn_on_large_env=False,
 ):
     """
     Spawns a given command.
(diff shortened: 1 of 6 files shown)


lib/portage/process.py
```diff
@@ -325,6 +326,31 @@ def spawn(
 
     env = os.environ if env is None else env
 
+    env_size = 0
+    env_largest_name = None
+    env_largest_size = 0
+    for env_name, env_value in env.items():
+        env_name_size = len(env_name)
+        env_value_size = len(env_value)
+        total_size = env_name_size + env_value_size
```
mgorny, line 335: Shouldn't you be counting `=` and null terminators too?

Flowdalic: Probably. I wonder if it would even be more accurate to convert the strings to UTF-8 bytes and count the length (if this is the right encoding). Right now, it is probably an approximation, but still sufficient to perform its use case. That said, I am always happy to improve on that. And adding `+ 2` to the `total_size` calculation seems trivial.

mgorny:
Yeah, I suppose encoding it would make the result more correct. Perhaps the encoding used by current locale would be more correct (with `replace` error handler, I guess). I'm a little worried about performance — though I guess you don't want to repeat the warning a lot anyway, so you'd probably have to figure out how to limit the number of times the check is run.

floppym:
> Yeah, I suppose encoding it would make the result more correct. Perhaps the encoding used by current locale would be more correct (with `replace` error handler, I guess). I'm a little worried about performance — though I guess you don't want to repeat the warning a lot anyway, so you'd probably have to figure out how to limit the number of times the check is run.

Internally, the subprocess module uses the `os.fsencode()` function to convert from unicode strings to bytes when preparing the environment of the child process.

thesamesam:
I like the concept of this, provided we cache it so we don't keep checking. It's fine if we're off by a bit. The main concern is it being initially too large.

Flowdalic:
I've added a cache for the encoding operation and added logic to emit the warning three times at most.

thesamesam:
Why do we need to show the warning more than once?

floppym:
Having thought about it, adding this warning feels a bit weird to me.

If the environment is too large, calling exec will fail with errno equal to E2BIG. We could just catch that and write a friendly error message.

If the environment is small enough to not cause exec to fail, we are warning the user about a possible problem that isn't immediately relevant.

I guess this warning is aimed at ebuilds where the environment is between 96k and the actual kernel limit? I suppose it might be helpful to catch ebuilds as their environment slowly grows due to additional files being added to SRC_URI.

Flowdalic:
> If the environment is too large, calling exec will fail with errno equal to E2BIG. We could just catch that and write a friendly error message.

The motivation was to provide users with a better idea why `exec` failed and to help finding the culprit by naming the largest environment variable.

Your comment made me reconsider the current approach. I also do not like that it causes a warning when there is a chance that nothing will break.

So what about only printing a message with environment diagnostics if `exec` failed with `E2BIG`?

floppym:
> So what about only printing a message with environment diagnostics if exec failed with E2BIG?

That sounds like a good plan to me.

floppym:
In case it isn't obvious, when I say "exec", I mean the underlying `execve` syscall. That's used by Python's `os.execve`, or any of the `subprocess` functions that execute a program.

A failure due to a large environment will raise `OSError`, with the `errno` member set to `E2BIG`.

Flowdalic:
See https://github.com/gentoo/portage/pull/978 for the "print diag message if exec returned E2BIG" approach.

Personally, I currently like PR #978 more than this approach. However, one could argue also for using this approach, because
1. it may be a good idea to warn developers of the 128 KiB limit is approached (we may enable this check could only for developers, e.g. via sys-app/portage's gentoo-dev USE flag).
2. even though I believe the limit is currently 128 KiB on all (current) Linux systems, that may not be true in the future, so just a reactive action if E2BIG is encountered may not be the best option

thesamesam:
What do we want to do here now #978 is merged?

I agree this warning might still be useful as an opt-in (e.g. gentoo-dev).

Flowdalic:
> I agree this warning might still be useful as an opt-in (e.g. gentoo-dev).

First untested draft pushed.

Flowdalic:
Tested locally and ready for review.

thesamesam:
Any chance of a test for this, even if primitive?

Flowdalic:
> Any chance of a test for this, even if primitive?

Test added. From my side, this is now ready to get merged.

lib/portage/process.py
```diff
@@ -241,6 +245,37 @@ def cleanup():
...
+        total_size = env_name_size + env_value_size + 2
+        if total_size > env_largest_size:
+            env_largest_name = env_name
+            env_largest_size = total_size
+        env_size += total_size
+
+    return EnvStats(env_size, env_largest_name, env_largest_size)
+
+
+env_to_large_warnings = 0
```
thesamesam, line 276: Typo:
```suggestion
env_too_large_warnings = 0
```
(and propagate the fixed spelling)

thesamesam (approved):
lgtm with that var renamed, thanks for persevering

EDIT: ... and CI fixed, I suppose :smile:

lib/portage/process.py
Flowdalic: fixed

Flowdalic:
> EDIT: ... and CI fixed, I suppose smile

Yep, juicy green here :)

PR #965, opened 2022-12-27
Rework signal handling in entry scripts

floppym:
Introduce a new exception SignalInterrupt which inherits from KeyboardInterrupt and adds a 'signum' member. When a signal is received, raise SignalInterrupt.

At the end of the script, catch KeyboardInterrupt, and look for the signum member. Reset the signal handler to SIG_DFL, and re-raise the signal to kill the process.

This ensures that the invoking shell sees that we got killed by a signal
instead of calling exit.

Alternative approach to #964. Best reviewed with whitespace changes hidden.

Bug: https://bugs.gentoo.org/887817

### DIFF
--- a/bin/ebuild
+++ b/bin/ebuild
@@ -2,434 +2,441 @@
 # Copyright 1999-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-import argparse
-import signal
-import sys
-import textwrap
-
-# This block ensures that ^C interrupts are handled quietly.
-try:
-
-    def exithandler(signum, _frame):
-        signal.signal(signal.SIGINT, signal.SIG_IGN)
-        signal.signal(signal.SIGTERM, signal.SIG_IGN)
-        sys.exit(128 + signum)
-
-    signal.signal(signal.SIGINT, exithandler)
-    signal.signal(signal.SIGTERM, exithandler)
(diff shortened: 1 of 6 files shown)


thesamesam (approved):
I find this latest revision pretty easy to follow and verify it's doing the right thing.

I'd still like @zmedico to double check it's okay based on the historical context for the old behaviour though.

zmedico (approved):

floppym:
In further testing, I found that this technique does not work very well for SIGPIPE: Python is unable to raise the SignalInterrupt exception while writing to a broken pipe on stdout.
```
% emerge --help | true
Exception ignored in: <module 'threading' from '/usr/lib/python3.11/threading.py'>
  File "/usr/lib/python3.11/threading.py", line 1534, in _shutdown
    def _shutdown():

  File "/home/floppym/src/portage/bin/emerge", line 24, in signal_interrupt
    raise SignalInterrupt(signum)
SignalInterrupt: 13
Exception ignored in sys.unraisablehook: <built-in function unraisablehook>
  File "/home/floppym/src/portage/bin/emerge", line 24, in signal_interrupt
    raise SignalInterrupt(signum)
SignalInterrupt: 13
```
I'm going to switch SIGPIPE back SIG_DFL directly to avoid this issue.

PR #966, opened 2022-12-29
portage.util: Add C implementation of Whirlpool hash

mgorny:
Add an extension module providing Whirlpool hash based on the reference C implementation.  This is necessary since Python hashlib no longer provides Whirlpool by default with OpenSSL 3, and the Python implementation used by Portage is unusably slow.


zmedico (approved):

floppym (approved):
Tested locally, and it seems to work well.

Without the C implementation:
```
ebuild systemd-252.4.ebuild digest  295.93s user 0.11s system 99% cpu 4:56.32 total
```
With the C implementation:
```
ebuild systemd-252.4.ebuild digest  2.03s user 0.10s system 99% cpu 2.129 total
```
The output is identical.

thesamesam:
Big thanks @mgorny and thanks to you both for reviewing (and so quickly)!

PR #967, opened 2022-12-30
Whirlpool use cleanup + warning

mgorny:
Bunch of fixes to:

1. Test both C and Python implementations of Whirlpool separately.
2. Consider only the Python implementation to be unaccelerated (i.e. don't skip Whirlpool hashes if we're using the C implementation).
3. Emit a `UserWarning` when the Python implementation is used, to have them expect very slow hashing and help finding the few ebuilds using that hash.


zmedico (approved):

thesamesam:
Missing signoff in first commit!

mgorny:
Fixed.

PR #968, opened 2022-12-31
git.py: fix "is-shallow-repository" check by stripping \n

Flowdalic:
The output of git commands has a trailing newline character which we usually strip.

Fixes: 92f8d27e0345 ("sync: git: only clobber changes if repository is marked as non-volatile")

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -11,6 +11,8 @@ Bug fixes:
   Whirlpool is considered deprecated within Portage and we recommend that
   repository maintainers remove it from `metadata/layout.conf`!
 
+* Fix "is shallow git repository" check.
+
 portage-3.0.42 (2022-12-26)
 --------------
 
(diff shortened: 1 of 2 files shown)


thesamesam:
Shouldn't this be `Fixes:` for da8e7d193563a2b0c88a523083b2a3466ff4526e at least in addition?

Flowdalic:
> Shouldn't this be `Fixes:` for [da8e7d1](https://github.com/gentoo/portage/commit/da8e7d193563a2b0c88a523083b2a3466ff4526e) at least in addition?

It appears you are right. I am sorry for that. It was  not my intention to falsely attribute the faulty commit.

thesamesam:
No bother, I didn't think it was at all, I just wanted to make sure I wasn't missing a mistake I made. Thanks!

PR #969, opened 2022-12-31
_doebuild_path: simplify logic used to set PATH

floppym:
Remove logic to incoporate PREROOTPATH and ROOTPATH. I'm not sure where PREROOTPATH was ever used, and ROOTPATH has been deprecated in baselayout for a while.

Remove logic to add hard-coded directories like
${EPREFIX}/{,usr{,/local}/{sbin,bin}. Adding these paths is unnecessary if env.d or the calling environment have a valid PATH setting.

Add logic to ignore PATH from the calling environment if PATH is set in env.d. This ensures that packages can update PATH by installing files in /etc/env.d and this will work without having to restart Portage with an updated environment.

Bug: https://bugs.gentoo.org/607696
Bug: https://bugs.gentoo.org/693308
Bug: https://bugs.gentoo.org/888543

### DIFF
--- a/lib/portage/package/ebuild/doebuild.py
+++ b/lib/portage/package/ebuild/doebuild.py
@@ -284,20 +284,8 @@ def _doebuild_path(settings, eapi=None):
     if portage_bin_path[0] != portage.const.PORTAGE_BIN_PATH:
         # Add a fallback path for restarting failed builds (bug 547086)
         portage_bin_path.append(portage.const.PORTAGE_BIN_PATH)
-    prerootpath = [x for x in settings.get("PREROOTPATH", "").split(":") if x]
-    rootpath = [x for x in settings.get("ROOTPATH", "").split(":") if x]
-    rootpath_set = frozenset(rootpath)
-    overrides = [
-        x for x in settings.get("__PORTAGE_TEST_PATH_OVERRIDE", "").split(":") if x
-    ]
 
-    prefixes = []
-    # settings["EPREFIX"] should take priority over portage.const.EPREFIX
-    if portage.const.EPREFIX != settings["EPREFIX"] and settings["ROOT"] == os.sep:
-        prefixes.append(settings["EPREFIX"])
-    prefixes.append(portage.const.EPREFIX)
-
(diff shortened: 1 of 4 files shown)


lib/portage/package/ebuild/doebuild.py
```diff
@@ -317,24 +305,15 @@ def _doebuild_path(settings, eapi=None):
...
-                path.append(x_abs)
 
-    path.extend(rootpath)
+    # If PATH is set in env.d, ignore PATH from the calling environment.
+    # This allows packages to update our PATH as they get installed.
+    if "PATH" in settings.configdict["env.d"]:
+        settings.configdict["env"].pop("PATH", None)
+
+    if "PATH" in settings:
+        path.extend(settings["PATH"].split(":"))
```
zmedico, line 315: This looks like it could lead to feedback, causing unbounded growth of the PATH setting with duplicate values.

floppym: It does seem that way, especially if env.d is empty and we skip line 312.

However, I wasn't able to cause PATH to be duplicated, even after commenting out line 311 and 312. I assume that the _doebuild_path function simply isn't called more than once with the same config instance.

I could add a failsafe If you think it makes sense. Maybe something like `settings["__doebuild_path"] = "done"`, and make `_doebuild_path` a noop if it finds that setting.

floppym: I changed the code to avoid adding duplicate entries. That's probably less error-prone.

thesamesam: I think I'd feel better if we had one, to be robust to future changes. Or a test to try check that duplicates get filtered if it doesn't already?

floppym: > However, I wasn't able to cause PATH to be duplicated, even after commenting out line 311 and 312.

The PATH was actually getting duplicated during the "merge" phase, so I didn't see it in my ebuild-based tests. The revised code with duplicate elimination should fix that.

zmedico (approved):

PR #970, opened 2023-01-01
make.globals: default-on FEATURES=pkgdir-index-trusted

ajakk:
Bug: https://bugs.gentoo.org/889300

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,9 @@
 portage-3.0.52 (UNRELEASED)
 --------------
 
+Breaking changes:
+* FEATURES=pkgdir-index-trusted is now on by default.
+
 Features:
 * bintree: Add new API member (invalid_paths) to allow gentoolkit to later
   clean up invalid binpkgs (bug #900224).
--- a/cnf/make.globals
+++ b/cnf/make.globals
@@ -78,7 +78,7 @@ FEATURES="assume-digests binpkg-docompress binpkg-dostrip binpkg-logs
           binpkg-multi-instance buildpkg-live
           config-protect-if-modified distlocks ebuild-locks
           fixlafiles ipc-sandbox merge-sync multilib-strict
-          network-sandbox news parallel-fetch pid-sandbox
+          network-sandbox news parallel-fetch pkgdir-index-trusted pid-sandbox
           preserve-libs protect-owned qa-unresolved-soname-deps
           sandbox sfperms strict
           unknown-features-warn unmerge-logs unmerge-orphans userfetch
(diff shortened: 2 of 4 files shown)


floppym:
Regarding the test failures, I'm guessing the ResolverPlayground class does not generate a valid package index.

ajakk:
Ah, missed your comment, but thanks! It *seems* simple to fix it to generate the Packages file when generating the binpkgs in the ResolverPlayground:

```
--- a/lib/portage/tests/resolver/ResolverPlayground.py
+++ b/lib/portage/tests/resolver/ResolverPlayground.py
@@ -17,6 +17,7 @@ from portage.const import (
 )
 from portage.process import find_binary
 from portage.dep import Atom, _repo_separator
+from portage.dbapi.bintree import binarytree
 from portage.package.ebuild.config import config
 from portage.package.ebuild.digestgen import digestgen
 from portage._sets import load_default_config
@@ -397,6 +398,9 @@ class ResolverPlayground:
             else:

+            bintree = binarytree(pkgdir=self.pkgdir, settings=self.settings)
+            bintree.populate()
+
     def _create_installed(self, installed):
         for cpv in installed:
             a = Atom("=" + cpv, allow_repo=True)
```

But, there's a [short-circuit in the `binarytree` class](https://github.com/gentoo/portage/blob/master/lib/portage/dbapi/bintree.py#L826) that prevents the index from being regenerated when `pkgdir-index-trusted` is in `FEATURES`:

```
            update_pkgindex = self._populate_local(
                reindex="pkgdir-index-trusted" not in self.settings.features
            )
```

I've added another boolean parameter to that method to make it possible to force the reindex despite our new default `FEATURES`. Does this sound like a sane fix?

mattst88 (approved):
I've been using this since it was added and have run into no issues.

thesamesam:
Asking Zac to take a look just in case there's any other locations we've missed. I hit that snag before with the immutable repo stuff.

thesamesam:
Can you rebase + add NEWS (call it a breaking change), then good to go?

thesamesam:
I fear that we should avoid this with getbinpkg/any remote indexes, and only do this for local indexes, until indexes are signed (https://bugs.gentoo.org/877241)...

zmedico (approved):

zmedico:
> I fear that we should avoid this with getbinpkg/any remote indexes, and only do this for local indexes, until indexes are signed (https://bugs.gentoo.org/877241)...

This feature only affects the local index. Reindexing a remote index is not really feasible since we don't have a mechanism to fetch the remote package metadata unless we fetch the whole package. In lib/portage/getbinpkg.py there's an old implementation that fetched remote package metadata, but we should not support it anymore at this point (it only supported xpak format).

thesamesam:
Thank you for the explanation Zac! I didn't realise that at all. Let's merge then 🎉

ajakk:
Thank you!

PR #972, opened 2023-01-05
depgraph: Print dependency resolution time

mgorny:
Print the wall clock time it took emerge to calculate the dependency graph if spinner is used (i.e. we're in verbose mode).

Screenshot_2023-01-05_16-28-19


thesamesam:
Please update NEWS.

floppym (approved):

PR #976, opened 2023-01-14
SCHEDULING_POLICY: switch to scheduler policy IDs rather than aliases.

slashbeast:
The os.SCHED_* aliases are not reliable enough, the mainline Python lacks deadline, and the pypy lacks them all together. The IDs are not going to change, at most new would arrive, so it is safe to use it as is.

By extension, the deadline policy was also added, since those originally were not present in aliases.

Closes: https://bugs.gentoo.org/867031
Signed-off-by: KARBOWSKI Piotr <slashbeast@gentoo.org>

### DIFF
--- a/cnf/make.conf.example
+++ b/cnf/make.conf.example
@@ -292,12 +292,12 @@
 #PORTAGE_IONICE_COMMAND="ionice -c 3 -p \${PID}"
 #
 # PORTAGE_SCHEDULING_POLICY allows changing the current scheduling policy. The
-# supported options are 'other', 'batch', 'idle', 'fifo' and 'round-robin'. When
-# unset, the scheduling policy remains unchanged, by default Linux uses 'other'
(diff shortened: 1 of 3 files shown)


floppym (approved):
Seems reasonable. Maybe add a comment mentioning the source of the values: `linux/sched.h`

thesamesam:
Could you also file a bug with cpython and pypy to add the missing constants, so we can drop this one day? Thanks!

slashbeast:
@floppym Good idea, added comment and rewritten history so it is still single commit.

@thesamesam I will look into chasing them, there were some bug reported about it in pypy, 5 years ago and then bumped 3 years ago so I wouldn't hold breath for them to get it sorted out anytime soon even if I poke them -- https://foss.heptapod.net/pypy/pypy/-/issues/2375

thesamesam:
Thanks! (And oh dear, thanks for finding that..)

PR #977, opened 2023-01-14
bin/quickpkg: loosen default umask

ajakk:
This default was originally changed in bug #182428 to prevent config files with secrets being leaked into binary packages that were readable by everyone, but given the default is to not include configs, it is safe to use a less restrictive umask by default. Also, warn the user if there is a possibility that secrets end up in a world-readable binpkg.

Bug: https://bugs.gentoo.org/182428
Bug: https://bugs.gentoo.org/890823

### DIFF
--- a/bin/quickpkg
+++ b/bin/quickpkg
@@ -6,6 +6,7 @@ import argparse
 import errno
 import math
 import signal
+import stat
 import subprocess
 import sys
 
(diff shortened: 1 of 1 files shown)


mattst88 (commented):
I wonder if we should use a different umask depending on `--include-unmodified-config` or not.

ajakk:
> I wonder if we should use a different umask depending on `--include-unmodified-config` or not.

Sorry for the delay, refreshing my memory here, but I'm not sure that a variable default is going to be worth the cost of having to convey this (probably confusingly) in the help text for the argument?

PR #979, opened 2023-01-15
emerge: add --onlydeps-with-ideps=<y|n> option (bug 890777)

tgbugs:
Add --onlydeps-with-ideps option in order to include install-time dependencies with --onlydeps and --onlydeps-with-rdeps=n. The dependencies that get pulled in are those that are necessary for emerge --nodeps to succeed when run after the equivalent --onlydeps. The default --onlydeps --onlydeps-with-rdeps=n behavior is unchanged.

Bug: https://bugs.gentoo.org/890777

### DIFF
--- a/lib/_emerge/main.py
+++ b/lib/_emerge/main.py
@@ -162,6 +162,7 @@ def insert_optional_args(args):
         "--jobs": valid_integers,
         "--keep-going": y_or_n,
         "--load-average": valid_floats,
+        "--onlydeps-with-ideps": y_or_n,
         "--onlydeps-with-rdeps": y_or_n,
         "--package-moves": y_or_n,
         "--quiet": y_or_n,
@@ -573,8 +574,12 @@ def parse_opts(tmpcmdline, silent=False):
             + "Emerge will ignore matching binary packages. ",
             "action": "append",
         },
+        "--onlydeps-with-ideps": {
+            "help": "modify interpretation of dependencies to include IDEPEND",
+            "choices": true_y_or_n,
+        },
         "--onlydeps-with-rdeps": {
-            "help": "modify interpretation of depedencies",
+            "help": "modify interpretation of dependencies",
             "choices": true_y_or_n,
         },
         "--rebuild-exclude": {
@@ -671,7 +676,7 @@ def parse_opts(tmpcmdline, silent=False):
             "action": "store",
         },
         "--root-deps": {
-            "help": "modify interpretation of depedencies",
+            "help": "modify interpretation of dependencies",
             "choices": ("True", "rdeps"),
         },
         "--search-index": {
--- a/lib/portage/tests/resolver/test_onlydeps_ideps.py
+++ b/lib/portage/tests/resolver/test_onlydeps_ideps.py
@@ -1,0 +1,172 @@
+# Copyright 2023 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+from portage.tests import TestCase
+from portage.tests.resolver.ResolverPlayground import (
+    ResolverPlayground,
+    ResolverPlaygroundTestCase,
+)
+
+
+class OnlydepsIdepsTestCase(TestCase):
+    def testOnlydepsIdepsEAPI7(self):
+        ebuilds = {
+            "dev-libs/A-1": {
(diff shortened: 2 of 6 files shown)


tgbugs:
This seems to be working as expected when I test manually, but for some reason the tests I added do not pass. Possibly because the test machinery is not aware of IDEPEND or something like that?

floppym:
Wow, that was fast!

Some administrative notes:

- We don't use X-Gentoo-Bug{,-url} anymore. Replace those with `Bug: https://bugs.gentoo.org/890777`.
- Please review our[ copyright policy](https://www.gentoo.org/glep/glep-0076.html) and add Signed-off-by to your commit(s).

tgbugs:
Fixed the issues with the commit and signed off. Still need help on the failing tests though I suspect.

thesamesam:
1. Please fix the linting issues (see CI).
2. I think ResolverPlayground defaults to EAPI 0. Try specifying EAPI 8 for the relevant tests (see e.g. lib/portage/tests/resolver/test_merge_order.py).

tgbugs:
I've pushed the linting changes. I'm still fighting with the tests. Changing the EAPI does seem to produce correct results when outside the test harness, but still causes issues inside it.

tgbugs:
Alright. After getting the issue with the EAPI sorted out it was easier to see that `"default": "n"` was having no effect on the actual default value and that the test in depgraph.py needs check for "n" or None. Now the tests pass and the behavior is as expected.

lib/_emerge/main.py
```diff
@@ -573,6 +574,10 @@ def parse_opts(tmpcmdline, silent=False):
             + "Emerge will ignore matching binary packages. ",
             "action": "append",
         },
+        "--onlydeps-with-ideps": {
+            "help": "modify interpretation of depedencies to include IDEPEND",
```
thesamesam, line 578:
```suggestion
            "help": "modify interpretation of dependencies to include IDEPEND",
```

lib/_emerge/main.py
```diff
@@ -573,6 +574,10 @@ def parse_opts(tmpcmdline, silent=False):
...
+            "choices": true_y_or_n,
+        },
         "--onlydeps-with-rdeps": {
             "help": "modify interpretation of depedencies",
```
thesamesam, line 582:
```suggestion
            "help": "modify interpretation of dependencies",
```

lib/portage/tests/resolver/test_onlydeps_ideps.py
```diff
@@ -0,0 +1,93 @@
+# Copyright 2017 Gentoo Foundation
```
thesamesam, line 1: It's not 2017, just do 2023 Gentoo Authors instead.

lib/portage/tests/resolver/test_onlydeps_ideps.py
```diff
@@ -0,0 +1,93 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+from portage.tests import TestCase
+from portage.tests.resolver.ResolverPlayground import (
+    ResolverPlayground,
+    ResolverPlaygroundTestCase,
+)
+
+
+class OnlydepsIdepsTestCase(TestCase):
```
thesamesam, line 11: Can you add another set of tests where it's EAPI 7 to verify it makes no difference to the results?

It's partly about testing your change but also about providing a bit better EAPI/IDEPEND coverage.

tgbugs: Yep. Will roll it in.

lib/_emerge/main.py
tgbugs: This one caught a few others that probably came from the same source and have been copy/pasted around.

tgbugs:
Anything else needed on my end for this?

zmedico:
Please update man/emerge.1 to include the new option.

tgbugs:
Added the changes to `man/emerge.1`.

zmedico (approved):

floppym (approved):

thesamesam:
Please update NEWS.

thesamesam:
Thank you!

PR #980, opened 2023-01-19
GPKG: Terminate writing if space is full

syu-nya:
Closes: https://bugs.gentoo.org/891391


thesamesam:
Thank you!

PR #981, opened 2023-01-19
Changed default action for emaint target sync to 'auto'.

nyback:
When using 'sync' module of emaint the default option is '--check' when none is provided which spits out an ERROR about '--check' not being an option for module 'sync'.
```
terminal ~ # emaint sync

ERROR: module 'sync' does not have option '--check'

sync module options:
-A, --allrepos (sync module only): -A, --allrepos  Sync all repos that have a sync-url defined
-a, --auto    (sync module only): -a, --auto  Sync auto-sync enabled repos only
-r, --repo    (sync module only): -r, --repo  Sync the specified repo
```

My idea was to change the default action for the 'sync' module to '--auto'. To me '--auto' seems to be the most useful default option for the 'sync' module since I would imagine that is the most common use case and presumably most users typing 'emaint sync' would be looking for this outcome. 

I'm not sure if I missed something but if defaulting to '--auto' is not good for some reason then perhaps changing the error message to something roughly along the lines of "please use an option" when using the 'sync' module without an option would be a good idea. 

This is my first time creating a pull request so by all means feel free to critique any mistakes. I've already tried to correct a few.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,8 @@ Bug fixes:
 
 * portage(5): List volatile option as a separate entry (bug #888585).
 
+* emaint: Changed default action for emaint target sync to 'auto'.
+
 portage-3.0.44 (2023-01-15)
 --------------
 
--- a/lib/portage/emaint/main.py
+++ b/lib/portage/emaint/main.py
@@ -210,10 +210,16 @@ def emaint_main(myargv):
             long_action = opt.long.lstrip("-")
 
     if long_action is None:
-        # print("DEBUG: long_action is None: setting to 'check'")
-        long_action = "check"
-        func = check_opt.func
-        status = check_opt.status
+        if args[0] == "sync":
+            # print("DEBUG: long_action is None: setting to 'auto'")
+            long_action = "auto"
+            func = "auto_sync"
+            status = "Syncing %s"
+        else:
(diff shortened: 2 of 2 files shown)


thesamesam:
It's ok to force push to branches used for PRs, by the way, to fix any issues.

This looks like a neat idea, although I'm wondering if I'm missing something about why it wasn't done originally.

nyback:
> It's ok to force push to branches used for PRs, by the way, to fix any issues.

I was a bit worried if changing a commit with force would mess something up by changing the hash of the original commit and thereby cause some kind of error with the pointer from the pull request.

thesamesam:
Understandable - but don't worry over that. gh handles it fine.

thesamesam:
I like it. Could you update NEWS?

nyback:
Sure, I'll get on that.

nyback:
Hey, I wasn't quite sure if I should add it as another bugfix or as a feature update. I added it on the bug fixes list for now since it kind of fixes a bug or at least removes an odd error. I can switch it if you think adding it as a feature update is better since it doesn't actually have a bug report to reference.

thesamesam:
I think it's probably more of a bug fix because it's a bit disturbing as a user when you hit this, but if you want to call it a feature so it's maybe more noticeable, that works for me too :)

Thanks a bunch for this!

PR #988, opened 2023-02-16
emerge: add --update-if-installed

thesamesam:
This adds a new emerge option '--update-if-installed'.

The use case for such an option is as follows:
- User finds out libfoo-1.2 is buggy.
- They want to upgrade all their systems if libfoo is installed.
- They don't want to install libfoo if it's not already installed.

  Unfortunately, --update fails this last point, hence the need for a new option.

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -1,4 +1,4 @@
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2023 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 import errno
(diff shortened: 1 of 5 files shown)


lib/_emerge/depgraph.py
```diff
@@ -5018,6 +5018,21 @@ def _resolve(self, myfavorites):
                     pkg, existing_node = self._select_package(
                         myroot, atom, onlydeps=onlydeps
                     )
+
+                    # Is the package installed (at any version)?
+                    if pkg:
+                        package_is_installed = False
+                        if "update_if_installed" in self._dynamic_config.myparams:
+                            for any_match in self._iter_match_pkgs_any(
```
zmedico, line 5026: We can used the `_iter_match_pkgs` method here with pkg_type "installed".

thesamesam: oh, nice!

zmedico (approved):

thesamesam:
Thanks Zac!

PR #989, opened 2023-02-16
tests: news: refactor, add cases

thesamesam:
This would've caught the various news regressions we had over the last year.

### DIFF
--- a/lib/portage/tests/news/test_NewsItem.py
+++ b/lib/portage/tests/news/test_NewsItem.py
@@ -139,6 +139,7 @@ class NewsItemTestCase(TestCase):
 
         item = self._processItem(str(tmpItem))
         try:
+            self.assertTrue(item.isValid())
             self.assertTrue(
                 item.isRelevant(self.vardb, self.settings, self.profile),
                 msg=f"Expected {tmpItem} to be relevant, but it was not!",
@@ -151,6 +152,7 @@ class NewsItemTestCase(TestCase):
 
         try:
             item = self._processItem(str(tmpItem))
+            self.assertTrue(item.isValid())
             self.assertTrue(
                 item.isRelevant(self.vardb, self.settings, self.profile),
                 msg=f"Expected {tmpItem} to be relevant, but it was not!",
@@ -163,6 +165,7 @@ class NewsItemTestCase(TestCase):
 
         try:
             item = self._processItem(str(tmpItem))
+            self.assertTrue(item.isValid())
             self.assertTrue(
                 item.isRelevant(self.vardb, self.settings, self.profile),
                 msg=f"Expected {tmpItem} to be relevant, but it was not!",
--- a/lib/portage/tests/news/test_NewsItem.py
+++ b/lib/portage/tests/news/test_NewsItem.py
@@ -11,18 +11,16 @@ from tempfile import mkstemp
 from dataclasses import dataclass
 from string import Template
 from typing import Optional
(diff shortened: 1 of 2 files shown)


lib/portage/tests/news/test_NewsItem.py
```diff
@@ -1,109 +1,347 @@
...
+                ]
+            }
+        )
+        self.assertTrue(item.isValid())
+        self.assertTrue(
+            item.isRelevant(self.vardb, self.settings, self.profile),
+            msg=f"Expected {item} to be relevant, but it was not!",
+        )
+
+        # What about several packages and we have all of them installed?
```
rndxelement, line 269: For `Display-If-Installed` you check the case in which you have all the packages installed, but for `Display-If-Profile` and `Display-If-Keyword` you don't check that case, which seems a bit inconsistent. Is there are reason for that?

lib/portage/tests/news/test_NewsItem.py
```diff
@@ -1,109 +1,347 @@
...
+                ]
+            }
+        )
+        self.assertTrue(item.isValid())
+        self.assertTrue(
+            item.isRelevant(self.vardb, self.settings, self.profile),
+            msg=f"Expected {item} to be relevant, but it was not!",
+        )
+
+        # What if we have a newer version of the listed package?
```
rndxelement, line 286: Is it maybe worth it to also check the case in which the required version is newer than our installed version?

lib/portage/tests/news/test_NewsItem.py
```diff
@@ -1,109 +1,347 @@
...
+
+        # What if several keywords are listed and we match none of them?
+        item = self._createNewsItem({"display_if_keyword": ["amd64", "~hppa"]})
+        self.assertTrue(item.isValid())
+        self.assertFalse(
+            item.isRelevant(self.vardb, self.settings, self.profile),
+            msg=f"Expected {item} to be irrelevant, but it was relevant!",
+        )
+
+        # What if the ~keyword (testing) keyword is listed but we're keyword (stable)?
```
rndxelement, line 335: Maybe it's worth it to also add the case in which we're testing and the stable keyword is listed in the news item?

rndxelement (commented):
Greetings, the news item test rework looks great! I came across a few questions and thought that I might share them through this review

lib/portage/tests/news/test_NewsItem.py
thesamesam: I don't think we can do that for Display-If-Profile (one can't have multiple selected), or am I missing something?

Good point on keywords. We should also add some tests for multiple constraints (mixed DIF) to ensure they interact correctly.

lib/portage/tests/news/test_NewsItem.py
thesamesam: Sure, why not? Adding it in.

lib/portage/tests/news/test_NewsItem.py
thesamesam: Good idea, done!

lib/portage/tests/news/test_NewsItem.py
rndxelement: Yes, you are right, I missed that multiple profiles don't make sense.

Having mixed constraints tested sounds good.

thesamesam:
Thanks for the review @rndxelement!

PR #991, opened 2023-02-21
mergeme: Don't overwrite files if the content matches

gcarq:
Uses `filecmp.cmp(..., shallow=False)` to compare file contents and doesn't replace them if they are equal. This results in less disk churn and helps to keep filesystem snapshots as small as possible.

Closes: https://bugs.gentoo.org/722270

Below are some benchmarks on btrfs with CoW enabled. I will do some more benchmarks with ext4 over the next days, so until then this is a draft get some comments and suggestions.

There seems to be no regression for CoW file systems, but a small performance improvement, especially for large files.

#### 1) Large files with PORTAGE_TMPDIR on tmpfs
To get a more realistic use case I first emerged an older version of linux-firmware and timed the update.

**portage-3.0.44:**

```
# export PORTAGE_TMPDIR="/tmp"
# emerge -1v =sys-kernel/linux-firmware-20220310 && time emerge -1v =sys-kernel/linux-firmware-20230210
...
real    0m55.446s
user    1m16.033s
sys     0m26.503s
```

**mergeme-filecmp branch:**

```
# export PORTAGE_TMPDIR="/tmp"
# bin/emerge -1v =sys-kernel/linux-firmware-20220310 && time bin/emerge -1v =sys-kernel/linux-firmware-20230210
...
real    0m44.532s
user    1m11.348s
sys     0m17.755s
```

#### 2) Large files with PORTAGE_TMPDIR on the same disk

**portage-3.0.44:**

```
# export PORTAGE_TMPDIR="/var/tmp"
# emerge -1v =sys-kernel/linux-firmware-20220310 && time emerge -1v =sys-kernel/linux-firmware-20230210
...
real    0m55.697s
user    1m14.798s
sys     0m26.915s
```

**mergeme-filecmp branch:**

```
# export PORTAGE_TMPDIR="/var/tmp"
# bin/emerge -1v =sys-kernel/linux-firmware-20220310 && time bin/emerge -1v =sys-kernel/linux-firmware-20230210
...
real    0m47.278s
user    1m15.744s
sys     0m18.229s
```

#### 3) Lots of small files with PORTAGE_TMPDIR on tmpfs
papirus-icon-theme provides a lot of small files which might be useful as benchmark to represent most use cases.

**portage-3.0.44:**

```
# export PORTAGE_TMPDIR="/tmp"
# time emerge -1v =x11-themes/papirus-icon-theme-20230104
...
real    2m1.555s
user    1m35.551s
sys     0m55.645s
```

**mergeme-filecmp branch:**

```
# export PORTAGE_TMPDIR="/tmp"
# time bin/emerge -1v =x11-themes/papirus-icon-theme-20230104
...
real    1m57.153s
user    1m34.715s
sys     0m50.894s
```

#### 4) Lots of small files with PORTAGE_TMPDIR on the same disk

**portage-3.0.44:**

```
# export PORTAGE_TMPDIR="/var/tmp"
# time emerge -1v =x11-themes/papirus-icon-theme-20230104
...
real    2m7.899s
user    1m33.625s
sys     1m1.591s
```

**mergeme-filecmp branch:**

```
# export PORTAGE_TMPDIR="/var/tmp"
# time bin/emerge -1v =x11-themes/papirus-icon-theme-20230104
...
real    2m3.284s
user    1m33.874s
sys     0m58.363s
```

### DIFF
--- a/lib/portage/dbapi/vartree.py
+++ b/lib/portage/dbapi/vartree.py
@@ -6267,9 +6267,10 @@ class dblink:
         if mydmode is None or not stat.S_ISREG(mydmode) or mymode != mydmode:
             return True
 
-        excluded_xattrs = self.settings.get("PORTAGE_XATTR_EXCLUDE", "")
-        if not _cmpxattr(mysrc, mydest, exclude=excluded_xattrs):
-            return True
+        if "xattr" in self.settings.features:
+            excluded_xattrs = self.settings.get("PORTAGE_XATTR_EXCLUDE", "")
+            if not _cmpxattr(mysrc, mydest, exclude=excluded_xattrs):
+                return True
 
         return not filecmp.cmp(mysrc, mydest, shallow=False)
 
--- a/lib/portage/dbapi/vartree.py
+++ b/lib/portage/dbapi/vartree.py
@@ -3,6 +3,8 @@
 
 __all__ = ["vardbapi", "vartree", "dblink"] + ["write_contents", "tar_contents"]
 
+import filecmp
+
 import portage
 
 portage.proxy.lazyimport.lazyimport(
@@ -5800,28 +5802,33 @@ class dblink:
                 # whether config protection or not, we merge the new file the
                 # same way.  Unless moveme=0 (blocking directory)
                 if moveme:
-                    # Create hardlinks only for source files that already exist
-                    # as hardlinks (having identical st_dev and st_ino).
-                    hardlink_key = (mystat.st_dev, mystat.st_ino)
+                    # only replace the existing file if it differs, see #722270
+                    already_merged = os.path.exists(mydest)
+                    if already_merged and filecmp.cmp(mysrc, mydest, shallow=False):
+                        zing = "==="
+                    else:
+                        # Create hardlinks only for source files that already exist
+                        # as hardlinks (having identical st_dev and st_ino).
+                        hardlink_key = (mystat.st_dev, mystat.st_ino)
 
-                    hardlink_candidates = self._hardlink_merge_map.get(hardlink_key)
-                    if hardlink_candidates is None:
-                        hardlink_candidates = []
-                        self._hardlink_merge_map[hardlink_key] = hardlink_candidates
+                        hardlink_candidates = self._hardlink_merge_map.get(hardlink_key)
+                        if hardlink_candidates is None:
+                            hardlink_candidates = []
+                            self._hardlink_merge_map[hardlink_key] = hardlink_candidates
 
-                    mymtime = movefile(
-                        mysrc,
-                        mydest,
-                        newmtime=thismtime,
-                        sstat=mystat,
-                        mysettings=self.settings,
-                        hardlink_candidates=hardlink_candidates,
-                        encoding=_encodings["merge"],
-                    )
-                    if mymtime is None:
-                        return 1
-                    hardlink_candidates.append(mydest)
-                    zing = ">>>"
+                        mymtime = movefile(
+                            mysrc,
+                            mydest,
+                            newmtime=thismtime,
+                            sstat=mystat,
+                            mysettings=self.settings,
+                            hardlink_candidates=hardlink_candidates,
+                            encoding=_encodings["merge"],
+                        )
+                        if mymtime is None:
+                            return 1
+                        hardlink_candidates.append(mydest)
+                        zing = ">>>"
 
                     try:
                         self._merged_path(mydest, os.lstat(mydest))
--- a/lib/portage/dbapi/vartree.py
+++ b/lib/portage/dbapi/vartree.py
@@ -3,8 +3,6 @@
 
 __all__ = ["vardbapi", "vartree", "dblink"] + ["write_contents", "tar_contents"]
 
-import filecmp
-
 import portage
 
 portage.proxy.lazyimport.lazyimport(
@@ -34,7 +32,7 @@ portage.proxy.lazyimport.lazyimport(
     "portage.util.env_update:env_update",
     "portage.util.install_mask:install_mask_dir,InstallMask,_raise_exc",
     "portage.util.listdir:dircache,listdir",
-    "portage.util.movefile:movefile",
+    "portage.util.movefile:movefile,_cmpxattr",
     "portage.util.path:first_existing,iter_parents",
     "portage.util.writeable_check:get_ro_checker",
     "portage.util._xattr:xattr",
@@ -95,6 +93,7 @@ from ._ContentsCaseSensitivityManager import ContentsCaseSensitivityManager
 
 import argparse
 import errno
+import filecmp
 import fnmatch
 import functools
 import gc
@@ -5803,10 +5802,7 @@ class dblink:
                 # same way.  Unless moveme=0 (blocking directory)
                 if moveme:
                     # only replace the existing file if it differs, see #722270
-                    already_merged = os.path.exists(mydest)
-                    if already_merged and filecmp.cmp(mysrc, mydest, shallow=False):
-                        zing = "==="
-                    else:
+                    if self._needs_move(mysrc, mydest, mymode, mydmode):
                         # Create hardlinks only for source files that already exist
                         # as hardlinks (having identical st_dev and st_ino).
                         hardlink_key = (mystat.st_dev, mystat.st_ino)
@@ -5829,6 +5825,8 @@ class dblink:
                             return 1
                         hardlink_candidates.append(mydest)
                         zing = ">>>"
+                    else:
+                        zing = "==="
 
                     try:
                         self._merged_path(mydest, os.lstat(mydest))
@@ -6253,6 +6251,26 @@ class dblink:
 
         finally:
             self.unlockdb()
+
+    def _needs_move(self, mysrc, mydest, mymode, mydmode):
+        """
+        Checks whether the given file at |mysrc| needs to be moved to |mydest| or if
+        they are identical.
+
+        Takes file mode and extended attributes into account.
+        Should only be used for regular files.
+        """
+        if not os.path.exists(mydest):
+            return True
+
+        if mymode != mydmode:
+            return True
+
+        excluded_xattrs = self.settings.get("PORTAGE_XATTR_EXCLUDE", "")
+        if not _cmpxattr(mysrc, mydest, exclude=excluded_xattrs):
+            return True
+
+        return not filecmp.cmp(mysrc, mydest, shallow=False)
 
 
 def merge(
--- a/lib/portage/dbapi/vartree.py
+++ b/lib/portage/dbapi/vartree.py
@@ -5544,6 +5544,8 @@ class dblink:
                     destmd5,
                     mydest_link,
                 )
+                if protected and moveme:
+                    mydmode = None
 
             zing = "!!!"
             if not moveme:
@@ -5584,6 +5586,7 @@ class dblink:
                         msg.append("")
                         self._eerror("preinst", msg)
                         mydest = newdest
+                        mydmode = None
 
                 # if secondhand is None it means we're operating in "force" mode and should not create a second hand.
                 if (secondhand is not None) and (not os.path.exists(myrealto)):
@@ -5797,6 +5800,7 @@ class dblink:
                     msg.append("")
                     self._eerror("preinst", msg)
                     mydest = newdest
+                    mydmode = None
 
                 # whether config protection or not, we merge the new file the
                 # same way.  Unless moveme=0 (blocking directory)
@@ -6260,10 +6264,7 @@ class dblink:
         Takes file mode and extended attributes into account.
         Should only be used for regular files.
         """
-        if not os.path.exists(mydest):
-            return True
-
-        if mymode != mydmode:
+        if mydmode is None or not stat.S_ISREG(mydmode) or mymode != mydmode:
             return True
 
         excluded_xattrs = self.settings.get("PORTAGE_XATTR_EXCLUDE", "")
--- a/lib/portage/dbapi/vartree.py
+++ b/lib/portage/dbapi/vartree.py
@@ -5422,6 +5422,7 @@ class dblink:
         srcroot = normalize_path(srcroot).rstrip(sep) + sep
         destroot = normalize_path(destroot).rstrip(sep) + sep
         calc_prelink = "prelink-checksums" in self.settings.features
+        ignore_mtime = "ignore-mtime" in self.settings.features
 
         protect_if_modified = (
             "config-protect-if-modified" in self.settings.features
@@ -5830,6 +5831,13 @@ class dblink:
                         hardlink_candidates.append(mydest)
                         zing = ">>>"
                     else:
+                        if not ignore_mtime:
+                            mymtime = thismtime if thismtime is not None else mymtime
+                            try:
+                                os.utime(mydest, ns=(mymtime, mymtime))
+                            except OSError:
+                                # utime can fail here with EPERM
+                                pass
                         zing = "==="
 
                     try:
--- a/lib/portage/util/movefile.py
+++ b/lib/portage/util/movefile.py
@@ -105,6 +105,35 @@ def _copyxattr(src, dest, exclude=None):
             )
 
 
+def _cmpxattr(src, dest, exclude=None):
+    """
+    Compares extended attributes between |src| and |dest| and returns True
+    if they are equal or xattrs are not supported, False otherwise
+    """
+    try:
+        src_attrs = xattr.list(src)
+        dest_attrs = xattr.list(dest)
+    except OSError as e:
+        if e.errno != OperationNotSupported.errno:
+            raise
+        return True
+
+    if src_attrs:
+        if exclude is not None and isinstance(src_attrs[0], bytes):
+            exclude = exclude.encode(_encodings["fs"])
+    exclude = _get_xattr_excluder(exclude)
+
+    src_attrs = {attr for attr in src_attrs if not exclude(attr)}
+    dest_attrs = {attr for attr in dest_attrs if not exclude(attr)}
+    if src_attrs != dest_attrs:
+        return False
+
+    for attr in src_attrs:
+        if xattr.get(src, attr) != xattr.get(dest, attr):
+            return False
+    return True
+
+
 def movefile(
     src,
     dest,
--- a/man/make.conf.5
+++ b/man/make.conf.5
@@ -533,6 +533,12 @@ If your GPG is auto unlocked on login, you do not need this.
 .B icecream
 Enable portage support for the icecream package.
 .TP
+.B ignore\-mtime
+Do not update mtime if the target file is equal. This can be useful on some
+filesystems to better utilize features like snapshots and data deduplication,
+however this violates the preservation of file modification times as stipulated
+in the PMS.
+.TP
 .B installsources
 Install source code into /usr/src/debug/${CATEGORY}/${PF} (also see
 \fBsplitdebug\fR). This feature works only if debugedit is installed, CFLAGS
--- a/lib/portage/const.py
+++ b/lib/portage/const.py
@@ -159,6 +159,7 @@ SUPPORTED_FEATURES = frozenset(
         "getbinpkg",
         "gpg-keepalive",
         "icecream",
+        "ignore-mtime",
         "installsources",
         "ipc-sandbox",
         "keeptemp",


thesamesam:
I wonder how the benchmarks contrast with @Chewi's suggested patch at https://github.com/chewi/portage/commit/f7e3e3753ab3a808865c2e351898b3054fd638ec.

Thanks for looking into this!

gcarq:
Thanks for the quick response. I will try with the suggested patch. I dismissed it too early, because after looking into it I figured the md5 would be calculated based on the path and not the file content itself, which is actually not the case for regular files.

gcarq:
I cherry-picked https://github.com/chewi/portage/commit/f7e3e3753ab3a808865c2e351898b3054fd638ec it seems to perform better than the filecmp version for the same benchmarks (see below). Its also fair to assume this doesn't regress on non CoW filesystems as the hash has been already calculated beforehand as being pointed out.

#### 1) Large files with PORTAGE_TMPDIR on tmpfs
```
real    0m42.785s
user    1m4.546s
sys     0m17.000s
```

#### 2) Large files with PORTAGE_TMPDIR on the same disk
```
real    0m42.703s
user    1m4.639s
sys     0m15.930s
```

#### 3) Lots of small files with PORTAGE_TMPDIR on tmpfs
```
real    1m48.801s
user    1m24.563s
sys     0m48.230s
```

#### 4) Lots of small files with PORTAGE_TMPDIR on the same disk
```
real    1m54.730s
user    1m24.774s
sys     0m55.111s
```

floppym (changes requested):
I gave this a test locally, and found a problem:

If the contents of the installed file have been modified, but the size and mtime are the same, Portage will fail to replace the file on subsequent merges.

I assume this is because it is using the md5sum from /var/db/pkg/.../CONTENTS, which is out of sync with the actual installed file.

This sort of thing might realistically happen if a bit flips on a storage device that isn't completely reliable.

mattst88:
> This sort of thing might realistically happen if a bit flips on a storage device that isn't completely reliable.

I'm certainly guilty of modifying installed python scripts instead of patching and re-emerging.

chewi:
> If the contents of the installed file have been modified, but the size and mtime are the same, Portage will fail to replace the file on subsequent merges.
> 
> I assume this is because it is using the md5sum from /var/db/pkg/.../CONTENTS, which is out of sync with the actual installed file.

This is what Zac was concerned about. 

> This sort of thing might realistically happen if a bit flips on a storage device that isn't completely reliable.

He didn't mention this, and I hadn't considered it, but I'm not sure whether that's something we should worry about. I've never personally witnessed it. It'll happen whether a package is updated or not.

> I'm certainly guilty of modifying installed python scripts instead of patching and re-emerging.

That's not a problem, as long as you don't somehow do it without changing the mtime, which would be weird.

floppym:
I'm not comfortable treating /var/db/pkg/.../CONTENTS as an authoritative on installed files. We should compare the file to be merged with what is actually on disk, not what CONTENTS has.

thesamesam:
It feels a shame to throw away the information given in some cases, we know that CONTENTS will be correct (think an otherwise read-only filesystem for example), but I get the point.

gcarq:
> If the contents of the installed file have been modified, but the size and mtime are the same, Portage will fail to replace the file on subsequent merges.
> 
> I assume this is because it is using the md5sum from /var/db/pkg/.../CONTENTS, which is out of sync with the actual installed file.
> 
> This sort of thing might realistically happen if a bit flips on a storage device that isn't completely reliable.

That is a very good point, now I see the initial concern about this.
I will revert it to the initial proposed patch which should not run into this issue, because the mtime is ignored and if the size differs they are immediately considered not equal, otherwise file contents will be checked.

gcarq:
As promised I did some tests of the latest commit 339d5be on a different system with ext4 and dm-crypt on a NVMe SSD (Toshiba XG3 Series THNSN5256GPU7).
This is the same patch as used for the initial numbers in the first comment, so they should be comparable for CoW vs Non-CoW.

Overall it seems always faster with 339d5be except for testcase 4) when the portage tmpdir is on the same disk, but even then the performance penalty is negligible.

#### 1) Large files with PORTAGE_TMPDIR on tmpfs

**portage-3.0.44:**

```
# export PORTAGE_TMPDIR="/tmp"
# emerge -1v =sys-kernel/linux-firmware-20220310 && time emerge -1v =sys-kernel/linux-firmware-20230210
...
real    0m46.741s
user    0m34.952s
sys     0m12.595s
```

**mergeme-filecmp branch:**

```
# export PORTAGE_TMPDIR="/tmp"
# bin/emerge -1v =sys-kernel/linux-firmware-20220310 && time bin/emerge -1v =sys-kernel/linux-firmware-20230210
...
real    0m37.551s
user    0m31.964s
sys     0m6.678s
```

#### 2) Large files with PORTAGE_TMPDIR on the same disk

**portage-3.0.44:**

```
# export PORTAGE_TMPDIR="/var/tmp"
# emerge -1v =sys-kernel/linux-firmware-20220310 && time emerge -1v =sys-kernel/linux-firmware-20230210
...
real    0m46.602s
user    0m34.800s
sys     0m13.045s

```

**mergeme-filecmp branch:**

```
# export PORTAGE_TMPDIR="/var/tmp"
# bin/emerge -1v =sys-kernel/linux-firmware-20220310 && time bin/emerge -1v =sys-kernel/linux-firmware-20230210
...
real    0m40.115s
user    0m32.003s
sys     0m7.198s
```

#### 3) Lots of small files with PORTAGE_TMPDIR on tmpfs
papirus-icon-theme provides a lot of small files which might be useful as benchmark to represent most use cases.


**portage-3.0.44:**

```
# export PORTAGE_TMPDIR="/tmp"
# time emerge -1v =x11-themes/papirus-icon-theme-20230104
...
real    1m26.034s
user    1m10.425s
sys     0m34.236s
```

**mergeme-filecmp branch:**

```
# export PORTAGE_TMPDIR="/tmp"
# time bin/emerge -1v =x11-themes/papirus-icon-theme-20230104
...
real    1m21.028s
user    1m8.786s
sys     0m30.296s
```

#### 4) Lots of small files with PORTAGE_TMPDIR on the same disk


**portage-3.0.44:**

```
# export PORTAGE_TMPDIR="/var/tmp"
# time emerge -1v =x11-themes/papirus-icon-theme-20230104
...
real    1m24.852s
user    1m7.403s
sys     0m34.894s
```

**mergeme-filecmp branch:**

```
# export PORTAGE_TMPDIR="/var/tmp"
# time bin/emerge -1v =x11-themes/papirus-icon-theme-20230104
...
real    1m25.832s
user    1m8.388s
sys     0m34.312s
```

lib/portage/dbapi/vartree.py
```diff
@@ -5800,28 +5802,33 @@ def mergeme(
                 # whether config protection or not, we merge the new file the
                 # same way.  Unless moveme=0 (blocking directory)
                 if moveme:
-                    # Create hardlinks only for source files that already exist
-                    # as hardlinks (having identical st_dev and st_ino).
-                    hardlink_key = (mystat.st_dev, mystat.st_ino)
+                    # only replace the existing file if it differs, see #722270
+                    already_merged = os.path.exists(mydest)
+                    if already_merged and filecmp.cmp(mysrc, mydest, shallow=False):
```
zmedico, line 5807: I see that filecmp.cmp does not compare xattrs, which could be an issue given that movefile copies xattrs.

lib/portage/dbapi/vartree.py
```diff
@@ -5800,28 +5802,33 @@ def mergeme(
                 # whether config protection or not, we merge the new file the
                 # same way.  Unless moveme=0 (blocking directory)
                 if moveme:
-                    # Create hardlinks only for source files that already exist
-                    # as hardlinks (having identical st_dev and st_ino).
-                    hardlink_key = (mystat.st_dev, mystat.st_ino)
+                    # only replace the existing file if it differs, see #722270
+                    already_merged = os.path.exists(mydest)
```
zmedico, line 5806: Instead of  ` os.path.exists(mydest)` which adds a redundant stat call, we might use something like this:
```
already_merged = mydstat is not None and stat.S_ISREG(mydstat.st_mode)
```

gcarq: The testcase `portage.tests.emerge.test_config_protect.ConfigProtectTestCase.testConfigProtect` fails when relying on `mydstat`. I'm not sure if the test needs some adaptions or if I didn't handle some special case
```
  File "/usr/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/usr/lib/python3.11/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/michi/Workspace/portage/lib/portage/util/_async/ForkProcess.py", line 162, in _bootstrap
    sys.exit(self._run())
             ^^^^^^^^^^^
  File "/home/michi/Workspace/portage/lib/portage/dbapi/_MergeProcess.py", line 237, in _run
    rval = mylink.merge(
           ^^^^^^^^^^^^^
  File "/home/michi/Workspace/portage/lib/portage/dbapi/vartree.py", line 1906, in wrapper
    return f(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/michi/Workspace/portage/lib/portage/dbapi/vartree.py", line 6037, in merge
    retval = self.treewalk(
             ^^^^^^^^^^^^^^
  File "/home/michi/Workspace/portage/lib/portage/dbapi/vartree.py", line 5022, in treewalk
    rval = self._merge_contents(srcroot, destroot, cfgfiledict)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/michi/Workspace/portage/lib/portage/dbapi/vartree.py", line 5322, in _merge_contents
    if self.mergeme(
       ^^^^^^^^^^^^^
  File "/home/michi/Workspace/portage/lib/portage/dbapi/vartree.py", line 5809, in mergeme
    if already_merged and filecmp.cmp(mysrc, mydest, shallow=False):
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/filecmp.py", line 54, in cmp
    s2 = _sig(os.stat(f2))
              ^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/portage.runtests.vv22c1ju/tmpzol2s1nj/etc/A/._cfg0000_regular_a'

```

lib/portage/dbapi/vartree.py
gcarq: I tackled the xattrs handling and file mode comparison in 3ff0d5a. The xattr comparison has been done in a similar way as `_copyxattr(...)` to respect `PORTAGE_XATTR_EXCLUDE`.

lib/portage/util/movefile.py
```diff
@@ -105,6 +105,35 @@ def _copyxattr(src, dest, exclude=None):
...
+        return True
+
+    if src_attrs:
+        if exclude is not None and isinstance(src_attrs[0], bytes):
+            exclude = exclude.encode(_encodings["fs"])
+    exclude = _get_xattr_excluder(exclude)
+
+    src_attrs = tuple(attr for attr in src_attrs if not exclude(attr))
+    dest_attrs = tuple(attr for attr in dest_attrs if not exclude(attr))
+    if src_attrs != dest_attrs:
```
floppym, line 128: Is it possible for `xattr.list` to return the attr names in a different order? If so, it might be better to use a set instead of a tuple.

thesamesam:
Could you rebase (at your leisure) for fixed CI? Thanks! (An issue has been fixed on master.)

lib/portage/util/movefile.py
gcarq: Good point. I'm not sure if there exists a fixed order, but to be safe I changed it to a set with 81a3ec3

lib/portage/dbapi/vartree.py
thesamesam: @zmedico Could you take a look when you get a chance?

lib/portage/util/movefile.py
floppym: I'm not seeing that change in the referenced commit or in this PR currently.

floppym: Sorry, I was looking for the word "set" and missed that you used the more compact brace notation.

lib/portage/dbapi/vartree.py
```diff
@@ -6247,6 +6252,26 @@ def _quickpkg_dblink(self, backup_dblink, background, logfile):
...
+        """
+        if not os.path.exists(mydest):
+            return True
+
+        if mymode != mydmode:
+            return True
+
+        excluded_xattrs = self.settings.get("PORTAGE_XATTR_EXCLUDE", "")
+        if not _cmpxattr(mysrc, mydest, exclude=excluded_xattrs):
+            return True
```
zmedico, line 6271: We need to make the xattr logic conditional on `"xattr" in self.settings.features`.

lib/portage/dbapi/vartree.py
zmedico: It works like this:
```diff
diff --git a/lib/portage/dbapi/vartree.py b/lib/portage/dbapi/vartree.py
index 317cf327a..13b201a3b 100644
--- a/lib/portage/dbapi/vartree.py
+++ b/lib/portage/dbapi/vartree.py
@@ -5544,6 +5544,8 @@ class dblink:
                     destmd5,
                     mydest_link,
                 )
+                if protected and moveme:
+                    mydmode = None
 
             zing = "!!!"
             if not moveme:
@@ -5584,6 +5586,7 @@ class dblink:
                         msg.append("")
                         self._eerror("preinst", msg)
                         mydest = newdest
+                        mydmode = None
 
                 # if secondhand is None it means we're operating in "force" mode and should not create a second hand.
                 if (secondhand is not None) and (not os.path.exists(myrealto)):
@@ -5797,6 +5800,7 @@ class dblink:
                     msg.append("")
                     self._eerror("preinst", msg)
                     mydest = newdest
+                    mydmode = None
 
                 # whether config protection or not, we merge the new file the
                 # same way.  Unless moveme=0 (blocking directory)
@@ -6260,7 +6264,7 @@ class dblink:
         Takes file mode and extended attributes into account.
         Should only be used for regular files.
         """
-        if not os.path.exists(mydest):
+        if not (mydmode is not None and stat.S_ISREG(mydmode)):
             return True
 
         if mymode != mydmode:

```

thesamesam:
@zmedico @mgorny @floppym I'll likely commit this after tagging .46 unless any objections, so we can all give it some testing?

gcarq:
I did a world rebuild with this PR on top of `sys-apps/portage-3.0.46` and verified the file integrity with gentoolkit, so far everything looks fine.

The md5sums seem to match besides some config and temporary files:
```
$ equery -N check -o '*' 2>&1 | grep 'incorrect MD5sum'
!!! /etc/rkhunter.conf has incorrect MD5sum
!!! /usr/share/texmf-dist/tex/generic/config/language.dat has incorrect MD5sum
!!! /usr/share/texmf-dist/tex/generic/config/language.dat.lua has incorrect MD5sum
!!! /usr/share/texmf-dist/tex/generic/config/language.def has incorrect MD5sum
!!! /usr/lib/python3.10/site-packages/twisted/plugins/dropin.cache has incorrect MD5sum
!!! /usr/lib/python3.11/site-packages/twisted/plugins/dropin.cache has incorrect MD5sum
!!! /usr/lib64/qt5/mkspecs/qconfig.pri has incorrect MD5sum
!!! /usr/lib64/qt5/mkspecs/qmodule.pri has incorrect MD5sum
!!! /usr/include/qt5/Gentoo/gentoo-qconfig.h has incorrect MD5sum
!!! /var/lib/texmf/web2c/metafont/mf.base has incorrect MD5sum
!!! /var/lib/texmf/web2c/metafont/mf.log has incorrect MD5sum
!!! /var/lib/texmf/web2c/luahbtex/luahbtex.fmt has incorrect MD5sum
!!! /var/lib/texmf/web2c/luahbtex/luahbtex.log has incorrect MD5sum
!!! /var/lib/texmf/web2c/pdftex/etex.fmt has incorrect MD5sum
!!! /var/lib/texmf/web2c/pdftex/etex.log has incorrect MD5sum
!!! /var/lib/texmf/web2c/pdftex/pdfetex.fmt has incorrect MD5sum
!!! /var/lib/texmf/web2c/pdftex/pdfetex.log has incorrect MD5sum
!!! /var/lib/texmf/web2c/pdftex/pdftex.fmt has incorrect MD5sum
!!! /var/lib/texmf/web2c/pdftex/pdftex.log has incorrect MD5sum
!!! /var/lib/texmf/web2c/tex/tex.fmt has incorrect MD5sum
!!! /var/lib/texmf/web2c/tex/tex.log has incorrect MD5sum
!!! /var/lib/texmf/web2c/luatex/dviluatex.fmt has incorrect MD5sum
!!! /var/lib/texmf/web2c/luatex/dviluatex.log has incorrect MD5sum
!!! /var/lib/texmf/web2c/luatex/luatex.fmt has incorrect MD5sum
!!! /var/lib/texmf/web2c/luatex/luatex.log has incorrect MD5sum
!!! /etc/ufw/ufw.conf has incorrect MD5sum
!!! /etc/proxychains.conf has incorrect MD5sum
!!! /etc/systemd/system.conf has incorrect MD5sum
!!! /etc/systemd/user.conf has incorrect MD5sum
!!! /etc/systemd/resolved.conf has incorrect MD5sum
!!! /etc/systemd/timesyncd.conf has incorrect MD5sum
!!! /etc/systemd/journald.conf has incorrect MD5sum
!!! /etc/default/grub has incorrect MD5sum
!!! /etc/dracut.conf has incorrect MD5sum
!!! /etc/locale.gen has incorrect MD5sum
!!! /etc/host.conf has incorrect MD5sum
!!! /usr/lib64/gtk-3.0/3.0.0/immodules.cache has incorrect MD5sum
```

However I'm not sure why `/usr/include/qt5/Gentoo/gentoo-qconfig.h` changed as this doesn't seem temporary to me.

Also I'm not sure how to proceed with the [mtime check](https://github.com/gentoo/gentoolkit/blob/master/pym/gentoolkit/equery/check.py#L155) in gentoolkit as it no longer seems that useful if this has been merged.

```
$ equery -N check -o '*' 2>&1 | grep 'mtime' | wc -l
196467
```

thesamesam:
wrt gentoo-qconfig.h: could you diff the old/new? fwiw, it's created in kind of a weird way and is a bit fragile, so I'm not super surprised.

Not sure what to do about gentoolkit, I guess it is indeed not super useful now. @dol-sen (see the whole thread first)?

gcarq:
I merged it again w/o this PR but there is no difference in the old / new `gentoo-qconfig.h` and its still being reported as md5sum mismatch. Just checked and on another machine its also reported as mismatch, so this might be another issue as you suggested.

floppym:
I didn't realize we were storing mtime in CONTENTS. This change ensures CONTENTS will not match the live filesystem after package rebuilds, which sucks a little bit.

Overall, I don't really like this change; it feels like over-optimization to me. However, I don't really feel like I have a strong argument against it, so I'm going to sit this one out.

zmedico:
> I didn't realize we were storing mtime in CONTENTS. This change ensures CONTENTS will not match the live filesystem after package rebuilds, which sucks a little bit.

I think it should always record the mtime from the live filesystem in CONTENTS, so that they match.

Maybe we could have an option to update the mtime of files even when their content has not changed. If we don't update the mtime, then it's clearly a violation of PMS preservation of modification times:

https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-146001r1

I think it might be okay to violate PMS if the user wants to, but it would be nice to provide an option for PMS compliance.

Flowdalic:
> I think it might be okay to violate PMS if the user wants to, but it would be nice to provide an option for PMS compliance.

That is what I am thinking too: configurable behavior seems to be what we the right way to go. There appear to be some, including me, who think that this feature has the potential for a noticeable improvement and hence who want to experiment with it. While others seem to be skeptical of the whole concept.

mgorny:
There's also the option of copying mtimes without rewriting the contents. I think that'd be a better option since it would guarantee that outputs are newer than sources.

gcarq:
I'm also gravitating towards the option to update the mtimes without rewriting the file contents per default to keep the disk churn low and having a feature flag to leave the mtime unchanged if wanted

man/make.conf.5
```diff
@@ -533,6 +533,10 @@ If your GPG is auto unlocked on login, you do not need this.
 .B icecream
 Enable portage support for the icecream package.
 .TP
+.B ignore\-mtime
+Do not update mtime if the target file is equal. This can be useful on some
+filesystems to better utilize features like snapshots and data deduplication.
```
zmedico, line 538: We should probably note here that this feature violates the PMS stipulation about [preservation of file modification times](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-146001r1).

gcarq: Good point. I added a note with 29d4c2e

thesamesam:
I think I'm happy now. @zmedico?

zmedico (approved):
LGTM. Thanks!

thesamesam:
Thank you everyone!

thesamesam:
(And I hope to see you again!)

PR #992, opened 2023-02-22
sync: git: fix code comment regarding clobbering the repository

Flowdalic:
The logic the comment stated was simply inverted. The actual implementation is correct.


thesamesam:
Could you check if the NEWS mention from a few releases ago is correct, and if not, fix that too? Thanks!

Flowdalic:
Good point, the logic was also inverted in the NEWS item.

PR #993, opened 2023-02-22
sync: git: report error if volatile repository

Flowdalic:
Previously portage would not report a merge_cmd error if the repository was volatile. This could potentially be the cause of bug #895526, however if it unclear, at the time of writing this, if the repository in the bug is a volatile one.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -33,6 +33,9 @@ Bug fixes:
 
 * tests: news: significantly improved test coverage (bug #889330).
 
+* git: also report sync errors for volatile repos. Portage would previously simply
+  report success if an volatile repository failed to sync
+
 portage-3.0.44 (2023-01-15)
 --------------
 
--- a/lib/portage/sync/modules/git/git.py
+++ b/lib/portage/sync/modules/git/git.py
@@ -376,15 +376,16 @@ class GitSync(NewBase):
             **self.spawn_kwargs,
         )
 
-        if exitcode != os.EX_OK and not self.repo.volatile:
-            # HACK - sometimes merging results in a tree diverged from
-            # upstream, so try to hack around it
-            # https://stackoverflow.com/questions/41075972/how-to-update-a-git-shallow-clone/41081908#41081908
-            exitcode = portage.process.spawn(
-                f"{self.bin_command} reset --hard refs/remotes/{remote_branch}",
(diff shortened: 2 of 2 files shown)


thesamesam:
If it might be related to the bug, use `Bug:`.

thesamesam:
I suppose we still want to deal with the bit at https://github.com/gentoo/portage/pull/993/files#diff-76e83ad764f507d6a1da5c1dfeaf5fa79ad191929660f1509f1d1b12cd903d02R358 (given it doesn't make much sense) but separately?

Flowdalic:
> If it might be related to the bug, use `Bug:`.

Right now, I don't see how it is related to the bug, given that the repo in the bug seems to be non-volatile.

> I suppose we still want to deal with the bit at https://github.com/gentoo/portage/pull/993/files#diff-76e83ad764f507d6a1da5c1dfeaf5fa79ad191929660f1509f1d1b12cd903d02R358 (given it doesn't make much sense) but separately?

Yes, baby steps please (if possible). :)

thesamesam:
I generally, on principle, think that if something is worth mentioning in a commit message, or if you did it _because_ of a bug in the course of investigating it, you should tag it, even if in the commit message you make clear you don't think it fixes that exact problem. It's related.

Flowdalic:
While this change was initiated by looking at the code while investigating the bug, I would not say it is related to the bug. At least we do not have any indication at the moment that it caused nor fixes the bug. Hence I would personally not use the `Bug:` trailer, to avoid linking the bug with the commit, causing noise at the bug, leading people on the wrong trace, or, at least wasting the time of people looking at the bug and following the link.

However if you feel strongly about it, then I am happy to add it.

thesamesam:
`Bug` is appropriate for when you're not sure but it's related. If it was worth mentioning in the commit message, it's worth tagging it properly - which is how GLEP 66 says we reference a bug. It's not about indicating that it fixes it.

In any case, it's yet another git-handle-state-issue and it specifically touches on git merge, so I think it's worth doing.

Flowdalic:
> Sorry, I'm not trying to be stubborn

No worries. Arguments where exchanged and I indicated in my previous commit that this not a hill do die on for me.

thesamesam:
thanks Flow & thanks for humouring me :)

PR #994, opened 2023-02-22
sync: git: also print merge_cmd

Flowdalic:
It felt strange that the sync operation would print the "git fetch" command, but not the command used to merge the remote's changes into the local working tree.

### DIFF
--- a/lib/portage/sync/modules/git/git.py
+++ b/lib/portage/sync/modules/git/git.py
@@ -169,7 +169,7 @@ class GitSync(NewBase):
             }
             self.spawn_kwargs["env"].update(pull_env)
 
-        if self.settings.get("PORTAGE_QUIET") == "1":
+        if quiet:
             git_cmd_opts += " --quiet"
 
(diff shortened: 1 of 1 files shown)


lib/portage/sync/modules/git/git.py
```diff
@@ -370,6 +370,8 @@ def update(self):
         if quiet:
             merge_cmd.append("--quiet")
 
+        writemsg_level(' '.join(merge_cmd) + "\n")
```
thesamesam, line 373: Maybe only if not quiet? What do you think?

Flowdalic: Sounds sensible. Should do the same for the "git fetch" message above?

thesamesam: I think so yeah

Flowdalic:
New version pushed.

thesamesam:
Thanks!

PR #1003, opened 2023-03-01
closes bug 898224

siddhanthrathod:


bin/ebuild
```diff
@@ -129,13 +128,16 @@ try:
...
         or portage.settings.get("NOCOLOR") in ("yes", "true")
+        or portage.settings.get("NO_COLOR") != ""
         or portage.settings.get("TERM") == "dumb"
         or not sys.stdout.isatty()
     ):
         portage.output.nocolor()
         portage.settings.unlock()
-        portage.settings["NOCOLOR"] = "true"
-        portage.settings.backup_changes("NOCOLOR")
+#         portage.settings["NOCOLOR"] = "true"
```
thesamesam, line 137: Please drop these commented out lines.

bin/ebuild
```diff
@@ -129,13 +128,16 @@ try:
     if not opts.color == "y" and (
         opts.color == "n"
         or portage.settings.get("NOCOLOR") in ("yes", "true")
```
thesamesam, line 130: I think I'd prefer this refactored into a function that we can call in other places too.

bin/isolated-functions.sh
```diff
@@ -443,6 +443,13 @@ case "${NOCOLOR:-false}" in
 		;;
 esac
 
+if [ -z "${NO_COLOR}" ] ; then
```
thesamesam, line 446: We're in Bash here, so [[ ]] please.

bin/portageq
```diff
@@ -1451,9 +1451,11 @@ try:
 
     def main(argv):
         argv = portage._decode_argv(argv)
-
-        nocolor = os.environ.get("NOCOLOR")
-        if nocolor in ("yes", "true"):
+        if os.environ.get("NO_COLOR"):
```
thesamesam, line 1454: If we put all of this in one function, we can add a nice comment about how we can cleanup NOCOLOR support (not NO_COLOR) in a few years.

lib/portage/_emirrordist/main.py
```diff
@@ -238,7 +238,7 @@ def emirrordist_main(args):
 
     if not sys.stdout.isatty():
         portage.output.nocolor()
-        env["NOCOLOR"] = "true"
+        env["NO_COLOR"] = "true"
```
thesamesam, line 241: good

thesamesam (changes requested):
Please also adjust the commit message to look like:
```
emerge: respect NO_COLOR

Bug: https://bugs.gentoo.org/123456 # replace the number here
Signed-off-by: ...
```

Please aldo add an entry to NEWS.

bin/ebuild
siddhanthrathod: Okay will do

bin/ebuild
siddhanthrathod: Yes that would be nice I'll do that

bin/isolated-functions.sh
siddhanthrathod: Okay I'll change it

bin/portageq
```diff
@@ -1451,9 +1451,11 @@ try:
...
+            nocolor = os.environ.get("NO_COLOR") != ""
```
ulm, line 1455: I think comparison with `""` is not correct, because `NO_COLOR` could be unset. Both `None` and `""` are logically false though:

```suggestion
            nocolor = os.environ.get("NO_COLOR")
```

(Same issue in more places below.)

siddhanthrathod: Oh sorry it should have been == right not !=

siddhanthrathod: It should have been == not != right ? I'm sorry I'll rectify it

siddhanthrathod: Wait wouldn't nocolor be "" like you mentioned above and it won't be a boolean right ? Is that an expected behaviour ?

siddhanthrathod: But in the condition if it's not unset it can be an empty str or srt of len > 0

siddhanthrathod: Should I add a condition to deal with None too ?

ulm: > Should I add a condition to deal with None too ?

I think you can check for both `None` and `""` at once, because in Python a string has boolean value true if (and only if) it exists and is non-empty. 

See here for the test in pkgcheck:
https://gitweb.gentoo.org/proj/pkgcore/pkgcheck.git/tree/src/pkgcheck/scripts/pkgcheck_scan.py#n404

siddhanthrathod: Yes I understand

bin/isolated-functions.sh
```diff
@@ -443,6 +443,13 @@ case "${NOCOLOR:-false}" in
 		;;
 esac
 
+if [[ -z "${NO_COLOR}" ]] ; then
+	__set_colors
+else
+	__unset_colors
+fi
```
ulm, line 450: I think we don't want to undo a previous `NOCOLOR=true` with an unset or empty `NO_COLOR` here.

lib/_emerge/AbstractEbuildProcess.py
```diff
@@ -390,9 +391,12 @@ def _elog(self, elog_funcname, lines):
         elog_func = getattr(elog_messages, elog_funcname)
         global_havecolor = portage.output.havecolor
         try:
-            portage.output.havecolor = self.settings.get(
-                "NOCOLOR", "false"
-            ).lower() in ("no", "false")
+            if not self.settings.get("NOCOLOR"):
+                portage.output.havecolor = self.settings.get("NO_COLOR") == ""
```
ulm, line 395: What happens if `NO_COLOR` is unset?

ulm (commented):
It would definitely make sense to collect the logic (at least for the Python part) into one function that deals with `NOCOLOR` and `NO_COLOR`.

siddhanthrathod:
Yes I'm on it

lib/_emerge/AbstractEbuildProcess.py
siddhanthrathod: I need to change it here too sorry but havecolor would be true. I had this in the back of my mind too ( then I convinced myself that it might be an expected outcome ) Should I change that ?

bin/isolated-functions.sh
siddhanthrathod: Ah yes I'll refactor it accordingly and put nocolor inside the if block of no_color. So if no_color is empty or unset then case nocolor will happen with a default of false

siddhanthrathod:
I should have made a function for it from the start things got a lot less confusing

bin/isolated-functions.sh
siddhanthrathod: Does the new implementation work ?

bin/egencache
```diff
@@ -1105,12 +1105,11 @@ try:
         if "PATH" in os.environ:
             env["PATH"] = os.environ["PATH"]
 
-        if not sys.stdout.isatty() or os.environ.get("NOCOLOR", "").lower() in (
-            "yes",
-            "true",
+        if not sys.stdout.isatty() or portage.util.no_color(
+            os.environ.get("NOCOLOR", "false").lower(), os.environ.get("NO_COLOR")
```
ulm, line 1109: Probably I am missing something, but couldn't the two `os.environ.get` calls be moved into the `no_color` function?

siddhanthrathod: but then we wouldn't be able to use it with portage.settings.get or settings.get

ulm: I see. I thought that I was missing something. :smile:

bin/isolated-functions.sh
```diff
@@ -434,14 +434,21 @@ RC_INDENTATION=''
 RC_DEFAULT_INDENT=2
 RC_DOT_PATTERN=''
 
-case "${NOCOLOR:-false}" in
+
+
+if [[ -z "${NO_COLOR}" ]] ; then
+	case "${NOCOLOR:-false}" in
```
ulm, line 440: I think the logic is correct now.

Very minor point, the quotes are not necessary in `[[ ]]` and `case`:
```suggestion
if [[ -z ${NO_COLOR} ]] ; then
	case ${NOCOLOR:-false} in
```

lib/_emerge/actions.py
```diff
@@ -2757,23 +2757,28 @@ def adjust_config(myopts, settings):
...
-            settings["NOCOLOR"] = "true"
-        settings.backup_changes("NOCOLOR")
+            # settings["NOCOLOR"] = "true"
+            settings["NO_COLOR"] = "true"
+        settings.backup_changes("NO_COLOR")
     elif settings.get("TERM") == "dumb" or not sys.stdout.isatty():
         portage.output.havecolor = 0
-        settings["NOCOLOR"] = "true"
-        settings.backup_changes("NOCOLOR")
+        # settings["NOCOLOR"] = "true"
```
thesamesam, line 2779: Why are these debugging lines here?

lib/portage/package/ebuild/_ipc/QueryCommand.py
```diff
@@ -149,9 +149,10 @@ def _elog(self, elog_funcname, lines):
         elog_func = getattr(elog_messages, elog_funcname)
         global_havecolor = portage.output.havecolor
         try:
-            portage.output.havecolor = self.settings.get(
-                "NOCOLOR", "false"
-            ).lower() in ("no", "false")
+            portage.output.havecolor = not no_color(
```
thesamesam, line 152: Can we have a function which takes self.settings as input and handles all of this? I don't want to have to keep repeating the same logic in a few places, and it's hard to read because of the double negative.

lib/portage/util/__init__.py
```diff
@@ -2006,3 +2008,9 @@ def read_ld_so_conf(path):
     rval.append("/lib")
 
     return [normalize_path(x) for x in rval if x]
+
+
+def no_color(nocolor: str, no__color: Optional[str]) -> bool:
```
thesamesam, line 2013: Why __? Can we just use a different function name to make this less confusing? "has_color"? "allow_color"?

bin/ebuild
```diff
@@ -128,14 +127,16 @@ try:
...
     if not opts.color == "y" and (
         opts.color == "n"
-        or portage.settings.get("NOCOLOR") in ("yes", "true")
+        or portage.util.no_color(portage.settings)
         or portage.settings.get("TERM") == "dumb"
         or not sys.stdout.isatty()
     ):
         portage.output.nocolor()
         portage.settings.unlock()
         portage.settings["NOCOLOR"] = "true"
```
thesamesam, line 136: Do we need to set NOCOLOR here for internal use, I wonder?

lib/_emerge/AbstractEbuildProcess.py
```diff
@@ -390,9 +391,7 @@ def _elog(self, elog_funcname, lines):
         elog_func = getattr(elog_messages, elog_funcname)
         global_havecolor = portage.output.havecolor
         try:
-            portage.output.havecolor = self.settings.get(
-                "NOCOLOR", "false"
-            ).lower() in ("no", "false")
+            portage.ouput.havecolor = not no_color(self.settings)
```
thesamesam, line 394: Typo.

lib/portage/util/__init__.py
```diff
@@ -2006,3 +2008,11 @@ def read_ld_so_conf(path):
     rval.append("/lib")
 
     return [normalize_path(x) for x in rval if x]
+
+
+def no_color(nocolor) -> bool:
```
thesamesam, line 2013: I was expecting a comment here to say we can clean up NOCOLOR support everywhere in future.

Also, 'nocolor' is not a good variable name here, it should be settings (with a type annotation)?

bin/ebuild
siddhanthrathod: Yes I think I should remove it ... as eventually support for NOCOLOR is gonna go

lib/_emerge/AbstractEbuildProcess.py
siddhanthrathod: Omg 😳

lib/portage/util/__init__.py
```diff
@@ -2006,3 +2008,12 @@ def read_ld_so_conf(path):
     rval.append("/lib")
 
     return [normalize_path(x) for x in rval if x]
+
+
+def no_color(settings: Optional[dict]) -> bool:
+    """This will enable us to safely clean up support for NOCOLOR"""
```
thesamesam, line 2014: This is still vague/missing the point, it also shouldn't be a docstring.

```suggestion
    # In several years (2026+), we can cleanup NOCOLOR support, and just support NO_COLOR.
```

lib/portage/tests/resolver/ResolverPlayground.py
```diff
@@ -587,8 +587,9 @@ def _create_profile(
             "PORTAGE_TMPDIR": os.path.join(self.eroot, "var/tmp"),
         }
 
-        if os.environ.get("NOCOLOR"):
+        if portage.util.no_color(os.environ):
             make_conf["NOCOLOR"] = os.environ["NOCOLOR"]
```
thesamesam, line 591: Why is this still here?

PR #1005, opened 2023-03-10
repository: config: absorb all exceptions when determining volatility

thesamesam:
Not ideal, but we don't have much else of a choice. It's a very small section and trying to granularly specify the relevant exceptions here is clearly a waste of time (see ref'd bugs).

Besides, the consequence is just failing-safe and assuming the repo is volatile anyway.

Bug: https://bugs.gentoo.org/899208
Bug: https://bugs.gentoo.org/900683

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,12 +1,19 @@
+portage-3.0.45.3 (UNRELEASED)
+----------------
+
+Bug fixes:
+* repository: config: Handle more error cases when determining repository
+  volatility (bug #900683).
+
 portage-3.0.45.2 (2023-03-04)
---------------
+----------------
 
 Bug fixes:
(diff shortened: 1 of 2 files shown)


thesamesam:
cc @tastytea

floppym:
I think it would suffice to catch OSError and KeyError. I'm trying to think of an error we wouldn't want to catch, but nothing comes to mind immediately.

floppym (approved):

thesamesam:
> I think it would suffice to catch OSError and KeyError. I'm trying to think of an error we wouldn't want to catch, but nothing comes to mind immediately.

I guess what makes me worried about this is that I didn't think to check KeyError originally, but maybe that's just something to accept?

(It wasn't obvious to me KeyError might be thrown, so I wonder if I'd be missing something else.)

PR #1007, opened 2023-03-20
checksum: drop STREEBOG{256,512} & pygost support

thesamesam:
checksum: drop STREEBOG{256,512} & pygost support

For STREEBOG{256,512}, we had three options that we "supported":

1. using pygost;

   pygost is unpackaged in Gentoo and appears to, as of March 2023, been yanked
   from pypi too.

   Hence not only can we test it on the distribution where we do our development
   (Gentoo), but we also can't acquire it for CI easily either.

2. using pygcrypt;

   No longer in Gentoo and the codepath for this has been disabled since 2017(!)
   because of https://bugs.gentoo.org/615620.

3. a slow fallback implementation.

   This is not something we can rely on anyway, but this is especially wrong
   given the issues we've had recently with WHIRLPOOL. See https://bugs.gentoo.org/885909.

As for whether we should support STREEBOG at all:
- Per above, we don't have a decent implementation for it available.
- We're not testing it very well.
- Jugging obscure hash algorithms has become a pain because some of them have
  ended up being marked "legacy" or removed from OpenSSL by default.
- I don't think we should have hash algorithms supported just for the sake of it.
  We're not in the business of supporting every single hash algorithm that a
  library with Python bindings out there provides.

Bug: https://bugs.gentoo.org/597736
Bug: https://bugs.gentoo.org/615620
Bug: https://bugs.gentoo.org/885909
Bug: https://bugs.gentoo.org/888884

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,54 @@
+portage-3.0.46 (UNRELEASED)
+---------------
+
+Breaking changes:
+* Drop STREEBOG{256,512} support
+
+  Support was originally added in https://bugs.gentoo.org/597736, but
+  exclusively via two modules (pygost and pygcrypt) which aren't packaged in
+  Gentoo - or seemingly any other major distribution.
+
+  We're dropping both pygost and pygcrypt compatibility: pygost has been
+  removed from pypi and pygcrypt is inactive upstream.
+
+Cleanups:
+* Drop support for pygcrypt
+
+  The codepath has been disabled since 2017 because of https://bugs.gentoo.org/615620,
+  so let's finally clean it up.
+
+  pygcrypt *prior to 2017* was used for RMD160/WHIRLPOOL/SHA3_256/SHA3_512
+  if hashlib didn't provide these *and* the sha3 module didn't either.
+
+  pygcrypt last received a release in 2017 too.
+
+* Drop support for pyblake2, sha3
+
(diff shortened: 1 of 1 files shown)


lib/portage/checksum.py
```diff
@@ -207,101 +141,26 @@ def hexdigest(self):
...
-    except ImportError:
-        pass
+        # Try to use mhash if available
+        # mhash causes GIL presently, so it gets less priority than hashlib and
+        # pycrypto. However, it might be the only accelerated implementation of
+        # WHIRLPOOL available.
+        try:
+            import mhash
+
+            for local_name, hash_name in (("RMD160", "RIPEMD160"),):
```
mgorny, line 151: I'd say you can drop RIPEMD as well.

thesamesam: From mhash (leaving pycrypto* as the provider) or entirely? I'm trying to avoid dropping things entirely just yet, but want to do that later on.

mgorny: I was thinking of mhash since it's its only use.

mgorny: Hmm, though I guess after this commit it's also only use of pycrypto.

thesamesam:
I'll merge this later today probably given CI is currently red.

PR #1012, opened 2023-03-21
testSimple: debugging flakey test

robbat2:
Per IRC, tweaks to make it easier to debug flakey behavior in testSimple.
```
testSimple (portage.tests.emerge.test_simple.SimpleEmergeTestCase.testSimple) ... emerge: superuser access is required
```

This PR should show:
1. What user the tests actually run as
2. Individual parts of `testSimple`, because there are a LOT of them, despite the name.

Signed-off-by: Robin H. Johnson <robbat2@gentoo.org>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,5 +44,7 @@ jobs:
         printf "[build_ext]\nportage_ext_modules=true" >> setup.cfg
         ./setup.py install --root=/tmp/install-root
     - name: Run tox targets for ${{ matrix.python-version }}
+      # Capture who this runs as before the run starts
       run: |
+        id
         tox -vv
--- a/lib/portage/tests/emerge/test_simple.py
+++ b/lib/portage/tests/emerge/test_simple.py
@@ -692,7 +692,7 @@ move dev-util/git dev-vcs/git
                 # triggered by python -Wd will be visible.
                 stdout = subprocess.PIPE
 
-            for args in test_commands:
+            for idx, args in enumerate(test_commands):
                 if hasattr(args, "__call__"):
                     args()
                     continue
@@ -704,21 +704,22 @@ move dev-util/git dev-vcs/git
                 else:
                     local_env = env
 
-                proc = await asyncio.create_subprocess_exec(
-                    *args, env=local_env, stderr=None, stdout=stdout
-                )
+                with self.subTest(cmd=args, i=idx):
+                    proc = await asyncio.create_subprocess_exec(
+                        *args, env=local_env, stderr=None, stdout=stdout
+                    )
 
-                if debug:
-                    await proc.wait()
-                else:
-                    output, _err = await proc.communicate()
-                    await proc.wait()
-                    if proc.returncode != os.EX_OK:
-                        portage.writemsg(output)
+                    if debug:
+                        await proc.wait()
+                    else:
+                        output, _err = await proc.communicate()
+                        await proc.wait()
+                        if proc.returncode != os.EX_OK:
+                            portage.writemsg(output)
 
-                self.assertEqual(
-                    os.EX_OK, proc.returncode, f"emerge failed with args {args}"
-                )
+                    self.assertEqual(
+                        os.EX_OK, proc.returncode, f"emerge failed with args {args}"
+                    )
         finally:
             binhost_server.__exit__(None, None, None)
             playground.cleanup()


thesamesam:
@palao Any chance you'd mind taking this, rebasing it, and splitting into multiple pytest subtests?

Right now, `lib/portage/tests/emerge/test_simple.py::SimpleEmergeTestCase::testSimple` (when run via the ebuild at least) is pretty slow with no visual feedback.

palao:
Sure. I'll have a look later

palao:
I'm looking at it now. I have two questions:

- Sorry for the stupid question, but what do you mean by "taking this"? Something like "git fetch" @robbat2's branch into mine, rebase and do the work?
- I see two options to solve the problem: 

  1. keep using subtests. That implies that we need to add another dependency for the tests: [pytest-subtests](https://github.com/pytest-dev/pytest-subtests), or
  2. parametrize the tests *à la* pytest. That will produce hopefully a cleaner result, from the point of view of pytest, but will break compatibility with runtests, I guess, unless a big effort is done to avoid it

  What do you think?

thesamesam:
> I'm looking at it now. I have two questions:
> 
>     * Sorry for the stupid question, but what do you mean by "taking this"? Something like "git fetch" @robbat2's branch into mine, rebase and do the work?

Yeah, or tbh I'd just run `pram` (`app-portage/pram`) on this PR and import it that way, then work on top of it.

> 
>     * I see two options to solve the problem:
>       
>       1. keep using subtests. That implies that we need to add another dependency for the tests: [pytest-subtests](https://github.com/pytest-dev/pytest-subtests), or
>       2. parametrize the tests _à la_ pytest. That will produce hopefully a cleaner result, from the point of view of pytest, but will break compatibility with runtests, I guess, unless a big effort is done to avoid it
>       

Given how pytest has gone so far (no complaints and seems to work well for me apart from the output issue with this test), I'd say we can just break compatibility?

palao:
I have a question.

In the current implementation there is a single test function that runs multiple portage commands in sequence. Some commands seem independent of the rest, and others seem to depend on previous ones. There are also post-checks. And preparation steps. The general pattern looks like: setup, exercise, assert, cleanup. In other words, the typical structure of a unit test. BUT in this case it applies to commands, and all the commands, as I wrote, are part of a single big test. It looks like the order is important and the shared context too.

I have rewritten almost all the tests in `test_simple.py` as a parametrized pytest test. During the whole process I was telling myself that this is not the best structure. It would be better if each command with its setup/cleanup steps was an individual test. What has prevented me to do this was only that I believe(d) that this arrangement has a reason. But as I go on, the idea that this structure is wrong becomes stronger and stronger.

And this brings me to the question, finally: Is there a reason why all the commands in `test_simple.py` must be run within the same test? Maybe due to performance? Or are they part of, for some reason, a monolithic unbreakable test?

If the test can be split in individual pieces, where do I cut? To be honest, I cannot see what is the intent of most commands, if there is a hidden wish to couple some of them or not, etc.

palao:
Ok. I'll need to think deeper about that...

In the meantime I have pushed to

https://github.com/palao/portage/tree/pytest/parametrize-testSimple

This divides the old test_simple into 184 cases.

I would make a PR now, but I can also keep working on this until it is a bit cleaner. I think the importance of these tests clearly justify more refactoring work on them. But the goal is achieved already.

Suggestions?

thesamesam:
> Ok. I'll need to think deeper about that...
> 
> In the meantime I have pushed to
> 
> [palao/portage@`pytest`/parametrize-testSimple](https://github.com/palao/portage/tree/pytest/parametrize-testSimple?rgh-link-date=2023-07-07T16%3A23%3A18Z)
> 
> This divides the old test_simple into 184 cases.
> 
> I would make a PR now, but I can also keep working on this until it is a bit cleaner. I think the importance of these tests clearly justify more refactoring work on them. But the goal is achieved already.
> 

Go for it (PR) - iterative work is no bad thing.

Also, this test is so huge that I can't really get my head around it until it's split up a bit.

thesamesam:
Also, @palao, it looks like your branch works when rebased on master, but it fails with `pytest-xdist` (e.g. `pytest -n 32`):

```
$ pytest --durations=0 -n 32 --maxfail 1
[...]
________________________________________________________________________________ test_portage_baseline[one protected file-xpak] ________________________________________________________________________________
[gw29] linux -- Python 3.12.0 /usr/bin/python3.12

async_loop = <portage.util._eventloop.asyncio_event_loop.AsyncioEventLoop object at 0x7f3200fe8380>, playground = <portage.tests.resolver.ResolverPlayground.ResolverPlayground object at 0x7f3200fe83e0>
binhost = {'dir': '/tmp/pytest-of-sam/pytest-4/popen-gw29/eprefix0/binhost', 'server': <portage.tests.util.test_socks5.AsyncHTTPServer object at 0x7f32010cca10>, 'uri': 'http://127.0.0.1:38853/binhost'}
simple_command = <function _generate_all_simple_commands.<locals>.<lambda> at 0x7f3200ee7240>

    @pytest.mark.ft
    def test_portage_baseline(async_loop, playground, binhost, simple_command):
>       async_loop.run_until_complete(
            asyncio.ensure_future(
                _async_test_simple(
                    playground,
                    binhost,
                    simple_command,
                    _METADATA_XML_FILES,
                    loop=async_loop,
                ),
                loop=async_loop,
            )
        )

lib/portage/tests/emerge/test_simple.py:61:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
lib/portage/util/_eventloop/asyncio_event_loop.py:129: in _run_until_complete
    return self._loop.run_until_complete(future)
/usr/lib/python3.12/asyncio/base_events.py:664: in run_until_complete
    return future.result()
lib/portage/tests/emerge/test_simple.py:198: in _async_test_simple
    command()
lib/portage/tests/emerge/conftest.py:631: in <lambda>
    test_commands["one protected file"] = lambda: _check_number_of_protected_files(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

must_have = 1, eroot = '/tmp/pytest-of-sam/pytest-4/popen-gw29/eprefix0/', config_protect = '/etc'

    def _check_number_of_protected_files(must_have, eroot, config_protect) -> None:
>       assert must_have == len(
            list(find_updated_config_files(eroot, shlex_split(config_protect)))
        )
E       AssertionError: assert 1 == 0
E        +  where 0 = len([])
E        +    where [] = list(<generator object find_updated_config_files at 0x7f32014ebdf0>)
E        +      where <generator object find_updated_config_files at 0x7f32014ebdf0> = find_updated_config_files('/tmp/pytest-of-sam/pytest-4/popen-gw29/eprefix0/', ['/etc'])
E        +        where ['/etc'] = shlex_split('/etc')

lib/portage/tests/emerge/conftest.py:323: AssertionError
[...]
```

palao:
These problems with xdist point to the problem I mentioned above: the tests have hidden couplings. There is no alternative that I can see apart from properly dividing the huge test into small ones. The refactor work will take some time, though. But I think it'll pay off.

thesamesam:
> These problems with xdist point to the problem I mentioned above: the tests have hidden couplings. There is no alternative that I can see apart from properly dividing the huge test into small ones. The refactor work will take some time, though. But I think it'll pay off.
> 
> PS Thank you, @thesamesam , for the test.

Yeah, I was really worried about the same. At least this seems to be the only instance of it (even though it's a rather big one).

I'm in two minds about what to do for now. I don't think xdist really gains us much for now until this test is fixed, so I'm leaning towards merging the branch as it is but just dropping xdist from CI + the ebuild for now until the refactoring is done. What do you think?

(The alternative being to wait until it's done and merge it all together, but right now, we have pretty poor visibility into what that huge test is doing so I feel like it's worth merging what we have now?)

palao:
I don't know what to say. From my perspective there is no alternative: we need a clear understanding of this code. And by *we* I mean anyone looking at the file. So I'm working to improve that code. The *side effect* of this will be that the xdist problem is solved.

But, of course, I see your point...

My intention was to finish it during this week, but I couldn't. I'll try hard to find time to solve it during the next week. If we can wait to merge until I'm done, I'd say it makes sense; to avoid bureaucracy.
But if it's important to have a working thing now, I'd say we should drop xdist and merge: we are too far to come back, but still not there, and I cannot promise a date.

palao:
After much longer than expected, I think I have a working thing. It has been reduced to 28 tests. Some of those tests can be further splitted.
I'm working in the rebase now.

thesamesam:
Wow, thank you! Please let me know if you need me.

palao:
> Wow, thank you! Please let me know if you need me.

I finished with the rebase. Could you please check if it's all right?
It was a bit long because I made a little mess with the conflicts. I think all is OK now. The only thing is a spurious commit at the end (I had renamed test_simple.py to test_baseline.py but, in the meantime, in master test_simple.py was modified, so I have removed again).

palao:
...and, BTW, there are more improvements to be done: as I said, I'd like this test to be as readable as possible. It is very important.

thesamesam:
I'm going to try finish looking at it today or tomorrow

palao:
Do you mean comments in the files? It might be a good idea,  yes

thesamesam:
yes please (for review comments of mine, I mean)

palao:
> yes please (for review comments of mine, I mean)

Sorry, but aren't comments on? How Do I enable them?

thesamesam:
Sorry, I mean it's easier to review stuff like this as a draft PR than just individual commits on your branch.

palao:
Should I just create a PR? Is there a way to relate it to this one?

thesamesam:
Yeah, just create a PR, and then in the second commit you cherry-pick from this one, add: `Closes: https://github.com/gentoo/portage/pull/1012` as a trailer above `Signed-off-by`.

palao:
As usual, not sure if I did what you meant... Just let me know.

PR #1013, opened 2023-03-22
More fstrings

xxc3nsoredxx:
This has a good bit more than last time, so if you want me to split the PR even further, just say so.

### DIFF
--- a/lib/_emerge/BinpkgExtractorAsync.py
+++ b/lib/_emerge/BinpkgExtractorAsync.py
@@ -5,7 +5,6 @@ import logging
 
 from _emerge.SpawnProcess import SpawnProcess
 import portage
-from portage.localization import _
 from portage.util.compression_probe import (
     compression_probe,
     _compressors,
@@ -21,6 +20,7 @@ from portage.binpkg import get_binpkg_format
 import signal
 import subprocess
 import tarfile
+import textwrap
 
 
 class BinpkgExtractorAsync(SpawnProcess):
@@ -72,9 +72,7 @@ class BinpkgExtractorAsync(SpawnProcess):
             decomp_cmd = None
         if decomp_cmd is None:
             self.scheduler.output(
-                "!!! %s\n"
-                % _("File compression header unrecognized: %s")
-                % self.pkg_path,
+                f"!!! File compression header unrecognized: {self.pkg_path}\n",
                 log_path=self.logfile,
                 background=self.background,
                 level=logging.ERROR,
@@ -104,15 +102,9 @@ class BinpkgExtractorAsync(SpawnProcess):
             if find_binary(decompression_binary) is None:
                 missing_package = decomp.get("package")
                 self.scheduler.output(
-                    "!!! %s\n"
-                    % _(
-                        "File compression unsupported %s.\n Command was: %s.\n Maybe missing package: %s"
-                    )
-                    % (
-                        self.pkg_path,
-                        varexpand(decomp_cmd, mydict=self.env),
-                        missing_package,
-                    ),
+                    f"!!! File compression unsupported {self.pkg_path}.\n"
+                    f" Command was: {varexpand(decomp_cmd, mydict=self.env)}.\n"
+                    f" Missing package: {missing_package}\n",
                     log_path=self.logfile,
                     background=self.background,
                     level=logging.ERROR,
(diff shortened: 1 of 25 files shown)


lib/_emerge/BinpkgExtractorAsync.py
```diff
@@ -129,26 +120,18 @@ def _xpak_start(self):
...
-            )
-            % (
-                pkg_xpak.filestat.st_size - pkg_xpak.xpaksize,
-                portage._shell_quote(self.pkg_path),
-                decomp_cmd,
-                tar_options,
-                portage._shell_quote(self.image_dir),
-                128 + signal.SIGPIPE,
-            ),
+            f"cmd0=(head -c {pkg_xpak.filestat.st_size - pkg_xpak.xpaksize} -- {portage._shell_quote(self.pkg_path)}) "
```
thesamesam, line 123: Some of the quoting here makes me a bit nervous but I don't have a better suggestion right now either.

thesamesam: @arthurzam @mgorny any ideas?

lib/_emerge/actions.py
```diff
@@ -1052,8 +1046,8 @@ def _calc_depclean(settings, trees, ldpath_mtimes, myopts, action, args_set, spi
             if not pkgs_for_cp or pkg not in pkgs_for_cp:
                 raise AssertionError(
                     "package expected in matches: "
-                    + "cp = %s, cpv = %s matches = %s"
-                    % (pkg.cp, pkg.cpv, [str(x) for x in pkgs_for_cp])
+                    f"cp = {pkg.cp}, cpv = {pkg.cpv} "
```
thesamesam, line 1049: wonder if we want a comma

lib/_emerge/actions.py
```diff
@@ -1065,8 +1059,8 @@ def _calc_depclean(settings, trees, ldpath_mtimes, myopts, action, args_set, spi
             if len(pkgs_for_cp) <= 1:
                 raise AssertionError(
                     "more packages expected: "
-                    + "cp = %s, cpv = %s matches = %s"
-                    % (pkg.cp, pkg.cpv, [str(x) for x in pkgs_for_cp])
+                    f"cp = {pkg.cp}, cpv = {pkg.cpv} "
```
thesamesam, line 1062: comma

lib/_emerge/actions.py
xxc3nsoredxx: After the `cpv = ...` bit, so that it looks like `cp =, cpv =, matches =`?

thesamesam: That's it, yes please! (Sorry for being vague)

lib/_emerge/BinpkgExtractorAsync.py
arthurzam: @thesamesam if we are nervous about quoting, we can use `shlex.quote`, but I see we are using `portage._shell_quote` in this line, so we should be fine.

Can you explain what makes you nervous (like the full code block)

arthurzam: Also, since this here is a single long string, I would use something like:

```python
textwrap.dedent(
    """\
        cmds
    """
)
```

In my opinion we have less python quotes, so easier to see the bash code.

xxc3nsoredxx: IMO, that would make it more complex/hard to read since you have to ensure all `{` get translated to `{{` (and vice versa), and it would be harder to pick out which ones are substitutions and which ones aren't.

xxc3nsoredxx: Perhaps something in between the two suggestions, where consecutive fstrings are multi-line and consecutive plain strings are multi-line?

lib/_emerge/BinpkgExtractorAsync.py
```diff
@@ -129,26 +121,30 @@ def _xpak_start(self):
...
-            % (
-                pkg_xpak.filestat.st_size - pkg_xpak.xpaksize,
-                portage._shell_quote(self.pkg_path),
-                decomp_cmd,
-                tar_options,
-                portage._shell_quote(self.image_dir),
-                128 + signal.SIGPIPE,
-            ),
+            .replace("\n", " ")
+            .strip(),
```
xxc3nsoredxx, line 147: @thesamesam @arthurzam What are your opinions on this?

xxc3nsoredxx:
First was just a rebase onto master (no content changes), second was just moving the NEWS change up under the 3.0.47 section.

lib/_emerge/BinpkgExtractorAsync.py
thesamesam: I think I can live with the ${{}} and it's more readable this way, but I wouldn't say we should necessarily do it _everywhere_.

lib/_emerge/BinpkgExtractorAsync.py
thesamesam: Coming back to this, I think it's OK as it is.

PR #1017, opened 2023-04-02
tests: fix util/test_shelve.py

xxc3nsoredxx:
Remove the `os.unlink()` call which can fail if the backend lib, eg, adds a suffix to the filename. When debugging this failed run [1], the `shelve_file` was being created as `shelve_file.db`.

From the Python docs:

    The filename parameter is the base filename for the underlying
    database.  As a side-effect, an extension may be added to the
    filename and more than one file may be created.

[1]: https://builds.sr.ht/~xxc3nsoredxx/job/966549

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -56,7 +56,8 @@ Features:
   for a time for compatibility.
 
 Bug fixes:
-* TODO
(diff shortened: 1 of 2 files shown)


thesamesam:
Thanks!

PR #1018, opened 2023-04-03
Enable automated tests on sourcehut builds

xxc3nsoredxx:
Run tests on push for Portage clones hosted on sourcehut. Currently skips testing on PyPy.

Example run: https://builds.sr.ht/~xxc3nsoredxx/job/967249

### DIFF
--- a/.builds/ci.yml
+++ b/.builds/ci.yml
@@ -1,0 +1,39 @@
+# Maintainer: Oskari Pirhonen <xxc3ncoredxx@gmail.com>
+
+image: ubuntu/jammy
+shell: true
+repositories:
+  # For more versions than just the default python3
+  deadsnakes: https://ppa.launchpadcontent.net/deadsnakes/ppa/ubuntu jammy main "BA6932366A755776"
+  # Because Ubuntu have yet to fix: https://foss.hetapod.net/pypy/pypy/-/issues/3741
+  # pypy: https://ppa.launchpadcontent.net/pypy/ppa/ubuntu jammy main "251104D968854915"
+environment:
+  PYTHON_VERSIONS:
+    - '3.7'
+    - '3.8'
+    - '3.9'
+    - '3.10'
+    - '3.11'
+    # Testing PyPy is currently broken, see bug #903709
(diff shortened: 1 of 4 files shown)


mgorny:
How many random third-party hostings are we going to support in Portage codebase?

xxc3nsoredxx:
Sourcehut is hardly random. And having multiple environments to run tests in is useful. I already fixed one broken test because the CPython was slightly different from GH's.

Then there's the whole thing with some people slowly getting tired of GH's centralization of free software on a non-free platform. Me included, and I plan on migrating my repos to sourcehut and using GH as a mirror. Gentoo GH repos are just mirrors, but if I want the automated testing, it currently has to go through GH.

Since I'm adding it, it's on me to keep it working (potentially with the help of others in the future if they use sourcehut too).

thesamesam:
It's fine with me as one alternative to GH as we're currently a bit too GH centric, but I wouldn't want to add one for every single non-GH site.

Could you add a comment at the top of the files indicating you're the maintainer?

xxc3nsoredxx:
Can do.

thesamesam:
> It's fine with me as one alternative to GH as we're currently a bit too GH centric, but I wouldn't want to add one for every single non-GH site.

(In particular, I don't feel like I'm abiding by the social contract when I tell people "yes, we accept ML submissions, but you should really do GH so we can get CI run on it", so I'm fine with this as we're giving an alternative in this instance, but wouldn't want to necessarily approve adding another third service.)

mgorny:
I disagree. The only reason we're providing GitHub config is because 1) we're officially accepting PRs on GH, and 2) we're running our CI on GH.

This is a slippery slope. Today sr.ht, tomorrow GitLab, Gitea, that fork of Gitea now that Gitea suddenly started being bad, and a dozen other incompatible CI configurations that nobody's going to keep in sync.

thesamesam:
I just said I wouldn't take a multitude of them.

xxc3nsoredxx:
Yep, if I can't keep up with it, I have no problems with dropping the files.

PR #1021, opened 2023-04-06
emerge-webrsync: support PGP verification

thesamesam:
Note that a lot of this is better reviewed with `-w`.

---
Introduce PGP verification of the webrsync snapshot tarballs
using app-portage/gemato - which is already a dependency of Portage
for verifying normal rsync.

This is the same method Portage uses (see below).

Technically, the fix is very straightforward, but getting to
the fix was the slightly painful bit. What I've concluded
happened is:
- Portage starts getting reworked to gain proper sync module support;

- Someone gets the idea of implementing emerge-webrsync fully in Python
  as a Portage sync module (which is a not-unreasonable idea);

  [This ultimately hasn't gone anywhere.]

- Some folks get the idea of deprecating emerge-webrsync in favour of
  letting Portage call it via its sync module with sync-type=webrsync,
  presumably with the ultimate goal of it transparently one day using
  the aforementioned (yet-non-existent) Python implementation as its
  backend, and not the shell script.

  [To this day, Portage's webrsync implementation shells out to
  the emerge-webrsync shell script, but it has the abstraction to
  switch that out.]

- People are discussing PGP verification in general of the Gentoo
  repository, especially now we'd migrated to git which makes it
  way easier, unlike CVS.

- A bug is filed for PGP verification in emerge-webrsync. People
  decide it doesn't matter too much, because Portage is going to
  Real Soon Now (TM) have its own backend (replacing the shell script)
  and/or Portage's sync module support obsoletes emerge-webrsync entirely
  (i.e. the idea being nobody should call emerge-webrsync, just use
  emerge to sync).

- Then nothing happens with the "replacement" ideas and the good
  ol' trusty emerge-webrsync ends up with the same problems sitting
  there because nobody saw the point in working on it if it was to
  be replaced soon. But that didn't happen.

Bug: https://bugs.gentoo.org/597800

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,11 @@
 portage-3.0.47 (UNRELEASED)
 ---------------
 
+Security:
+* emerge-webrsync (the standalone tool) now verifies PGP signatures, see
+  bug #597800. Note that 'sync-type = webrsync' in repos.conf already
+  handled PGP verification when configured to do so (and it is by default).
+
 Features:
 * install-qa-check.d: 90gcc-warnings: Add additional code quality warnings:
   - -Wrestrict
--- a/bin/emerge-webrsync
+++ b/bin/emerge-webrsync
@@ -24,14 +24,49 @@
 # gpg --homedir /etc/portage/gnupg --edit-key ${KEY_ID} trust
 #
 
-# Only echo if in verbose mode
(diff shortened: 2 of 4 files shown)


thesamesam:
Note that this soft-requires >=gemato-20.1 for nicer output, but it's not a hard requirement: https://github.com/projg2/gemato/pull/30.

thesamesam:
Merged.

PR #1023, opened 2023-04-09
econf: Match configure --help output better

ulm:
Check for proper end of string for all option names beginning with "with", "disable" or "enable".


SoapGentoo (approved):
and pkgcore too plox

PR #1024, opened 2023-04-19
Add a install log check for missing dev-python/cython dep

mgorny:


bin/install-qa-check.d/90cython-dep
```diff
@@ -0,0 +1,41 @@
+# Check for missing dev-python/cython dep based on output
+# from setuptools cythonize() support
+
+cython_dep_check() {
+	[[ ${BDEPEND} == *dev-python/cython* ]] && return
```
thesamesam, line 5:
```suggestion
	[[ ${BDEPEND}${DEPEND} == *dev-python/cython* ]] && return
```

thesamesam: This check doesn't look at whether distutils-r1 is inherited, though.

mgorny: Do we really care more about a potential false positive of an ebuild using ancient EAPI, or that people put it wrongly in `DEPEND` instead of `BDEPEND`?

thesamesam: No. :)

thesamesam:
Lgtm but do you want an inherit check for distutils or not? I would prefer one to avoid penalising other installs.

mgorny:
> Lgtm but do you want an inherit check for distutils or not? I would prefer one to avoid penalising other installs.

I don't mind either way. My original thought was that people could trigger setuptools without using distutils-r1 but I suppose grepping the log a few times can be expensive.

thesamesam:
Let's make it conditional + add a NEWS entry, then LGTM.

mgorny:
Done.

thesamesam (approved):

PR #1025, opened 2023-04-24
dispatch-conf: Avoid race when accessing log file

hannob:
First creating the file and then running chmod creates a security risk where a user could access the file. Avoid this by enforcing the file permissions via umask.


floppym (approved):

thesamesam:
Thanks!

PR #1026, opened 2023-04-27
bin: Rename all _E_*DESTTREE_ variables to __E_*DESTTREE

ulm:
The __* namespace is reserved for package manager use.


floppym (approved):
Looks good to me, pending CI results.

PR #1027, opened 2023-04-28
fowners, fperms: Fix handling of relative pathnames

ulm:
PMS section 12.3.9 specifies that fowners and fperms "[take] paths relative to the image directory."

Currently, these helpers interpret any pathnames not beginning with a slash as relative to the current working directory (after printing a QA warning). This contradicts the spec.

It is also inconsistent with other helpers like insinto or dostrip, which expand their pathnames relative to ${ED}.

Closes: https://bugs.gentoo.org/905223

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,6 @@
+Bug fixes:
+* fowners, fperms: Fix handling of relative pathnames (bug #905223).
+
 portage-3.0.47 (2023-04-30)
 ---------------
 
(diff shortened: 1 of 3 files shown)


ulm:
See also #364.

mgorny:
I don't mind changing the behavior but I'd prefer if you left The QA warning in (and we wouldn't rely on either behavior, at least for some time).

ulm:
> I don't mind changing the behavior but I'd prefer if you left The QA warning in (and we wouldn't rely on either behavior, at least for some time).

Re-added QA warnings, with slightly different wording. Is is o.k. to remove them by the end of 2024?

mgorny:
Well, now it suggests that it's fine to use it, except it isn't because people may still be using old Portage versions that exhibit the old behavior. The QA warning should make it clear "don't use it until <some-date> because old Portage doesn't handle it".

ulm:
Better now?

mgorny (approved):
LGTM (untested).

PR #1028, opened 2023-04-29
doebuild.py: PVR includes -r0 if explicitly specified

ulm:
PMS says that PVR is "Package version and revision (if any)". https://projects.gentoo.org/pms/8/pms.html#x1-109001r1

Bug: https://bugs.gentoo.org/875362

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -28,6 +28,8 @@ Bug fixes:
 * All _E_*DESTTREE_ variables to renamed to __E_*DESTTREE, in order to move
   them to the __* namespace which is reserved for the package manager.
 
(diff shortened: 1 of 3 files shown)


thesamesam:
Would you mind mentioning the paludis+pkgcore research in the commit message, as well as the commit you dug up originally adding this?

Thank you for this!

ulm:
> Would you mind mentioning the paludis+pkgcore research in the commit message, as well as the commit you dug up originally adding this?

Commit message updated. (No change of the code.)

thesamesam:
Thanks!

PR #1031, opened 2023-04-30
type annotate backtrack_depgraph in depgraph.py

berinaniesh:


lib/_emerge/depgraph.py
```diff
@@ -11385,7 +11387,7 @@ def _spinner_stop(spinner):
     portage.writemsg_stdout(f"Dependency resolution took {darkgreen(time_fmt)} s.\n\n")
 
 
-def backtrack_depgraph(settings, trees, myopts, myparams, myaction, myfiles, spinner):
+def backtrack_depgraph(settings: portage.package.ebuild.config.config, trees: dict, myopts: dict, myparams: dict, myaction, myfiles: [str], spinner: stdout_spinner) -> (bool, depgraph, [str]):
```
thesamesam, line 11390: You should be able to quote the type to avoid an import here:
```suggestion
def backtrack_depgraph(settings: "portage.package.ebuild.config.config", trees: dict, myopts: dict, myparams: dict, myaction, myfiles: [str], spinner: stdout_spinner) -> (bool, depgraph, [str]):
```

Also, when you run mypy before & afterwards, do you get many more errors?

berinaniesh: Quotes to avoid import is very useful I guess. So, we can remove the import of stdout_spinner and use quotes again. 
`mypy` says "Found 115 errors in 43 files (checked 1 source file)", when checking depgraph.py.

thesamesam: How many errors did it give _before_ your change, vs _after_?

berinaniesh: I did not run `mypy` before. Ran after the commit and it gives 115 errors.

berinaniesh: I did not run `mypy` before. After the commit, it says 115 errors. I ran `mypy` now without the commit, it says "Found 111 errors in 42 files (checked 1 source file)".

thesamesam: Ok, but run it before _now_ and compare?

berinaniesh: This is the output of `mypy depgraph.py` before the commit. https://dpaste.com/2X6BTB24X
This is the output of `mypy depgraph.py` after the commit. https://dpaste.com/BK8MNCWFN

thesamesam: Okay, so this introduces 4 new mypy errors. Specifically:
```

lib/_emerge/depgraph.py:11390: error: Bracketed expression "[...]" is not valid as a type  [valid-type]
lib/_emerge/depgraph.py:11390: note: Did you mean "List[...]"?
lib/_emerge/depgraph.py:11390: error: Syntax error in type annotation  [syntax]
lib/_emerge/depgraph.py:11390: note: Suggestion: Use Tuple[T1, ..., Tn] instead of (T1, ..., Tn)

berinaniesh:
@thesamesam I *think* I rebased my fork with gentoo/portage. 
The new commit brings in two new errors from `mypy`. https://dpaste.com/2CU8ZU7UC
The diff can be viewed here. https://dpaste.com/2CU8ZU7UC/diff/2X6BTB24X

thesamesam:
Looks like it fixes 2 errors which is good!

berinaniesh:
> Looks like it fixes 2 errors which is good!

Is there anything else I should change?

berinaniesh:
> 

You said these last time, I changed them. Won't the new commits automatically come into the pull request?

thesamesam:
I think however you're pushing isn't right.

`git rebase origin/master` then `git push -f` should work.

berinaniesh:
I did and it shows "Everything up-to-date". I didn't make a branch before making changes. That hinders my other works as well. I'll close this pull request, make a branch, make few more type annotations and submit a new one @thesamesam.

thesamesam:
OK, sounds good.

I recommend that:
- you have `origin` as github.com/gentoo/portage (or anongit.gentoo.org etc)
- you have `berin` or `me` or whatever as github.com/berinaniesh/portage
- you run `git config --global pull.rebase true` to avoid merge commits being generated

PR #1034, opened 2023-05-10
[RFC] install-qa-check.d/90config-impl-decl: Add infrastructure for skipping functions globally

mattst88:
The first patch makes it possible to specify wildcards in `QA_CONFIG_IMPL_DECL_SKIP`. The next patch ignores `__builtin_*`, `__sync_*`, and `__atomic_*` built-ins globally. The subsequent patches add functions that are not present on Linux as examples.

This series aims to provide a single location to list functions

1. whose availability might differ per-platform (e.g. listing `acl` in `app-editors/vim` unconditionally is wrong for Solaris).
2. that are never available on a platform (i.e., it makes little sense to list `res_getservers` in many separate ebuilds when it is never present on Linux)

### DIFF
--- a/bin/install-qa-check.d/90config-impl-decl
+++ b/bin/install-qa-check.d/90config-impl-decl
@@ -19,6 +19,18 @@
 #
 # See also: bug 892651
 
+# Same as the "has" function, but allows wildcards in the array
+is_in() {
+	local needle=$1
+	shift
+
+	local x
+	for x in "$@"; do
+		[[ "${needle}" = ${x} ]] && return 0
+	done
+	return 1
+}
+
 find_log_targets() {
 	local log_targets=(
 		'config.log'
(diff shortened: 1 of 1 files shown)


thesamesam:
cc @xxc3nsoredxx

xxc3nsoredxx:
I like this.

Side note: `is_in` sounds like something that might be useful for other things too. Maybe wildcard support can be added to a future EAPI?

mattst88:
> Side note: is_in sounds like something that might be useful for other things too. Maybe wildcard support can be added to a future EAPI?

Yeah, seems potentially valuable. I'm not in love with the name, so if you have any suggestions I'd be happy to hear them.

xxc3nsoredxx:
That is always the hard part isn't it. Maybe `hasw` for "has wildcard" or something? Alternatively, since it's not _just_ wildcards that it unlocks, `hasp` for "has pattern"?

thesamesam:
I'll merge this now in any case but ofc feel free to discuss more, it's just beyond whether or not the PR is suitable - so let's go!

thesamesam:
Thank you!

PR #1037, opened 2023-05-13
portageq-wrapper: New wrapper script

ulm:
Move the internal portageq wrapper script out of the ebuild-helpers directory. Call the new script from best_version and has_version in the case when IPC is disabled.

The goal is to ban portageq from the ebuild environment after some transition time.


ulm:
Tested at runtime with `USE=-ipc`:

- `best_version` works
- `portageq` (in ebuild environment) works and outputs a QA warning

PR #1040, opened 2023-05-17
WIP: add support for parallel lzip (plzip)

Flowdalic:
Using `plzip` (parallel lzip) over `lzip` provides a significant speed improvement, especially on machines with many cores.

However, due the design of portage's `_compressor` map it is not straight forward to make the decision between lzip and plzip configurable. This is because the map assumes that there can only be one compressor/decompressor for every format. And lzip appears to be the only format known in the map where the parallel (de)compressor is actually a different binary.

Potential design options include:
- simply always use plzip instead of lzip
- add a FEATURE parallel-zip. but the existence of this feature may not be easily determinable with the current code design?
- make it configurable via an environment variable (and default to plzip?)

Personally, I am leaning towards simply using plzip instead of lzip.  But I am happy about any other suggestions how to make portage plzip.

### DIFF
--- a/lib/portage/util/compression_probe.py
+++ b/lib/portage/util/compression_probe.py
@@ -9,6 +9,15 @@ import re
 from portage import _encodings, _unicode_encode
 from portage.exception import FileNotFound, PermissionDenied
 
+parallel_lzip = True
+
+if parallel_lzip:
+    lzip_cmd = "plzip -n {JOBS}"
+    lzip_pkg = "app-arch/plzip"
+else:
+    lzip_cmd = "lzip"
+    lzip_pkg = "app-arch/lzip"
+
 _compressors = {
     "bzip2": {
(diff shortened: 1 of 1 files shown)


lib/portage/util/compression_probe.py
```diff
@@ -9,6 +9,15 @@
...
 from portage.exception import FileNotFound, PermissionDenied
 
+parallel_lzip = True
+
+if parallel_lzip:
+    lzip_cmd = "plzip -n {JOBS}"
+    lzip_pkg = "app-arch/plzip"
+else:
+    lzip_cmd = "lzip"
+    lzip_pkg = "app_arch/lzip"
```
xxc3nsoredxx, line 19:
```suggestion
    lzip_pkg = "app-arch/lzip"
```

Small typo?

PR #1045, opened 2023-05-23
Revert FEATURES=ignore-mtime

ulm:
The ignore-mtime feature violates PMS:
https://projects.gentoo.org/pms/8/pms.html#x1-14500013.3.2

Allowing the package manager to arbitrarily modify timestamps of installed files will lead to problems with several languages:

- Lisp (*.fasl must be newer than corresponding *.lisp)
- Emacs (*.elc must not be older than *.el)
- ghdl/VHDL (libraries check their mtimes)
- Python (*.pyc vs *.py, not sure if this one is still an issue)

If the PM does not preserve timestamps, there is no reasonable way to work around this on the ebuild level. (Some horrible hacks existed in the past, see for example impl-*-timestamp-hack in previous versions of common-lisp-common*.eclass which used to overwrite its installed files in pkg_postinst.)

This partially reverts commit 89703c688868c9eb8cd6115cb42ff92f0b9668b8.

Closes: https://bugs.gentoo.org/906978
See-also: https://bugs.gentoo.org/264130
Signed-off-by: Ulrich Müller <ulm@gentoo.org>

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,10 +9,6 @@ Features:
   contents have changed. This is particularly beneficial on CoW filesystems
   with snapshots. This is always enabled. See bug #722270.
 
-  This also adds an additional optional optimization to overwrite only if
-  the hashes have changed, via FEATURES="ignore-mtime". Note that this violates
-  a condition in PMS so should not be relied upon. It is off by default.
-
 Bug fixes:
 * fowners, fperms: Fix handling of relative pathnames (bug #905223).
 
--- a/lib/portage/const.py
+++ b/lib/portage/const.py
@@ -159,7 +159,6 @@ SUPPORTED_FEATURES = frozenset(
         "getbinpkg",
         "gpg-keepalive",
         "icecream",
-        "ignore-mtime",
         "installsources",
         "ipc-sandbox",
         "keeptemp",
--- a/lib/portage/dbapi/vartree.py
+++ b/lib/portage/dbapi/vartree.py
@@ -5422,7 +5422,6 @@ class dblink:
         srcroot = normalize_path(srcroot).rstrip(sep) + sep
         destroot = normalize_path(destroot).rstrip(sep) + sep
         calc_prelink = "prelink-checksums" in self.settings.features
-        ignore_mtime = "ignore-mtime" in self.settings.features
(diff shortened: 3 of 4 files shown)


ulm:
> Would you mind putting some concrete example of where it's likely to break in the commit message?

I've augmented the commit message of the revert.

gcarq:
I think just reverting 29d4c2e76c2c590c15c67c2bd3adede9fc5746eb isn't enough here. In the original PR #991 the mtime hasn't been handled properly until this commit. So if I'm not mistaken the following should be added to preserve the upstream package mtime as expected (even if the file has not been modified otherwise)
```
--- a/lib/portage/dbapi/vartree.py
+++ b/lib/portage/dbapi/vartree.py
@@ -5830,6 +5830,12 @@ class dblink:
                         hardlink_candidates.append(mydest)
                         zing = ">>>"
                     else:
+                        mymtime = thismtime if thismtime is not None else mymtime
+                        try:
+                            os.utime(mydest, ns=(mymtime, mymtime))
+                        except OSError:
+                            # utime can fail here with EPERM
+                            pass
                         zing = "==="
 
                     try:
```

ulm:
> I think just reverting [29d4c2e](https://github.com/gentoo/portage/commit/29d4c2e76c2c590c15c67c2bd3adede9fc5746eb) isn't enough here. In the original PR #991 the mtime hasn't been handled properly until this commit. So if I'm not mistaken the following should be added to preserve the upstream package mtime as expected (even if the file has not been modified otherwise)
> 
> ```
> --- a/lib/portage/dbapi/vartree.py
> +++ b/lib/portage/dbapi/vartree.py
> @@ -5830,6 +5830,12 @@ class dblink:
>                          hardlink_candidates.append(mydest)
>                          zing = ">>>"
>                      else:
> +                        mymtime = thismtime if thismtime is not None else mymtime
> +                        try:
> +                            os.utime(mydest, ns=(mymtime, mymtime))
> +                        except OSError:
> +                            # utime can fail here with EPERM
> +                            pass
>                          zing = "==="
>  
>                      try:
> ```

Oh sorry, I got the condition reverted.

PR #1046, opened 2023-05-24
New codepath for emerge --version

berinaniesh:
New separate and simpler functions for emerge --version. This reduces the time taken for emerge --version from ~0.7 seconds to ~0.25 seconds. Large functions are refactored into smaller ones for code reusability. Also converted format to f strings in getportagever().

### DIFF
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -772,6 +772,89 @@ def action_config(settings, trees, myopts, myfiles):
     return retval
 
 
+def get_libc_version(vardb):
+    libcver = []
+    libclist = set()
+    for atom in expand_new_virt(vardb, portage.const.LIBC_PACKAGE_ATOM):
+        if not atom.blocker:
+            libclist.update(vardb.match(atom))
+    if libclist:
+        for cpv in sorted(libclist):
+            libc_split = portage.catpkgsplit(cpv)[1:]
+            if libc_split[-1] == "r0":
+                libc_split = libc_split[:-1]
+            libcver.append("-".join(libc_split))
+        return libcver
+    return ["unavailable"]
+
+
+def get_profilever(portdir, profile, vardb):
+    profilever = None
+    repositories = vardb.settings.repositories
+    if profile:
+        profilever = relative_profile_path(portdir, profile)
+        if profilever is None:
+            try:
+                for parent in portage.grabfile(os.path.join(profile, "parent")):
+                    profilever = relative_profile_path(
+                        portdir, os.path.join(profile, parent)
+                    )
+                    if profilever is not None:
+                        break
+                    colon = parent.find(":")
+                    if colon != -1:
+                        p_repo_name = parent[:colon]
+                        try:
+                            p_repo_loc = repositories.get_location_for_name(p_repo_name)
+                        except KeyError:
+                            pass
+                        else:
+                            profilever = relative_profile_path(
+                                p_repo_loc,
+                                os.path.join(
+                                    p_repo_loc, "profiles", parent[colon + 1 :]
+                                ),
+                            )
+                            if profilever is not None:
+                                break
+            except portage.exception.PortageException:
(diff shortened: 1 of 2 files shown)


lib/_emerge/actions.py
```diff
@@ -772,6 +772,89 @@ def action_config(settings, trees, myopts, myfiles):
...
+
+    if profilever is None:
+        profilever = "unavailable"
+
+    return profilever
+
+
+def emerge_version(emerge_config):
+    portagever = portage.VERSION
+    pythonver = (
```
thesamesam, line 837: Why don't we write this inline?

python=f"python{sys.version_info[0]}..."

thesamesam (commented):
Could you also explain in the commit message why we're able to do this and how it works?

lib/_emerge/actions.py
```diff
@@ -772,6 +772,89 @@ def action_config(settings, trees, myopts, myfiles):
...
+
+def get_profilever(portdir, profile, vardb):
+    profilever = None
+    repositories = vardb.settings.repositories
+    if profile:
+        profilever = relative_profile_path(portdir, profile)
+        if profilever is None:
+            try:
+                for parent in portage.grabfile(os.path.join(profile, "parent")):
+                    profilever = relative_profile_path(
```
afalko, line 799: We have >2 levels of nesting where, I think it would make it more readable to refactor the contents of this for loop into a function

lib/_emerge/actions.py
```diff
@@ -772,6 +772,89 @@ def action_config(settings, trees, myopts, myfiles):
...
+def get_libc_version(vardb):
+    libcver = []
+    libclist = set()
+    for atom in expand_new_virt(vardb, portage.const.LIBC_PACKAGE_ATOM):
+        if not atom.blocker:
+            libclist.update(vardb.match(atom))
+    if libclist:
+        for cpv in sorted(libclist):
+            libc_split = portage.catpkgsplit(cpv)[1:]
+            if libc_split[-1] == "r0":
```
afalko, line 784: Should r0 be a constant?

lib/_emerge/main.py
```diff
@@ -1257,6 +1259,10 @@ def emerge_main(args=None):
     os.umask(0o22)
     emerge_config = load_emerge_config(action=myaction, args=myfiles, opts=myopts)
 
+    if myaction == "version":
+        writemsg_stdout(emerge_version(emerge_config) + "\n", noiselevel=-1)
```
afalko, line 1263: Could be cool to make noiselevel an enumeration

lib/_emerge/actions.py
```diff
@@ -772,6 +772,89 @@ def action_config(settings, trees, myopts, myfiles):
...
+        profilever = "unavailable"
+
+    return profilever
+
+
+def emerge_version(emerge_config):
+    portagever = portage.VERSION
+    pythonver = (
+        "python"
+        f" {sys.version_info[0]}"
```
afalko, line 839: Should we do a len check to make sure we never hit index out of bounds?

thesamesam: I don't see a need for that, given https://docs.python.org/3/library/sys.html#sys.version_info.

lib/_emerge/main.py
thesamesam: I think you're right, although it's likely to be a job to do in a followup (and quite a big one, even if it is mostly mechanical).

berinaniesh:
Till now, I don't see how we can get rid of `gcc-config` call without affecting behaviour. I'm cherry picking (manually, because all commits were squashed) the useful parts of this pull request into a new pull request and closing this. I'll try to use the lessons learnt from making this PR to improve `emerge --version` in the future.

berinaniesh:
The useful changes are present in [this](https://github.com/gentoo/portage/pull/1058) PR.

PR #1047, opened 2023-05-25
Feature/testing with pytest

palao:
This PR brings changes that permit the test suite to run both under `pytest` and under the old `runtests` script.
A minimalist approach has been followed in the sense that some work can/must be done to make the transition to `pytest` cleaner and more solid. This PR is a solid step in that direction.


thesamesam:
This looks like a pretty solid start (thank you!)

Once CI is happy, I'm good w/ merging it so we can build on it.

Would you mind tweaking the commit messages to the style we tend to use?

e.g. `tests: runtests: ...`, `tests: whirlpool: ...`, but feel free to add your own flair like e.g. `tests: whirlpool: ... (chore)` or similar

(All I really care about is the `component:` bit because it's what we tend to use across Gentoo.)

palao:
Thank you, Sam, for the feedback!
So, there are two things to fix:

1) The commit messages. Sorry about that! Do you think that something like this is OK?
```
tests: pytest: added session scoped autoused fixture to prepare env
tests: re-introduced runTests.py
tests: WhirlpoolTestCase.testCWhirlpool skipped conditionally with pytest
tests: OrChoicesLibpostprocTestCase.testOrChoicesLibpostproc marked xfail
tests: UnshareNetTestCase.testUnshareNet uses skipif
tests: the order of skipif conditions changed
tests: AutounmaskMultilibUseTestCase.testAutounmaskMultilibUse marked xfail
tests: AutounmaskUseSlotConflictTestCase.testAutounmaskUseSlotConflict marked xfail
tests: last part of CopyFileSparseTestCase.testCopyFileSparse marked as xfail
tests: test about spawning with large env var converted to pytest
tests: pytest: the last four failing tests with pytest fixed
tests: runtests: ensure diff test logic for pytest and runtests
tests: PipeReaderArrayTestCase is marked xfail
tests: adjusted the copyright years in some files
tests: tox: added pytest dependency
```
I tried to make it a bit terser and more consistent. But maybe you meant something different? In any case, are there some guidelines for this?

2) the failing pre-commit. pytest is missing, but not sure where is the best place to include its installation. An easy solution I think it would be to add it to `.builds/lint.yml`, replacing `pip install pylint` by `pip install pylint pytest`. Ideally I'd have this in `pyproject.toml` as part of an optional `test` dependency, or something like that. But that would require to revamp some other pieces related to the management of the project.
Should I simply hack the `.builds/lint.yml` file?

thesamesam:
No, thank _you_!

1. That looks good to me, although I'd maybe do something like:
```
tests: pytest: added session scoped autoused fixture to prepare env
tests: re-introduced runTests.py
tests: whirlpool: skip C test conditionally w/ pytest
tests: resolver: use XFAIL w/ pytest
tests: process: port testUnshareNet to pytest
[...]
```

The idea being, roughly:
`component: component2: action`
with component2 (or maybe even component3) being optional, only there if the first component is too ambiguous.

2. tbh I think it's fine to just unconditionally have it in the lint job via hacking `.builds/lint.yml`. We _do_ need to sort out the build system at some point, but that's for another time.

thesamesam:
I should've said this as well, in case it helps:
1. If in doubt, pick the directory (or if needed, filename) as a component
2. See https://www.gentoo.org/glep/glep-0066.html#commit-messages, we kind of extend that to other projects (obviously it doesn't apply directly because ebuilds aren't in all repos, but hopefully it helps give the idea)

thesamesam:
Huh, I don't get why it's failing still..

thesamesam:
Oh, maybe we need https://pre-commit.com/#config-additional_dependencies?

palao:
Oh, the joy of yaml...

thesamesam:
Many, many thanks for this. So chuffed to see us on the road to pytest!

Thank you David!

thesamesam:
We should have a pytest job in CI during this transition period. Maybe we can just do one for Python 3.11, given it's temporary?

palao:
Thanks a lot for the patience and help!
I'm very glad it's working now.

PR #1048, opened 2023-06-05
main.py, portage.util: Add docstrings and type annotations

berinaniesh:
added a few more items to `.gitignore` also.

### DIFF
--- a/lib/portage/util/__init__.py
+++ b/lib/portage/util/__init__.py
@@ -130,7 +130,7 @@ def writemsg_stdout(mystr: str, noiselevel: int = 0) -> None:
     writemsg(mystr, noiselevel=noiselevel, fd=sys.stdout)
 
 
-def writemsg_level(msg: str, level: int = 0, noiselevel: int = 0):
+def writemsg_level(msg: str, level: int = 0, noiselevel: int = 0) -> None:
     """
     Show a message for the given level as defined by the logging module
     (default is 0).
--- a/lib/portage/util/__init__.py
+++ b/lib/portage/util/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2004-2020 Gentoo Authors
+# Copyright 2004-2023 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 from portage.cache.mappings import UserDict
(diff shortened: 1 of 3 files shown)


berinaniesh:
I ran black on the modified files before committing, but the CI fails here. I am guessing it is because of the python 3.9+ style type annotations. What should I do @thesamesam ?

thesamesam (changes requested):

lib/portage/util/__init__.py
```diff
@@ -126,8 +133,8 @@ def writemsg_stdout(mystr: str, noiselevel: int = 0) -> None:
 def writemsg_level(msg: str, level: int = 0, noiselevel: int = 0):
```
thesamesam, line 133: Shouldn't we annotate this function while we're here? It's very similar to the other.

berinaniesh: Did you mean to add return type to `writemsg_level`? My bad, just did. Pushing it.

thesamesam:
lgtm, will merge when CI is done, thanks!

thesamesam:
Missing signoff, please `git rebase --signoff -f` and push

berinaniesh:
> Missing signoff, please `git rebase --signoff -f` and push

But I have signoffs in all three commits. Git push says everything up to date.

thesamesam:
Ah, you manually wrote the signoff: it's `Signed-off-by`, not `Signed-Off-by`, so some tools don't recognise it, like pram:
```
2023-06-06 07:47:06 (119 MB/s) - ‘/tmp/tmp.jqur4VRL4a/all.patch’ saved [7316]

Commit no. 0001 was not signed off by the author!
```

berinaniesh:
> Ah, you manually wrote the signoff: it's `Signed-off-by`, not `Signed-Off-by`, so some tools don't recognise it, like pram:
> 
> ```
> 2023-06-06 07:47:06 (119 MB/s) - ‘/tmp/tmp.jqur4VRL4a/all.patch’ saved [7316]
> 
> Commit no. 0001 was not signed off by the author!
> ```

I did. I had git do it automatically in the local repo because I don't want to sign off other commits I do. I often delete the source repo and resync. It resets. So, I manually add Sign offs. Next time I'll take care of the case. Is that fine?

thesamesam:
>I often delete the source repo and resync.

Please don't do this, it's an antipattern. You should never need to. If you think you're in a situation where you do, ask for help.

thesamesam:
Anyway, yes, fine, I'll fix it here, although I think you could've with `git rebase -i` quickly?

ajakk:
Thanks for working on this!

PR #1051, opened 2023-06-06
Adjust write tests for DISTDIR and PORTAGE_TMPDIR to work with automount directories

floppym:
Bug: https://bugs.gentoo.org/485100
Bug: https://bugs.gentoo.org/890812

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,9 @@ Bug fixes:
 * emerge-webrsync: Give a nicer error message if the requested key isn't
   on the filesystem (bug #905868).
 
+* Adjust write tests for DISTDIR and PORTAGE_TMPDIR to work with automount
+  directories (bug #485100, bug #890812).
+
 portage-3.0.48.1 (2023-06-06)
 ----------------
 
--- a/lib/portage/package/ebuild/doebuild.py
+++ b/lib/portage/package/ebuild/doebuild.py
@@ -1608,7 +1608,22 @@ def _check_temp_dir(settings):
     # for those people.
     checkdir = first_existing(os.path.join(settings["PORTAGE_TMPDIR"], "portage"))
 
-    if not os.access(checkdir, os.W_OK):
+    try:
+        with tempfile.NamedTemporaryFile(prefix="exectest-", dir=checkdir) as fd:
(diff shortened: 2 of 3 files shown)


floppym:
If anyone wants to test this, you can set up an automount using systemd via an fstab entry like this:
```
tmpfs /path/to/mount tmpfs mode=0770,uid=root,gid=portage,x-systemd.automount
```
Then run `systemctl daemon-reload` and `systemctl start path-to-mount.automount.

thesamesam:
I'll test this out tomorrow, thank you!

thesamesam:
> If anyone wants to test this, you can set up an automount using systemd via an fstab entry like this:
> 
> ```
> tmpfs /path/to/mount tmpfs mode=0770,uid=root,gid=portage,x-systemd.automount 0 0
> ```
> 
> Then run `systemctl daemon-reload` and `systemctl start path-to-mount.automount`.

The patch lgtm, but I couldn't reproduce the bug with:
```
tmpfs /mnt/testing-this-bug tmpfs mode=0770,uid=root,gid=portage,x-systemd.automount 0 0
```

I could see this when I start the automount:
```
systemd-1 on /mnt/testing-this-bug type autofs (rw,relatime,fd=60,pgrp=1,timeout=0,minproto=5,maxproto=5,direct)
```

And after Portage tries to use it with `PORTAGE_TMPDIR=/mnt/testing-this-bug emerge -v1 google-chrome`:
```
tmpfs on /mnt/testing-this-bug type tmpfs (rw,relatime,mode=770,gid=250,x-systemd.automount)
```

floppym:
To reproduce the original issue, I guess you would need to include "/portage" in the mount point. For example:
```
tmpfs /mnt/testing-this-bug/portage tmpfs mode=0770,uid=root,gid=portage,x-systemd.automount 0 0
```
Then, set `PORTAGE_TMPDIR=/mnt/testing-this-bug`

thesamesam:
Ah, thanks! I should've clocked that. Let me try.

thesamesam (approved):
Reproduced, thanks!

It might be nice to put that in the commit message but it's not a big deal if you object.

floppym:
Can do.

PR #1052, opened 2023-06-09
Remove dummy threading

berinaniesh:
`dummy_threading` is deprecated since python 3.7 and portage's minimum supported python version is 3.7. So, all code related to `dummy_threading` is removed.

### DIFF
--- a/lib/_emerge/AsynchronousLock.py
+++ b/lib/_emerge/AsynchronousLock.py
@@ -1,19 +1,10 @@
-# Copyright 2010-2020 Gentoo Authors
+# Copyright 2010-2023 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 import fcntl
 import logging
 import sys
-
-try:
-    import dummy_threading
-except ImportError:
-    dummy_threading = None
-
-try:
-    import threading
-except ImportError:
-    threading = dummy_threading
+import threading
 
 import portage
 from portage import os
@@ -39,7 +30,6 @@ class AsynchronousLock(AsynchronousTask):
     __slots__ = ("path",) + (
         "_imp",
         "_force_async",
-        "_force_dummy",
         "_force_process",
         "_force_thread",
         "_unlock_future",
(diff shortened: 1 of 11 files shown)


thesamesam:
Oh, very nice.

lib/_emerge/AsynchronousLock.py
```diff
@@ -61,14 +51,11 @@ def _start(self):
...
-            not self._force_thread
-            and (self._use_process_by_default or threading is dummy_threading)
+            not self._force_thread and self._use_process_by_default
         ):
             self._imp = _LockProcess(path=self.path, scheduler=self.scheduler)
         else:
-            self._imp = _LockThread(
-                path=self.path, scheduler=self.scheduler, _force_dummy=self._force_dummy
-            )
+            self._imp = _LockThread(path=self.path, scheduler=self.scheduler)
```
thesamesam, line 58: These two codepaths are now identical (both sides of the `if`/`else`).

I haven't checked, but it might be force_process or at least force_thread are obsolete now. (Verify!)

thesamesam (changes requested):

lib/_emerge/AsynchronousLock.py
```diff
@@ -50,13 +46,7 @@ def _start(self):
...
                 return
 
-        if self._force_process or (
-            not self._force_thread and self._use_process_by_default
-        ):
-            self._imp = _LockProcess(path=self.path, scheduler=self.scheduler)
-        else:
-            self._imp = _LockThread(path=self.path, scheduler=self.scheduler)
-
+        self._imp = _LockProcess(path=self.path, scheduler=self.scheduler)
```
berinaniesh, line 49: Please hold merge.

lib/_emerge/AsynchronousLock.py
berinaniesh: One is `_LockProcess` and the other is `_LockThread`. I am not sure what differences they have. Maybe we should keep that change for a later time?

floppym:
Please squash the last two commits; it makes no sense to make a change and revert it in the same PR.

thesamesam:
I suspect all of these could be squashed into one, or at least structured differently. Right now, there's e.g. "Remove dummy_threading completely" and then _afterwards_, there's "Remove all traces of ...".

If you want to split it into components/touched files, that might work, but otherwise one commit seems best.

PR #1054, opened 2023-06-12
depgraph.py: type annotations

berinaniesh:


thesamesam:
LGTM & adds no new mypy errors.

lib/_emerge/depgraph.py
```diff
@@ -100,6 +100,10 @@
 from _emerge.resolver.circular_dependency import circular_dependency_handler
 from _emerge.resolver.output import Display, format_unmatched_atom
 
+# Type annotation imports
+from typing import Any, Optional, Dict, List, Tuple, Union
+from _emerge.stdout_spinner import stdout_spinner
```
thesamesam, line 105: For these, I would really just quote the full type in references to avoid an unnecessary import.

thesamesam: Done in a5b4eb5a69be4b0abe0c0ee682c40f1881be084f. The motivation here being imports are slow and Portage already does a looooot of them.

PR #1057, opened 2023-06-14
Drop FEATURES=cgroup, i.e., v1 cgroup usage

Flowdalic:
Remove portage's usage of Linux version 1 cgroups, which are itself superseded by version 2 cgroups. This basically reverts b01a1b90d8c5 ("Add FEATURES=cgroup to isolate phase processes.").

Portage's usage of version 1 cgroups has caused some issues in the past. For example https://bugs.gentoo.org/894398, where LXD got confused by the existence of the version 1 cgroup created by portage.

Arguably, this could be considered a bug in LXD, but with FEATURES=pid-sandbox, as better alternative to FEATURES=cgroup exists. And removing the code for FEATURES=cgroup reduces portage's code size, which is also a plus.

Bug: https://bugs.gentoo.org/894398

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,11 @@
 portage-3.0.49 (UNRELEASED)
 --------------
 
+Breaking changes:
+* FEATURES=cgroup was removed since it was based on version 1 cgroups, which
+  caused some issues and version 1 cgroups are obsolete. Use
+  FEATURES=pid-sandbox instead.
+
 Bug fixes:
 * Adjust write tests for DISTDIR and PORTAGE_TMPDIR to work with automount
   directories (bug #485100, bug #890812).
--- a/lib/_emerge/AbstractEbuildProcess.py
+++ b/lib/_emerge/AbstractEbuildProcess.py
@@ -3,10 +3,7 @@
 
 import functools
 import io
-import platform
 import stat
-import subprocess
-import tempfile
 import textwrap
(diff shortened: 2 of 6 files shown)


Flowdalic:
> Yeah, the presence of this causes endless confusion. I'm not aware of a reason to use this at all now. We've had pid-sandbox since.. 2017? Earlier?

1. FEATURES=cgroup was introduced with b01a1b90d8c5 ("Add FEATURES=cgroup to isolate phase processes.") with author date 2013-08-19. 
2. FEATURES=pid-sandbox was introduced with 88ba0f45b546 ("Support FEATURES=pid-sandbox") with author date 2018-11-11. 
3. FEATURES=mound-sandbox was introduced with 8aa33981d839 ("Add FEATURES=mount-sandbox to take advantage of mount ns") with author date 2018-11-12

So FEATURE=cgroup predates the other features that are based on modern kernel namespace-based isolation mechanisms by 5 years.

floppym:
Dropping this feature allows for more consistent handling of the cgroup hierarchy by other tools (like systemd). I support this.

berinaniesh:
I am so glad seeing code getting deleted from portage. :)

Flowdalic:
@mgorny friendly ping

Flowdalic:
@mgorny another friendly ping

thesamesam:
Thanks! mgorny acked on IRC.

PR #1058, opened 2023-06-18
Some refactoring / tidying

berinaniesh:
~~Only now, I am noticing the first commit doesn't have a signoff message and the third commit has two signoff messages. I'll fix them soon~~. Done.

### DIFF
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -2789,6 +2789,7 @@ def relative_profile_path(portdir, abs_profile):
         profilever = None
     return profilever
 
+
 def get_libc_version(vardb):
     libcver = []
     libclist = set()
@@ -2804,6 +2805,7 @@ def get_libc_version(vardb):
     else:
         libcver = ["unavailable"]
     return libcver
+
 
 def get_profile_version(portdir, profile, vardb):
     profilever = None
@@ -2845,8 +2847,9 @@ def get_profile_version(portdir, profile, vardb):
 
     if profilever is None:
         profilever = "unavailable"
-    
+
     return profilever
+
 
 def getportageversion(portdir, _unused, profile, chost, vardb):
     pythonver = (
@@ -2863,14 +2866,7 @@ def getportageversion(portdir, _unused, profile, chost, vardb):
     gccver = getgccversion(chost)
     unameout = platform.release() + " " + platform.machine()
 
-    return "Portage {} ({}, {}, {}, {}, {})".format(
-        portage.VERSION,
-        pythonver,
-        profilever,
-        gccver,
-        ",".join(libcver),
-        unameout,
-    )
+    return f"Portage {portage.VERSION} ({pythonver}, {profilever}, {gccver}, {','.join(libcver)}, {unameout})"
 
 
 class _emerge_config(SlotObject):
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -1908,7 +1908,6 @@ def action_info(settings, trees, myopts, myfiles):
     append(
         getportageversion(
             settings["PORTDIR"],
-            None,
             settings.profile_path,
             chost,
             trees[settings["EROOT"]]["vartree"].dbapi,
@@ -2858,7 +2857,6 @@ def get_profile_version(
 
 def getportageversion(
     portdir: str,
-    _unused: None,
     profile: str,
     chost: str,
     vardb: portage.dbapi.vartree.vardbapi,
@@ -3527,7 +3525,6 @@ def run_action(emerge_config):
         writemsg_stdout(
             getportageversion(
                 emerge_config.target_config.settings["PORTDIR"],
-                None,
                 emerge_config.target_config.settings.profile_path,
                 emerge_config.target_config.settings.get("CHOST"),
                 emerge_config.target_config.trees["vartree"].dbapi,
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -2789,16 +2789,23 @@ def relative_profile_path(portdir, abs_profile):
         profilever = None
     return profilever
 
+def get_libc_version(vardb):
+    libcver = []
+    libclist = set()
+    for atom in expand_new_virt(vardb, portage.const.LIBC_PACKAGE_ATOM):
+        if not atom.blocker:
+            libclist.update(vardb.match(atom))
+    if libclist:
+        for cpv in sorted(libclist):
+            libc_split = portage.catpkgsplit(cpv)[1:]
+            if libc_split[-1] == "r0":
+                libc_split = libc_split[:-1]
+            libcver.append("-".join(libc_split))
+    else:
+        libcver = ["unavailable"]
+    return libcver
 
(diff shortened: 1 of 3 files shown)


lib/portage/dep/__init__.py
```diff
@@ -96,7 +97,11 @@ def _get_slot_dep_re(eapi_attrs):
     return slot_re
 
 
-def _match_slot(atom, pkg):
+def _match_slot(atom, pkg) -> bool:
+    """
+    @type atom: portage.dep.Atom
+    @type pkg: _emerge.Package.Package
```
ajakk, line 103: Do these annotations actually mean anything in terms of PEP 484?

lib/_emerge/actions.py
```diff
@@ -2839,31 +2848,25 @@ def getportageversion(portdir, _unused, profile, chost, vardb):
...
     unameout = platform.release() + " " + platform.machine()
 
-    return "Portage {} ({}, {}, {}, {}, {})".format(
-        portage.VERSION,
-        pythonver,
-        profilever,
-        gccver,
-        ",".join(libcver),
-        unameout,
-    )
```
ajakk, line 2866: Are we gaining anything by refactoring these functions, or just aesthetic?

lib/_emerge/actions.py
```diff
@@ -2790,15 +2790,24 @@ def relative_profile_path(portdir, abs_profile):
...
-def getportageversion(portdir, _unused, profile, chost, vardb):
-    pythonver = (
-        "python"
-        f" {sys.version_info[0]}"
-        f".{sys.version_info[1]}"
-        f".{sys.version_info[2]}"
-        f"-{sys.version_info[3]}"
-        f"-{sys.version_info[4]}"
-    )
+def get_libc_version(vardb):
```
ajakk, line 2793: Any reason to not add typing here?

lib/_emerge/actions.py
```diff
@@ -2790,15 +2790,24 @@ def relative_profile_path(portdir, abs_profile):
...
+            libc_split = portage.catpkgsplit(cpv)[1:]
+            if libc_split[-1] == "r0":
+                libc_split = libc_split[:-1]
+            libcver.append("-".join(libc_split))
+    else:
+        libcver = ["unavailable"]
+    return libcver
+
+
+def get_profile_version(portdir, profile, vardb):
```
ajakk, line 2810: Any reason to not add typing here?

lib/_emerge/actions.py
```diff
@@ -2839,31 +2848,25 @@ def getportageversion(portdir, _unused, profile, chost, vardb):
...
-            libc_split = portage.catpkgsplit(cpv)[1:]
-            if libc_split[-1] == "r0":
-                libc_split = libc_split[:-1]
-            libcver.append("-".join(libc_split))
-    else:
-        libcver = ["unavailable"]
+    return profilever
 
+
+def getportageversion(portdir, _unused, profile, chost, vardb):
```
ajakk, line 2854: Here too?

lib/_emerge/actions.py
thesamesam: I don't mind this one, it's cleaner the new way, I'd say.

ajakk: Sure, I agree it looks cleaner, I just wasn't sure if I was missing some magic optimization or something

lib/portage/dep/__init__.py
berinaniesh: PEP 484 way of adding type hints to this function requires reordering. I thought the original ordering makes sense in a logical way. Shall I reorder the functions?

lib/_emerge/actions.py
berinaniesh: The original function was very big. It was hard to read and keep track of. That's why split into three functions in the spirit of "single responsibility". `get_libc_version()` gets libc's version, `get_profile_version()` gets profile's version etc. No functionality or logic is changed.

berinaniesh:
Should I be concerned of GitHub CI minutes? I have made all required changes except [this](https://github.com/gentoo/portage/pull/1058#discussion_r1233402046). I am waiting on your reply. I want to push the other changes, but if I push, CI will run.

lib/portage/dep/__init__.py
ajakk: If it's what's required to get proper typing working, I think so. I think it might be even better to pop out the classes from the `__init__` to their own dedicated file, for modularity. I'm not sure if there's an undesired side-effect that I'm not seeing, but it will help pre-empt these circular dependency issues (since you don't need all of the dependencies in `$CLASS` just to import the module the class is in).

berinaniesh: Even if we move the position of `_match_slot`, the annotation is not possible because of partial initialization of the module / circular imports. Like @ajakk said, we might need to move the class to a separate file. Can we merge this as it is and keep that for a separate pull request?

thesamesam:
> Should I be concerned of GitHub CI minutes? I have made all requested changes except [this](https://github.com/gentoo/portage/pull/1058#discussion_r1233402046) for which I am waiting on your reply. I want to push the other changes, but if I push, CI will run.

Don't be actively wasteful obviously, but don't stress over it either.

lib/portage/dep/__init__.py
thesamesam: OK, but please follow it up.

lib/_emerge/actions.py
```diff
@@ -2790,15 +2792,26 @@ def relative_profile_path(portdir, abs_profile):
...
-def getportageversion(portdir, _unused, profile, chost, vardb):
-    pythonver = (
-        "python"
-        f" {sys.version_info[0]}"
-        f".{sys.version_info[1]}"
-        f".{sys.version_info[2]}"
-        f"-{sys.version_info[3]}"
-        f"-{sys.version_info[4]}"
-    )
+def get_libc_version(vardb: portage.dbapi.vartree.vardbapi) -> List[str]:
```
thesamesam, line 2795: I have no idea why this returns a list, to be honest. I don't think multiple providers of `virtual/libc` should ever be installed. Can you check git history for this to see why/when it was added and do a followup if appropriate? Thanks.

thesamesam (approved):

lib/_emerge/actions.py
berinaniesh: Yeah, I had the same thought when committing it too. Then thought why change logic and continued with the flow. Will check.

berinaniesh: The implementation seems like multiple libc versions was considered and it was written originally by Zac.
```py
if libclist:
    for cpv in sorted(libclist):
        libc_split = portage.catpkgsplit(cpv)[1:]
        if libc_split[-1] == "r0":
            libc_split = libc_split[:-1]
        libcver.append("-".join(libc_split))
else:
    libcver = ["unavailable"]
return libcver
```
[This](https://stackoverflow.com/questions/847179/multiple-glibc-libraries-on-a-single-host) SO question also talks about multiple libc versions coexisting and `patchelf`, etc. Will someone downstream or someone cross compiling have multiple libc versions?  What should we do @thesamesam?

PR #1062, opened 2023-06-29
actions.py: Remove unreachable code and type annotations

berinaniesh:
`action_build()` is called in one place and it is called in the newer way. Checks for older way of calling is removed.


berinaniesh:
Stale

PR #1063, opened 2023-07-07
digraph.py: Type annotations

berinaniesh:


lib/portage/util/digraph.py
```diff
@@ -1,6 +1,11 @@
 # Copyright 2010-2014 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 
+# Type annotation imports
```
thesamesam, line 4: Please guard this with if type checking

berinaniesh: Seems tests set `TYPE_CHECKING` to `False`. What to do @thesamesam ?

lib/portage/util/digraph.py
```diff
@@ -76,6 +83,9 @@ def remove(self, node):
     def update(self, other):
         """
         Add all nodes and edges from another digraph instance.
+
+        other: another digraph instance,
+        can't be type annotated according to PEP 484 due to cyclic imports
```
thesamesam, line 88: Is there really no way around that?

berinaniesh: Only recently discovered "forward reference" type hints. Fixed.

thesamesam:
Please squash.

berinaniesh:
> Please squash.

done

PR #1065, opened 2023-07-09
depgraph.py: fix "no ebuilds/binpkgs" message

xxc3nsoredxx:
The "there are no binary packages to satisfy" was being unconditionally output for packages that could not be found. Fix the logic for choosing between the "binary packages" and "ebuilds" form of the message.

This is a temporary stopgap as alluded to by me in the bug, but the tl;dr is that some entries in the `myopts` dict have "y"/"n" values whereas some are True/unset, and this discrepancy should be sorted out.

Bug: https://bugs.gentoo.org/909853

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -6456,7 +6456,7 @@ class depgraph:
                     cp_exists = True
                     break
 
-            if self._frozen_config.myopts.get("--usepkgonly", "y"):
+            if self._frozen_config.myopts.get("--usepkgonly", False):
                 writemsg(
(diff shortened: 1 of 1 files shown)


thesamesam:
This whole thing is again making me feel like we need more primitive tests, even if they're not great, to just assert basic things about Portage's behaviour. Bleh.

Thank you!

xxc3nsoredxx:
Yeh. Unfortunate thing is testing seemingly basic stuff like "does the output look like what it should?" sounds like a lot of manual work because scripting generation of example output sounds non-trivial.

thesamesam:
Almost as if we also need to make Portage more machine-readable, too. Crazy!

PR #1066, opened 2023-07-09
_global_updates.py: Type hints

berinaniesh:


lib/portage/_global_updates.py
```diff
@@ -54,7 +70,7 @@ def _do_global_updates(trees, prev_mtimes, quiet=False, if_mtime_changed=True):
     world_list = grabfile(world_file)
     world_modified = False
     world_warnings = set()
-    updpath_map = {}
+    updpath_map: Dict[Any, Any] = {}
```
thesamesam, line 73: I don't think this needs to be Any, from looking at how it gets used below...

berinaniesh:
These are stale now. Closing.

PR #1067, opened 2023-07-10
depgraph.py: fix wrong type annotations

berinaniesh:
bug: https://bugs.gentoo.org/910171

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -101,7 +101,11 @@ from _emerge.resolver.circular_dependency import circular_dependency_handler
 from _emerge.resolver.output import Display, format_unmatched_atom
 
 # Type annotation imports
-from typing import Any, Optional, Dict, List, Tuple, Union
+from typing import Any, Optional, Dict, List, Tuple, Union, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import _emerge.stdout_spinner.stdout_spinner
+
 
 # Exposes a depgraph interface to dep_check.
 _dep_check_graph_interface = collections.namedtuple(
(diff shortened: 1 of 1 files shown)


thesamesam:
cc @xxc3nsoredxx

lib/_emerge/depgraph.py
```diff
@@ -11536,7 +11540,7 @@ def resume_depgraph(
     settings: portage.package.ebuild.config.config,
     trees: portage._trees_dict,
     mtimedb: Any,
-    myopts: Dict[str, str],
+    myopts: Dict[str, Union[str, int, bool]],
     myparams: Dict[str, Any],
```
xxc3nsoredxx, line 11544: While we're at it...
```suggestion
    myparams: Dict[str, Union[int, str, bool]],
```
based on lines 11417 and 11439?

Am I correct in assuming the `Dict[str, Any]` was meant as a placeholder while figuring out a better/more exact type, but you forgot to replace them here?

lib/_emerge/depgraph.py
```diff
@@ -11554,7 +11558,7 @@ def _resume_depgraph(
...
```
xxc3nsoredxx, line 11562: Ditto.

xxc3nsoredxx (changes requested):

lib/_emerge/depgraph.py
berinaniesh: This file was my first ever type annotation. I had no clue on what's going on. I'll check the whole file once again @xxc3nsoredxx. Having said that,
```py
myparams = {
    "recurse": True,
    "binpkg_respect_use": "auto",
    "autounmask": True,
    "autounmask_keep_use": False,
    "autounmask_keep_license": True,
    "autounmask_keep_keywords": True,
    "autounmask_keep_masks": True,
    "bdeps": "auto",
    "ignore_soname_deps": "y",
    "dynamic_deps": True,
    "implicit_system_deps": True,
    "selective": True,
    "deep": True,
    "binpkg_changed_deps": "auto",
}

myopts = {
    "--newuse": True,
    "--pretend": True,
    "--update": True,
    "--deep": True,
    "--keep-going": True,
    "--quiet": True,
    "--regex-search-auto": "y",
    "--verbose": True,
}
```

PR #1068, opened 2023-07-15
lib: env_update: check ldconfig exit code

thesamesam:
Bug: https://bugs.gentoo.org/910376

### DIFF
--- a/lib/portage/util/env_update.py
+++ b/lib/portage/util/env_update.py
@@ -361,19 +361,17 @@ def _env_update(makelinks, target_root, prev_mtimes, contents, env, writemsg_lev
             # an older package installed ON TOP of a newer version will cause ldconfig
             # to overwrite the symlinks we just made. -X means no links. After 'clean'
             # we can safely create links.
-            writemsg_level(
-                _(">>> Regenerating %setc/ld.so.cache...\n") % (target_root,)
-            )
+            writemsg_level(_(f">>> Regenerating {target_root}etc/ld.so.cache...\n"))
             ret = subprocess.run(
-                [f"{ldconfig}", "-X", "-r", f"{target_root}"], cwd="/"
+                [ldconfig, "-X", "-r", target_root], cwd="/"
             ).returncode
         elif ostype in ("FreeBSD", "DragonFly"):
             writemsg_level(
-                _(">>> Regenerating %svar/run/ld-elf.so.hints...\n") % target_root
+                _(f">>> Regenerating {target_root}var/run/ld-elf.so.hints...\n")
             )
             ret = subprocess.run(
                 [
-                    f"{ldconfig}",
+                    ldconfig,
                     "-elf",
                     "-i",
                     "-f",
--- a/lib/portage/util/env_update.py
+++ b/lib/portage/util/env_update.py
@@ -6,7 +6,7 @@ __all__ = ["env_update"]
 import errno
 import glob
 import stat
-import sys
+import subprocess
 import time
 
 import portage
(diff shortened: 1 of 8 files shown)


bmenrigh:
Looks good to me!

    epsilon /tmp # env-update 
    >>> Regenerating /etc/ld.so.cache...
    !!! ldconfig failed with exit code -6

    epsilon /tmp # ldconfig; echo $?
    Aborted
    134

thesamesam:
Thanks for testing! I guess I need to fix that exit code mangling..

bmenrigh:
I was going to make a comment about the exit codes not matching but then I thought perhaps that's just how they work? I don't think I've ever seen an actual negative code in $? and 128 - 134 = -6 so I just assumed negative codes were getting a +128 bias?

Really the only problem seems to be that env-update uses a different encoding for negative codes (looks like standard C-style two's-compliment)

    # env-update; echo $?
    >>> Regenerating /etc/ld.so.cache...
    !!! ldconfig failed with exit code -6
    250

I'd call this a non-issue.

lib/portage/util/env_update.py
```diff
@@ -363,19 +364,24 @@ def _env_update(makelinks, target_root, prev_mtimes, contents, env, writemsg_lev
...
-            os.system(
+            ret = os.system(
                 (
                     "cd / ; %s -elf -i "
                     + "-f '%svar/run/ld-elf.so.hints' '%setc/ld.so.conf'"
                 )
                 % (ldconfig, target_root, target_root)
             )
 
+        ret = os.waitstatus_to_exitcode(ret)
```
floppym, line 380: This function was added in Python 3.9, but we at least pretend to support 3.7 and 3.8.

floppym: To support older Pythons, I think some combination of `os.WIFSIGNALED`, `os.WIFEXITED`, `os.WEXITSTATUS`, and `os.WTERMSIG` will be needed.

https://docs.python.org/3.8/library/os.html#os.WIFSIGNALED

floppym: Something like this should work.
```suggestion
        if os.WIFEXITED(ret):
            ret = os.WEXITSTATUS(ret)
            writemsg(f"!!! ldconfig failed with exit status {ret}\n", noiselevel=-1)
        elif os.WIFSIGNALED(ret):
            ret = os.WTERMSIG(ret)
            writemsg(f"!!! ldconfig was killed with signal {ret}\n", noiselevel=-1)
```

bmenrigh:
A bit more research into exit codes. the high bit (bit 7 with a value of 128) indicates the process was killed by a signal and the low 6 bits are the signal code.  SIGABRT is code 6 hence a return of 134.

    $ ./abort_test; echo 'exit code' $?
    SIGABRT code: 6 (should result in return 134)
    Aborted
    exit code 134

Test code:

    $ cat abort_test.c 
    #include <stdlib.h>
    #include <stdio.h>
    #include <signal.h>
    
    int main(void) {
    
        fprintf(stderr, "SIGABRT code: %d (should result in return %u)\n", SIGABRT, (1 << 7) | SIGABRT);
    
        abort();
    
        return 0;
    }

So here Python is (improperly?) interpreting 134 as -6.

floppym:
> So here Python is (improperly?) interpreting 134 as -6.

This is simply the documented behavior for [os.waitstatus_to_exitcode](https://docs.python.org/3.11/library/os.html#os.waitstatus_to_exitcode). It returns -signum when the process was terminated by a signal.

lib/portage/util/env_update.py
thesamesam: Ah, thanks. Maybe it's time to stop pretending though?

lib/portage/util/env_update.py
```diff
@@ -360,21 +362,30 @@ def _env_update(makelinks, target_root, prev_mtimes, contents, env, writemsg_lev
...
-                (
-                    "cd / ; %s -elf -i "
-                    + "-f '%svar/run/ld-elf.so.hints' '%setc/ld.so.conf'"
-                )
-                % (ldconfig, target_root, target_root)
+                _(f">>> Regenerating {target_root}var/run/ld-elf.so.hints...\n")
             )
+            ret = subprocess.run(
+                [
+                    f"{ldconfig}",
```
mgorny, line 375: Any reason not to use `ldconfig` straight, i.e. without f-string?

lib/portage/util/env_update.py
floppym: Python 3.7 is officially end-of-life now, so no objection from me if you want to drop it.

```suggestion
        ret = os.waitstatus_to_exitcode(ret)
        if ret > 0:
            writemsg(f"!!! ldconfig failed with exit status {ret}\n", noiselevel=-1)
        if ret < 0:
            writemsg(f"!!! ldconfig was killed with signal {-ret}\n", noiselevel=-1)
```

thesamesam: Thanks (on both counts).

thesamesam:
Merged.

PR #1069, opened 2023-07-16
Add some typing

Tatsh:
Someday this will help with static analysis...

### DIFF
--- a/lib/portage/dbapi/__init__.py
+++ b/lib/portage/dbapi/__init__.py
@@ -5,6 +5,7 @@ __all__ = ["dbapi"]
 
 import re
 import warnings
+from typing import Any, Dict, List, Optional, Sequence, Tuple
 
 import portage
 
@@ -29,7 +30,7 @@ from _emerge.Package import Package
 
 class dbapi:
     _category_re = re.compile(r"^\w[-.+\w]*$", re.UNICODE)
-    _categories = None
+    _categories: Optional[Tuple[str, ...]] = None
     _use_mutable = False
     _known_keys = frozenset(auxdbkeys)
     _pkg_str_aux_keys = ("EAPI", "KEYWORDS", "SLOT", "repository")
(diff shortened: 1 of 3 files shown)


lib/portage/dbapi/__init__.py
```diff
@@ -38,7 +39,7 @@ def __init__(self):
         pass
 
     @property
-    def categories(self):
+    def categories(self) -> Sequence[str]:
```
eli-schwartz, line 42: This always returns a tuple so why not just say so? :)

The annotations you are using instead, allow `return 'a simple string'` because the str type is a sequence of str (if you try to act on a str as a sequence, it acts like `['a', ' ', 's', 'i', 'm', 'p', 'l', 'e', ' ', 's', 't', 'r', 'i', 'n', 'g']`).

lib/portage/dbapi/__init__.py
```diff
@@ -76,7 +77,7 @@ def _cpv_sort_ascending(cpv_list):
             # dict to map strings back to their original values.
             cpv_list.sort(key=cmp_sort_key(dbapi._cmp_cpv))
 
-    def cpv_all(self):
+    def cpv_all(self) -> Sequence[str]:
```
eli-schwartz, line 80: Similarly, this is always a list of str, never an actual str, but the type annotations allow the latter.

lib/portage/dbapi/__init__.py
Tatsh: I kind of forgot that that can be the case, specifically for `Sequence[str]`. Fixing this.

lib/portage/dbapi/__init__.py
Tatsh: Yes. It is implied by the docs that `Sequence` should not be used for return type annotations. `List`/`list` should be used instead.

thesamesam:
Thanks!

PR #1071, opened 2023-07-21
Migrate from setuptools to Meson and meson-python

chewi:
This makes Portage PEP 517 compliant.

When building via meson-python, the man pages and logrotate config are no longer included as there seems little point.


.builds/ci.yml
```diff
@@ -31,7 +31,8 @@ tasks:
       for py in "${PYTHON_VERSIONS[@]}"; do
         source ".venv-$py/bin/activate"
         pushd portage
-          time ./setup.py clean install
```
thesamesam, line 34: Could you bang in a single muon job (look at pax-utils)?

thesamesam (commented):
Thank you so much for tackling this!

.builds/ci.yml
eli-schwartz: Unfortunately, muon does not (yet? patches welcome, usually) support the `import('python')` module for non-trivial uses -- that is, anything other than using find_installation() as an inefficient replacement for `find_program('python3')`.

chewi: I guess not then. :cry:

meson.build
```diff
@@ -0,0 +1,69 @@
...
+        'NEWS',
+        'RELEASE-NOTES'
+    ],
+    install_dir : docdir
+)
+
+test(
+    'python',
+    py,
+    args : ['-bWd', meson.current_source_dir() / 'lib' / 'portage' / 'tests' / 'runTests.py'],
```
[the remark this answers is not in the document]
chewi: Done.

tox.ini
```diff
@@ -30,4 +30,4 @@ allowlist_externals =
 	./run-pylint
 commands =
 	pylint: ./run-pylint
-	test: python -b -Wd setup.py test
+	test: python -bWd lib/portage/tests/runTests.py
```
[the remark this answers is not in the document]
chewi: Done.

.builds/ci.yml
eli-schwartz: To be fair it's not that bad... You need python in this case anyway so meson can be bootstrapped with just that python (as a pyz "zipapp" for example). The value of muon is a bit less since you can't knock python out of the dependency loop.

Still better than setuptools. :D

chewi:
After much thrashing around and head-desking, I'm nearly there now. Don't know why Meson isn't liking pypy-3.9. If it was pypy-3.8, I'd have just dropped it.

eli-schwartz:
> After much thrashing around and head-desking, I'm nearly there now. Don't know why Meson isn't liking pypy-3.9. If it was pypy-3.8, I'd have just dropped it.

Sorry! The reason for this is because the CI environment is testing using the Ubuntu-packaged version of meson, which is 0.61.2, and building extensions with PyPy was broken until meson 0.62. See upstream commit https://github.com/mesonbuild/meson/commit/2540ad6e9e08370ddd0b6753fdc9314945a672f0

src/meson.build
```diff
@@ -0,0 +1,59 @@
+# We create symlinks to the native extensions in the source tree for the tests
+# and for development. Meson does not allow you to build in-place and Python
+# cannot create a single namespace from two identically-named paths.
+
+c_args = [
+    '-D_FILE_OFFSET_BITS=64',
```
eli-schwartz, line 6: This is actually a bit funny, because meson gives you this  for free: https://github.com/mesonbuild/meson/blob/c2ed846b64f02653d1460855c0aadef20308fc1f/mesonbuild/compilers/compilers.py#L1131-L1148

https://github.com/mesonbuild/meson/commit/853634a48da025c59eef70161dba0d150833f60d


Doing anything is probably an ABI break, in the meantime meson sort of forces you into it.

floppym: Yeah, we could drop `-D_FILE_OFFSET_BITS` given that meson already enables it.

We never actually needed `_LARGEFILE_SOURCE` or `_LARGEFILE64_SOURCE`, so those can be dropped as well.

chewi: I knew `-D_FILE_OFFSET_BITS=64` was already included, at least in recent versions, but I thought I'd include it with the others for the avoidance of doubt. If we don't even need the others though then I guess I'll just drop them all.

chewi:
@eli-schwartz I don't think there's anything I can do about it here, but I tested cross-compiling and found that while it built the native modules with the right compiler and flags, it had the wrong filename, matching that of the build host. I think it used pkg-config for the flags but relied on the build host's Python to determine the filename. Is this a known issue?

eli-schwartz:
Cross-compiling is the bugbear of python. The only way to get this information is to execute python and run a python code snippet that imports `sysconfig` to get some variables.

The state of the art is therefore to do one of:
- use the ***undocumented*** `$_PYTHON_SYSCONFIGDATA_NAME` environment variable to make a non-cross python load a data file from a cross python and print sysconfig information as if it were the cross python
- run the cross python using qemu or wine, depending

For meson, we only need the correct information here to know which python installation to pass -I flags for, which may or may not wreak havoc in a cross build (only pyconfig.h should differ between installations), and to know what file extension to use. The compiler is always the correct cross compiler, since it uses meson's own understanding of cross targets.

floppym (approved):
This seems like a reasonable choice to me. I'm glad I won't have to learn another Python build system. :)

I'll give this some testing locally and report back.

src/meson.build
```diff
@@ -0,0 +1,50 @@
...
+
+whirlpool_ext = py.extension_module(
+    '_whirlpool',
+    'portage_util__whirlpool.c',
+    dependencies : py.dependency(),
+    subdir : 'portage' / 'util',
+    install : true
+)
+
+run_command(
```
floppym, line 21: I like this workaround! It probably warrants a comment with an explanation though.

chewi: Thanks! I'm not sure what to say about it beyond what I said at the top of the file though. Any suggestions?

floppym: Oh, I somehow overlooked that. Nevermind!

chewi:
See gentoo/gentoo#32098 for the new ebuild.

lib/portage/package/ebuild/meson.build
```diff
@@ -0,0 +1,23 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/package/meson.build
```diff
@@ -0,0 +1,9 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/proxy/meson.build
```diff
@@ -0,0 +1,9 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/repository/meson.build
```diff
@@ -0,0 +1,10 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/repository/storage/meson.build
```diff
@@ -0,0 +1,11 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/sync/meson.build
```diff
@@ -0,0 +1,14 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/sync/modules/cvs/meson.build
```diff
@@ -0,0 +1,8 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/sync/modules/git/meson.build
```diff
@@ -0,0 +1,8 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/sync/modules/mercurial/meson.build
```diff
@@ -0,0 +1,8 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/sync/modules/meson.build
```diff
@@ -0,0 +1,14 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/sync/modules/rsync/meson.build
```diff
@@ -0,0 +1,8 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/sync/modules/svn/meson.build
```diff
@@ -0,0 +1,8 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/sync/modules/webrsync/meson.build
```diff
@@ -0,0 +1,8 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/bin/meson.build
```diff
@@ -0,0 +1,14 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/dbapi/meson.build
```diff
@@ -0,0 +1,12 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/dep/meson.build
```diff
@@ -0,0 +1,28 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/ebuild/meson.build
```diff
@@ -0,0 +1,17 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/emerge/meson.build
```diff
@@ -0,0 +1,14 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/env/config/meson.build
```diff
@@ -0,0 +1,12 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/glsa/meson.build
```diff
@@ -0,0 +1,9 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/gpkg/meson.build
```diff
@@ -0,0 +1,15 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/lafilefixer/meson.build
```diff
@@ -0,0 +1,9 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/lazyimport/meson.build
```diff
@@ -0,0 +1,10 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/lint/meson.build
```diff
@@ -0,0 +1,12 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/locks/meson.build
```diff
@@ -0,0 +1,10 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/meson.build
```diff
@@ -0,0 +1,31 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/news/meson.build
```diff
@@ -0,0 +1,9 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/process/meson.build
```diff
@@ -0,0 +1,16 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/resolver/binpkg_multi_instance/meson.build
```diff
@@ -0,0 +1,10 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/resolver/meson.build
```diff
@@ -0,0 +1,96 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/resolver/soname/meson.build
```diff
@@ -0,0 +1,19 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/sets/base/meson.build
```diff
@@ -0,0 +1,10 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/sets/files/meson.build
```diff
@@ -0,0 +1,10 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/sets/meson.build
```diff
@@ -0,0 +1,12 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/sets/shell/meson.build
```diff
@@ -0,0 +1,9 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/sync/meson.build
```diff
@@ -0,0 +1,9 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/unicode/meson.build
```diff
@@ -0,0 +1,9 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/update/meson.build
```diff
@@ -0,0 +1,11 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/util/dyn_libs/meson.build
```diff
@@ -0,0 +1,9 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/util/eventloop/meson.build
```diff
@@ -0,0 +1,9 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/util/file_copy/meson.build
```diff
@@ -0,0 +1,9 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/util/futures/asyncio/meson.build
```diff
@@ -0,0 +1,15 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/util/futures/meson.build
```diff
@@ -0,0 +1,15 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/util/meson.build
```diff
@@ -0,0 +1,31 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/versions/meson.build
```diff
@@ -0,0 +1,10 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/tests/xpak/meson.build
```diff
@@ -0,0 +1,9 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/util/_async/meson.build
```diff
@@ -0,0 +1,20 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/util/_dyn_libs/meson.build
```diff
@@ -0,0 +1,14 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/util/_eventloop/meson.build
```diff
@@ -0,0 +1,9 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/util/elf/meson.build
```diff
@@ -0,0 +1,9 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/util/endian/meson.build
```diff
@@ -0,0 +1,8 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/util/file_copy/meson.build
```diff
@@ -0,0 +1,7 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/util/futures/_asyncio/meson.build
```diff
@@ -0,0 +1,8 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/util/futures/executor/meson.build
```diff
@@ -0,0 +1,8 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/util/futures/meson.build
```diff
@@ -0,0 +1,17 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/util/iterators/meson.build
```diff
@@ -0,0 +1,8 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/util/meson.build
```diff
@@ -0,0 +1,49 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

lib/portage/xml/meson.build
```diff
@@ -0,0 +1,8 @@
+py.install_sources(
+    [
```
SoapGentoo, line 2:
```suggestion
    files([
```

man/meson.build
```diff
@@ -0,0 +1,31 @@
+man_pages_out = []
+man_pages_in = [
```
SoapGentoo, line 2:
```suggestion
man_pages_in = files([
```

man/ru/meson.build
```diff
@@ -0,0 +1,19 @@
+man_pages_out = []
+man_pages_in = [
```
SoapGentoo, line 2:
```suggestion
man_pages_in = files([
```

meson.build
```diff
@@ -0,0 +1,121 @@
...
+
+subdir('cnf')
+
+if system_wide
+    subdir('doc')
+    subdir('man')
+endif
+
+install_data(
+    [
```
SoapGentoo, line 116:
```suggestion
    files([
```

SoapGentoo (changes requested):

SoapGentoo (changes requested):
My suggestion" let's get rid of 3.7 and 3.8 before we merge this, it makes the reviewing a bit simpler.

cnf/meson.build
```diff
@@ -0,0 +1,133 @@
+install_data(
+    [
```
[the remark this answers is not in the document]
eli-schwartz: I'm not sure what you mean by this. The first thing that install_data does when receiving a "list of strings or list of files() objects" parameter is internally convert all bare strings into files() objects using the same code that files() uses to check if they exist up front.

Are you aware of cases where this does not happen? If so, then it is a meson bug, because the configure stage should *never* succeed when the build or install stages can be trivially detected as impossible to succeed, such as for example when rule inputs are missing.

The difference between using strings and using files() objects is supposed to be limited to the fact that files() can be defined in one subdir and used in another subdir while remembering which subdir they were defined in (and therefore always resolving relative to the subdir they were defined in rather than the subdir they were used in).

bin/meson.build
```diff
@@ -0,0 +1,56 @@
+bins = [
+    'ebuild',
+    'egencache',
+    'emerge-webrsync',
+    'emerge',
+    'emirrordist',
+    'glsa-check',
+    'gpkg-sign',
+    'portageq',
+    'quickpkg'
```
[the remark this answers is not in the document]
eli-schwartz: Although it isn't specified anywhere, `muon fmt` (an autoformatter provided by the c99 implementation of meson) uses four spaces by default.

Personally I've always been a fan of four spaces, for the same reason I like them in python (and also bash and C/C++).

chewi:
> My suggestion: let's get rid of 3.7 and 3.8 before we merge this, it makes the reviewing a bit simpler.

We've already done that since I opened this.

SoapGentoo (approved):

thesamesam:
As far as I'm concerned, we can merge this now, and pick up any pieces once it gets a bit more testing. Any objections?

chewi:
> As far as I'm concerned, we can merge this now, and pick up any pieces once it gets a bit more testing. Any objections?
> 
> I've been running it for a few days and everything seems fine, although there were 2 missing files I commented about on the ebuild side ([gentoo/gentoo#32098 (comment)](https://github.com/gentoo/gentoo/pull/32098#issuecomment-1657512310)). I've not investigated that at all and it might've been fixed since.

Sorry, I missed that comment. I wasn't aware of that issue so I'll take a look.

.github/workflows/ci.yml
```diff
@@ -24,20 +24,37 @@ jobs:
...
           python-version: ${{ matrix.python-version }}
-      - name: Install python dependencies
+      - name: Install dependencies
         run: |
           set -xe
           sudo apt-get update
-          sudo apt-get install -y --no-install-recommends libxslt-dev libxml2-dev libxml2-utils zstd
+          sudo apt-get install -y --no-install-recommends libxslt-dev libxml2-dev libxml2-utils meson zstd
+
+          # Patch Meson to fix pypy-3.9 detection.
```
mgorny, line 33: Link to a PR or a bug, or something we could follow?

chewi: It's already fixed upstream, this is just backporting the fix into Ubuntu's old Meson. I'll clarify though.

mgorny: You may also try installing a newer meson via `pip`.

chewi: Yeah, fetching the patch is a fair suggestion, so I've done that now. I did consider getting a newer Meson with pip, but testing with an older Meson isn't such a bad idea.

chewi:
> I've been running it for a few days and everything seems fine, although there were 2 missing files I commented about on the ebuild side ([gentoo/gentoo#32098 (comment)](https://github.com/gentoo/gentoo/pull/32098#issuecomment-1657512310)). I've not investigated that at all and it might've been fixed since.

`elog-save-summary` was one I simply missed. Fixed that now.

`bashrc-functions.sh` was always there, but the mode has changed, along with some other files in portage-bindir. They were previously all marked executable. Is that what your report was flagging?

.github/workflows/ci.yml
eli-schwartz: Testing against older meson was my original assumption for why you wanted to patch it. Conditionally using a newer version of meson only for pypy would be pretty awkward to do with a matrix job.

thesamesam:
I've merged this so we can get it some more testing - that doesn't mean more comments and such aren't welcome, obviously. Thank you!!

PR #1072, opened 2023-07-24
ebuild/doebuild.py: also set --load-average in GNUMAKEFLAGS default

Flowdalic:
In the presence of multiple parallel emerge jobs, i.e., emerge -j<num>, it is good to limit the overall parallelism in the system via its load average. This avoids overprovisioning the system with tasks and losing performance due to an increased number of context switches.

note: untested

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,7 @@
+Features:
+* GNUMAKEFLAGS: also specifiy "-l $(nproc)", that is, limit by load average,
+  per default.
+
 portage-3.0.52 (2023-10-03)
 --------------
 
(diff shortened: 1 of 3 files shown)


zmedico (approved):

thesamesam:
I like this idea, but AIUI we're blocked on fixing `multiprocessing.eclass`: https://marc.info/?l=gentoo-dev&m=169022500126237&w=2.

Flowdalic:
Glad that you like it. :)

Unfortunately, I have no idea what to do unblock it as [my last question wasn't answered](https://marc.info/?l=gentoo-dev&m=169080052823215&w=2).

thesamesam:
I'm not sure what that question means. Just handle MAKEFLAGS as well like mgorny said?

Flowdalic:
Oh, I've read "we're missing other variables" as that there are multiple other variables that are missing. Hence my question.

Would just adding `${MAKEFLAGS}` behind `${GNUMAKEFLAGS}` do the trick for you?

Flowdalic:
> Would just adding `${MAKEFLAGS}` behind `${GNUMAKEFLAGS}` do the trick for you?

https://github.com/gentoo/gentoo/pull/32385 is what I have so far. I turns out that there are multiple design options. Happy about any feedback, and I'll probably send it to list in a few days.

PR #1076, opened 2023-08-03
Fix Meson CPU detection for make.conf

chewi:
I used the CPU name rather than the family by mistake.


floppym:
It looks like "riscv32" and "riscv64" are being mapped to "risc". That should probably be "riscv" instead.

chewi:
Well spotted. Fixed.

floppym (approved):

thesamesam (approved):

PR #1077, opened 2023-08-06
Define BROOT in pkg_* phases following PMS change

chewi:
Closes: https://bugs.gentoo.org/911797
Bug: https://bugs.gentoo.org/911574

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,7 @@ portage-3.0.50 (UNRELEASED)
 
 Breaking changes:
 * The minimum supported Python version is now >= Python 3.9.
+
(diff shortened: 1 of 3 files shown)


ulm (approved):
It's a bit wider than the spec says (as it includes all `pkg_*` phases) but that's fine.

chewi:
Once this is merged, I can revert gentoo/gentoo@9e4dc068a3066434bfbfa3a0f979ec5d6d8020eb, which fixed cross-prefix but broke regular prefix. This Portage fix addresses both cases, but regular prefix users will be unable to take an update that revert.

floppym (approved):

PR #1078, opened 2023-08-10
BROOT is defined in pkg_* phases only when the EAPI has IDEPEND

ulm:
Bug: https://bugs.gentoo.org/911797#c13


chewi:
This will break cross-prefix under EAPI 7. I thought you were happy to keep it as it was? cross-prefix is very niche, but I'm trying to make it work better.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

ulm:
OK, closing then.

PR #1079, opened 2023-08-13
-Oline fix

xxc3nsoredxx:
Add a check to see if MAKEFLAGS is set (alongside the existing MAKEOPTS
and GNUMAKEFLAGS checks) before enabling output synchronization.

This is only tangentially related to the bug below via being a (likely
rare) case of MAKEFLAGS being set in the wild.

Bug: https://bugs.gentoo.org/909009

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,9 @@
+portage-3.0.51 (UNRELEASED)
+--------------
+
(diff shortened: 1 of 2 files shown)


thesamesam:
Thanks!

PR #1082, opened 2023-08-20
Allow building of pure Python wheel by setting "pure" appropriately

chewi:
It will only build a pure wheel when pure is true in all cases.

To get a pure wheel, run:
```
python -m build --wheel -Csetup-args="-Dnative-extensions=false"
```

### DIFF
--- a/DEVELOPING
+++ b/DEVELOPING
@@ -238,7 +238,8 @@ and commit.
 	- python -m venv .venv
 	- . .venv/bin/activate
 	- pip install build twine
-	- python -m build
+	- python -m build # build the sdist and arch/Python-specific wheel
+	- python -m build --wheel -Csetup-args="-Dnative-extensions=false" # build the pure wheel
 	- twine upload dist/<filenames>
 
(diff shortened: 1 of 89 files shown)


thesamesam:
Thanks! Could you update the `DEVELOPING` file too?

chewi:
> Thanks! Could you update the `DEVELOPING` file too?

What do you want it to say exactly? Only building a pure wheel is easy, otherwise you need to consider the arch and Python versions.

thesamesam:
> > Thanks! Could you update the `DEVELOPING` file too?
> 
> What do you want it to say exactly? Only building a pure wheel is easy, otherwise you need to consider the arch and Python versions.

I figured it was worth changing the instructions to mention:
>python -m build --wheel -Csetup-args="-Dnative-extensions=false"
and then mention that if wanting to build non-pure wheels, `python -m build --wheel`.

I can chuck that in though.

chewi:
Okay, I added something.

thesamesam:
Thank you!

PR #1083, opened 2023-08-23
 lib/portage/gpg.py: send gpg output to stdout, then send stdout to /dev/null

AndrewAmmerlaan:
GnuPG removes the --output file on failure. Removing /dev/null breaks all kinds of things. So lets use a temporary file for the output instead.

If the file already exists then we remove it to prevent GnuPG from complaining about the file already existing and asking us if we want to choose a different --output file.

Closes: https://bugs.gentoo.org/912808

### DIFF
--- a/lib/portage/gpg.py
+++ b/lib/portage/gpg.py
@@ -35,7 +35,7 @@ class GPG:
             f"--homedir {self.signing_gpg_home} "
             f"--digest-algo {self.digest_algo} "
             f"--local-user {self.signing_gpg_key} "
-            "--output /dev/null /dev/null",
+            "--output - /dev/null",
         )
 
         if "gpg-keepalive" in self.settings.features:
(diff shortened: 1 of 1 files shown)


lib/portage/gpg.py
```diff
@@ -22,6 +23,7 @@ def __init__(self, settings):
         """
         Portage settings are needed to run GPG unlock command.
         """
+        self.GPG_tmp_file = "/tmp/portage_gpg_output"
```
floppym, line 26: Using a fixed file name in /tmp is a really bad idea, especially for a program that often runs as root.

I would suggest using [NamedTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) to securely create a unique filename on every invocation of `unlock()`.

floppym:
Upon further thought, I think maybe we should just pass `--stdout -` and redirect stdout to /dev/null as suggested by upstream. That can be accomplished by passing `stdout=subprocess.DEVNULL` to `subprocess.Popen`.

AndrewAmmerlaan:
> Upon further thought, I think we should just pass `--output -` and redirect stdout to /dev/null as suggested by gnupg upstream. That can be accomplished by passing `stdout=subprocess.DEVNULL` to `subprocess.Popen`.

Sure, this is also fine by me. 
This certainly makes the patch smaller, and avoids having to deal with the situation where the file might already exist.

floppym (approved):

PR #1085, opened 2023-08-30
Run PORTAGE_TRUST_HELPER before remote binary package operations

akhuettel:
Tested to do what it's supposed to do...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,9 @@ jobs:
           python -m pip install --upgrade pip
           # setuptools needed for 3.12+ because of https://github.com/mesonbuild/meson/issues/7702.
           python -m pip install pytest setuptools
+
+          # symlink /bin/true to /usr/bin/getuto (or do we want to grab the script from github?)
+          sudo ln -s /bin/true /usr/bin/getuto
       - name: Test meson install --destdir /tmp/install-root
         run: |
           echo -e "[binaries]\npython = '$(command -v python)'" > /tmp/native.ini
--- a/lib/portage/dbapi/bintree.py
+++ b/lib/portage/dbapi/bintree.py
@@ -1236,6 +1236,9 @@ class binarytree:
 
     def _run_trust_helper(self):
         portage_trust_helper = self.settings.get("PORTAGE_TRUST_HELPER", "true")
+        if portage_trust_helper == "true":
+            return 0
+
         # getuto is a shell script...
         ret = os.waitstatus_to_exitcode(os.system(portage_trust_helper))
         if ret == 127:
--- a/lib/portage/dbapi/bintree.py
+++ b/lib/portage/dbapi/bintree.py
@@ -1243,12 +1243,14 @@ class binarytree:
         if ret.returncode == 127:
             raise OSError(
                 _(
-                    "Did not find trust helper. Install app-portage/getuto or set PORTAGE_TRUST_HELPER=\"\""
+                    'Did not find trust helper. Install app-portage/getuto or set PORTAGE_TRUST_HELPER=""'
                 )
             )
         elif ret.returncode != 0:
             raise OSError(
-                _("Failed to run trust helper executable for binary package verification: Error ")
+                _(
+                    "Failed to run trust helper executable for binary package verification: Error "
+                )
                 + str(ret.returncode)
             )
 
--- a/lib/portage/dbapi/bintree.py
+++ b/lib/portage/dbapi/bintree.py
@@ -1239,8 +1239,7 @@ class binarytree:
         if portage_trust_helper == "true":
             return 0
 
-        # getuto is a shell script...
-        ret = subprocess.run(portage_trust_helper, shell=True)
+        ret = subprocess.run(portage_trust_helper)
         if ret.returncode == 127:
             raise OSError(
                 _(
@@ -1249,7 +1248,7 @@ class binarytree:
             )
         elif ret.returncode != 0:
             raise OSError(
-                _("Failed to run trust helper for binary package verification: Error ")
+                _("Failed to run trust helper executable for binary package verification: Error ")
                 + str(ret.returncode)
             )
 
--- a/lib/portage/dbapi/bintree.py
+++ b/lib/portage/dbapi/bintree.py
@@ -1234,11 +1234,32 @@ class binarytree:
 
         return pkgindex if update_pkgindex else None
 
+    def _run_trust_helper(self):
+        portage_trust_helper = self.settings.get("PORTAGE_TRUST_HELPER", "true")
+        # getuto is a shell script...
+        ret = os.waitstatus_to_exitcode(os.system(portage_trust_helper))
+        if ret == 127:
+            raise OSError(
+                _(
+                    "Did not find trust helper. Install app-portage/getuto or set PORTAGE_TRUST_HELPER=true"
+                )
+            )
+        elif ret != 0:
+            raise OSError(
+                _("Failed to run trust helper for binary package verification: Error ")
+                + str(ret)
+            )
+
     def _populate_remote(self, getbinpkg_refresh=True):
         self._remote_has_index = False
         self._remotepkgs = {}
 
         if "binpkg-request-signature" in self.settings.features:
+            # This is somewhat broken, we *should* run the trust helper always
+            # when binpackages are involved, not only when we refuse unsigned
+            # ones. (If the keys have expired we end up refusing signed but
+            # technically invalid packages...)
+            self._run_trust_helper()
             gpkg_only = True
         else:
             gpkg_only = False
--- a/lib/portage/dbapi/bintree.py
+++ b/lib/portage/dbapi/bintree.py
@@ -1235,15 +1235,15 @@ class binarytree:
         return pkgindex if update_pkgindex else None
 
     def _run_trust_helper(self):
-        portage_trust_helper = self.settings.get("PORTAGE_TRUST_HELPER", "true")
-        if portage_trust_helper == "true":
-            return 0
+        portage_trust_helper = self.settings.get("PORTAGE_TRUST_HELPER", "")
+        if portage_trust_helper == "":
+            return
 
         ret = subprocess.run(portage_trust_helper)
         if ret.returncode == 127:
             raise OSError(
                 _(
-                    "Did not find trust helper. Install app-portage/getuto or set PORTAGE_TRUST_HELPER=true"
+                    "Did not find trust helper. Install app-portage/getuto or set PORTAGE_TRUST_HELPER=\"\""
                 )
             )
         elif ret.returncode != 0:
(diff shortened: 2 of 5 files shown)


lib/portage/dbapi/bintree.py
```diff
@@ -1234,7 +1234,16 @@ def _populate_local(self, reindex=True):
...
 
+    def _run_trust_helper(self):
+        portage_trust_helper = self.settings.get("PORTAGE_TRUST_HELPER")
+        ret = os.system(portage_trust_helper)
+        if ret != 0:
+            raise OSError(_("Failed to run trust helper for binary package verification"))
+        return
+
     def _populate_remote(self, getbinpkg_refresh=True):
+        self._run_trust_helper()
```
thesamesam, line 1245: Let's conditionalise this on `FEATURES="binpkg-request-signature"` (maybe `FEATURES="binpkg-request-signature"` or `FEATURES="binpkg-signing"`, but I don't see that much point there)?

If you agree with that, we can shove it under the `if` on line 1250.

lib/portage/dbapi/bintree.py
```diff
@@ -1234,7 +1234,16 @@ def _populate_local(self, reindex=True):
 
         return pkgindex if update_pkgindex else None
 
+    def _run_trust_helper(self):
+        portage_trust_helper = self.settings.get("PORTAGE_TRUST_HELPER")
+        ret = os.system(portage_trust_helper)
+        if ret != 0:
```
thesamesam, line 1240: Let's maybe give a more informative error if it doesn't exist, and tell the user to either install getuto or set PORTAGE_TRUST_HELPER to `true`?

lib/portage/dbapi/bintree.py
```diff
@@ -1234,7 +1234,16 @@ def _populate_local(self, reindex=True):
 
         return pkgindex if update_pkgindex else None
 
+    def _run_trust_helper(self):
+        portage_trust_helper = self.settings.get("PORTAGE_TRUST_HELPER")
+        ret = os.system(portage_trust_helper)
```
thesamesam, line 1239: Try `subprocess.run`?

lib/portage/dbapi/bintree.py
```diff
@@ -1234,7 +1234,16 @@ def _populate_local(self, reindex=True):
 
         return pkgindex if update_pkgindex else None
 
+    def _run_trust_helper(self):
+        portage_trust_helper = self.settings.get("PORTAGE_TRUST_HELPER")
```
thesamesam, line 1238: Maybe we should set a default value here too.

If we want it to succeed if the user nukes it from `make.globals` (I think I want to do this, because users might just delete it because they want something to work NOW!!!):
```suggestion
        portage_trust_helper = self.settings.get("PORTAGE_TRUST_HELPER", "true")
```

If we want it to default to getuto even in the source (not so sure about this - would also need some Prefix magic, but not sure if there's much point in this at all):
```suggestion
        portage_trust_helper = self.settings.get("PORTAGE_TRUST_HELPER", "/usr/bin/getuto")
```

lib/portage/dbapi/bintree.py
akhuettel: getuto is a shell script, so (I suspect) we'd need to set `shell=True` ?

We could capture the output into a log file (do we want to?)

lib/portage/dbapi/bintree.py
akhuettel: Let's use `true`. Makes for one less point of failure (and this is only a "second line of defense" anyway).

lib/portage/dbapi/bintree.py
thesamesam: Bleh, okay. Let's put it under that `if` I mentioned, but put a comment there. I think we should circle back and change that behaviour you mention at some point, not sure.

akhuettel: OK done.

lib/portage/dbapi/bintree.py
akhuettel: Done

lib/portage/dbapi/bintree.py
akhuettel: Added a comment that os.system is fine here.

lib/portage/dbapi/bintree.py
```diff
@@ -1234,11 +1234,33 @@ def _populate_local(self, reindex=True):
 
         return pkgindex if update_pkgindex else None
 
+    def _run_trust_helper(self):
+        portage_trust_helper = self.settings.get("PORTAGE_TRUST_HELPER", "true")
+        # getuto is a shell script...
+        ret = os.system(portage_trust_helper)
+        if ret == 32512:
+            # lsb 0 (no error in shell process), msb 127 (command not found)
```
thesamesam, line 1242: Thanks for the comment (people often like to put magic numbers like they're obvious)

lib/portage/dbapi/bintree.py
thesamesam: ok

lib/portage/dbapi/bintree.py
```diff
@@ -1234,11 +1234,33 @@ def _populate_local(self, reindex=True):
 
         return pkgindex if update_pkgindex else None
 
+    def _run_trust_helper(self):
+        portage_trust_helper = self.settings.get("PORTAGE_TRUST_HELPER", "true")
+        # getuto is a shell script...
+        ret = os.system(portage_trust_helper)
+        if ret == 32512:
```
thesamesam, line 1241: Could you try https://docs.python.org/3/library/os.html#os.waitstatus_to_exitcode?

thesamesam (approved):
Please add it to the make.conf man page, make that return code tweak, then lgtm - thank you!

lib/portage/dbapi/bintree.py
akhuettel: Done

lib/portage/dbapi/bintree.py
```diff
@@ -1234,11 +1234,32 @@ def _populate_local(self, reindex=True):
 
         return pkgindex if update_pkgindex else None
 
+    def _run_trust_helper(self):
+        portage_trust_helper = self.settings.get("PORTAGE_TRUST_HELPER", "true")
+        # getuto is a shell script...
+        ret = os.waitstatus_to_exitcode(os.system(portage_trust_helper))
```
floppym, line 1240: New code should use [subprocess.run](https://docs.python.org/3/library/subprocess.html#subprocess.run).

If the command needs to be passed to a shell, pass `shell=True` as a keyword argument.

man/make.conf.5
```diff
@@ -1208,6 +1208,15 @@ Defaults to /var/tmp.
 
 This should not be set to point anywhere under location of any repository.
 .TP
+\fBPORTAGE_TRUST_HELPER\fR = \fI[path]\fR
+Defines an executable file which initializes and maintains
```
floppym, line 1212: This says PORTAGE_TRUST_HELPER is an "executable file", but it actually seems to be a shell command line.

lib/portage/dbapi/bintree.py
```diff
@@ -1234,11 +1234,32 @@ def _populate_local(self, reindex=True):
...
+        if ret == 127:
+            raise OSError(
```
floppym, line 1242: Application code should almost never raise `OSError`. This is only meant to be raised from system calls.

lib/portage/dbapi/bintree.py
```diff
@@ -1234,11 +1234,32 @@ def _populate_local(self, reindex=True):
 
         return pkgindex if update_pkgindex else None
 
+    def _run_trust_helper(self):
+        portage_trust_helper = self.settings.get("PORTAGE_TRUST_HELPER", "true")
```
floppym, line 1238: Instead of executing "true" when PORTAGE_TRUST_HELPER is unset, we should just return early.

.github/workflows/ci.yml
```diff
@@ -38,6 +38,9 @@ jobs:
           python -m pip install --upgrade pip
           # setuptools needed for 3.12+ because of https://github.com/mesonbuild/meson/issues/7702.
           python -m pip install pytest setuptools
+
+          # symlink /bin/true to /usr/bin/getuto (or do we want to grab the script from github?)
+          sudo ln -s /bin/true /usr/bin/getuto
```
floppym, line 43: Another possible alternative would be to set PORTAGE_TRUST_HELPER in the environment.

lib/portage/dbapi/bintree.py
akhuettel: So what hell should I raise instead?

lib/portage/dbapi/bintree.py
akhuettel: Please resolve that with @thesamesam ...

.github/workflows/ci.yml
akhuettel: Yes. 
In either case, my logic here was that getuto has its own tests (thanks sam); here we test portage and want to avoid spurious failures.

lib/portage/dbapi/bintree.py
akhuettel: Done

man/make.conf.5
akhuettel: Fixed

lib/portage/dbapi/bintree.py
akhuettel: Done

lib/portage/dbapi/bintree.py
floppym: I see OSError being raised elsewhere in this file, so I guess it's ok.

A more specific option would be [FileNotFoundError](https://docs.python.org/3/library/exceptions.html#FileNotFoundError), which would normally be raised if `subprocess.run` cannot find the program to be executed. Using `shell=True` changes the behavior because we get bash's return code instead.

floppym: If we wanted a more application-specific exception, it looks like common exception classes are defined in lib/portage/exception.py. Several of them are imported at the top of this file.

.github/workflows/ci.yml
floppym: Right. I'm just saying that instead of a symlink, we could set PORTAGE_TRUST_HELPER=true to turn it into a noop.

If the symlink is easier to manage, that's fine.

lib/portage/dbapi/bintree.py
```diff
@@ -1234,11 +1234,35 @@ def _populate_local(self, reindex=True):
...
+        if portage_trust_helper == "true":
+            return 0
```
floppym, line 1240:
```suggestion
        portage_trust_helper = self.settings.get("PORTAGE_TRUST_HELPER", None)
        if portage_trust_helper is None:
            return
```

floppym: This would allow the user to disable the setting by setting it to empty string.

akhuettel: Done, good idea.

PR #1086, opened 2023-09-01
emerge: ensure paths are UTF-8 encoded in _needs_move()

gcarq:
Bug: https://bugs.gentoo.org/913103


floppym (approved):
Looks good, thanks!

PR #1087, opened 2023-09-04
emerge: allow --onlydeps merge if root package(s) masked by license

tgbugs:
This commit makes it possible to merge dependencies for packages that are masked by license since the masked package is not actually going to be installed.

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -5492,7 +5492,13 @@ class depgraph:
         # set below is reserved for cases where there are *zero* other
         # problems. For reference, see backtrack_depgraph, where it skips the
         # get_best_run() call when success_without_autounmask is True.
-        if self._have_autounmask_changes():
+
+        if onlydeps and self._dynamic_config._needed_license_changes:
+            # needed license changes should only be fatal for packages that
+            # would actually be installed
+            return True, myfavorites
+
+        elif self._have_autounmask_changes():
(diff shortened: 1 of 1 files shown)


tgbugs:
Hrm. For some reason when I test this in a musl docker image I don't see the same behavior I see on my development system with `emerge --onlydeps -ep media-fonts/corefonts`. Looks like I will have to do a bit more digging :/.

lib/_emerge/depgraph.py
```diff
@@ -5250,7 +5250,13 @@ def _resolve(self, myfavorites):
         # set below is reserved for cases where there are *zero* other
         # problems. For reference, see backtrack_depgraph, where it skips the
         # get_best_run() call when success_without_autounmask is True.
-        if self._have_autounmask_changes():
+
+        if onlydeps and self._dynamic_config._needed_license_changes:
+            # needed license changes should only be fatal for packages that
+            # would actually be installed
+            return True, myfavorites
```
zmedico, line 5257: You've ignored non-license autounmask changes here. It would only make sense to return True here if there were no non-license autounmask changes. Ultimately, I think it will be much cleaner to set empty LICENSE metadata for the onlydeps package instance, rather than trigger autoumask for a LICENSE that you intend to ignore anyway.

tgbugs: Setting license metadata to empty is an excellent suggestion. I'll look into it. Thanks!

PR #1088, opened 2023-09-04
ebuild: inject implicit libc RDEPEND

thesamesam:
commit 739202e405cc0bb7ae8e99766e875efac639f398 (HEAD -> binpkg-dep-hack, sam/binpkg-dep-hack)
Author: Sam James <sam@gentoo.org>
Date:   Mon Sep 4 16:38:18 2023 +0100

    ebuild: inject implicit libc RDEPEND

    Inject >=${LIBC_PROVIDER}-${VERSION_OF_LIBC_PROVIDER} into RDEPEND.

    We already try to upgrade the virtual/libc provider and its deps aggressively
    but that's not so helpful if there's a binpkg for, say, one of its deps available
    built against a newer glibc - which ends up breaking the system because our
    glibc isn't new enough (symbol versioning).

    This is a long-standing problem for binpkg users and one of the last showstoppers
    for being able to use them reliably.

    Note that this applies to source installs too, although it matters less there;
    it'll make downgrading libc a bit harder for people who want to do that, but users
    are already prevented from doing that (pkg_* check) for glibc, and I don't really
    see it as a bad thing to effectively propagate this to other libcs.

    To fully solve the problem, we should arguably do at least one of the following:
    1) Implicit >= on anything (not just libc) which installs ELF (at least a PROVIDES);
    2) Implement the suggestion in bug #753500 based on analysis of used versioned symbols (*).

    But libc is really the critical one and the one where things explode pretty badly,
    especially combined with us trying to Do The Right Thing for non-binpkg cases
    (aggressively upgrading libc before anything else). The other cases don't matter
    so much as they get upgraded later and by that point, the library is usually done.

    (It's not really clear if downgrading musl works and even if it is supported,
    I'm not really sure it's a supported case at all, so I'm not bothering to carve
    out an exception here. It'd make this far less elegant and I don't see any benefit
    to doing so.)

    (*) util-linux is one of the examples I have in mind here as justification
    for either point.

    Bug: https://bugs.gentoo.org/753500
    Bug: https://bugs.gentoo.org/913628

commit 7cdb42307e285320e20d2e8c977bcbdff316b2a8
Author: Sam James <sam@gentoo.org>
Date:   Mon Sep 4 16:37:24 2023 +0100

    ebuild: refactor flushing vdb keys

    Refactor how we flush these VDB keys in build-info/ to make it easier
    to implement bug #913628: it was a pain with the forced-append of a newline
    even if we might tamper with the contents later on, so just postpone flushing
    to disk until the end. It saves some repetition too, so double win.

    Bug: https://bugs.gentoo.org/913628

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
           set -xe
           echo "force-unsafe-io" | sudo tee /etc/dpkg/dpkg.cfg.d/force-unsafe-io
           sudo apt-get update -q
-          sudo apt-get install -qy --no-install-recommends libxslt-dev libxml2-dev libxml2-utils meson zstd
+          sudo apt-get install -qy --no-install-recommends libxslt-dev libxml2-dev libxml2-utils meson pax-utils zstd
 
           # Patch Ubuntu's old Meson to fix pypy-3.9 detection.
           curl -s -f https://github.com/mesonbuild/meson/commit/2540ad6e9e08370ddd0b6753fdc9314945a672f0.patch | sudo patch -d /usr/lib/python3/dist-packages -p1 --no-backup-if-mismatch
--- a/lib/portage/tests/emerge/test_libc_dep_inject.py
+++ b/lib/portage/tests/emerge/test_libc_dep_inject.py
@@ -35,7 +35,8 @@ class LibcDepInjectEmergeTestCase(TestCase):
 
         src_install() {
             insinto /usr/bin
-            # We need an ELF binary for the injection to trigger
+            # We need an ELF binary for the injection to trigger, so
+            # use ${BASH} given we know it must be around for running ebuilds.
             cp "${BASH}" "${ED}"/usr/bin/${PN} || die
         }
         """


thesamesam:
I'm going to see about finding a cleanish way to handle the NEEDED bit.

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2560,17 +2552,51 @@ def _post_src_install_write_metadata(settings):
...
-                ),
-                mode="w",
-                encoding=_encodings["repo.content"],
+
+            metadata_buffer[k] = v
+
+    # We could skip this for non-binpkgs but there doesn't seem to be much
+    # value in that, as users shouldn't downgrade libc anyway.
+    #
+    # TODO: We could skip the injection for packages installing no ELF files
```
thesamesam, line 2561: I'm going to see about doing this (the conditional part, not the "start doing everything in Python" bit), although I'm worried it may be messier than the PR currently is.

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2560,17 +2552,51 @@ def _post_src_install_write_metadata(settings):
...
+                "vartree"
+            ].dbapi.match(atom)[0]
+        except IndexError:
+            continue
+
+        injected_libc_depstring.append(f">={libc_realized_atom}")
+
+    # For RDEPEND, we want an implicit dependency on >=${PROVIDER_OF_LIBC}
+    # to avoid runtime breakage when merging binpkgs, see bug #753500.
+    if injected_libc_depstring and settings.configdict["pkg"].get("RDEPEND"):
```
thesamesam, line 2583: I'm going to see about only doing this for binpkgs to avoid the changed-deps issue Zac mentions in the bug.

akhuettel:
Sorry, first test failed rather quickly.

`
(amd64 chroot) amd64-stable ~ # emerge -eav --buildpkg world
`

`
Total: 323 packages (323 reinstalls), Size of downloads: 0 KiB

Would you like to merge these packages? [Yes/No] y
`

`
>>> Emerging (1 of 323) sys-devel/gnuconfig-20230731::gentoo
>>> Emerging (2 of 323) virtual/libintl-0-r2::gentoo
>>> Emerging (3 of 323) sys-libs/ncurses-6.4_p20230401::gentoo
>>> Emerging (4 of 323) virtual/libiconv-0-r2::gentoo
>>> Emerging (5 of 323) acct-group/root-0-r1::gentoo
>>> Jobs: 0 of 323 complete, 5 running              Load avg: 2.50, 2.38, 2.38Exception in callback AsynchronousTask._exit_listener_cb(<bound method...7fab6923fa30>>)
handle: <Handle AsynchronousTask._exit_listener_cb(<bound method...7fab6923fa30>>)>
  File "/usr/lib/python3.11/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/lib/python3.11/site-packages/_emerge/AsynchronousTask.py", line 209, in _exit_listener_cb
    listener(self)
  File "/usr/lib/python3.11/site-packages/_emerge/EbuildPhase.py", line 288, in _ebuild_exit
    self._ebuild_exit_unlocked(ebuild_process)
  File "/usr/lib/python3.11/site-packages/_emerge/EbuildPhase.py", line 352, in _ebuild_exit_unlocked
    _post_src_install_write_metadata(settings)
  File "/usr/lib/python3.11/site-packages/portage/package/ebuild/doebuild.py", line 2584, in _post_src_install_write_metadata
    metadata_buffer["RDEPEND"] += f" {' '.join(injected_libc_depstring)}"
    ~~~~~~~~~~~~~~~^^^^^^^^^^^
KeyError: 'RDEPEND'
Terminated
(amd64 chroot) amd64-stable ~ # 
`

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2553,6 +2555,37 @@ def _post_src_install_write_metadata(settings):
...
```
akhuettel, line 2583: I don't really know what I'm doing here, but I'm continuing testing with
```
if injected_libc_depstring and "RDEPEND" in metadata_buffer:
```

akhuettel: And now it's happily adding the dependency in the VDB:
```
pinacolada /var/lib/machines/amd64-stable/var/db/pkg/virtual/perl-Encode-3.190.0 # cat RDEPEND 
|| ( =dev-lang/perl-5.38* ~perl-core/Encode-3.190.0 ) dev-lang/perl:0/5.38= !<perl-core/Encode-3.190.0 !>perl-core/Encode-3.190.0-r999 >=sys-libs/glibc-2.37-r3
```

akhuettel:
What happens when glibc RDEPENDs on itself? 🤔

thesamesam:
> What happens when glibc RDEPENDs on itself? 🤔

Good point! We need to skip adding it if it resolves to any of the (actualised) providers of virtual/libc.

lib/portage/package/ebuild/doebuild.py
thesamesam: Does it work correctly if the package has no RDEPEND beforehand?

akhuettel: Yes, checked that too
```
pinacolada /var/lib/machines/amd64-stable/var/db/pkg/dev-libs/libpipeline-1.5.7 # cat RDEPEND 
>=sys-libs/glibc-2.37-r3
```

akhuettel:
> > What happens when glibc RDEPENDs on itself? 🤔
> 
> Good point! We need to skip adding it if it resolves to any of the (actualised) providers of virtual/libc.

Easier is maybe to skip if it resolves to the package itself?

lib/portage/package/ebuild/doebuild.py
akhuettel: Hmm, I built a binary package of `app-arch/tar` with this, and now the `Packages` file has the following in the tar entry:
```
RDEPEND: virtual/acl >=sys-libs/glibc-2.37-r3 >=sys-libs/glibc-2.37-r3
```
So somehow the glibc entry duplicated.

akhuettel: Confirmed with `zstd`, building it again with binpkg creation doubles the entry.
Does *not* become 3 or 4 entries if I repeat that though.

akhuettel:
Additional problem...
```
(amd64 chroot) amd64-stable ~ # emerge -uDNav world

 * IMPORTANT: 4 news items need reading for repository 'gentoo'.
 * Use eselect news read to view new items.


Dependency resolution took 4.51 s.


Total: 0 packages, Size of downloads: 0 KiB

Nothing to merge; quitting.

(amd64 chroot) amd64-stable ~ # emerge -uDNav --changed-deps world

 * IMPORTANT: 4 news items need reading for repository 'gentoo'.
 * Use eselect news read to view new items.


Dependency resolution took 3.25 s.

[ebuild   R    ] sys-devel/gnuconfig-20230731::gentoo  0 KiB
[ebuild   R    ] virtual/libintl-0-r2::gentoo  ABI_X86="(64) -32 (-x32)" 0 KiB
[ebuild   R    ] sys-libs/ncurses-6.4_p20230401:0/6::gentoo  USE="cxx (split-usr) stack-realign (tinfo) -ada -debug -doc -gpm -minimal -profile -static-libs -test -trace -verify-sig" ABI_X86="(64) -32 (-x32)" 0 KiB
[ebuild   R    ] virtual/libiconv-0-r2::gentoo  ABI_X86="(64) -32 (-x32)" 0 KiB
[ebuild   R    ] acct-group/root-0-r1::gentoo  0 KiB
[ebuild   R    ] acct-group/man-0-r2::gentoo  0 KiB
[ebuild   R    ] acct-group/sshd-0-r2::gentoo  0 KiB
(...)
[ebuild   R    ] app-alternatives/bc-0::gentoo  USE="gnu -gh" 0 KiB

Total: 323 packages (323 reinstalls), Size of downloads: 0 KiB

Would you like to merge these packages? [Yes/No] n

Quitting.

(amd64 chroot) amd64-stable ~ # 
```

akhuettel:
* Made a snapshot of the whole system,
* Updated glibc to 2.38,
* Made a binpackage of iptables,
* Copied that binpackage into the glibc-2.37 snapshot

```
(amd64 chroot) amd64-stable-01-glibc237 ~ # emerge -a1 iptables

 * IMPORTANT: 4 news items need reading for repository 'gentoo'.
 * Use eselect news read to view new items.


Dependency resolution took 1.78 s.

[ebuild   R    ] net-firewall/iptables-1.8.9:0/1.8.3::gentoo  USE="(split-usr) -conntrack -netlink -nftables -pcap -static-libs" 0 KiB

Total: 1 package (1 reinstall), Size of downloads: 0 KiB

Would you like to merge these packages? [Yes/No] n

Quitting.

(amd64 chroot) amd64-stable-01-glibc237 ~ # emerge -a1k iptables

 * IMPORTANT: 4 news items need reading for repository 'gentoo'.
 * Use eselect news read to view new items.


Dependency resolution took 0.83 s.

[ebuild   R    ] net-firewall/iptables-1.8.9:0/1.8.3::gentoo  USE="(split-usr) -conntrack -netlink -nftables -pcap -static-libs" 0 KiB

Total: 1 package (1 reinstall), Size of downloads: 0 KiB

!!! The following binary packages have been ignored due to changed dependencies:

     net-firewall/iptables-1.8.9::gentoo

NOTE: The --binpkg-changed-deps=n option will prevent emerge
      from ignoring these binary packages if possible.
      Using --binpkg-changed-deps=y will silence this warning.

Would you like to merge these packages? [Yes/No] n

Quitting.

(amd64 chroot) amd64-stable-01-glibc237 ~ # emerge -a1k --binpkg-changed-deps=n iptables

 * IMPORTANT: 4 news items need reading for repository 'gentoo'.
 * Use eselect news read to view new items.


Dependency resolution took 1.01 s.

[ebuild   R    ] net-firewall/iptables-1.8.9:0/1.8.3::gentoo  USE="(split-usr) -conntrack -netlink -nftables -pcap -static-libs" 0 KiB

Total: 1 package (1 reinstall), Size of downloads: 0 KiB

Would you like to merge these packages? [Yes/No] n

Quitting.

(amd64 chroot) amd64-stable-01-glibc237 ~ # emerge -a1K --binpkg-changed-deps=n iptables

 * IMPORTANT: 4 news items need reading for repository 'gentoo'.
 * Use eselect news read to view new items.


Dependency resolution took 0.58 s.


emerge: there are no binary packages to satisfy ">=sys-libs/glibc-2.38-r1".
(dependency required by "net-firewall/iptables-1.8.9::gentoo" [binary])
(dependency required by "iptables" [argument])
(amd64 chroot) amd64-stable-01-glibc237 ~ # 
```
So, apart from the fact that we also need to handle `--binpkg-changed-deps` somehow, everything works as expected! :+1:

akhuettel:
Looks good.
```
(amd64 chroot) amd64-stable-01-glibc237 ~ # emerge -a1 iptables

 * IMPORTANT: 4 news items need reading for repository 'gentoo'.
 * Use eselect news read to view new items.


Dependency resolution took 2.70 s.

[ebuild   R    ] net-firewall/iptables-1.8.9:0/1.8.3::gentoo  USE="(split-usr) -conntrack -netlink -nftables -pcap -static-libs" 0 KiB

Total: 1 package (1 reinstall), Size of downloads: 0 KiB

Would you like to merge these packages? [Yes/No] n

Quitting.
```
Normal source merge would rebuild iptables.
```
(amd64 chroot) amd64-stable-01-glibc237 ~ # emerge -a1 --changed-deps iptables

 * IMPORTANT: 4 news items need reading for repository 'gentoo'.
 * Use eselect news read to view new items.


Dependency resolution took 1.35 s.


Total: 0 packages, Size of downloads: 0 KiB

Nothing to merge; quitting.

```
With `--changed-deps` no rebuild is done. :+1: 
```
(amd64 chroot) amd64-stable-01-glibc237 ~ # emerge -a1k iptables

 * IMPORTANT: 4 news items need reading for repository 'gentoo'.
 * Use eselect news read to view new items.


Dependency resolution took 1.04 s.

[ebuild   R    ] net-firewall/iptables-1.8.9:0/1.8.3::gentoo  USE="(split-usr) -conntrack -netlink -nftables -pcap -static-libs" 0 KiB

Total: 1 package (1 reinstall), Size of downloads: 0 KiB

Would you like to merge these packages? [Yes/No] n

Quitting.
```
Doesn't try to use the binpackage with >=2.38 dependency. Good. :+1: 
```
(amd64 chroot) amd64-stable-01-glibc237 ~ # emerge -a1K iptables

 * IMPORTANT: 4 news items need reading for repository 'gentoo'.
 * Use eselect news read to view new items.


Dependency resolution took 1.03 s.


emerge: there are no binary packages to satisfy ">=sys-libs/glibc-2.38-r1".
(dependency required by "net-firewall/iptables-1.8.9::gentoo" [binary])
(dependency required by "iptables" [argument])
```
Binary only installation complains about missing binpkg for glibc-2.38. Good. :+1: 
```
(amd64 chroot) amd64-stable-01-glibc237 ~ # emerge -a1k --binpkg-changed-deps=y iptables

 * IMPORTANT: 4 news items need reading for repository 'gentoo'.
 * Use eselect news read to view new items.


Dependency resolution took 1.76 s.

[ebuild   R    ] net-firewall/iptables-1.8.9:0/1.8.3::gentoo  USE="(split-usr) -conntrack -netlink -nftables -pcap -static-libs" 0 KiB

Total: 1 package (1 reinstall), Size of downloads: 0 KiB

Would you like to merge these packages? [Yes/No] n

Quitting.

(amd64 chroot) amd64-stable-01-glibc237 ~ # emerge -a1k --binpkg-changed-deps=n iptables

 * IMPORTANT: 4 news items need reading for repository 'gentoo'.
 * Use eselect news read to view new items.


Dependency resolution took 1.05 s.

[ebuild   R    ] net-firewall/iptables-1.8.9:0/1.8.3::gentoo  USE="(split-usr) -conntrack -netlink -nftables -pcap -static-libs" 0 KiB

Total: 1 package (1 reinstall), Size of downloads: 0 KiB

Would you like to merge these packages? [Yes/No] n

Quitting.

(amd64 chroot) amd64-stable-01-glibc237 ~ # 
```
These both seem to have no effect, which is OK.

thesamesam:
Only thing left for now is to suppress the injection for packages not installing any ELF.

akhuettel:
FWIW I went through some test procedure (similar to the ones described above) for the current state together with @thesamesam and everything worked exactly as expected and intended.

akhuettel (approved):

lib/portage/dep/libc.py
```diff
@@ -0,0 +1,69 @@
...
+    # --changed-deps results far too often.
+    #
+    # This penalizes a bit the case where someone adds a
+    # minimum (or maximum) version of libc explicitly in an ebuild
+    # without a new revision, but that's extremely rare, and doesn't
+    # feel like it changes the balance for what we prefer here.
+    for libc_dep in libc_deps:
+        for i, x in enumerate(dep_struct):
+            if isinstance(x, list):
+                strip_libc_deps(x, libc_deps)
```
zmedico, line 66: Maybe skip recursion here, since generated deps are always on the top-level.

lib/portage/dep/libc.py
```diff
@@ -0,0 +1,69 @@
...
+            elif isinstance(x, Atom) and libc_dep.without_slot.match(x):
```
zmedico, line 67: We could move the libc_deps loop down here, like:
```python
any(libc_dep.without_slot.match(x) for libc_dep in libc_deps)
```

lib/portage/dep/libc.py
thesamesam: Oh yes, done!

lib/portage/dep/libc.py
thesamesam: Nice! I've dropped the Atom check too because I don't think we need it now, but let me know if it looks off to you.

lib/portage/dep/libc.py
```diff
@@ -0,0 +1,65 @@
...
+    # We can't go off versions, even though it'd be more precise
+    # (see below), because we'd end up with FPs and unnecessary
+    # --changed-deps results far too often.
+    #
+    # This penalizes a bit the case where someone adds a
+    # minimum (or maximum) version of libc explicitly in an ebuild
+    # without a new revision, but that's extremely rare, and doesn't
+    # feel like it changes the balance for what we prefer here.
+    for i, x in enumerate(dep_struct):
+        if any(libc_dep.without_slot.match(x) for libc_dep in libc_deps):
```
zmedico, line 63: Here, x could be a list, an Atom, or an operator like ||.

thesamesam: Oops, sorry, I get it now (I made a mistake at https://github.com/gentoo/portage/pull/1088#discussion_r1337748901 and wasn't thinking).

lib/portage/dep/libc.py
```diff
@@ -0,0 +1,70 @@
...
+    #
+    # This penalizes a bit the case where someone adds a
+    # minimum (or maximum) version of libc explicitly in an ebuild
+    # without a new revision, but that's extremely rare, and doesn't
+    # feel like it changes the balance for what we prefer here.
+    for i, x in enumerate(dep_struct):
+        # We only need to bother if x is an Atom because we know the deps
+        # we inject are simple & flat.
+        if isinstance(x, Atom) and any(
+            libc_dep.without_slot.match(x) for libc_dep in libc_deps
```
zmedico, line 67: I've just realized that it's only by luck that this hasn't triggered an exception due to some assumption in match_from_list, since match_from_list is intended to match an atom against packages, not atoms against atoms.  For example, if you look through match_from_list you'll see that it assumes packages have a cpv_split attribute which atoms do not have. The function originally operated on simple "cpv" packages, and later gained support for Package and _pkg_str instances.

lib/_emerge/depgraph.py
```diff
@@ -2935,6 +2936,13 @@ def _changed_deps(self, pkg):
             else:
                 depvars = Package._runtime_keys
 
+            libc_deps = find_libc_deps(
```
zmedico, line 2939: It's wasteful to repeat this over an over, so we really should cache the result someplace like self._frozen_config.

lib/portage/dep/libc.py
zmedico: The Atom.intersects method is the closest we have to what you want here.

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2877,6 +2911,11 @@ def _post_src_install_soname_symlinks(mysettings, out):
         if f is not None:
             f.close()
 
+    # We do RDEPEND mangling here instead of the natural location
+    # in _post_src_install_write_metadata because NEEDED hasn't been
+    # written yet at that point.
+    _inject_libc_dep(build_info_dir, mysettings)
```
zmedico, line 2917: I feel like this should at least be conditional on the presence of an ELF file. Even better if we check for an soname dependency that is satisfied by libc.

thesamesam: It should be already here -- we only get to that point if there's a `NEEDED` -> we have an ELF file. The function returns early above it otherwise.

(But the previous behaviour of this PR was to not do that - you are right it's important).

Let me know if I'm missing something here though

lib/portage/dep/libc.py
thesamesam: I'm wondering what to do about this. The intention is to only call `strip_libc_deps` with "realized" dependencies, which should be simple atoms.

So, the current approach is okay only if `find_libc_deps` can never return a complex dependency with an operator like '=*'.

Can we rely on `expand_new_virt` with portdb to only return things we can use as an atom, or no?

zmedico: The intention is to only call strip_libc_deps with "realized" dependencies, which should be simple atoms.

You say "realized" but strip_libc_deps is currently called with realized=False just prior to strip_libc_deps.

`expand_new_virt` with portdb will always return a sequence of valid atoms, which may include atoms with any operators including '=*'.

lib/portage/package/ebuild/doebuild.py
zmedico: Okay, looks good. Thanks!

lib/portage/dep/libc.py
akhuettel: @zmedico do you have a suggestion on how do do this correctly?

zmedico: I think we can do it with the `intersects` method instead of the `match` method , at least until someone comes up with a better idea.

thesamesam: Thanks. I have a bunch of local changes I've been chipping away at but gone around in circles a bit. I'll try tidy them up over the weekend.

I also need to see if we have a test I can borrow for merging a binpkg so I can inspect its metadata to check the whole thing works OK.

thesamesam: I think we're good here now, fingers crossed.

lib/_emerge/depgraph.py
thesamesam: This is the last TODO, will do that now.

thesamesam:
Playing with something else, I noticed:
```
$ cat /tmp/tmpcm2inalt/var/db/pkg/dev-libs/A-1/RDEPEND
[]
```

I'll check what's going on there.

thesamesam:
Now with more tests! 🎉

thesamesam:
OK, really ready, I think.

akhuettel:
I think there's still something missing, sorry.

Test env: nspawn from a stage3, stable portage 3.0.56-r1, with the 6 commits from this PR added as patches
Tests done: re-installed portage with the patches, then `emerge -a1 --buildpkg bash`

Result:
* Local VDB entry of bash has glibc in RDEPEND, good
* Packages cache file shows glibc in RDEPEND of the binary package, good
* `emerge -a1 --changed-deps world` wants to rebuild bash :(

Going outside now, but can provide more info later...

thesamesam:
ok, i can reproduce that, lemme try make a testcase

thesamesam:
ok, so at this point, I'm back to "intersects doesn't quite do what I want", but I'm not really sure how to fix it. I've pushed some failing test cases as well.

@zmedico could I ask for your help here?

lib/portage/dep/libc.py
```diff
@@ -0,0 +1,84 @@
...
+    # minimum (or maximum) version of libc explicitly in an ebuild
+    # without a new revision, but that's extremely rare, and doesn't
+    # feel like it changes the balance for what we prefer here.
+    for i, x in enumerate(dep_struct):
+        # We only need to bother if x is an Atom because we know the deps
+        # we inject are simple & flat.
+        if isinstance(x, Atom) and any(
+            x.intersects(libc_dep) for libc_dep in libc_deps
+        ):
+            del dep_struct[i]
```
zmedico, line 84: Tests pass with this change:
```diff
--- a/lib/portage/dep/libc.py
+++ b/lib/portage/dep/libc.py
@@ -75,10 +75,10 @@ def strip_libc_deps(dep_struct: list, libc_deps: set):
     # minimum (or maximum) version of libc explicitly in an ebuild
     # without a new revision, but that's extremely rare, and doesn't
     # feel like it changes the balance for what we prefer here.
-    for i, x in enumerate(dep_struct):
+    for i, x in reversed(list(enumerate(dep_struct))):
         # We only need to bother if x is an Atom because we know the deps
         # we inject are simple & flat.
         if isinstance(x, Atom) and any(
-            x.intersects(libc_dep) for libc_dep in libc_deps
+            x.cp == libc_dep.cp for libc_dep in libc_deps
         ):
             del dep_struct[i]
```

lib/_emerge/depgraph.py
```diff
@@ -2984,6 +2985,15 @@ def _changed_deps(self, pkg):
...
                 depvars = Package._runtime_keys
 
+            if not hasattr(self._frozen_config, "_libc_deps_cache"):
+                self._frozen_config._libc_deps_cache = find_libc_deps(
+                    self._frozen_config._trees_orig[pkg.root_config.settings["EROOT"]][
+                        "vartree"
+                    ].dbapi,
+                    False,
+                )
+            libc_deps = self._frozen_config._libc_deps_cache
```
zmedico, line 2995: Should make _libc_deps_cache a dict with key `pkg.root`.

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2856,6 +2850,48 @@ def _reapply_bsdflags_to_image(mysettings):
...
+        injected_libc_depstring.append(f">={libc_realized_atom}")
+
+    rdepend_file = os.path.join(build_info_dir, "RDEPEND")
+    # Slurp the existing contents because we need to mangle it a bit
+    # It'll look something like (if it exists):
+    # ```
+    # app-misc/foo dev-libs/bar
+    # <newline>
+    # ````
+    rdepend = []
```
zmedico, line 2875: Can initialize this to an empty string, None, or False since it's always reassigned later.

lib/portage/dep/libc.py
thesamesam: Kicking myself! Thank you Zac

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2856,6 +2850,48 @@ def _reapply_bsdflags_to_image(mysettings):
...
+    #
+    if injected_libc_depstring:
+        if rdepend:
+            rdepend += f" {' '.join(injected_libc_depstring).strip()}"
+        else:
+            # The package doesn't have an RDEPEND, so make one up.
+            rdepend = " ".join(injected_libc_depstring)
+
+        with open(rdepend_file, "w") as f:
+            f.write(f"{rdepend}\n")
```
zmedico, line 2892: Should read and write this file with utf8 encoding.

akhuettel:
Looks good now, thanks!

* Same test procedure as above, 
* the dependency is in both local VDB and binpkgs/Packages, 
* and --changed-deps does not trigger any rebuild.

I'll poke it a bit more, but right now I see no reason why it shouldnt work.

zmedico (approved):

thesamesam:
Thank you very much Zac for the help & dilfridge for the testing!

PR #1089, opened 2023-09-04
vartree: Remove unused variables and parameters

gcarq:
Removes unused local variable `cache_incomplete` in `aux_get()` and removes unused parameters for methods around `counter_tick()`.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -12,6 +12,9 @@ Bug fixes:
 
 * emerge: ensure paths are UTF-8 encoded in _needs_move() (bug #913103).
 
+Cleanups:
+* vartree: Remove unused variables and parameters
+
 portage-3.0.51 (2023-08-20)
 --------------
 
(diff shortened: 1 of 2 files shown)


NEWS
```diff
@@ -12,6 +12,9 @@ Bug fixes:
 
 * emerge: ensure paths are UTF-8 encoded in _needs_move() (bug #913103).
 
+Cleanups:
+* vartree: Remove unused variables and parameters
```
floppym, line 16: I wonder if we need a NEWS entry for an internal cleanup that users wont really care about.

floppym (approved):
Looks ok to me.

NEWS
gcarq: Makes sense, I figured all code changes should be reflected in NEWS. Should I remove it?

floppym: Let's see what @thesamesam thinks.

thesamesam: I think it's nice fodder if not much else happened in the release, it's somehow notable (e.g. PEP8 compliance, type annotations (because it's useful externally too), or some risk of breakage), but I wouldn't worry about including it if it's trivial _and_ internal.

floppym: Thinking about this more, I suspect the useless parameters were left in place to avoid breaking any dbapi consumers outside of the Portage project.

However I doubt any external projects are actually using these functions. I checked genkernel, and it does not.

That said, we could leave the NEWS entry in place as a heads-up to any external projects.

thesamesam (approved):

PR #1090, opened 2023-09-05
vartree: keep build dir if postinst fails

xxc3nsoredxx:
Bug: https://bugs.gentoo.org/704866

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -12,6 +12,8 @@ Bug fixes:
 
 * emerge: ensure paths are UTF-8 encoded in _needs_move() (bug #913103).
 
+* vartree: keep build dir if postinst fails (bug #704866).
+
 portage-3.0.51 (2023-08-20)
 --------------
 
--- a/lib/portage/dbapi/vartree.py
+++ b/lib/portage/dbapi/vartree.py
@@ -6070,8 +6070,11 @@ class dblink:
                 ebuild_phase.wait()
                 self._elog_process()
 
-                if "noclean" not in self.settings.features and (
-                    retval == os.EX_OK or "fail-clean" in self.settings.features
+                # Keep the build dir around if postinst fails (bug #704866)
+                if (
+                    not self._postinst_failure
+                    and "noclean" not in self.settings.features
+                    and (retval == os.EX_OK or "fail-clean" in self.settings.features)
                 ):
                     if myebuild is None:
                         myebuild = os.path.join(inforoot, self.pkg + ".ebuild")


xxc3nsoredxx:
```
Download action repository 'actions/checkout@v3' (SHA:f43a0e5ff2bd294095638e18286ca9a3d1956744)
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/f43a0e5ff2bd294095638e18286ca9a3d1956744'. Error: Response status code does not indicate success: 503 (Service Unavailable).
Warning: Back off 12.587 seconds before retry.
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/f43a0e5ff2bd294095638e18286ca9a3d1956744'. Error: Response status code does not indicate success: 503 (Service Unavailable).
Warning: Back off 28.428 seconds before retry.
Error: Response status code does not indicate success: 503 (Service Unavailable).
```
Huh.

mgorny:
Oh, nice, I wouldn't have thought it were that simple. Does that also cover preinst failures? (I think Portage proceeds to install in that case.) Perhaps it'd be good to also cover `*rm` failures.

xxc3nsoredxx:
I thought about `preinst` when I was tracking down what needed to be changed, and included that in my test ebuild. Turns out it already works because of how `treewalk()` exits (it also aborts the install):

```python
# run preinst script
showMessage(
    _(">>> Merging %(cpv)s to %(destroot)s\n")
    % {"cpv": self.mycpv, "destroot": destroot}
)
phase = EbuildPhase(
    background=False,
    phase="preinst",
    scheduler=self._scheduler,
    settings=self.settings,
)
phase.start()
a = phase.wait()

# XXX: Decide how to handle failures here.
if a != os.EX_OK:
    showMessage(
        _("!!! FAILED preinst: ") + str(a) + "\n",
        level=logging.ERROR,
        noiselevel=-1,
    )
    return a
```

vs how `postinst` is handled:

```python
if a != os.EX_OK:
    # It's stupid to bail out here, so keep going regardless of
    # phase return code.
    self._postinst_failure = True
    self._elog(
        "eerror",
        "postinst",
        [
            _("FAILED postinst: %s") % (a,),
        ],
    )

# ... snip ...

return os.EX_OK
```

Interestingly, a failed `prerm` does not leave anything in `/var/tmp/portage/._unmerge_` (despite it saying it does), but a failed `postrm` does. Neither of which actually aborts the unmerge.

mgorny (approved):
LGTM

thesamesam:
Thank you! See also https://bugs.gentoo.org/913905 for the followup bit.

PR #1092, opened 2023-09-10
depgraph: improve backtracking performance

living180:
Avoid applying `@functools.lru_cache()` to the `depgraph._slot_operator_check_reverse_depencencies()` method, as applying the `lru_cache` decorator to a method can prevent the class instances from being garbage collected [due to `self` being stored in the cache as part of the `args` key][1].  Instead, set up a per-instance cache in the `__init__()` method.  Also change the cache size from a `maxsize` of `100` to unlimited to avoid blowing out the cache and destroying performance.

Testing with a recent dev-lang/perl update which required a significant amount of backtracking showed a nearly 40% reduction in dependency resolution time (235s instead of 385s) along with reduced memory usage (max RSS size after dependency resolution of ~432MiB instead of ~466MiB).  Maximum per-instance cache sizes were measured in the low- to mid-hundreds, and the number of cache hits was generally 6x-7x the number of misses.

[1]: https://stackoverflow.com/q/33672412/3573385


thesamesam:
@mattst88 Could you take a look and also possibly ping the author of 839ab46be1777e5886da28b98b53a462b992c5bf (https://bugs.gentoo.org/883071)?

mattst88:
I asked Pin-yen for his opinion. He said

- The rationale is sound
- He's unsure whether making the cache size unlimited is safe

I would suggest making those changes in separate patches, if that's not difficult to do. I wouldn't ask you to recollect performance data, because I know what a pain that is.

living180:
@mattst88 I've updated the PR, splitting the change into two commits, one that makes the caching per-instance, and another that increase the cache size from 100 to 1000 (instead of unlimited).  Unfortunately since I first started working on these changes I went ahead an upgraded my dev-lang/perl version so I no longer have the pathological case that motivated this change, but I did some basic testing, and saw no difference in max RSS size with cache usage of 100 versus cache usage of 558.

thesamesam:
Thanks! Could you add `Signed-off-by` to the last commit then I'll merge?

living180:
> Thanks! Could you add `Signed-off-by` to the last commit then I'll merge?

Sorry about that - PR updated!

thesamesam:
Thank you!

PR #1093, opened 2023-09-15
_emerge: use binpkg coloring for pkg_pretend

thesamesam:
For the 'Running pre-merge checks' message, we write:
```
>>> Running pre-merge checks for X
```

X is currently always in green, while the emerge list above might have the atom X listed in purple if it's a binpkg.

Change X to be colored based on if it's a binpkg or not.

Bug: https://bugs.gentoo.org/914159

### DIFF
--- a/lib/_emerge/Scheduler.py
+++ b/lib/_emerge/Scheduler.py
@@ -865,12 +865,13 @@ class Scheduler(PollScheduler):
             if self._terminated_tasks:
                 raise asyncio.CancelledError
 
-            out_str = "Running pre-merge checks for " + colorize("INFORM", x.cpv)
-            self._status_msg(out_str)
-
             root_config = x.root_config
             settings = self._allocate_config(root_config.root)
             settings.setcpv(x)
+
+            color = "PKG_BINARY_MERGE" if x.built else "INFORM"
+            self._status_msg(f"Running pre-merge checks for {colorize(color, x.cpv)}")
+
             if not x.built:
                 # Get required SRC_URI metadata (it's not cached in x.metadata
                 # because some packages have an extremely large SRC_URI value).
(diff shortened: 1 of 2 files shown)


lib/_emerge/Scheduler.py
```diff
@@ -865,13 +865,19 @@ async def _run_pkg_pretend(self, loop=None):
...
-            if not x.built:
+
+            if x.built:
+                self._status_msg(
+                    f"Running pre-merge checks for {colorize('PKG_BINARY_MERGE', x.cpv)}"
+                )
+            else:
+                self._status_msg(
+                    f"Running pre-merge checks for {colorize('INFORM', x.cpv)}"
+                )
```
mgorny, line 879: I'm not saying this is wrong but I'd just make it something like `"INFORM" if x.built else "PKG_BINARY_MERGE"` above to avoid repeating the whole message.

lib/_emerge/Scheduler.py
```diff
@@ -1259,12 +1265,16 @@ def sighandler(signum, frame):
             and not mod_echo_output
         ):
             for mysettings, key, logentries in self._failed_pkgs_die_msgs:
+                color = "INFORM"
+                if pkg.built:
+                    color = "PKG_BINARY_MERGE"
```
mgorny, line 1270:
```suggestion
                color = "PKG_BINARY_MERGE" if pkg.built else "INFORM"
```

lib/_emerge/Scheduler.py
thesamesam: sorry - tbh, I'd completely forgot Python allowed that, I'm used to it in Perl. Agreed.

mgorny (approved):
LGTM. You may consider adding a `colorize_pkg(x)` helper that would print `x.cpv` with color based on `x.built`.

thesamesam:
Yeah, I was wondering the same. I'll leave it for now though.

Thanks!

thesamesam:
Merged.

PR #1094, opened 2023-09-15
man: use groff escape sequence for tilde characters

floppym:
groff replaces normal tilde (~) with a weird "modifier" on output. Use the escape sequence from groff_char(7) to avoid this.


ulm (approved):
Good catch.

PR #1095, opened 2023-09-16
fetch: ignore force-mirror feature when try_mirrors is false

floppym:
When fetching layout.conf, we call fetch(try_mirrors=False) to force direct use of the primary URI being passed and ignore mirrors.

The force-mirror feature causes fetch to ignore the primary URI.

Ignoring both the primary URI and mirrors leads to an unfetchable file. Give precedence to the try_mirrors parameter since we really only set it to False when fetching layout.conf from a mirror.

Bug: https://bugs.gentoo.org/877793

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -14,6 +14,8 @@ Bug fixes:
 
 * vartree: keep build dir if postinst fails (bug #704866).
 
+* fetch: fix fetching of layout.conf when FEATURES=force-mirror (bug #877793).
+
 portage-3.0.51 (2023-08-20)
 --------------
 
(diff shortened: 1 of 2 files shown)


thesamesam (approved):
I find `fetch.py` kind of hard to follow, but this sounds right. I'd like mgorny to look ideally given he implemented the GLEP 75 bits.

mgorny (approved):
LGTM. I was a little worried this could affect `RESTRICT=mirror`-ed packages but now I see that the logic for detecting that is inlined ugly into `fetch()`, sigh.

floppym:
Yeah, fetch could definitely use a refactor.

PR #1096, opened 2023-09-17
estrip: update for modern elfutils by passing -R

thesamesam:
Since elfutils 0.170, eu-strip supports -R / --remove-section option [0] """
strip: Add -R, --remove-section=SECTION and --keep-section=SECTION. """

"--remove-comment" is the same as "-R .comment". (eu-strip still does not support -N / --strip-symbol option, which would be good to use when it is implemented (Bug #686282).)

<elfutils-0.170 was deleted almost 2 years ago:
https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=7a0474559d4361a75a90aadc2b0f5ec46d85f359

Call eu-strip with the same -R options as for GNU strip / llvm-strip:
  -R .comment -R .GCC.command.line -R .note.gnu.gold-version

[0] https://sourceware.org/git/?p=elfutils.git;a=blob;f=NEWS;h=3d097c68438489b028e2eb6931f8287fd9a46652;hb=HEAD#l197

Bug: https://bugs.gentoo.org/728100

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,7 +5,10 @@ Features:
 * bintree: Add new API member (invalid_paths) to allow gentoolkit to later
   clean up invalid binpkgs (bug #900224).
 
-* Run PORTAGE_TRUST_HELPER before remote binary package operations.
+* gpkg: Run PORTAGE_TRUST_HELPER before remote binary package operations (bug #913070).
+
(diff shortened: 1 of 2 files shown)


thesamesam:
cc @Arfrever

thesamesam:
Not tested yet.

PR #1098, opened 2023-09-19
Fix some nondeterminism when generating binpkgs

ismell:
- config: Don't directly modify FEATURES
- bin/phase-functions: Move du stats into subshell

### DIFF
--- a/lib/portage/package/ebuild/config.py
+++ b/lib/portage/package/ebuild/config.py
@@ -1746,7 +1746,7 @@ class config:
             # Preserve the order of var_split because it can matter for things
             # like LINGUAS.
             var_split = [x for x in var_split if x in expand_flags]
-            var_split.extend(expand_flags.difference(var_split))
+            var_split.extend(sorted(expand_flags.difference(var_split)))
             has_wildcard = "*" in expand_flags
             if has_wildcard:
                 var_split = [x for x in var_split if x != "*"]
--- a/lib/portage/package/ebuild/config.py
+++ b/lib/portage/package/ebuild/config.py
@@ -2204,7 +2204,7 @@ class config:
                 # "test" is in IUSE and USE=test is masked, so execution
                 # of src_test() probably is not reliable. Therefore,
                 # temporarily disable FEATURES=test just for this package.
-                self["FEATURES"] = " ".join(x for x in self.features if x != "test")
+                self.features.remove("test")
 
         # Allow _* flags from USE_EXPAND wildcards to pass through here.
         use.difference_update(
--- a/bin/phase-functions.sh
+++ b/bin/phase-functions.sh
@@ -620,12 +620,11 @@ __dyn_install() {
 
 	# record build & installed size in build log
 	if type -P du &>/dev/null; then
-		local nsz=( $(du -ks "${WORKDIR}") )
-		local isz=( $(du -ks "${D}") )
-
 		# subshell to avoid polluting the caller env with the helper
 		# functions below
 		(
+			local nsz=( $(du -ks "${WORKDIR}") )
+			local isz=( $(du -ks "${D}") )
 			# align $1 to the right to the width of the widest of $1 and $2
 			padl() {
 				local s1=$1


lib/portage/package/ebuild/config.py
```diff
@@ -2204,7 +2204,7 @@ def setcpv(self, mycpv, use_cache=None, mydb=None):
                 # "test" is in IUSE and USE=test is masked, so execution
                 # of src_test() probably is not reliable. Therefore,
                 # temporarily disable FEATURES=test just for this package.
-                self["FEATURES"] = " ".join(x for x in self.features if x != "test")
+                self.features.remove("test")
```
floppym, line 2207: This code is workaround from a time when ebuilds were not setting `RESTRICT="!test? ( test )` consistently. I think we could simply drop this hack at this point.

vapier: that doesn't really make sense, nor is it PMS compliant

PMS states that `RESTRICT` can only be a [simple string](https://projects.gentoo.org/pms/8/pms.html#x1-730008.2) with [specific tokens](https://projects.gentoo.org/pms/8/pms.html#x1-680007.3.6).  anything else is a PM extension and doesn't belong in the Gentoo overlay.  specifically, writing `flag? ( ... )` is only permitted in `specification style variables` which `RESTRICT` is not.

PMS also does not guarantee that `USE=test` is set when running tests -- it doesn't mention that flag anywhere that i can see.  that's a portage-specific convention.  any PM that wants to implement `src_test` support in a reasonable/automatic way for the user has to do the same, but it'd also be perfectly acceptable for the PM to require the user explicitly set `USE=test` and be compliant.

should PMS do better ?  sure.  but since we're squarely in the land of "portage-defined behavior", portage should be doing the heavy lifting here.  otherwise you're suggesting every single ebuild that has `IUSE=test` also has to write `RESTRICT="!test? ( test )"`.  no one is writing ebuilds, let alone verifying them, with the idea that `src_test` will be executed when `USE=-test`.

Raul's change is maintaining status quo.  if you want to change the status quo, do it holistically, and in a separate change.

vapier (approved):

lib/portage/package/ebuild/config.py
floppym: > otherwise you're suggesting every single ebuild that has IUSE=test also has to write RESTRICT="!test? ( test )".

This is already the case. See the Gentoo [policy](https://projects.gentoo.org/qa/policy-guide/other-metadata.html#pg0703).

floppym (approved):

floppym (changes requested):
Please review our [copyright policy](https://www.gentoo.org/glep/glep-0076.html) and add Signed-off-by.

lib/portage/package/ebuild/config.py
floppym: I created https://github.com/gentoo/portage/pull/1116 to remove this logic.

ismell:
Oops. I added the signed off line and also added an additional patch for some non-determinism when generating USE_EXPAND variables.

floppym (approved):

thesamesam:
Thank you!

thesamesam:
@ismell Please take a look at the issue mentioned in b150419d28bd7afb98404a829c639584d34efc03.

zmedico:
> @ismell Please take a look at the issue mentioned in [b150419](https://github.com/gentoo/portage/commit/b150419d28bd7afb98404a829c639584d34efc03).

The issue is that `self.features.remove("test")` creates a persistent setting in `self._features_overrides` which is not reverted by `self.reset()`. Meanwhile, direct modification of the FEATURES variable just changes `self.configdict["env"]` which is reverted by each call to `self.reset()`.

ismell:
Oh interesting. I guess we need to keep the open coded removal. I can add a sorted(x) instead.

thesamesam:
Thanks!

ismell:
Thanks for pushing the new commit!

thesamesam:
i think that's on one of @palao's branches but not sure where...

PR #1099, opened 2023-09-21
etc-update: set SELinux security labels on merged files

0xC0ncord:
For files merged with etc-update, also set their SELinux security labels. Without this, merged files will have the type user_tmp_t and cause issues on enforcing SELinux systems.

### DIFF
--- a/bin/dispatch-conf
+++ b/bin/dispatch-conf
@@ -12,6 +12,7 @@
 #
 
 import atexit
+import errno
 import re
 import subprocess
 import sys
@@ -398,6 +399,8 @@ class dispatch:
                     mystat = os.lstat(conf["new"])
                     os.chmod(mrgconf, mystat[ST_MODE])
                     os.chown(mrgconf, mystat[ST_UID], mystat[ST_GID])
+                    if "selinux" in portage.settings.features:
+                        self.copy_selinux_label(conf["current"], mrgconf)
                     newconf = mrgconf
                     continue
                 elif c == "l":
@@ -433,6 +436,30 @@ class dispatch:
             print()
 
         perform_conf_update_session_hooks("post-session")
+
+    def copy_selinux_label(self, curconf, newconf):
+        """Copy the SELinux security label from the current config file to
+        the new/merged config file."""
+        try:
+            label = os.getxattr(curconf, "security.selinux")
+        except OSError as e:
+            if e.errno == errno.ENOTSUP:
+                # Filesystem does not support xattrs
+                return
+            writemsg(
+                f"dispatch-conf: Failed getting SELinux label on {curconf}; ignoring...\n",
+                noiselevel=-1,
+            )
+            return
+
+        if label:
+            try:
+                os.setxattr(newconf, "security.selinux", label)
+            except OSError:
+                writemsg(
+                    f"dispatch-conf: Failed setting SELinux label on {newconf}; ignoring...\n",
+                    noiselevel=-1,
+                )
 
     def replace(self, newconf, curconf):
         """Replace current config with the new/merged version.  Also logs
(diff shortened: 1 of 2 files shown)


bin/etc-update
```diff
@@ -631,6 +631,10 @@ do_merge() {
 					else
 						chown --reference="${ofile}" "${mfile}"
 						chmod --reference="${ofile}" "${mfile}"
+						${selinux} && {
+							chcon --reference="${ofile}" "${mfile}" || \
+								warn "Failed setting SELinux security label"
```
Flowdalic, line 636: Maybe mention the file where it failed
```suggestion
								warn "Failed setting SELinux security label on ${myfile}"
```

0xC0ncord: Sure, that makes sense.

bin/etc-update
```diff
@@ -631,6 +631,10 @@ do_merge() {
 					else
 						chown --reference="${ofile}" "${mfile}"
 						chmod --reference="${ofile}" "${mfile}"
+						${selinux} && {
```
floppym, line 634: Please use an `if` block instead. Or get rid of the special warning and consolidate this on a single line.

bin/etc-update
```diff
@@ -631,6 +631,10 @@ do_merge() {
...
+							chcon --reference="${ofile}" "${mfile}" || \
+								warn "Failed setting SELinux security label on ${mfile}"
```
floppym, line 636: This feels redundant. Doesn't chcon output an error message on failure? We don't add a special warning when chmod fails above.

floppym (changes requested):

floppym:
Does dispatch-conf suffer from the same issue?

0xC0ncord:
> Does dispatch-conf suffer from the same issue?

Yes, in fact. I will get in a fix for it as well. Thanks.

bin/etc-update
0xC0ncord: Good point. I'll remove this extraneous warning.

0xC0ncord:
OK, I've made the requested changes and also added a similar fix for `dispatch-conf`.

bin/dispatch-conf
```diff
@@ -434,6 +436,33 @@ class dispatch:
 
         perform_conf_update_session_hooks("post-session")
 
+    def copy_selinux_label(self, curconf, newconf):
+        """Copy the SELinux security label from the current config file to
+        the new/merged config file."""
+        try:
+            _ = os.listxattr(curconf)
```
floppym, line 443: This call isn't really needed. I would check errno for ENOTSUP when you do the getxattr call instead.

0xC0ncord: Good idea. Thanks!

bin/dispatch-conf
```diff
@@ -434,6 +436,30 @@ class dispatch:
...
 
+    def copy_selinux_label(self, curconf, newconf):
+        """Copy the SELinux security label from the current config file to
+        the new/merged config file."""
+        try:
+            label = os.getxattr(curconf, "security.selinux")
+        except OSError as why:
+            if why.errno == 95:
+                # Filesystem does not support xattrs
+                return
```
floppym, line 447: It would be better to use the symbolic ENOTSUP value from the `errno` module. Add `import errno` at the top of the file and make this change:
```suggestion
        except OSError as e:
            if e.errno == errno.ENOTSUP:
                # Filesystem does not support xattrs
                return
```
Also, would it make sense to skip the warning if getxattr sets errno = ENODATA? The man page describes that as "The named attribute does not exist, or the process has no access to this attribute."

I'm not sure if that's a typical data condition on SELinux or not.

0xC0ncord: >I'm not sure if that's a typical data condition on SELinux or not.

The only situation I think this would realistically happen is if someone is in the middle of switching to an SELinux profile but decides to do `etc-update` or `dispatch-conf` before doing a full relabel of all files on the system, or if for some reason they updated the original file without being on an SELinux kernel. In either case this isn't typical during day-to-day SELinux usage, so I think it's worth emitting the warning if it happens.

PR #1102, opened 2023-09-26
Set gpkg as BINPKG_FORMAT default

akhuettel:
**Untested so far**, but pretty straightforward given existing mechanisms

### DIFF
--- a/lib/portage/_compat_upgrade/meson.build
+++ b/lib/portage/_compat_upgrade/meson.build
@@ -1,6 +1,7 @@
 py.install_sources(
     [
         'binpkg_compression.py',
+	'binpkg_format.py',
         'binpkg_multi_instance.py',
         'default_locations.py',
         '__init__.py',
--- a/cnf/make.globals
+++ b/cnf/make.globals
@@ -38,9 +38,9 @@ PORTAGE_TMPDIR="/var/tmp"
 # existing installs where bzip2 is used for backward compatibility.
 BINPKG_COMPRESS="zstd"
 
-# The format used for binary packages. The default is use old "xpak" format.
-# Set to "gpkg" to use new gentoo binary package format.
-BINPKG_FORMAT="xpak"
+# The format used for binary packages. The default is to use the new "gpkg" format.
+# Set to "xpak" to use the old gentoo binary package format.
+BINPKG_FORMAT="gpkg"
 
 # The binary package default GPG signing command.
 # flock is used to avoid a racing condition of gnupg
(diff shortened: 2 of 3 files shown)


thesamesam:
Did you have a chance to test this yet?

akhuettel:
Needed an entry in meson.build, but otherwise works now as intended. Here's the required patch for portage-9999.ebuild : 

```
diff --git a/sys-apps/portage/portage-9999.ebuild b/sys-apps/portage/portage-9999.ebuild
index 5bf30b190b87e..abf843edb7357 100644
--- a/sys-apps/portage/portage-9999.ebuild
+++ b/sys-apps/portage/portage-9999.ebuild
@@ -205,6 +205,10 @@ pkg_preinst() {
                env -u FEATURES -u PORTAGE_REPOSITORIES \
                        PYTHONPATH="${D}${sitedir}${PYTHONPATH:+:${PYTHONPATH}}" \
                        "${PYTHON}" -m portage._compat_upgrade.binpkg_multi_instance || die
+
+               env -u BINPKG_FORMAT \
+                       PYTHONPATH="${D}${sitedir}${PYTHONPATH:+:${PYTHONPATH}}" \
+                       "${PYTHON}" -m portage._compat_upgrade.binpkg_format || die
        fi
 
        # elog dir must exist to avoid logrotate error for bug #415911.
```

akhuettel:
Just as a remark, this has, ahem, interesting side effects if binpkg-request-signature is enabled but local binpackage signing is not set up. You suddenly can't generate any binary packages locally anymore!

thesamesam:
> Just as a remark, this has, ahem, interesting side effects if binpkg-request-signature is enabled but local binpackage signing is not set up. You suddenly can't generate any binary packages locally anymore! (Trying to do so results in a backtrace and invalid files in the package directory.)

I think this was always a problem if you set `binpkg-request-signature` -- i.e. it's not new to the default change, it just makes it more prominent? Or am I misunderstanding?

lib/portage/_compat_upgrade/meson.build
```diff
@@ -1,6 +1,7 @@
 py.install_sources(
     [
         'binpkg_compression.py',
+	'binpkg_format.py',
```
zmedico, line 4: Indentation is abnormal here.

zmedico (approved):

lib/portage/_compat_upgrade/meson.build
thesamesam: Fixed on merge

PR #1103, opened 2023-09-30
Support Python UTF-8 Mode via portage.utf8_mode (bug 914722)

zmedico:
When a UTF-8 locale, or UTF-8 mode is detected, set
portage.utf8_mode to True, and do not wrap file access
with _unicode_func_wrapper. This is intended to mitgate
issues with byte string handling in python libraries
like shutil, as reported in bug 914722.

This patch is intended to be a simple and minimal
implementation that can be optimized later through
the elimination of unecessary encoding/decoding.

The str() wrapping in the unit tests is for lazily
evaluated instances of lazy_value, which is used to
account for mock portage.const.EPREFIX values that
exist during unit tests.

Upgrade pypy unit tests from pypy-3.9 to pypy-3.10
in order to solve [unexplainable unit test failures that
only affected pypy-3.9](https://github.com/gentoo/portage/actions/runs/6463627312/job/17547082772).

Bug: https://bugs.gentoo.org/914722

### DIFF
--- a/lib/portage/__init__.py
+++ b/lib/portage/__init__.py
@@ -186,6 +186,7 @@ except ImportError as e:
     sys.stderr.write(f"    {e}\n\n")
     raise
 
+utf8_mode = sys.getfilesystemencoding() == "utf-8"
 
 # We use utf_8 encoding everywhere. Previously, we used
 # sys.getfilesystemencoding() for the 'merge' encoding, but that had
@@ -319,6 +320,9 @@ class _unicode_module_wrapper:
         object.__setattr__(self, "_cache", cache)
 
     def __getattribute__(self, attr):
+        if utf8_mode:
+            return getattr(object.__getattribute__(self, "_mod"), attr)
+
         cache = object.__getattribute__(self, "_cache")
         if cache is not None:
             result = cache.get(attr)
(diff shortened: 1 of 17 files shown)


lib/portage/__init__.py
```diff
@@ -186,6 +187,7 @@
     sys.stderr.write(f"    {e}\n\n")
     raise
 
+utf8_mode = locale.getpreferredencoding(do_setlocale=False).lower() == "utf-8"
```
zmedico, line 190: Maybe `sys.getfilesystemencoding()` would be better here (also support Python UTF-8 Mode).

floppym:
Seems like a good idea.

I don't quite understand all the `str()` wrapping in the unit tests; I guess it's a workaround for path variables that are currently stored as bytes?

zmedico:
The `str()` wrapping in the unit tests is for lazily evaluated instances of lazy_value, which is used to account for mock portage.const.EPREFIX values that exist during unit tests.

thesamesam:
I'm scared of regressions but the possibility of that is inevitable with a change like this.

We definitely need to do it and this seems like a nice way of doing it (even better than having to use a wrapper/re-exec, although that wouldn't have been so bad).

thesamesam (approved):
Let's ship it, I think - the whole thing is smaller than I thought it'd be and I think we just need to try it.

PR #1105, opened 2023-10-01
env-update: Write /usr/etc/ld.so.conf to fix bfd in some obscure cases

chewi:
This is only needed on prefixed systems. bfd currently reads `${EPREFIX}/etc/ld.so.conf` and adds the prefix to these paths, but these paths are already prefixed. We need them to stay prefixed for the runtime linker. bfd will use `${EPREFIX}/usr/etc/ld.so.conf` instead if that is present, so we can write the unprefixed paths there.

Other linkers do not use these files at all. We tried to patch bfd to not use them either, as it shouldn't really be necessary, but that broke some cases, so we are trying this safer approach instead.

env-update does not write the files under `/etc/ld.so.conf.d`, but we shouldn't need to handle these in any case, as all known instances are not affected by this issue.

See [Gentoo bug #892549](https://bugs.gentoo.org/892549).


floppym (changes requested):
As someone completely unfamiliar with this issue, this "fix" seems pretty hacky to me. Is this meant to be temporary?

Anyway, please add a comment to the code with a reference to the bug report to make finding it easier.

chewi:
> As someone completely unfamiliar with this issue, this "fix" seems pretty hacky to me. Is this meant to be temporary?

It's not temporary. What makes you think it's hacky? If it's because `/usr/etc/ld.so.conf` is an odd path, then I agree, but blame bfd, not me! We've never added this file before, but there's nothing inherently wrong with getting bfd to use it.

> Anyway, please add a comment to the code with a reference to the bug report to make finding it easier.

Done.

floppym:
> What makes you think it's hacky?

Coming at this from a place of mostly ignorance and skimming through the bug report:

- Essentially the same data is being written to 2 places (/etc/ld.so.conf, /usr/etc/ld.so.conf), to avoid some strange behavior in binutils? It's hard-coded to add `${EPREFIX}` to paths at runtime?
- Writing to `${EPREFIX}/usr/etc` seems really weird. We always pass `--sysconfdir=${EPREFIX}/etc` to configure scripts. /usr/etc does not even exist on my system.
- I'm guessing the check for `${prefix}/etc/ld.so.conf` was added to ld for some reason unrelated to this issue?
- If ld from binutils needs a separate config file for library search paths, it seems like it should have a different file name and not attempt to utilize a file intended for ld.so. In other words, it shouldn't be looking at "ld.so.conf" in the first place.
- Are we sure glibc's ld.so will not start using `${prefix}/etc/ld.so.conf` in the future?

floppym:
Anyway, don't interpret my asking questions as a blocker for merging this. I just want to be on the record as saying "this feels weird".

chewi:
Fair points. I was surprised when I learned about this behaviour. You can see for yourself what bfd does [here](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=ld/ldelf.c;h=2e27cf48a816dc78bd76d2f0185a601d2edfb392;hb=ef8f08ca13f6c111cc549a3e13be5c5e2d95ca82#l910).

```c
 910       tmppath = concat (ld_sysroot, prefix, "/etc/ld.so.conf",
 911                         (const char *) NULL);
 912       if (!ldelf_parse_ld_so_conf (&info, tmppath))
 913         {
 914           free (tmppath);
 915           tmppath = concat (ld_sysroot, "/etc/ld.so.conf",
 916                             (const char *) NULL);
 917           ldelf_parse_ld_so_conf (&info, tmppath);
 918         }
 919       free (tmppath);
 920 
 921       if (info.path)
 922         {
 923           ld_so_conf = ldelf_add_sysroot (info.path);
 924           free (info.path);
 925         }
```

`ld_sysroot` is the toolchain's sysroot. With regular prefix, this is just `${EPREFIX}`. With cross prefix, this is `${SYSROOT}${EPREFIX}` aka `${ESYSROOT}`, as you would set with the `--sysroot` flag. `prefix` is the `--prefix` given to `configure`, always `/usr` in our case. So as I said, bfd is to blame for the weird path.

It checks `${ESYSROOT}/usr/etc/ld.so.conf` first, and if that fails, it checks `${ESYSROOT}/etc/ld.so.conf`. Regardless of which it uses, it then prepends the sysroot. This may seem odd, given that `/etc/ld.so.conf` is also used by the runtime linker, but I guess this alternative file is upstream's way of avoiding the problem.

As for why bfd looks at these files at all, I think it goes a lot further than other linkers to ensure that libraries will be found at runtime, not just at build time. This is probably a large part of why it is the slowest linker. I initially patched out the ld.so.conf lookup to stop it searching in `${EPREFIX}/${EPREFIX}`, thinking I had already done enough to let it find any needed libraries without help from this file, but it turned out I hadn't. Trying to further patch out the behaviour led to unresolved symbols. I decided I did not like changing bfd's behaviour that much.

I believe it is actually `ld.so.cache` that glibc primarily reads from. `ld.so.conf` is only used by `ldconfig` to generate the cache. I can't predict what glibc will do, but I cannot see any evidence that it would look at `/usr/etc/ld.so.conf`. From ldconfig.c:

```c
#ifndef LD_SO_CONF
# define LD_SO_CONF SYSCONFDIR "/ld.so.conf"
#endif
```

floppym:
Ok, so it seems I'm not missing anything here. No real objections, especially since it wont effect non-prefix systems.

thesamesam:
I gave you git access as we discussed a while back. Please add yourself to the project as well on the wiki, thanks!

I think this seems better than what we were doing before at least. It might be nice to include some of the information in that comment you posted in a comment in the code.

PR #1106, opened 2023-10-01
search: fix 'mlen' counter for ambiguous package names

floppym:
self.mlen was being incremented in both addCP() and output().

Drop the increment in addCP(), and convert mlen to a function-local variable in output().

Bug: https://bugs.gentoo.org/915054

### DIFF
--- a/lib/_emerge/search.py
+++ b/lib/_emerge/search.py
@@ -57,7 +57,6 @@ class search:
         self.fuzzy = fuzzy
         self.search_similarity = 80 if search_similarity is None else search_similarity
         self.matches = {"pkg": []}
-        self.mlen = 0
 
         self._dbs = []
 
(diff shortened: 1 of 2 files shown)


lib/_emerge/search.py
```diff
@@ -407,8 +405,10 @@ def append(msg):
             # Do a normal search
             iterator = self._iter_search()
 
+        mlen = 0
         for mtype, match in iterator:
-            self.mlen += 1
+            mlen += 1
+            print(mtype, match)
```
kuraga, line 411:
```suggestion
```

@floppym , debug code?

floppym: Yep, good catch.

zmedico (approved):

PR #1107, opened 2023-10-02
depgraph: show backtrack information after dep resolution

Flowdalic:
To improve user's awareness of the backtracking performed by portage and the related limit, show briefly how much backtracking was done.


thesamesam (approved):
As discussed on IRC, we may want to colourise this in future as the limit is approached, but not a blocker for this at all of course.

thesamesam:
Thanks!

PR #1110, opened 2023-10-03
SpawnProcess: Make fd_pipes optional

zmedico:
Since fd_pipes is incompatible with the multiprocessing spawn start method, allow empty fd_pipes so that ForkProcess can inherit SpawnProcess and still be compatible with multiprocessing spawn.

Bug: https://bugs.gentoo.org/915128


thesamesam:
FWIW, once you're happy, unless you explicitly want review from one of us for these multiprocessing PRs, feel free to just commit them - you're really the expert on this part of portage and I don't feel like I can offer much review at all (but if you want me or someone else to look, please do ask ofc)

PR #1116, opened 2023-10-04
ebuild: do not drop FEATURES=test when USE=-test

floppym:
By convention the "test" USE flag is used to control dependencies required for execution of the src_test function.

The "test" USE flag has no special meaning in PMS.

The Gentoo Policy Guide [1] advises that ebuilds must make use of RESTRICT to disable tests based on USE flags.

Current versions of Portage will not run tests when RESTRICT=test is set, regardless of the FEATURES setting.

This code in Portage predates the Gentoo policy. It is unneeded since the policy was created and ebuilds have been updated with the necessary RESTRICT values.

[1] https://projects.gentoo.org/qa/policy-guide/other-metadata.html#pg0703

### DIFF
--- a/lib/portage/package/ebuild/config.py
+++ b/lib/portage/package/ebuild/config.py
@@ -2205,20 +2205,6 @@ class config:
                 "BASH_FUNC____in_portage_iuse%%"
             ] = "() { [[ $1 =~ ${PORTAGE_IUSE} ]]; }"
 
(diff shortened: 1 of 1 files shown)


floppym:
Hmm. I would would say that EBUILD_FORCE_TEST should only cause the test phase to be executed, possibly even overriding RESTRICT.

I don't see any reason to modify USE based on EBUILD_FORCE_TEST, especially since it is only meant for developers.

zmedico (approved):

PR #1123, opened 2023-10-05
Inroducing Invalid_errors

siddhanthrathod:
Introducing Invalid_errors var to enable suppression
of invalid binary error
see also: https://github.com/gentoo/gentoolkit/pull/35

### DIFF
--- a/lib/portage/dbapi/bintree.py
+++ b/lib/portage/dbapi/bintree.py
@@ -809,6 +809,7 @@ class binarytree:
         getbinpkg_refresh=False,
         add_repos=(),
         force_reindex=False,
+        invalid_errors=True,
     ):
         """
         Populates the binarytree with package metadata.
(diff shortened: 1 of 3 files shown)


lib/portage/dbapi/bintree.py
```diff
@@ -1087,18 +1095,20 @@ def _populate_local(self, reindex=True):
                         msg = []
                         if missing_keys:
                             missing_keys.sort()
+                            if invalid_errors:
+                                msg.append(
+                                    _("Missing metadata key(s): %s.")
+                                    % ", ".join(missing_keys)
+                                )
+                        if invalid_errors:
```
zmedico, line 1103: We could move the above msg variable definition below the invalid_errors condition like this:
```python
if invalid_errors:
    msg = []
    if missing_keys:
        missing_keys.sort()
        msg.append(
```

PR #1125, opened 2023-10-09
repos.conf: fix value of sync-rsync-verify-max-age for ::gentoo

Flowdalic:
The value sync-rsync-verify-max-age specifies the default age in days. The current value of 24 is too high for an repo with daily changes like ::gentoo, and was probably set under the assumption that max-age denotes hours.


floppym (approved):
Makes sense to me.

floppym:
Actually, I think 1 day is a bit too tight for this. There have been times where an infra or repo issue has caused the gentoo repository to be stale for more than 1 day.

floppym (changes requested):
I feel like maybe 7 days would be safer.

Flowdalic:
> Actually, I think 1 day is a bit too tight for this. There have been times where an infra or repo issue has caused the gentoo rsync repository to be stale for more than 1 day.

Right, and the idea is that users are made aware of such situations, which causes security fixes not to be delivered to them. Personally, I would got with a warning if ::gentoo is outdated by more than 24 hours. Just like the initial commit which added this feature, but unfortunately made the warning after 24 days, intended.

I have no very strong opinion here. Everything lower than 24 days is an improvement. But I do not see how 7 days are better than 1 day for ::gentoo.

But maybe wait about the thoughts that others may have on this topic.

thesamesam:
7 days wouldn't have helped enough with the case which prompted this, where someone had been on a mirror that was dead for about 3.5-4 days and therefore didn't yet pick up an important vulnerability fix.

I'm thinking 3. @robbat2 @mgorny: thoughts?

mgorny:
3 sounds good to me.

thesamesam:
wfm, let's do 3

floppym:
3 is fine by me.

Flowdalic:
> To elaborate on my thinking: if we set it to 1 day, I fear users will run into sync errors somewhat frequently.

FWIW, according to https://mirrorstats.gentoo.org/rsync/ the mirrors are either up-to-date within 12h or completely outdated. 

But since 3 is better than 24, let's get this in first. But I hope we can revisit it in the future and potentially lower it.

PR #1126, opened 2023-10-09
Detect and handle unnecessary package reinstall (bug 915494)

zmedico:
Compare package rebuilds/reinstalls to installed packages of the same
exact version, and eliminate unecessary rebuilds/reinstalls triggered
solely by the `@__auto_slot_operator_replace_installed__` set. This is
careful to obey the user's wishes if they have explicitly requested
for a package to be rebuilt or reinstalled for some reason.

The SonameSkipUpdateTestCase::testSonameSkipUpdateNoPruneRebuilds
test case shows that the new depgraph._eliminate_rebuilds method
eliminates a backtracking run that is needed for the
testSonameSkipUpdate test case to succeed via prune_rebuilds
backtracking which was added for bug 439688.

Bug: https://bugs.gentoo.org/915494

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -53,6 +53,7 @@ from portage.package.ebuild.config import _get_feature_flags
 from portage.package.ebuild.getmaskingstatus import _getmaskingstatus, _MaskReason
 from portage._sets import SETPREFIX
 from portage._sets.base import InternalPackageSet
+from portage.dep._slot_operator import evaluate_slot_operator_equal_deps
 from portage.util import ConfigProtect, shlex_split, new_protect_filename
 from portage.util import cmp_sort_key, writemsg, writemsg_stdout
 from portage.util import ensure_dirs, normalize_path
@@ -448,6 +449,8 @@ class _dynamic_depgraph_config:
 
     """
 
+    _ENABLE_PRUNE_REBUILDS = True
+
     def __init__(self, depgraph, myparams, allow_backtracking, backtrack_parameters):
         self.myparams = myparams.copy()
         self._vdb_loaded = False
@@ -459,7 +462,7 @@ class _dynamic_depgraph_config:
         self._filtered_trees = {}
         # Contains installed packages and new packages that have been added
         # to the graph.
-        self._graph_trees = {}
+        self._graph_trees = portage._trees_dict()
         # Caches visible packages returned from _select_package, for use in
         # depgraph._iter_atoms_for_pkg() SLOT logic.
         self._visible_pkgs = {}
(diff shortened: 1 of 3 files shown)


thesamesam:
I suppose you can't come up with a test case without say, andrew's vdb? (I keep thinking we should have some way of summarising the "important bits" of a vdb so users can just run a command and dump it for the purposes of reproducing bugs.)

zmedico:
Well, lib/portage/tests/resolver/soname/test_skip_update.py::SonameSkipUpdateTestCase::testSonameSkipUpdate can suffice as a test case if I add a parameter to disable prune_rebuilds backtracking. I'll do that now...

zmedico:
> Well, lib/portage/tests/resolver/soname/test_skip_update.py::SonameSkipUpdateTestCase::testSonameSkipUpdate can suffice as a test case if I add a parameter to disable prune_rebuilds backtracking. I'll do that now...

Now testSonameSkipUpdateNoPruneRebuilds will fail if _eliminate_rebuilds doesn't do the job that prune_rebuilds backtracking does in testSonameSkipUpdate.

lib/_emerge/depgraph.py
```diff
@@ -3588,7 +3597,148 @@ def _add_pkg_soname_deps(self, pkg, allow_unsatisfied=False):
...
+                    # --newuse, --changed-use
+                    continue
+
+                if self._dynamic_config.myparams.get("changed_slot") and (
+                    self._changed_slot(pkg) or self._changed_slot(installed_instance)
+                ):
+                    continue
+
+                unsatisfied_parent = False
+                for parent, atom in self._dynamic_config._parent_atoms[pkg]:
```
thesamesam, line 3632:
```suggestion
unsatisfied_parent = any(not atom.match(installed_instance) for parent, atom in self._dynamic_config._parent_atoms[pkg])
```

thesamesam (approved):
Thanks!

lib/_emerge/depgraph.py
zmedico: Applied, thanks!

lib/_emerge/depgraph.py
```diff
@@ -3588,7 +3597,147 @@ def _add_pkg_soname_deps(self, pkg, allow_unsatisfied=False):
...
+                            eapi=pkg.eapi,
+                            token_class=Atom,
+                        )
+                        installed_deps.append(dep_struct)
+                except InvalidDependString:
+                    continue
+                else:
+                    new_deps = []
+                    new_use = self._pkg_use_enabled(pkg)
+                    if pkg.built:
```
zmedico, line 3680: This condition looks backwards, and inverting it causes the unit test to fail. Checking into it...

zmedico: Fixed, with tests passing after an adjustment to the _eval_deps function called by evaluate_slot_operator_equal_deps.

PR #1128, opened 2023-10-10
man/make.conf.5: note locations with trust issues

floppym:
Bug: https://bugs.gentoo.org/915330


thesamesam (approved):
I've not checked the man formatting (please do locally if you haven't already) but content lgtm

PR #1129, opened 2023-10-14
Remove leftovers from {FEATURES->USE}=test removal

parona-source:
* FEATURES="test" was binded to USE="test" in 6b19f71, but this was effectively removed in 8c5598c.
* This change removes mentions of this now bygone feature and unused codepaths associated with it.


floppym:
I guess we should probably combine this change with #1116.

parona-source:
Seems reasonable. How would combining work in this case? Does this require action from me, if so what?

floppym:
> Does this require action from me, if so what?

No, I was just noting it. The changes can be merged separately.

zmedico (approved):
Looks good! Thanks!

PR #1131, opened 2023-10-15
slot_dict_class: Make instances picklable

zmedico:
This improves compatibility with the multiprocessing spawn start method, by eliminating this error for MergeProcess:
```
AttributeError: Can't pickle local object 'slot_dict_class.<locals>.LocalSlotDict'
```
Bug: https://bugs.gentoo.org/915834

### DIFF
--- a/lib/portage/cache/mappings.py
+++ b/lib/portage/cache/mappings.py
@@ -1,4 +1,4 @@
-# Copyright: 2005-2020 Gentoo Authors
+# Copyright: 2005-2023 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 # Author(s): Brian Harring (ferringb@gentoo.org)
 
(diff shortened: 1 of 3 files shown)


thesamesam:
Nit: would it work to split the hoisting into its own class into its own commit, as it's much easier to review then?

zmedico:
> Nit: would it work to split the hoisting into its own class into its own commit, as it's much easier to review then? (I suspect it might be ok as it is with `git diff -w` but not tried yet)

Sure, spit that into the first commit 67ee41a123770ae389f03abc68d65d3b4d963f30 now.

lib/portage/cache/mappings.py
```diff
@@ -439,7 +439,8 @@ def slot_dict_class(keys, prefix="_val_"):
         keys_set = keys
     else:
         keys_set = frozenset(keys)
-    v = _slot_dict_classes.get((keys_set, prefix))
+    cache_key = (keys_set, prefix)
+    v = _slot_dict_classes.get(cache_key)
```
thesamesam, line 466: Oh, good spot. We don't have any tests for slot_dict_class - we only seem to test portage.cache at all indirectly in `lib/portage/tests/emerge/test_simple.py`.

Any chance of writing up a quick one?

zmedico: There's lib/portage/tests/process/test_pickle.py which tests two different slot_dict_class types: _PackageMetadataWrapperBase (`prefix="_val_"`) and FifoIpcDaemon._files_dict (`prefix=""`). That gives us cases with both types of prefixes.

thesamesam (approved):

lib/portage/cache/mappings.py
thesamesam: Thanks!

PR #1132, opened 2023-10-16
_post_src_install_uid_fix: Pass unicode to write_atomic for utf8_mode

zmedico:
Bug: https://bugs.gentoo.org/915837

### DIFF
--- a/lib/portage/package/ebuild/doebuild.py
+++ b/lib/portage/package/ebuild/doebuild.py
@@ -2777,7 +2777,9 @@ def _post_src_install_uid_fix(mysettings, out):
                         # a normal write might fail due to file permission
                         # settings on some operating systems such as HP-UX
                         write_atomic(
-                            _unicode_encode(
+                            fpath
+                            if portage.utf8_mode
+                            else _unicode_encode(
                                 fpath, encoding=_encodings["merge"], errors="strict"
                             ),
                             new_contents,


lib/portage/package/ebuild/doebuild.py
```diff
@@ -2777,7 +2777,9 @@ def _post_src_install_uid_fix(mysettings, out):
                         # a normal write might fail due to file permission
                         # settings on some operating systems such as HP-UX
                         write_atomic(
-                            _unicode_encode(
+                            fpath
+                            if portage.utf8_mode
+                            else _unicode_encode(
```
zmedico, line 2782: 4f5f6f571e52af6d2703db760bad4e0ad7439d5a has a similar adjustment for write_atomic in lib/portage/dbapi/vartree.py.

thesamesam (approved):
Thanks for the quick fix! This woked for me. I had a quick look and couldn't find any other cases like this either.

lib/portage/package/ebuild/doebuild.py
thesamesam: Wondering if we could have an assert or something in write_atomic?

zmedico: The problem is this line inside atomic_ofstream:
```python
tmp_name = "%s.%i" % (canonical_path, portage.getpid())
```
When canonical_path is bytes, the string formatting misbehaves here.

zmedico: In a world of pid namespaces, we should also not use `portage.getpid()` there.

thesamesam: Oh yeah, I remember https://bugs.gentoo.org/851015 now..

PR #1133, opened 2023-10-17
ForkProcess: Support logging via multiprocessing.Pipe

zmedico:
Detect the multiprocessing spawn start method, and log via multiprocessing.Pipe because file descriptor inheritance via fd_pipes does not work with the spawn start method.

The included test cases set the multiprocessing start method to spawn in a subprocess, in order to prevent interference with the main process.

Bug: https://bugs.gentoo.org/915123


lib/portage/util/_async/ForkProcess.py
```diff
@@ -29,6 +30,41 @@ class ForkProcess(SpawnProcess):
     # (after the sentinel has become ready).
     _proc_join_interval = 0.1
 
+    def _start(self):
+        if self.fd_pipes or self.logfile:
```
thesamesam, line 34: I wish we could write this first part (up to retval = ...) in some way which feels less repetitive, but I think the current way is easiest to read, so lgtm

thesamesam (approved):

PR #1137, opened 2023-10-21
_emerge: Fix NameError 'pkg' bug 915551

zmedico:
Use `(pkg.cpv, pkg.root)` to lookup package
instances corresponding to die messages.

Bug: https://bugs.gentoo.org/915551


thesamesam (approved):
Thank you!

PR #1138, opened 2023-10-21
Ban non-PMS helpers

ulm:
Finally ban portageq, prepinfo, prepman, prepstrip and prepallstrip.

Closes: https://bugs.gentoo.org/899898
Closes: https://bugs.gentoo.org/906129
Closes: https://bugs.gentoo.org/906156

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,10 @@ portage-3.0.54 (UNRELEASED)
 
 TODO
 
+Breaking changes:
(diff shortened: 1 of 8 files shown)


NEWS
```diff
@@ -1,3 +1,7 @@
+Breaking changes:
```
thesamesam, line 1: Pop it underneath.

thesamesam (approved):

PR #1139, opened 2023-10-21
Add more general BSD-friendly changes

Connor-GH:
Portage is a package manager that allows the user the most choice on how to customize their system. Some users may want to use Portage in an environment where the GNU coreutils are not fully present or are not present at all. This patch allows for support for the linux version of the BSD find(1) and the BSD-native command to be used with etc-update and dispatch-conf.

### DIFF
--- a/bin/dispatch-conf
+++ b/bin/dispatch-conf
@@ -155,11 +155,25 @@ class dispatch:
             if not os.path.isdir(os.path.dirname(path)):
                 continue
 
+			# check whether find command is the GNU version
+			# if not, use BSD-style find arguments
+			s = subprocess.run(["find", "--version"], capture_output=True)
+			if s.returncode == 1:
+				gnu_find = False
+			else:
+				gnu_find = True
+
             basename = "*"
-            find_opts = ["-name", ".*", "-type", "d", "-prune", "-o"]
+			if gnu_find == True:
+            	find_opts = ["-name", ".*", "-type", "d", "-prune", "-o"]
+			else:
+				find_opts = ["--", "-name", ".*", "-type", "d", "-prune", "-o"]
             if not os.path.isdir(path):
                 path, basename = os.path.split(path)
+			if gnu_find == True:
                 find_opts = ["-maxdepth", "1"]
+			else:
+				find_opts = ["--", "-maxdepth", "1"]
             if "case-insensitive-fs" in portage.settings.features:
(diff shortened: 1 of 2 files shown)


bin/dispatch-conf
```diff
@@ -155,11 +155,25 @@ class dispatch:
             if not os.path.isdir(os.path.dirname(path)):
                 continue
 
+			# check whether find command is the GNU version
+			# if not, use BSD-style find arguments
+            s = subprocess.run(["find", "--version"], capture_output=True)
```
mgorny, line 160: A presence of `--version` option is hardly sufficient to guarantee that it's GNU find.

bin/dispatch-conf
```diff
@@ -155,11 +155,25 @@ class dispatch:
...
+                gnu_find = False
+            else:
+                gnu_find = True
+
             basename = "*"
-            find_opts = ["-name", ".*", "-type", "d", "-prune", "-o"]
+            if gnu_find == True:
+                find_opts = ["-name", ".*", "-type", "d", "-prune", "-o"]
+            else:
+                find_opts = ["--", "-name", ".*", "-type", "d", "-prune", "-o"]
```
mgorny, line 170: Requiring `--` before option names sounds like a bug, especially that it makes the invocation incompatible with POSIX.

Connor-GH:
You're right, but it's even worse than I expected. The BSD version of getopt follows POSIX semantics more closely than the GNU getopt, and the libc change/regression was only seen after porting to glibc. The people who maintain this version of the BSD utils, https://github.com/chimera-linux, use musl libc which possibly already follows the POSIX behavior. setting POSIXLY_CORRECT="1" environment variable fixes this behavior. Is it possible then that etc-update and dispatch-conf (or portage as a whole) set POSIXLY_CORRECT when using find, or would you rather leave it up to the user? I am thinking in terms of a potential app-alternatives/coreutils here.

floppym:
I would say that this BSD find should be patched to work properly with glibc without requiring callers to set POSIXLY_CORRECT in the environment.

Connor-GH:
> I would say that this BSD find should be patched to work properly with glibc without requiring callers to set POSIXLY_CORRECT in the environment.

I ended up doing this in the gentoo repo I maintain for chimerautils. It sets POSIXLY_CORRECT to 1 internally for find, which doesn't leak out into the user's environment.

PR #1141, opened 2023-10-22
MergeProcess: Support QueryCommand with spawn start method

zmedico:
Since QueryCommand._db does not propagate to the child process
in MergeProcess with the multiprocessing spawn start method,
handle it by passing QueryCommand._db as an explicit parameter.

Since self.mydbapi can be a bindbapi instance, override the
bindbapi `__getstate__` method so that it is picklable (omit
the unpicklable _aux_cache_slot_dict attribute and regenerate
it after unpickling).

Bug: https://bugs.gentoo.org/916106

### DIFF
--- a/lib/portage/dbapi/_MergeProcess.py
+++ b/lib/portage/dbapi/_MergeProcess.py
@@ -9,6 +9,7 @@ import platform
 import fcntl
 import portage
 from portage import os, _unicode_decode
+from portage.package.ebuild._ipc.QueryCommand import QueryCommand
 from portage.util._ctypes import find_library
 import portage.elog.messages
 from portage.util._async.ForkProcess import ForkProcess
@@ -180,6 +181,15 @@ class MergeProcess(ForkProcess):
         self._dblink = mylink
         self._elog_reader_fd = elog_reader_fd
 
+        # Since the entire QueryCommand._db is not required, only pass
+        # in tree types that QueryCommand specifically requires.
+        child_db = {}
+        parent_db = portage.db if QueryCommand._db is None else QueryCommand._db
+        for root in parent_db:
+            child_db[root] = {}
+            for tree_type in ("vartree", "porttree"):
+                child_db[root][tree_type] = parent_db[root][tree_type]
+
         self.target = functools.partial(
             self._target,
             self._counter,
@@ -192,6 +202,7 @@ class MergeProcess(ForkProcess):
             self.settings,
             self.unmerge,
             self.vartree.dbapi,
+            child_db,
         )
 
         pids = super()._spawn(args, fd_pipes, **kwargs)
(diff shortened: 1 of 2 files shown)


lib/portage/dbapi/_MergeProcess.py
```diff
@@ -224,9 +225,10 @@ def _target(
         unmerge,
         vardb,
     ):
-        """
-        TODO: Make all arguments picklable for the multiprocessing spawn start method.
-        """
+        if QueryCommand._db is None and not hasattr(portage, "db"):
+            # Create portage.db for AbstractEbuildProcess/EbuildIpcDaemon
+            # when not using the multiprocessing fork start method.
+            portage._reset_legacy_globals()
```
zmedico, line 231: Maybe it would be nicer to copy down `QueryCommand._db` from the parent. Using `portage._reset_legacy_globals()` could possibly lead to there being duplicate dbapi instances in memory, which would not be nice.

thesamesam: Let's try copy it down, I think? I wasn't sure when I was looking earlier as was on mobile and I couldn't look at _reset_legacy_globals and friends.

zmedico: There was some difficulty pickling all of `QueryCommand._db`, so I pruned it down to just the required vartree and porttree keys, and that worked:
```python
# Since QueryCommand._db is not entirely picklable, select only
# tree types that QueryCommand specifically requires.
child_db = {}
for root in QueryCommand._db:
	child_db[root] = {}
	for tree_type in ("vartree", "porttree"):
		child_db[root][tree_type] = QueryCommand._db[root][tree_type]
```

thesamesam (approved):
Thanks! I was a bit worried about whether hardcoding the tree types would be a problem but it shows up in tests immediately (and in a few places) if we break the list of types, so it's fine.

lib/portage/dbapi/_MergeProcess.py
```diff
@@ -180,6 +181,15 @@ def _spawn(self, args, fd_pipes, **kwargs):
         self._dblink = mylink
         self._elog_reader_fd = elog_reader_fd
 
+        # Since QueryCommand._db is not entirely picklable, select only
+        # tree types that QueryCommand specifically requires.
```
zmedico, line 185: It turns out the the bindbapi._aux_cache_slot_dict instance attribute is not picklable, which needs to be fixed anyway because self.mydbapi can be a bindbapi.

zmedico: Fixed bindbapi, and updated the comment here since QueryCommand._db is now entirely picklable but it's still desirable to filter it.

thesamesam (approved):

PR #1144, opened 2023-10-22
AsyncFunction: restore fd_pipes initialization

aggstam:
Due to the missing pw, UserQuery prompt fails when using --ask flag on emaint

### DIFF
--- a/lib/portage/util/_async/AsyncFunction.py
+++ b/lib/portage/util/_async/AsyncFunction.py
@@ -25,6 +25,8 @@ class AsyncFunction(ForkProcess):
 
     def _start(self):
         pr, pw = multiprocessing.Pipe(duplex=False)
+        self.fd_pipes = {} if self.fd_pipes is None else self.fd_pipes
+        self.fd_pipes[pw] = pw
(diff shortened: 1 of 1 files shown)


aggstam:
For reference, this is the breaking commit: 1c0b5ff6aabe99224b07f7fc2e09e4cb6c9c70c4

aggstam:
(Fixing tests)

lib/portage/util/_async/AsyncFunction.py
```diff
@@ -25,6 +25,8 @@ class AsyncFunction(ForkProcess):
 
     def _start(self):
         pr, pw = multiprocessing.Pipe(duplex=False)
+        self.fd_pipes = {} if self.fd_pipes is None else self.fd_pipes
+        self.fd_pipes[pw] = pw
```
zmedico, line 29: We actually need to pass stdin via fd_pipes. See https://github.com/gentoo/portage/pull/1145.

aggstam: aha good to know, I will close this PR since its obselete, good job!
KR

PR #1145, opened 2023-10-22
SpawnProcess/ForkProcess: pass stdin via fd_pipes if not background

zmedico:
Bug: https://bugs.gentoo.org/916116


thesamesam (approved):
looks fine but maybe a test?

zmedico:
> looks fine but maybe a test?

I figured out that we can actually handle it better in SpawnProcess and ForkProcess which AsyncFunction inherts.

Yeah, I'll create a test that overwrites stdin with a pipe, and then runs AsyncFunction which should be able to read from the pipe.

PR #1147, opened 2023-10-22
tests/resolver: test depclean order with IDEPEND and circular deps

the-horo:
Bug: https://bugs.gentoo.org/916135


zmedico (approved):
I've got a fix over in https://github.com/gentoo/portage/pull/1212. I go ahead and merge your commit in there.

PR #1148, opened 2023-10-22
emerge: avoid spamming too much info about unused binpkgs

eli-schwartz:
Specifically:

- the binhost might have multiple binpkgs for the same cpv, same USE flags, but with different build ids, in which case the same line is repeated twice

- many packages can be participating in the recursive deptree, but not get selected for actual merging, and may not even be installed at all; these were still reported as mismatched-USE although the packages themselves were not shown in show_merge_list

Both cases are useless noise. Prune them before printing a status report.

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -1232,25 +1232,38 @@ class depgraph:
             self._show_ignored_binaries_changed_deps(ignored_binaries["changed_deps"])
 
     def _show_ignored_binaries_respect_use(self, respect_use):
-        writemsg(
-            "\n!!! The following binary packages have been ignored "
-            + "due to non matching USE:\n\n",
-            noiselevel=-1,
-        )
-
+        seen = {}
+        messages = []
+        merging = {pkg.cpv for pkg in self._dynamic_config._displayed_list}
         for pkg, flags in respect_use.items():
+            # Don't include recursive deps which aren't in the merge list anyway.
+            if pkg.cpv not in merging:
+                continue
             flag_display = []
             for flag in sorted(flags):
                 if flag not in pkg.use.enabled:
                     flag = "-" + flag
                 flag_display.append(flag)
             flag_display = " ".join(flag_display)
-            # The user can paste this line into package.use
-            writemsg(f"    ={pkg.cpv} {flag_display}", noiselevel=-1)
-            if pkg.root_config.settings["ROOT"] != "/":
-                writemsg(f" # for {pkg.root}", noiselevel=-1)
-            writemsg("\n", noiselevel=-1)
+            # We don't want to list the same USE flags for multiple build IDs
+            if pkg.cpv not in seen or flag_display not in seen[pkg.cpv]:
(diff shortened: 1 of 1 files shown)


lib/_emerge/depgraph.py
```diff
@@ -1232,25 +1232,36 @@ def _show_ignored_binaries(self):
...
                 if flag not in pkg.use.enabled:
                     flag = "-" + flag
                 flag_display.append(flag)
             flag_display = " ".join(flag_display)
-            # The user can paste this line into package.use
-            writemsg(f"    ={pkg.cpv} {flag_display}", noiselevel=-1)
-            if pkg.root_config.settings["ROOT"] != "/":
-                writemsg(f" # for {pkg.root}", noiselevel=-1)
-            writemsg("\n", noiselevel=-1)
+            if pkg.cpv not in seen or seen[pkg.cpv] != flag_display:
```
thesamesam, line 1247:
```suggestion
			# We don't want to list the same USE flags for multiple build IDs
            if pkg.cpv not in seen or seen[pkg.cpv] != flag_display:
```

lib/_emerge/depgraph.py
```diff
@@ -1232,25 +1232,36 @@ def _show_ignored_binaries(self):
...
-            "\n!!! The following binary packages have been ignored "
-            + "due to non matching USE:\n\n",
-            noiselevel=-1,
-        )
-
+        seen = {}
+        messages = []
+        merging = {pkg.cpv for pkg in self._dynamic_config._displayed_list}
         for pkg, flags in respect_use.items():
+            if pkg.cpv not in merging:
```
thesamesam, line 1239:
```suggestion
			# Don't include recursive deps which aren't in the merge list anyway.
            if pkg.cpv not in merging:
```

thesamesam (commented):
Could you add the before/after output to the commit message too?

lib/_emerge/depgraph.py
```diff
@@ -1232,25 +1232,36 @@ def _show_ignored_binaries(self):
...
                     flag = "-" + flag
                 flag_display.append(flag)
             flag_display = " ".join(flag_display)
-            # The user can paste this line into package.use
-            writemsg(f"    ={pkg.cpv} {flag_display}", noiselevel=-1)
-            if pkg.root_config.settings["ROOT"] != "/":
-                writemsg(f" # for {pkg.root}", noiselevel=-1)
-            writemsg("\n", noiselevel=-1)
+            if pkg.cpv not in seen or flag_display not in seen[pkg.cpv]:
+                seen.setdefault(pkg.cpv, set()).add(flag_display)
```
eli-schwartz, line 1248: Updated this to use a set because we need to track multiple flag_display possibilities.

lib/_emerge/depgraph.py
eli-schwartz: Done.

lib/_emerge/depgraph.py
eli-schwartz: Done.

thesamesam (approved):

PR #1151, opened 2023-10-23
bindbapi: Fix pickling for MergeProcess

zmedico:
Both the `_aux_cache_slot_dict_cache` and `_instance_key` attributes are
unpicklable, so fix `__getstate__` to exclude them both from the state.
Override `__setstate__` to initialize _instance_key appropriately.
These changes prevent the following error:
```
AttributeError: Can't pickle local object 'slot_dict_class.<locals>.LocalSlotDict'
```
Fixes: b9a85ff987ea ("MergeProcess: Support QueryCommand with spawn start method")
Bug: https://bugs.gentoo.org/916106

### DIFF
--- a/lib/portage/dbapi/bintree.py
+++ b/lib/portage/dbapi/bintree.py
@@ -139,14 +139,19 @@ class bindbapi(fakedbapi):
         return self._aux_cache_slot_dict_cache
 
     def __getstate__(self):
-        # This attribute is not picklable, but it automatically
-        # regenerates after unpickling when set to None.
-        _aux_cache_slot_dict = self._aux_cache_slot_dict_cache
-        self._aux_cache_slot_dict_cache = None
-        try:
-            return super().__getstate__()
-        finally:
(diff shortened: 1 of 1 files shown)


lib/portage/dbapi/bintree.py
```diff
@@ -139,14 +139,19 @@ def _aux_cache_slot_dict(self):
...
-        finally:
-            self._aux_cache_slot_dict_cache = _aux_cache_slot_dict
+        state = self.__dict__.copy()
+        # These attributes are not picklable, so they are automatically
+        # regenerated after unpickling.
+        state["_aux_cache_slot_dict_cache"] = None
+        state["_instance_key"] = None
+        return state
+
+    def __setstate__(self, state):
```
thesamesam, line 149: We don't need to do this now, but at some point, maybe we should have some tests where we force pickling and make sure all the instance data is the same before/after

thesamesam (approved):

PR #1153, opened 2023-10-23
StaticFileSet: Fix os.walk for utf8_mode

zmedico:
Bug: https://bugs.gentoo.org/916182

### DIFF
--- a/lib/portage/_sets/files.py
+++ b/lib/portage/_sets/files.py
@@ -1,10 +1,11 @@
-# Copyright 2007-2020 Gentoo Authors
+# Copyright 2007-2023 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 import errno
 import re
 from itertools import chain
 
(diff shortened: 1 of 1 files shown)


lib/portage/_sets/files.py
```diff
@@ -176,6 +177,11 @@ def multiBuilder(self, options, settings, trees):
             directory = normalize_path(directory)
 
             for parent, dirs, files in os.walk(directory):
+                if portage.utf8_mode:
+                    parent = os.fsencode(parent)
+                    dirs = [os.fsencode(value) for value in dirs]
```
zmedico, line 182: There's a `dirs.remove(d)` that will be ineffective because this is not the dirs used by os.walk.

zmedico: Added an `omit_dir` lambda to handle it.

thesamesam (approved):

PR #1156, opened 2023-10-24
bin/emaint: Use __main__ for spawn compat

zmedico:
Bug: https://bugs.gentoo.org/916142


thesamesam (approved):
lgtm for all of these __main__ chnages but maybe copy the rationale from the bug into the commit message? (might even be worth a comment in the `finally:`)

zmedico:
> lgtm for all of these **main** chnages but maybe copy the rationale from the bug into the commit message? (might even be worth a comment in the `finally:`)

Yes, did both. Thanks!

PR #1161, opened 2023-10-24
_unshare_validator: multiprocessing spawn compat

zmedico:
Since ctypes objects are not picklable for the multiprocessing spawn start method, acquire them in the subprocess.

Bug: https://bugs.gentoo.org/916242

### DIFF
--- a/lib/portage/process.py
+++ b/lib/portage/process.py
@@ -1,5 +1,5 @@
 # portage.py -- core Portage functionality
-# Copyright 1998-2020 Gentoo Authors
+# Copyright 1998-2023 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 
@@ -943,6 +943,8 @@ class _unshare_validator:
         @rtype: int
         @returns: errno value, or 0 if no error occurred.
         """
+        # This ctypes library lookup caches the result for use in the
+        # subprocess when the multiprocessing start method is fork.
         filename = find_library("c")
         if filename is None:
             return errno.ENOTSUP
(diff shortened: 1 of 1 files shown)


thesamesam (approved):

floppym (changes requested):
Can't you drop the find_library and LoadLibrary calls from the _validate method now?

floppym:
> We can, but if we leave it there then those functions calls will cache the result for use in the subprocess when the multiprocesssing start method is fork.

That's not obvious to someone reading the code; it just looks like unused variables. Probably needs a comment if we keep it.

Is this code called frequently enough to warrant caching results in the parent process?

floppym (approved):

zmedico:
> That's not obvious to someone reading the code; it just looks like unused variables. Probably needs a comment if we keep it.

Comment added.

> Is this code called frequently enough to warrant caching results in the parent process?

It's called whenever we spawn a process with one of the sandboxes that uses unshare, so yes it's called very frequently. I've updated it to pass the find_library result down, since that's picklable and I think that is the most expensive since it calls ldconfig:
```
$ strace -f python -c 'import ctypes.util; ctypes.util.find_library("c")' |& grep ldconfig
[pid 60804] execve("/sbin/ldconfig", ["/sbin/ldconfig", "-p"], 0x7f20509e38d0 /* 2 vars */ <unfinished ...>
[pid 60804] openat(AT_FDCWD, "/sbin/ldconfig", O_RDONLY) = 3
[pid 60804] execve("/sbin/ldconfig.real", ["/sbin/ldconfig.real", "-p"], 0x5558b7fd2bf0 /* 3 vars */) = 0
[pid 60804] readlink("/proc/self/exe", "/usr/sbin/ldconfig.real", 4096) = 23
```

PR #1162, opened 2023-10-24
SonameAtom: multiprocessing spawn compat

zmedico:
Do not pickle the SonameAtom _hash_value attribute, since hash() results are not meaningfully picklable.

Bug: https://bugs.gentoo.org/916248


thesamesam (approved):
I'm surprised this didn't blow up when pickilng (rather than just in the tests) and I swear I've seen it happen before with hashes so ????

PR #1167, opened 2023-10-29
unpack: Limit memory usage when decompressing xz

swegener:
Each thread requires memory and it's quite easy to use too much memory. We already use --memlimit-compress=50% to restrict the memory usage when compressing, let's to the same for decompressing.

### DIFF
--- a/bin/phase-helpers.sh
+++ b/bin/phase-helpers.sh
@@ -544,7 +544,7 @@ unpack() {
 						"with EAPI '${EAPI}'. Instead use 'xz'."
 				fi
 				if ___eapi_unpack_supports_xz; then
-					__unpack_tar "xz -T$(___makeopts_jobs) -d"
+					__unpack_tar "xz -T$(___makeopts_jobs) --memlimit-mt-decompress=50% -d"
 				else
 					__vecho "unpack ${x}: file format not recognized. Ignoring."
 				fi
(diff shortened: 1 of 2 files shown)


bin/phase-helpers.sh
```diff
@@ -557,7 +557,7 @@ unpack() {
 						"with EAPI '${EAPI}'. Instead use 'txz'."
 				fi
 				if ___eapi_unpack_supports_txz; then
-					XZ_OPT="-T$(___makeopts_jobs)" tar xof "${srcdir}${x}" || die "${myfail}"
+					XZ_OPT="-T$(___makeopts_jobs) --memlimit-mt-decompress=50%" tar xof "${srcdir}${x}" || die "${myfail}"
```
thesamesam, line 560: 6a47cc7bcf49b7c17fdca31969b6d6aeff84332c calls out the decompress option -- but what I don't get yet is there's already a default limit for multithreaded decompression (which is part of why I didn't add it to begin with).

This is mentioned both in the man page and you can see it with:
```console
$ xz --info-memory
Hardware information:
  Amount of physical memory (RAM):  64,215 MiB (67,334,234,112 B)
  Number of processor threads:      32

Memory usage limits:
  Compression:                      Disabled
  Decompression:                    Disabled
  Multi-threaded decompression:     16,054 MiB (16,833,558,528 B)
  Default for -T0:                  16,054 MiB (16,833,558,528 B)
````

That's less than 50%. So, while we could add an option, adding it with a value of 50% doesn't seem to do much (if anything it increases the limit).

What am I missing?

swegener:
Ah, didn't check the default. I'v recently run into issues with decompressing the chromium tarball and xz being OOM-killed, but then my issue must have a different underlying reason.

thesamesam:
No problem - if it does end up being this, ofc feel free to reopen

PR #1169, opened 2023-10-30
emerge: fix binpkg-respect-use notice with blockers

thesamesam:
Items in _dynamic_config._displayed_list might be blockers, so filter those out.

Bug: https://bugs.gentoo.org/916336
Fixes: bb82666b48e18f448661a1a8bf6a39b773cc4b1c

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -1234,11 +1234,13 @@ class depgraph:
     def _show_ignored_binaries_respect_use(self, respect_use):
         seen = {}
         messages = []
+
         merging = {
             (pkg.root, pkg.cpv)
-            for pkg in self._dynamic_config._displayed_list
+            for pkg in self._dynamic_config._displayed_list or ()
             if isinstance(pkg, Package)
         }
+
         for pkg, flags in respect_use.items():
             # Don't include recursive deps which aren't in the merge list anyway.
             if (pkg.root, pkg.cpv) not in merging:
--- a/NEWS
+++ b/NEWS
@@ -15,6 +15,8 @@ Bug fixes:
 
 * Avoid crash with incomplete depgraph for binpkg-respect-use notice (bug #916614).
 
+* Avoid crash with blockers in depgraph for binpkg-respect-use notice (bug #916336).
+
 portage-3.0.54 (2023-10-25)
 --------------
 
--- a/lib/portage/tests/resolver/test_useflags.py
+++ b/lib/portage/tests/resolver/test_useflags.py
@@ -229,3 +229,92 @@ class UseFlagsTestCase(TestCase):
                         )
                 finally:
                     playground.cleanup()
+
+    def testNoMergeBinpkgRespectUse(self):
+        """
+        Testcase for bug #916614 where an incomplete depgraph may be fed into
+        _show_ignored_binaries_respect_use.
+
+        We use a mix of +/-abi_x86_32 to trigger the binpkg-respect-use notice
+        and depend on a non-existent package in one of the available ebuilds we
+        queue to reinstall to trigger an aborted calculation.
+        """
+        ebuilds = {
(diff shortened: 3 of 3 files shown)


lib/_emerge/depgraph.py
```diff
@@ -1234,7 +1234,11 @@ def _show_ignored_binaries(self):
     def _show_ignored_binaries_respect_use(self, respect_use):
         seen = {}
         messages = []
-        merging = {pkg.cpv for pkg in self._dynamic_config._displayed_list}
+        merging = {
+            pkg.cpv
```
zmedico, line 1238: Since this list can contain packages from multiple roots, it would make sense to make this a tuple that includes pkg.root (EROOT) as in 935d1432b2ea42fc0a49b9e3368a1766a673e1a9.

thesamesam: Ah, good catch. I've added a test for the original bug so now I'll look at this.

lib/_emerge/depgraph.py
```diff
@@ -1235,13 +1235,13 @@ def _show_ignored_binaries_respect_use(self, respect_use):
...
         merging = {
-            pkg.cpv
+            (pkg.root, pkg.cpv)
             for pkg in self._dynamic_config._displayed_list
             if isinstance(pkg, Package)
         }
         for pkg, flags in respect_use.items():
-            # Don't include recursive deps which aren't in the merge list anyway.
-            if pkg.cpv not in merging:
+            # Don't include recursive deps wohich aren't in the merge list anyway.
```
eli-schwartz, line 1243: Huh. Wonder how that happened. :D

thesamesam: (I've just realised I need to fix `seen` for root as well)

thesamesam:
@zmedico I spent a while trying to get a test case for https://bugs.gentoo.org/916614 (the other bug with this function not this one involving blockers) and I didn't have any luck. Do you have any guesses?

I can't see how we end up with a `None`  `_displayed_list` there..

thesamesam:
Thank you! 

I tried emulating failure with masks but I couldn't get there. Do you have any suggestions for how to capture that incomplete calculation for a test here? (I could mock it but I'm hoping to not have to)

zmedico:
> Thank you!
> 
> I tried emulating failure with masks but I couldn't get there. Do you have any suggestions for how to capture that incomplete calculation for a test here? (I could mock it but I'm hoping to not have to)

It typically happens due to an unsatisfied dependency that causes _create_graph to fail and also can't be resolved by backtracking. You can probably trigger it by adding a required package with an unsatisfied dependency. A dependency on a package that doesn't exist should do it.

eli-schwartz:
I think I've discovered one way to resolve the issue.

```
emerge -pe @installed
```

with intent to grep for/count the number of [binary ...] vs [ebuild ...], crashes with `TypeError: 'NoneType' object is not iterable` until I add:
```
if not self._dynamic_config._displayed_list:
    return
```

Then, I get a bit of data on why it didn't like the calculation:

```
Local copy of remote index is up-to-date and will be used.
 * The @installed set is not recommended when updating
 * packages because it will often introduce unsolved blocker
 * conflicts. Please refer to bug #387059 for details.

Local copy of remote index is up-to-date and will be used.

Dependency resolution took 1.93 s.


!!! All ebuilds that could satisfy "dev-lang/rapydscript-ng:0" have been masked.
!!! One of the following masked packages is required to complete your request:
- dev-lang/rapydscript-ng-9999::gentoo (masked by: missing keyword)
- dev-lang/rapydscript-ng-0.7.22::gentoo (masked by: ~amd64 keyword)

(dependency required by "@installed" [argument])
For more information, see the MASKED PACKAGES section in the emerge
man page or refer to the Gentoo Handbook.
```

Using `@installed` seems to trigger it, but... not when trying to install the one masked package? Fascinating.

PR #1170, opened 2023-10-30
pyproject.toml, NEWS: declare pytest markers.

palao:
Declare pytest markers. They can be used to select what tests (not) to run, e.g.:

  pytest -m ft  # to select tests marked with"ft"
  pytest -m "not ft"  # to skip tests marked with "ft"

to list the markers:

  pytest --markers

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -13,6 +13,9 @@ Bug fixes:
 * Convert portageq helper to a function to avoid breaking external callers
   (bug #916287, bug #916296).
 
+Cleanups:
+* Add pytest markers: ft, unit and stress.
+
 portage-3.0.54 (2023-10-25)
 --------------
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,3 +33,10 @@ gpkg-sign = 'portage.util.bin_entry_point:bin_entry_point'
 portageq = 'portage.util.bin_entry_point:bin_entry_point'
 quickpkg = 'portage.util.bin_entry_point:bin_entry_point'
 regenworld = 'portage.util.bin_entry_point:bin_entry_point'
+
+[tool.pytest.ini_options]
+markers = [
+    "ft: functional tests (select: '-m ft'; deselect with '-m \"not ft\"')",
+    "stress: stress tests (select: '-m stress'; deselect with '-m \"not stress\"')",
+    "unit: unit tests (select: '-m unit'; deselect with '-m \"not unit\"')",
+]


palao:
I think it would be good to make an effort to revamp and cleanup the tests in general, and also document it. I think I'd be glad to do it, but not sure if it makes sense from the "Bigger picture" point of view. Thoughts?

thesamesam:
> I think it would be good to make an effort to revamp and cleanup the tests in general, and also document it. I think I'd be glad to do it, but not sure if it makes sense from the "Bigger picture" point of view. Thoughts?

If you're happy to work on it, please do. The tests really do need more love and the recent improvements you've made have already improved the situation quite a bit.

My main thoughts on the tests are that we eventually need to get to the point where we have some bash-based ones where we invoke emerge and assert stuff happened for a full end-to-end check (just because some stuff will always be easier if we can just feed it a bunch of actual ebuild files, not just defined inline in Python, etc) but I know you want to do some of that in Python too.

thesamesam:
@zmedico is there anything you think he should focus on?

palao:
Let me explain better what I mean. What I was thinking is: if we are going to use the markers, it must be documented somewhere. Sure, the `NEWS` is a place to mention that (maybe I'm wrong here), but maybe also in `TEST-NOTES`? It might be the place, but it is obsolete now. We should rewrite it. But, to write something that makes sense we need a more defined structure for the tests and for that we need a clearer picture. That's why I didn't go beyond modifying the `NEWS` file and wrote my comment here.
And as I said I'll be glad to help.

palao:
> My main thoughts on the tests are that we eventually need to get to the point where we have some bash-based ones where we invoke emerge and assert stuff happened for a full end-to-end check (just because some stuff will always be easier if we can just feed it a bunch of actual ebuild files, not just defined inline in Python, etc) but I know you want to do some of that in Python too.
> 
> i.e. I want to be able to test more in Python, but I also think there's a place for some tests where we run `emerge`, `emaint`, and make sure they do the right thing too.
> 
> I welcome more tests of any sort, but especially if it leads to more refactoring, or better coverage in areas where we have very little or none. You _may_ want to look at some list of bugs to see where we didn't add a test as a way to direct efforts, but that's just an idea..

For those tests that you (and I!) want, the language we use to write them is irrelevant. We want to test `emerge`, `emaint` and so on, *from outside*. But *how* we *drive* the tests is just orthogonal to the way the actual code is working (and the language the code is written in). We could decide to write those tests in Bash, but I find writing tests in Bash a terribly bad idea (at least for my bash skills), whereas python is very versatile language that could help us. The fact that portage itself is written in Python has nothing to do with that. It could be that Portage was written in Rust and still we could decide to write those tests in Python.

The old `test_simple.py` and the new `test_baseline.py` (which is, at its core, the same thing) is doing the right thing (AFAICT): it is testing the actual commands. There is no interference in the internals. It's end-to-end testing.
Now, it it could be better: we could (as you suggest) have a more realistic setup with actual ebuilds being written to files, etc. And actually pytest is very useful for these things. But the tests (at least those ones) are conceptually the right thing.
[Note: With fixtures we can create the needed files to have a realistic scenario to run actual portage commands]

Anyway, I find your suggestion extremely practical: I'll look for low hanging fruits in the bugs, areas with little test coverage and try to figure out ways to test that functionality end-to-end.

I'll try to find ways to make more explicit what I wrote in this message.

Thank you very much for sharing your thoughts.

zmedico (approved):

zmedico:
> @zmedico is there anything you think he should focus on?

Hmm, that's a good question. Maybe implement [GLEP 63](https://www.gentoo.org/glep/glep-0073.html) in portage, or implement configurable lru cache scaling ([bug 747016](https://bugs.gentoo.org/747016)).

palao:
> > @zmedico is there anything you think he should focus on?
> 
> Hmm, that's a good question. Maybe implement [GLEP 63](https://www.gentoo.org/glep/glep-0073.html) in portage, or implement configurable lru cache scaling ([bug 747016](https://bugs.gentoo.org/747016)).

Thank you, @zmedico, for the suggestions.

I'm still reading [GLEP 73](https://www.gentoo.org/glep/glep-0073.html) (you meant 73, right?). It's really interesting and I think I'll like to take care of it.

The other thing, [bug 747016](https://bugs.gentoo.org/747016), is also interesting but easier. Maybe I start with that, if it's ok. In the meantime I'll keep studying [GLEP 73](https://www.gentoo.org/glep/glep-0073.html)...

PR #1173, opened 2023-11-01
Use documented os.register_at_fork function

zmedico:
The multiprocessing.util.register_after_fork function is undocumented, so use the documented os.register_at_fork function instead.


thesamesam (approved):

zmedico:
We can now remove manual _ForkWatcher hook invocations after os.fork() as in https://github.com/gentoo/portage/pull/1242.

PR #1177, opened 2023-11-06
ebuild: add some missing newlines

wh0:
messages would show up like this

```
!!! BINPKG_COMPRESS unsupported zstd. Missing package: app-arch/zstdCalculating dependencies... done!
Dependency resolution took 0.04 s.
```

in early bootstrapping before you have app-arch/zstd set up


thesamesam:
Thanks!

PR #1180, opened 2023-11-11
tests: resolver: test_use_dep_defaults.py: add test for || () dep int…

thesamesam:
…eraction

Bug: https://bugs.gentoo.org/917145

### DIFF
--- a/lib/portage/tests/resolver/test_use_dep_defaults.py
+++ b/lib/portage/tests/resolver/test_use_dep_defaults.py
@@ -1,4 +1,4 @@
-# Copyright 2010 Gentoo Foundation
+# Copyright 2010-2023 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 from portage.tests import TestCase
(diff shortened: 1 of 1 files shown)


lib/portage/tests/resolver/test_use_dep_defaults.py
```diff
@@ -55,3 +55,195 @@ def testUseDepDefaultse(self):
...
+            # !!! One of the following packages is required to complete your request:
+            # - sys-apps/systemd-253.11::test_repo (Change USE: +gnuefi)
+            # (dependency required by "sys-kernel/installkernel-systemd-2-r4::test_repo" [ebuild])
+            # (dependency required by "sys-kernel/installkernel-systemd" [argument])
+            ResolverPlaygroundTestCase(
+                ["sys-kernel/installkernel-systemd"],
+                success=False,
+                ambiguous_merge_order=True,
+                options={
+                    "--autounmask": "n",
```
thesamesam, line 139: I'm not sure if we have a way to test for the ".. is required to complete your request" bit, so maybe I'll drop this one.

lib/portage/tests/resolver/test_use_dep_defaults.py
```diff
@@ -55,3 +55,195 @@ def testUseDepDefaultse(self):
...
+                "EAPI": 8,
+                "RDEPEND": "!sys-kernel/installkernel-systemd",
+            },
+            "sys-kernel/installkernel-systemd-2-r4": {
+                "EAPI": 8,
+                "RDEPEND": """
+                    !sys-kernel/installkernel-gentoo
+                    || (
+                        sys-apps/systemd[gnuefi(-)]
+                        sys-apps/systemd[kernel-install(-)]
```
zmedico, line 99: Why not swap the order of preference for gnuefi vs. kernel-install flags? Since neither one is satisfied, dep_zapdeps will prefer the one on the left.

PR #1182, opened 2023-11-15
find_smallest_cycle: Do not return single node for asap mode

zmedico:
For asap mode, it is possible for gather_deps to prematurely select a single node instead of a cycle, as observed in bug
917259. Continue search for a true cycle in this case.

When not in asap mode, we sometimes need to select a single node here, as demonstrated by PerlRebuildBugTestCase which selects automake this way.

The included AlternativesGzipTestCase demonstrates correct behavior, though it fails to reproduce bug 917259. Even though this patch has not yet been proven to fix bug 917259, it is still useful because it limits the ability of find_smallest_cycle to misbehave.

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -9308,8 +9308,10 @@ class depgraph:
                 def find_smallest_cycle(mergeable_nodes, local_priority_range):
                     if prefer_asap and asap_nodes:
                         nodes = asap_nodes
+                        asap_mode = True
                     else:
                         nodes = mergeable_nodes
+                        asap_mode = False
                     # When gathering the nodes belonging to a runtime cycle,
                     # we want to minimize the number of nodes gathered, since
                     # this tends to produce a more optimal merge order.
@@ -9339,6 +9341,15 @@ class depgraph:
                             if gather_deps(
                                 priority, mergeable_nodes, selected_nodes, node
                             ):
+                                if len(selected_nodes) == 1 and asap_mode:
+                                    # For asap mode, it is possible for gather_deps to
+                                    # prematurely select a single node instead of a
+                                    # cycle, as observed in bug 917259. Continue search
+                                    # for a true cycle in this case. When not in asap
+                                    # mode, we sometimes need to select a single node
+                                    # here, as demonstrated by PerlRebuildBugTestCase
+                                    # which selects automake this way.
+                                    continue
                                 if smallest_cycle is None or len(selected_nodes) < len(
                                     smallest_cycle
                                 ):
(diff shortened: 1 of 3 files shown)


lib/portage/tests/resolver/test_alternatives_gzip.py
```diff
@@ -0,0 +1,82 @@
...
+)
+
+
+class AlternativesGzipTestCase(TestCase):
+    def testAlternativesGzip(self):
+        """
+        Test bug 917259, where app-alternatives/gzip was upgraded before
+        its pigz RDEPEND was installed.
+        """
+        ebuilds = {
```
thesamesam, line 17: Yeah, I imagine we need some deep `@system` deps in-between in order to get the bad behaviour to reproduce.

If I had to guess, maybe it involved pigz depending on zlib, but dunno. Not going to worry about it unless we can reproduce it, as the problem is pretty well-understood.

lib/_emerge/depgraph.py
```diff
@@ -9339,6 +9341,15 @@ def find_smallest_cycle(mergeable_nodes, local_priority_range):
                             if gather_deps(
                                 priority, mergeable_nodes, selected_nodes, node
                             ):
+                                if len(selected_nodes) == 1 and asap_mode:
```
thesamesam, line 9344: Thanks. I should've really thought of this before when we spoke about it. We should still obviously think about proper priorities but this is good enough.

thesamesam (approved):
Thank you! I'm not really that worried about the test because I've not seen any instances of this in the wild apart from that one report, and I feel like with the combination of this and the Perl test, we already have a lot better coverage for this than we did before.

thesamesam:
oh, I guess we should tag bug [756199](https://bugs.gentoo.org/756199) too, even if it's not fully fixed (or is it now?)

zmedico:
I'll just close this one for now, since I reproduced the bug and found that changing the topological sort to remove only one node at a time solved the problem:

```patch
diff --git a/lib/_emerge/depgraph.py b/lib/_emerge/depgraph.py
index 8d72c3241..1c99ade43 100644
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -9464,8 +9464,8 @@ class depgraph:
                             ignore_priority=ignore_priority
                         )
                         if leaves:
-                            cycle_digraph.difference_update(leaves)
-                            selected_nodes.extend(leaves)
+                            cycle_digraph.remove(leaves[0])
+                            selected_nodes.append(leaves[0])
                             break
                     else:
                         selected_nodes.extend(cycle_digraph)
```

PR #1183, opened 2023-11-16
depclean: return failure when no packages are selected for depclean

ajakk:
Based on the preceding writemsg, one might think that this function could be used for more than depcleans, but there's no other reference to it in the codebase.

With this patch:
```
$ sudo ./bin/emerge --debug --verbose --depclean foo; echo $? myaction depclean
myopts {'--debug': True, '--binpkg-respect-use': 'y', '--quiet-build': 'y', '--regex-search-auto': 'y', '--verbose': True} [DEBUG] Using selector: EpollSelector
--- Couldn't find 'foo' to depclean.
>>> No packages selected for removal by depclean
1
```
Without:
```
$ sudo ./bin/emerge --debug --verbose --depclean foo; echo $? myaction depclean
myopts {'--debug': True, '--binpkg-respect-use': 'y', '--quiet-build': 'y', '--regex-search-auto': 'y', '--verbose': True} [DEBUG] Using selector: EpollSelector
--- Couldn't find 'foo' to depclean.
>>> No packages selected for removal by depclean
0
```
Bug: https://bugs.gentoo.org/917120

### DIFF
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -843,7 +843,7 @@ def action_depclean(
                 )
         if not matched_packages:
             writemsg_level(f">>> No packages selected for removal by {action}\n")
-            return 0
+            return 1
 
     # The calculation is done in a separate function so that depgraph
     # references go out of scope and the corresponding memory
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,12 @@ Release notes take the form of the following optional categories:
 * Bug fixes
 * Cleanups
 
+portage-3.0.59 (UNRELEASED)
+
+Features:
+* emerge: depclean now returns with failure if no packages are matched
+  (bug #917120)
+
 portage-3.0.58 (2023-12-14)
 --------------
 


lib/_emerge/actions.py
```diff
@@ -842,7 +842,7 @@ def action_depclean(
                 )
         if not matched_packages:
             writemsg_level(f">>> No packages selected for removal by {action}\n")
-            return 0
+            return 1
```
thesamesam, line 845: What does this give with bare `emerge -vc` if there's nothing for it to do?

ajakk:
```
~/gentoo/portage $ sudo ./bin/emerge -vc; echo $?
[copious output]
  x11-wm/i3-4.22 pulled in by:
    x11-misc/polybar-3.6.3-r3 requires x11-wm/i3

>>> No packages selected for removal by depclean
Packages installed:   2356
Packages in world:    431
Packages in system:   49
Required packages:    2356
Number removed:       0
0
~/gentoo/portage $ git log -1
commit 3edfa9ced9cb0455239f6cc47867f1ebd9bcb203 (HEAD -> bug917120, fork/bug917120)
Author: John Helmert III <ajak@gentoo.org>
Date:   Wed Nov 15 19:45:14 2023 -0800

    depclean: return failure when no packages are selected for depclean
```

The code in question is:

```
    if myfiles:
        args_set.update(myfiles)
        matched_packages = False
        for x in args_set:
            if vardb.match(x):
                matched_packages = True
            else:
                writemsg_level(
                    f"--- Couldn't find '{x.replace('null/', '')}' to {action}.\n",
                    level=logging.WARN,
                    noiselevel=-1,
                )
        if not matched_packages:
            writemsg_level(f">>> No packages selected for removal by {action}\n")
            return 1

    # The calculation is done in a separate function so that depgraph
    # references go out of scope and the corresponding memory
    # is freed before we call unmerge().
    rval, cleanlist, ordered, req_pkg_count = calc_depclean(
        settings, trees, ldpath_mtimes, myopts, action, args_set, spinner
    )
```

So, it looks like it tries to resolve the packages specified on the command line first, and then returns if no packages were matched based on the command line arguments, but it doesn't follow that path at all if there's nothing on the command line. In that case, execution flows right into the below `calc_depclean` which does exit with success if there's nothing to do.

zmedico (approved):
Seems consistent with this part of 4e2abb474f0fc624c51948f0939e3123f6382992 with respect to `--unmerge`:
```diff
--- a/pym/_emerge/unmerge.py
+++ b/pym/_emerge/unmerge.py
@@ -515,7 +517,7 @@ def unmerge(root_config, myopts, unmerge_action,
                writemsg_level=writemsg_level)
 
        if rval != os.EX_OK:
-               return 0
+               return rval
 
        enter_invalid = '--ask-enter-invalid' in myopts
        vartree = root_config.trees["vartree"]
```

ajakk:
Merged (hm, missed that pram didn't add the `Closes:`): b7e35beaa4168b315ad3ae294b71706d24c795b0

PR #1185, opened 2023-11-18
find_smallest_cycle: Increase ignore_priority to find smaller cycles

zmedico:
Fix AlternativesGzipTestCase by increasing ignore_priority in
order to find smaller cycles. This causes some changes in merge
order for MergeOrderTestCase, but they appear to be acceptable
since they prevent temporarily unsatisfied RDEPEND by relying
on satisfied installed build-time dependencies.

Add a test case for [bug 690436](https://bugs.gentoo.org/690436), since the change to merge order
in MergeOrderTestCase is related (see commit 680276cc4d4f).

Bug: https://bugs.gentoo.org/690436

### DIFF
--- a/lib/portage/tests/resolver/test_merge_order.py
+++ b/lib/portage/tests/resolver/test_merge_order.py
@@ -382,10 +382,12 @@ class MergeOrderTestCase(TestCase):
                 ambiguous_merge_order=True,
                 # The following merge order assertion reflects optimal order for
                 # a circular relationship which is DEPEND in one direction and
-                # RDEPEND in the other.
-                merge_order_assertions=(
-                    ("app-misc/circ-buildtime-a-1", "app-misc/circ-buildtime-c-1"),
-                ),
+                # RDEPEND in the other. However, it is not respected because
+                # it would result in a temporarily broken RDEPEND, so we instead
+                # rely on satisfied installed build-time dependencies.
+                # merge_order_assertions=(
+                #    ("app-misc/circ-buildtime-a-1", "app-misc/circ-buildtime-c-1"),
+                # ),
                 mergelist=[
                     (
                         "app-misc/circ-buildtime-b-1",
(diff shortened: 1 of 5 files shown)


lib/portage/tests/resolver/test_merge_order.py
```diff
@@ -382,10 +382,12 @@ def testMergeOrder(self):
...
-                merge_order_assertions=(
-                    ("app-misc/circ-buildtime-a-1", "app-misc/circ-buildtime-c-1"),
-                ),
+                # RDEPEND in the other. However, it is not respected because
+                # it would result in a temporarily broken RDEPEND, so we instead
+                # rely on satisfied installed build-time dependencies.
+                # merge_order_assertions=(
+                #    ("app-misc/circ-buildtime-a-1", "app-misc/circ-buildtime-c-1"),
+                # ),
                 mergelist=[
```
thesamesam, line 391: I probably won't review this tonight because I want to dig up whatever bug we added that original test case for, just to make sure the change preserves the intent too. It probably does but I'm curious as to what it was.

zmedico: Seems this particular assertion was added as an xfail in 7dc4131eca73aee2a78fd5c4339cb41760f7a54f, in preparation for a later patch related to https://bugs.gentoo.org/690436. So, it's a thing that I made up myself, rather than being a response to something that happened in the field.

zmedico: Actually https://bugs.gentoo.org/703676 is linked from https://github.com/gentoo/portage/pull/491.

zmedico: A test case that tests for bug 703676 and has the cycle shown in https://bugs.gentoo.org/703676#c1 would be good to have.

zmedico: Added RebuildGhostscriptTestCase, and it passes.

thesamesam (approved):
Thank you!

lib/portage/tests/resolver/test_merge_order.py
thesamesam: Thank you for doing the digging! I was distracted with https://bugs.gentoo.org/917496 and didn't have a chance yet. Looks good.

thesamesam: By the way, I wonder if we should at some point think about some special `--debug` mode which dumps the depgraph in the form of (a large) `ResolverPlaygroundTestCase` input for playing around with.

zmedico: Yeah, I suppose we could save state to pickle or json file and use that to populate a ResolverPlayground and ResolverPlaygroundTestCase.

zmedico:
The change from this PR was reverted in cdc781349337fa755bc15773e2a87e4b41f5ff1e, and a new fix is in https://github.com/gentoo/portage/pull/1192.

PR #1186, opened 2023-11-19
_run_pkg_pretend: Refer to emerge-fetch.log for binpkg parallel-fetch

zmedico:
If there is a parallel-fetch task fetching a binpkg in the background, refer to emerge-fetch.log for status updates, just like the Binpkg class does prior to pkg_setup.

Bug: https://bugs.gentoo.org/760893


thesamesam (approved):

thesamesam:
Thank you! This has confused quite a few people.

PR #1188, opened 2023-11-26
Stop implicitly manipulating `NO_COLOR`/`NOCOLOR`

mgorny:
Stop implicitly forcing `NO_COLOR` and `NOCOLOR` in ebuild environment. This is undesired for several reasons:

1. It makes it impossible to control color for emerge output independently of command output, e.g. when one goes to a pty while the other goes to logs.

2. It makes it impossible to get color output in logs when running `emerge --jobs ...`.

3. Forcing `NO_COLOR=1` turns out to cause random test failures, and while fixing them is commendable, it is a pain for arch testing and it is currently blocking stabilization requests.

With the new approach, the color output in programs is consistent between using ``--jobs`` or ``--quiet-build``, and not.  Therefore, both cases generate uniform logs.  In order to obtain logs free of color codes, one can either filter them via `ansifilter(1)` (as the manpages already recommend) or explicitly set `NO_COLOR`.

Furthermore, one can combine color-free build output (for clean logs) with colorful emerge output by using:

    NO_COLOR=true emerge --color y ...

Bug: https://bugs.gentoo.org/918515

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,6 +9,11 @@ Release notes take the form of the following optional categories:
 portage-3.0.56 (UNRELEASED)
 --------------
 
+Breaking changes:
+* ebuild: no longer implicitly set NOCOLOR and NO_COLOR when color output
+  is disabled for emerge itself.  In order to obtain color-free logs, please
+  either explicitly set NO_COLOR environment variable or use e.g. ansifilter.
+
 Features:
 * sync: git: Add sync-git-verify-max-age-days option.
 
--- a/bin/ebuild
+++ b/bin/ebuild
@@ -132,10 +132,6 @@ def main():
         or not sys.stdout.isatty()
     ):
         portage.output.nocolor()
-        portage.settings.unlock()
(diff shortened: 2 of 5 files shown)


ajakk (approved):

zmedico (approved):

floppym (approved):
Sounds good to me.

zmedico:
We could also do the same in `bin/egencache` and `lib/portage/_emirrordist/main.py` while we're here.

thesamesam (approved):
Please document this in `NEWS` under "Breaking changes", then ok, thanks.

I think the total of the effects here are enough for us to do this even if it might surprise some people.

mgorny:
> We could also do the same in `bin/egencache` and `lib/portage/_emirrordist/main.py` while we're here.

I wasn't sure whether I ought to touch it since it doesn't really affect ebuilds and I don't know if `egencache` doesn't internally use any tools that would spew color.

thesamesam:
ok, wfm, ship it

PR #1189, opened 2023-11-26
Do not link NO_COLOR to isatty or TERM

zmedico:
Do not link NO_COLOR to isatty or TERM, in order to allow more explicit control. Also do not foce NO_COLOR for background jobs in AbstractEbuildProcess.

The link between NO_COLOR and the --color option is currently preserved, though it could be useful to decouple them, since the user may want to set NO_COLOR just for ebuild processes while keeping color enabled in the portage UI. However, this would drastically change the result of having NO_COLOR set to true in make.conf, so they remain coupled for now.

Bug: https://bugs.gentoo.org/918515

### DIFF
--- a/bin/ebuild
+++ b/bin/ebuild
@@ -126,10 +126,7 @@ def main():
         portage.debug.set_trace(True)
 
     if not opts.color == "y" and (
-        opts.color == "n"
-        or portage.util.no_color(portage.settings)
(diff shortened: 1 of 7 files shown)


zmedico:
Closing this for now since I assume https://github.com/gentoo/portage/pull/1188 is more in alignment with mgorny's wishes.

PR #1192, opened 2023-11-28
Increase ignore_priority during topological sort for runtime cycle

zmedico:
Fix AlternativesGzipTestCase by increasing ignore_priority in order to find smaller groups of leaf nodes during topological sort for runtime cycles. This causes some changes in merge order for MergeOrderTestCase, but they appear to be acceptable since they prevent temporarily unsatisfied RDEPEND by relying on satisfied installed build-time dependencies.


thesamesam (commented):
This seems like a decent compromise, but I'm wondering if the fact it doesn't fix a bug I reported just now (https://bugs.gentoo.org/918683#c6) means it's not worth doing given it's less elegant (which isn't your fault, just a consequence of not being able to pick the original way) and it doesn't fix that case, while the original patch seems to..

Perf is fine with this one too.

zmedico:
> This seems like a decent compromise, but I'm wondering if the fact it doesn't fix a bug I reported just now (https://bugs.gentoo.org/918683#c6) means it's not worth doing given it's less elegant (which isn't your fault, just a consequence of not being able to pick the original way) and it doesn't fix that case, while the original patch seems to..

For bug 918683 I think we can use "finer-grained dep priority" as I demonstrated in https://github.com/gentoo/portage/pull/1193#issuecomment-1829178126.

thesamesam (approved):

thesamesam:
Looks like CI started failing after one of the rebases?

zmedico:
> Looks like CI started failing after one of the rebases?

Yeah, 2e298ea7ba36801a1cfba6e4cbfc16a7c05ee73d causes the test case for 3487594cd8f46a5c83caaab3a9425321443e5efc to fail. Looking into it.

zmedico:
If we fix up DepPriority, to account for 2e298ea7ba36801a1cfba6e4cbfc16a7c05ee73d, for purposes of digraph.debug_print:
```diff
--- a/lib/_emerge/DepPriority.py
+++ b/lib/_emerge/DepPriority.py
@@ -19,24 +19,27 @@ class DepPriority(AbstractDepPriority):
 
         buildtime_slot_op                       0
         buildtime                              -1
-        runtime                                -2
-        runtime_post                           -3
-        optional                               -4
-        (none of the above)                    -5
+        runtime_slot_op                        -2
+        runtime                                -3
+        runtime_post                           -4
+        optional                               -5
+        (none of the above)                    -6
 
         """
 
         if self.optional:
-            return -4
+            return -5
         if self.buildtime_slot_op:
             return 0
         if self.buildtime:
             return -1
-        if self.runtime:
+        if self.runtime_slot_op:
             return -2
-        if self.runtime_post:
+        if self.runtime:
             return -3
-        return -5
+        if self.runtime_post:
+            return -4
+        return -6
 
     def __str__(self):
         if self.ignored:
```
Then the debug output shows runtime_slot_op coming into play for the dev-lang/perl-5.36.1-r3 -> virtual/libcrypt-2-r1 dep:
```
runtime cycle digraph (9 nodes):

(app-alternatives/gzip-1:0/0::test_repo, ebuild scheduled for merge) depends on
  (app-arch/pigz-2.8:0/0::test_repo, ebuild scheduled for merge) (runtime)
(sys-libs/zlib-1.3-r1:0/1::test_repo, ebuild scheduled for merge) depends on
  (sys-devel/automake-1.16.5-r1:0/0::test_repo, installed) (buildtime)
(sys-devel/automake-1.16.5-r1:0/0::test_repo, installed) depends on
  (dev-lang/perl-5.36.1-r3:0/0::test_repo, installed) (runtime)
  (app-alternatives/gzip-1:0/0::test_repo, ebuild scheduled for merge) (optional)
(app-arch/pigz-2.8:0/0::test_repo, ebuild scheduled for merge) depends on
  (sys-libs/zlib-1.3-r1:0/1::test_repo, ebuild scheduled for merge) (buildtime)
  (app-alternatives/gzip-1:0/0::test_repo, ebuild scheduled for merge) (runtime_post)
(dev-lang/perl-5.36.1-r3:0/0::test_repo, installed) depends on
  (sys-libs/zlib-1.3-r1:0/1::test_repo, ebuild scheduled for merge) (runtime)
  (virtual/libcrypt-2-r1:0/2::test_repo, installed) (runtime_slot_op)
(sys-libs/glibc-2.37-r7:0/0::test_repo, installed) depends on
  (dev-lang/perl-5.36.1-r3:0/0::test_repo, installed) (runtime)
  (sys-apps/locale-gen-2.23-r1:0/0::test_repo, installed) (runtime)
(sys-apps/locale-gen-2.23-r1:0/0::test_repo, installed) depends on
  (app-alternatives/gzip-1:0/0::test_repo, ebuild scheduled for merge) (runtime)
(virtual/libcrypt-2-r1:0/2::test_repo, installed) depends on
  (sys-libs/libxcrypt-4.4.36:0/0::test_repo, installed) (runtime)
(sys-libs/libxcrypt-4.4.36:0/0::test_repo, installed) depends on
  (sys-libs/glibc-2.37-r7:0/0::test_repo, installed) (runtime)
```

zmedico:
> I guess reverting the NormalRange part helps? We could add another tier to `DepPriorityNormalRange` for `runtime_slot_op`..

It's actually just that 3487594cd8f4 was very fragile, being highly dependent on the order of leaf node selection. I think https://github.com/gentoo/portage/pull/1195 is probably a good solution.

PR #1193, opened 2023-11-28
DepPrioritySatisfiedRange: strengthen _ignore_runtime for runtime slot dep operators

thesamesam:
In the reported bug, net-misc/curl gets merged (binary), then dev-util/cmake gets
bulit (from source) which fails because one of the built curl's dependencies
(net-libs/nghttp2) is missing:
```
[binary   R    ] net-misc/curl-8.4.0::test_repo  USE="http2%*" 0 KiB
[ebuild     U  ] dev-util/cmake-3.27.8::test_repo [3.26.5-r2::test_repo] 0 KiB
[ebuild  N     ] net-libs/nghttp2-1.57.0::test_repo  0 KiB
```

We should consider the existing dev-util/cmake as sufficient for nghttp2 and instead do:
1. net-libs/nghttp2-1.57.0
2. net-misc/curl-8.4.0 (binary)
3. dev-util/cmake-3.27.8 (upgrade, we didn't need to do this first as we already had a CMake installed)

Bug: https://bugs.gentoo.org/918683

### DIFF
--- a/lib/_emerge/DepPriorityNormalRange.py
+++ b/lib/_emerge/DepPriorityNormalRange.py
@@ -37,7 +37,13 @@ class DepPriorityNormalRange:
     def _ignore_runtime(cls, priority):
         if priority.__class__ is not DepPriority:
             return False
-        return bool(priority.optional or not priority.buildtime)
+        # If we ever allow "optional" runtime_slot_op, we'll need
+        # to adjust this appropriately. But only build time dependencies
+        # are optional right now, so it's not an issue as-is.
+        return bool(
+            not priority.runtime_slot_op
+            and (priority.optional or not priority.buildtime)
+        )
 
     ignore_medium = _ignore_runtime
     ignore_medium_soft = _ignore_runtime_post
--- a/lib/_emerge/DepPrioritySatisfiedRange.py
+++ b/lib/_emerge/DepPrioritySatisfiedRange.py
@@ -1,4 +1,4 @@
-# Copyright 1999-2013 Gentoo Foundation
+# Copyright 1999-2023 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 from _emerge.DepPriority import DepPriority
@@ -89,7 +89,16 @@ class DepPrioritySatisfiedRange:
     def _ignore_runtime(cls, priority):
         if priority.__class__ is not DepPriority:
             return False
-        return bool(priority.satisfied or priority.optional or not priority.buildtime)
+        # We could split this up into 2 variants (ignoring satisfied
+        # runtime_slot_op, and not) if we need more granularity for ignore_priority
+        # in future.
+        return bool(
+            (
+                (not priority.runtime_slot_op)
+                or (priority.satisfied and priority.runtime_slot_op)
+            )
+            and (priority.satisfied or priority.optional or not priority.buildtime)
+        )
 
     ignore_medium = _ignore_runtime
     ignore_medium_soft = _ignore_satisfied_buildtime_slot_op
--- a/lib/portage/tests/resolver/test_runtime_cycle_merge_order.py
+++ b/lib/portage/tests/resolver/test_runtime_cycle_merge_order.py
@@ -7,6 +7,8 @@ from portage.tests.resolver.ResolverPlayground import (
     ResolverPlaygroundTestCase,
 )
 
+import pytest
+
 
 class RuntimeCycleMergeOrderTestCase(TestCase):
     def testRuntimeCycleMergeOrder(self):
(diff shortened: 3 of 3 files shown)


zmedico:
It looks like this little tweak produces a workable merge order:
```diff
--- a/lib/_emerge/DepPrioritySatisfiedRange.py
+++ b/lib/_emerge/DepPrioritySatisfiedRange.py
@@ -89,7 +89,10 @@ class DepPrioritySatisfiedRange:
     def _ignore_runtime(cls, priority):
         if priority.__class__ is not DepPriority:
             return False
-        return bool(priority.satisfied or priority.optional or not priority.buildtime)
+        return (
+            bool(priority.satisfied or priority.optional or not priority.buildtime)
+            and not priority.runtime_slot_op
+        )
 
     ignore_medium = _ignore_runtime
     ignore_medium_soft = _ignore_satisfied_buildtime_slot_op
```
Gives this merge order:
```
[ebuild     U  ] dev-util/cmake-3.27.8::test_repo [3.26.5-r2::test_repo] 0 KiB
[ebuild  N     ] net-libs/nghttp2-1.57.0::test_repo  0 KiB
[binary   R    ] net-misc/curl-8.4.0::test_repo  USE="http2%*" 0 KiB
```
I got the idea from seeing the `runtime_slot_op` in this debug output:
```
runtime cycle digraph (3 nodes):

(dev-util/cmake-3.27.8:0/0::test_repo, ebuild scheduled for merge) depends on
  (net-misc/curl-8.4.0:0/0::test_repo, binary scheduled for merge) (buildtime)
(net-misc/curl-8.4.0:0/0::test_repo, binary scheduled for merge) depends on
  (net-libs/nghttp2-1.57.0:0/0::test_repo, ebuild scheduled for merge) (runtime_slot_op)
(net-libs/nghttp2-1.57.0:0/0::test_repo, ebuild scheduled for merge) depends on
  (dev-util/cmake-3.27.8:0/0::test_repo, ebuild scheduled for merge) (buildtime)
```

thesamesam:
Trying to figure out if the test change in `lib/portage/tests/resolver/test_merge_order.py::MergeOrderTestCase::testMergeOrder` is okay (wonder if it could be displayed better somehow?)

thesamesam:
> Trying to figure out if the test change in `lib/portage/tests/resolver/test_merge_order.py::MergeOrderTestCase::testMergeOrder` is okay (wonder if it could be displayed better somehow?)

```
>>> from difflib import unified_diff
>>> lines = list(unified_diff(expected, got))
>>> for line in lines: print(line)
--- 
+++ 
@@ -11,8 +11,6 @@
 !app-misc/installed-blocker-a
 app-misc/circ-direct-a-1
 app-misc/circ-direct-b-1
-x11-base/xorg-server-1.14.1
-media-libs/mesa-9.1.3
 app-misc/circ-buildtime-a-1
 app-misc/circ-buildtime-b-1
 app-misc/circ-buildtime-c-1
@@ -20,3 +18,5 @@
 app-misc/circ-satisfied-a-1
 app-misc/circ-satisfied-b-1
 app-misc/circ-satisfied-c-1
+x11-base/xorg-server-1.14.1
+media-libs/mesa-9.1.3
>>> 
```

I think this is fine as the order between the two is preserved and nothing else there depends on xorg-server or mesa.

lib/portage/tests/resolver/test_runtime_cycle_merge_order.py
```diff
@@ -74,3 +74,147 @@ def testRuntimeCycleMergeOrder(self):
                 self.assertEqual(test_case.test_success, True, test_case.fail_msg)
         finally:
             playground.cleanup()
+
+    def testBuildtimeRuntimeCycleMergeOrder(self):
```
thesamesam, line 78: I might be able to tighten this up or add a shorter test case but I'll look at that later.

lib/_emerge/DepPrioritySatisfiedRange.py
```diff
@@ -89,7 +89,10 @@ def _ignore_satisfied_buildtime_slot_op(cls, priority):
     def _ignore_runtime(cls, priority):
         if priority.__class__ is not DepPriority:
             return False
-        return bool(priority.satisfied or priority.optional or not priority.buildtime)
+        return (
+            bool(priority.satisfied or priority.optional or not priority.buildtime)
+            and not priority.runtime_slot_op
```
zmedico, line 94: We should allow it to be ignored here if priority.satisfied is True.

In DepPriorityNormalRange._ignore_runtime, we should also return False for priority.runtime_slot_op, and there it doesn't matter if it's satisfied or not.

lib/_emerge/DepPriorityNormalRange.py
```diff
@@ -37,7 +37,10 @@ def _ignore_runtime_post(cls, priority):
     def _ignore_runtime(cls, priority):
         if priority.__class__ is not DepPriority:
             return False
-        return bool(priority.optional or not priority.buildtime)
+        return bool(
+            not priority.runtime_slot_op
+            and (priority.optional or not priority.buildtime)
+        )
```
zmedico, line 43: This sort of perverts the meaning of `_ignore_runtime`, since it is no longer possible to ignore an unsatisfied runtime_slot_op dependency. The perversion is good in the sense that it gives a strict merge order guarantee for unsatisfied runtime_slot_op dependencies.

lib/_emerge/DepPrioritySatisfiedRange.py
```diff
@@ -89,7 +89,13 @@ def _ignore_satisfied_buildtime_slot_op(cls, priority):
...
         if priority.__class__ is not DepPriority:
             return False
-        return bool(priority.satisfied or priority.optional or not priority.buildtime)
+        return bool(
+            (
+                (not priority.runtime_slot_op)
+                or (priority.satisfied and priority.runtime_slot_op)
+            )
+            and (priority.satisfied or priority.optional or not priority.buildtime)
+        )
```
zmedico, line 98: We could conceivably split this method into one that ignores satisfied runtime_slot_op and one that does not, providing some finer granularity for ignore_priority loops. Maybe it's not worth the trouble until the need arises, though.

lib/_emerge/DepPrioritySatisfiedRange.py
zmedico: Looks good!

lib/_emerge/DepPriorityNormalRange.py
zmedico: If we ever added support for an "optional" runtime_slot_op then we would need to adjust this code to account for that, but it's fine for now since only buildtime dependencies can currently be optional. Maybe a --onlydeps runtime_slot_op could be considered optional.

lib/_emerge/DepPrioritySatisfiedRange.py
thesamesam: Ah, good point. I'll add a comment about it.

PR #1195, opened 2023-11-29
Prefer installed leaves in runtime cycle topological sort

zmedico:
In order to avoid possibly merging a package too early, prefer installed leaves in runtime cycle topological sort. This fixes an AlternativesGzipTestCase failure that arose after 2e298ea7ba36 caused leaves to be selected in a slightly different order.


thesamesam (approved):
Nice!

PR #1198, opened 2023-12-02
Set SYSROOT for has/best_version so that they work when IPC is disabled

chewi:
This issue broke cross-compiling.


chewi:
CC @immolo

PR #1199, opened 2023-12-05
BuildLogger: Close self._stdin after fork

zmedico:
In order to ensure that we can observe EOF on the read end of the pipe, close self._stdin after fork. Since portage.locks._close_fds() already does something similar for _lock_manager instances which have a close() method, it will also work with these _file_close_wrapper objects.

Bug: https://bugs.gentoo.org/919072


arkamar (approved):
LGTM, it solves the issue for me.

floppym:
I don't really understand how this code works; it's not obvious to me how the `close` function is invoked after a `fork`. Could you fill in the gaps for me?

floppym:
Thanks.

thesamesam:
fork semantics continue to give me the creeps but if this fixes the issue, it's fine with me.

Maybe add that last comment into the commit message as well?

zmedico:
> fork semantics continue to give me the creeps but if this fixes the issue, it's fine with me.

Yeah, it'll be nice when spawn is the default multiprocessing start method.

> Maybe add that last comment into the commit message as well?

Good idea, done.

PR #1200, opened 2023-12-05
EbuildFetchTestCase: Fix dict/shelve key order assumption

zmedico:
Bug: https://bugs.gentoo.org/919105


thesamesam (approved):
Thank you Zac!

PR #1201, opened 2023-12-06
circular_dependency: Handle SonameAtom

zmedico:
Treat soname deps as unconditional for now. In some
cases they can be avoided by a rebuild with changed
USE, but ebuilds sometimes do not specify the
corresponding conditional dependency (especially for
system packages like gcc which provides libstdc++.so.6
and libgcc_s.so.1).

Bug: https://bugs.gentoo.org/919311

### DIFF
--- a/lib/_emerge/resolver/circular_dependency.py
+++ b/lib/_emerge/resolver/circular_dependency.py
@@ -1,4 +1,4 @@
-# Copyright 2010-2020 Gentoo Authors
+# Copyright 2010-2023 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 import logging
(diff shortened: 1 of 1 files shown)


lib/_emerge/resolver/circular_dependency.py
```diff
@@ -132,9 +132,15 @@ def _find_suggestions(self):
...
                     changed_parent = ppkg
-                    parent_atom = atom.unevaluated_atom
+                    parent_atom = atom
                     break
 
+            if parent_atom.package:
+                parent_atom = parent_atom.unevaluated_atom
+            else:
+                # SonameAtom is always unconditional.
+                continue
```
zmedico, line 142: The bug was actually triggered by this cycle which only occurs with emerge --ignore-soname-deps=n because gmp does not specify a dependency on gcc:
```
(dev-libs/gmp-6.3.0-2:0/10.4::gentoo, binary scheduled for merge to '/tmp/testroot/') depends on
 (sys-devel/gcc-13.2.1_p20230826-2:13/13::gentoo, binary scheduled for merge to '/tmp/testroot/') (runtime)
  (dev-libs/gmp-6.3.0-2:0/10.4::gentoo, binary scheduled for merge to '/tmp/testroot/') (runtime_slot_op)
```
Also, in theory you can break the cycle by disabling USE=cxx for gmp, but since the dependency is unspecified it's kind of an odd thing to do.

zmedico: Adjusted comments and commit message for this.

thesamesam (approved):

PR #1202, opened 2023-12-06
DepPriority{Normal,Satisfied}Range: weaken _ignore_runtime for cross root

zmedico:
When the dependency parent is for a cross root (ROOT != /) package, weaken _ignore_runtime in order to tolerate runtime cycles that are less problematic for cross root packages.

The included test case fails with this error without the fix:
```
 * Error: circular dependencies:

(dev-libs/gmp-6.3.0:0/10.4::test_repo, binary scheduled for merge to '/tmp/tmp25nwdjn7/cross_root/') depends on
 (sys-devel/gcc-13.2.1_p20230826:0/0::test_repo, binary scheduled for merge to '/tmp/tmp25nwdjn7/cross_root/') (runtime)
  (dev-libs/gmp-6.3.0:0/10.4::test_repo, binary scheduled for merge to '/tmp/tmp25nwdjn7/cross_root/') (runtime_slot_op)

It might be possible to break this cycle
by applying the following change:
- dev-libs/gmp-6.3.0 (Change USE: -cxx)
```

Bug: https://bugs.gentoo.org/919174

### DIFF
--- a/lib/_emerge/DepPriority.py
+++ b/lib/_emerge/DepPriority.py
@@ -1,11 +1,11 @@
-# Copyright 1999-2013 Gentoo Foundation
+# Copyright 1999-2023 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 from _emerge.AbstractDepPriority import AbstractDepPriority
 
 
 class DepPriority(AbstractDepPriority):
-    __slots__ = ("satisfied", "optional", "ignored")
+    __slots__ = ("cross", "ignored", "optional", "satisfied")
 
     def __int__(self):
         """
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -3630,7 +3630,7 @@ class depgraph:
                     blocker=False,
                     depth=depth,
                     parent=pkg,
-                    priority=self._priority(runtime=True),
+                    priority=self._priority(cross=self._cross(pkg.root), runtime=True),
                     root=pkg.root,
                 )
                 if not self._add_dep(dep, allow_unsatisfied=allow_unsatisfied):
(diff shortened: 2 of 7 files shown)


lib/_emerge/DepPriority.py
```diff
@@ -1,11 +1,11 @@
-# Copyright 1999-2013 Gentoo Foundation
+# Copyright 1999-2023 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 from _emerge.AbstractDepPriority import AbstractDepPriority
 
 
 class DepPriority(AbstractDepPriority):
-    __slots__ = ("satisfied", "optional", "ignored")
+    __slots__ = ("cross", "ignored", "optional", "satisfied")
```
thesamesam, line 8: "cross" (as in cross-root) is fine with me, but maybe "altroot" or something would work if people are likely to get confused by it

lib/_emerge/depgraph.py
```diff
@@ -4616,6 +4633,12 @@ def _priority(self, **kwargs):
             priority_constructor = DepPriority
         return priority_constructor(**kwargs)
 
+    def _cross(self, eroot):
+        """
+        Returns True if the ROOT for  the given EROOT is not /.
```
thesamesam, line 4638:
```suggestion
        Returns True if the ROOT for the given EROOT is not /.
```

thesamesam (approved):
looks good if you've tested it (I haven't yet), it would be nice if we had more tests for the general `ROOT` situation. from a glance we have very few and it's an interesting case for dep resolution

lib/_emerge/DepPriority.py
zmedico: I've realized that this should also apply to the cross-prefix case where ROOT is /, if that matters for naming.

lib/_emerge/depgraph.py
zmedico: I've adjusted this method and docstring to also account for the cross-prefix case.

lib/_emerge/DepPriority.py
thesamesam: thanks! cross it is then

lib/_emerge/depgraph.py
thesamesam: thanks!

PR #1203, opened 2023-12-08
sets: VariableSet: flatten dependencies

thesamesam:
This is needed to pick up www-client/firefox's BDEPEND of "|| ( virtual/rust:0/a virtual/rust:0/b ...)" for @rust-rebuild.

### DIFF
--- a/lib/portage/_sets/dbapi.py
+++ b/lib/portage/_sets/dbapi.py
@@ -172,7 +172,7 @@ class VariableSet(EverythingSet):
             for include in self._includes:
                 include_atoms.append(Atom(include))
 
-            for x in use_reduce(values, token_class=Atom):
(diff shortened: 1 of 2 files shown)


lib/portage/_sets/dbapi.py
```diff
@@ -172,7 +172,7 @@ def _filter(self, atom):
             for include in self._includes:
                 include_atoms.append(Atom(include))
 
-            for x in use_reduce(values, token_class=Atom):
+            for x in use_reduce(values, token_class=Atom, flat=True):
                 if not isinstance(x, Atom):
```
thesamesam, line 176: might be able to just drop this with flat, not checked yet

thesamesam: Ah, thanks!

PR #1204, opened 2023-12-10
bindbapi: add SIZE to _pkg_str_aux_keys

zmedico:
The _pkg_str_aux_keys are used inside dbapi.update_ents,
and need to contain SIZE in order for the fakedbapi
_instance_key_multi_instance method to operate correctly.

Incorrect operation of _instance_key_multi_instance could
prevent binarytree.inject from removing an old instance
from its internal state. It could also trigger a KeyError
in bindbapi.aux_update as in bug 918597, since it could
cause binarytree.getname to return a non-existent path.
It could also cause binarytree.getname to return an
existing but incorrect path, which might trigger an
InvalidBinaryPackageFormat exception as in bug 906675.

Bug: https://bugs.gentoo.org/906675
Bug: https://bugs.gentoo.org/918597
Bug: https://bugs.gentoo.org/919668

### DIFF
--- a/lib/portage/dbapi/bintree.py
+++ b/lib/portage/dbapi/bintree.py
@@ -84,9 +84,12 @@ class bindbapi(fakedbapi):
     _known_keys = frozenset(
         list(fakedbapi._known_keys) + ["CHOST", "repository", "USE"]
     )
+    # Must include keys used to create _pkg_str attributes used in
+    # the fakedbapi _instance_key_multi_instance method.
     _pkg_str_aux_keys = fakedbapi._pkg_str_aux_keys + (
         "BUILD_ID",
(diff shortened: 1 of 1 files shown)


thesamesam (approved):
Good catch!

We may want to add a comment saying they should be kept in sync?

zmedico:
> We may want to add a comment saying they should be kept in sync?

Comment added:
```python
    # Must include keys used to create _pkg_str attributes used in
    # the fakedbapi _instance_key_multi_instance method.
    _pkg_str_aux_keys = fakedbapi._pkg_str_aux_keys + (
        "BUILD_ID",
        "BUILD_TIME",
        "SIZE",
        "_mtime_",
    )
```

PR #1206, opened 2023-12-11
sync: git: include signing key and git revision in log output

Flowdalic:


thesamesam:
Could you include old vs new output in the commit message?

Flowdalic:
> Could you include old vs new output in the commit message?

Done

PR #1207, opened 2023-12-18
ResolverPlayground: Create $EPREFIX/usr/bin/python symlink

zmedico:
This ensures that CrossDepPriorityTestCase uses the correct python implementation when it calls emerge, by overriding behavior of `#!/usr/bin/env python` shebangs.

Fixes: 1d856747ada4 ("DepPriority{Normal,Satisfied}Range: weaken _ignore_runtime for cross root")

### DIFF
--- a/lib/portage/tests/resolver/ResolverPlayground.py
+++ b/lib/portage/tests/resolver/ResolverPlayground.py
@@ -1,4 +1,4 @@
-# Copyright 2010-2021 Gentoo Authors
+# Copyright 2010-2023 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 import bz2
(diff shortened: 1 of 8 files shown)


lib/portage/tests/resolver/ResolverPlayground.py
```diff
@@ -683,7 +684,7 @@ def _load_config(self):
             create_trees_kwargs["target_root"] = self.target_root
 
         env = {
-            "PATH": os.environ["PATH"],
+            "PATH": f"{self.eprefix}/usr/sbin:{self.eprefix}/usr/bin:{os.environ['PATH']}",
```
thesamesam, line 687: Do the hacks from e.g. `lib/portage/tests/emerge/test_libc_dep_inject.py`, `lib/portage/tests/emerge/test_emerge_blocker_file_collision.py` need adjusting too?

zmedico: Yeah, updated other things using ResolverPlayground to inhert its PATH.

thesamesam: Thanks!

PR #1208, opened 2023-12-18
Scheduler: Reset current_task in _run_pkg_pretend

zmedico:
For completed tasks, reset current_task to None so that it does not cause the clean phase to execute with an invalid environment in the finally block.

Bug: https://bugs.gentoo.org/920258


thesamesam (approved):
This makes sense to me, thank you!

PR #1209, opened 2023-12-19
depgraph: Use strip_libc_deps in _eliminate_rebuilds

zmedico:
The included test case will fail without strip_libc_deps because the "injected" >=sys-libs/glibc-2.37 dependency will be interpreted as a dependency change, triggering unwanted reinstall of app-misc/A-1.

Bug: https://bugs.gentoo.org/915494

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -2966,6 +2966,23 @@ class depgraph:
                 return flags
         return None
 
+    def _installed_libc_deps(self, eroot):
+        """
+        Return find_libc_deps result for installed packages from the
+        given EROOT.
+        """
+        try:
+            return self._frozen_config._libc_deps_cache[eroot]
+        except (AttributeError, KeyError) as e:
+            if isinstance(e, AttributeError):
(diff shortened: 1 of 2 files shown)


lib/_emerge/depgraph.py
```diff
@@ -3726,6 +3726,18 @@ def _eliminate_rebuilds(self):
...
+                except (AttributeError, KeyError) as e:
+                    if isinstance(e, AttributeError):
+                        self._frozen_config._libc_deps_cache = {}
+
+                    self._frozen_config._libc_deps_cache[pkg.root] = find_libc_deps(
+                        self._frozen_config._trees_orig[pkg.root]["vartree"].dbapi,
+                        False,
+                    )
+                libc_deps = self._frozen_config._libc_deps_cache[pkg.root]
+
```
zmedico, line 3740: We can add a libc_deps method, so as not to duplicate this code from _changed_deps.

zmedico: Added _installed_libc_deps method.

thesamesam (approved):
Thanks!

PR #1210, opened 2023-12-20
Qa command not found improvements

eli-schwartz:
- check for dash as well
- don't special case and ignore errors in ./configure scripts


thesamesam (approved):
I'd prefix the commit msgs with `ebuild:` as that's the relevant component.

It looks good to me but I'd like @zmedico to take a quick look because he authored the original change to suppress it for configure.

eli-schwartz:
Will fix up the commit messages tomorrow as I've already gone to bed. :D

zmedico (approved):
The `command_not_found_exclude_re` traces back to 930bbbf31c10265b27825426f1eff6d7f17395e6 for https://bugs.gentoo.org/245716 and I guess it's fine to remove and see if anyone complains.

PR #1211, opened 2023-12-21
Add more pre-commit hooks

gothicVI:
Also, add the fixes coming with them.


lib/portage/tests/conftest.py
eli-schwartz, line 1: This feels like a sign that the hook is wrong. It shouldn't be marking this file as executable.

lib/portage/util/changelog.py
eli-schwartz, line 1: And here too.

lib/portage/tests/conftest.py
gothicVI: The file shouldn't have a shebang then.

lib/portage/util/changelog.py
gothicVI: Same here.

lib/portage/tests/conftest.py
eli-schwartz: That's a very brave claim to make.

Although I would perhaps use `#!hint/python` as the shebang instead.

gothicVI: In what sense though?

https://github.com/pre-commit/pre-commit-hooks/blob/main/pre_commit_hooks/check_executables_have_shebangs.py
checks if files with a shebang are executable.

If that's not desired they shouldn't have one, right?

gothicVI: Anyhow, should be rebased and resolved now.

eli-schwartz: In the sense that you're making the very brave and intimidating claim that just because the pre-commit-hooks project says something is "bad", that means it is bad.

Maybe the hook is wrong, like I originally said, and it's totally fine and desired for python files to have shebangs, whether they are marked executable or not. For example, text editors may use them as syntax highlighting hints. Okay, granted, that is more likely to actually *matter* for a python-formatted "rc" file than a .py file, but the general point remains.

On the other hand, I'm a pretty biased person -- I think that the python-pre-commit tool is a bad tool, badly designed, badly fitting into git, and encourages bad mishandling of software installation at a technical design level. I also think that it's a large collection of bad recommendations and incorrect ideas about what to check, so it's in bad taste to boot.

So I guess feel free to take everything I say with a grain of salt. ;)

gothicVI: Thanks for the explanation - could you live with the current state though?

zmedico:
Hi, could you rebase this on current master to verify that everything still passes?

gothicVI:
Hi @zmedico looks good to me.

gothicVI:
Giving this another shot. Please close if not desired.

PR #1213, opened 2023-12-24
ebuild.sh: Warn about colon-separated argument in sandbox functions

ulm:
According to PMS and the Devmanual, only a single item is allowed as argument for addread, addwrite, adddeny and addpredict:
https://projects.gentoo.org/pms/8/pms.html#x1-12300012.3.3
https://devmanual.gentoo.org/function-reference/sandbox-functions/

Bug: https://bugs.gentoo.org/920654

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -12,6 +12,10 @@ Features:
 * emerge: depclean now returns with failure if no packages are matched
   (bug #917120)
 
+Bug fixes:
+* addread, addwrite, adddeny, addpredict: Warn about passing a colon-separated
+  list of paths as argument (bug #920654).
+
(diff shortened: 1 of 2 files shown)


thesamesam:
Two comments:
1. Could you add a reference to the commit where vapier added this?
2. Could you file a bug (and reference it here) for pkgcheck support for detecting it?

Thank you!

ulm:
>     1. Could you add a reference to the commit where vapier added this?
>     2. Could you file a bug (and reference it here) for pkgcheck support for detecting it?

Both done.

thesamesam (approved):

PR #1214, opened 2023-12-24
bindbapi: Update state for package remove in aux_update

zmedico:
When removing a signed gpkg in aux_update, update internal state including $PKGDIR/Packages (important especially for FEATURES=pkgdir-index-trusted).

Bug: https://bugs.gentoo.org/920095
Fixes: a7bbb4fc4d38 ("Fix move_ent with signed binpkg")


thesamesam (approved):
Thanks, that looks right and sorry for missing this before. It feels obvious now I see this (and the test changes...)

PR #1217, opened 2023-12-26
Scheduler: Handle unpack_metadata SignatureException in _run_pkg_pretend

zmedico:
This will handle the InvalidSignature exception that triggered bug 920258, allowing emerge --keep-going to skip the package if possible. All of the gpg stderr output is dumped before the exception is raised, as shown in [bug 920258 comment #0](https://bugs.gentoo.org/920258#c0).

Bug: https://bugs.gentoo.org/920258


thesamesam (approved):
Thanks! Would you mind looking at the general lock-on-updates bit as well? I think it'll probably be worth making a release then with the existing binpkg fixes + that in master.

PR #1218, opened 2023-12-26
_global_updates: Acquire global vardbapi lock

zmedico:
Also use bindbapi.writable attribute in case the
PKGDIR is missing for some reason.

Bug: https://bugs.gentoo.org/587088


thesamesam (approved):
Thank you!

PR #1220, opened 2023-12-29
dbapi: KeyError tolerance during package moves

zmedico:
Raise a new CorruptionKeyError exception type instead of a plain KeyError when os.stat fails. Treat this type of exception as a warning during package moves.

Also fix this error that the test case triggered in the binarytree.remove method:
```
  NameError: name 'binpkg_path' is not defined
```
Bug: https://bugs.gentoo.org/920828


thesamesam (approved):
Thanks!

PR #1222, opened 2023-12-30
EbuildFetchTestCase: Fix dict/shelve key order assumption

zmedico:
Just like https://github.com/gentoo/portage/pull/1200.

Bug: https://bugs.gentoo.org/921107


thesamesam (approved):
Nit: maybe add "Just like 4de288937ad740d5373a0b77a84e224af636ce1e" to the commit message.

Works & tests pass on x86 now, thanks

PR #1227, opened 2024-01-03
tests: add getbinpkg file:// sync testcase for BUILD_ID

thesamesam:
We do a bit of a strange dance here where we:
1) create a binpkg;
2) shift the PKGDIR so that it becomes a "remote" one for use with file:///;
3) try to merge the binpkg but defer checking the exit status;
4) check if the downloaded filename was unnecessarily incremented (-2). if it is,
fail.
5) check the deferred exit status from 3) so that we fail if the merge failed
for another reason (e.g. failed injection).

Bug: https://bugs.gentoo.org/921208

### DIFF
--- a/lib/_emerge/BinpkgFetcher.py
+++ b/lib/_emerge/BinpkgFetcher.py
@@ -106,7 +106,8 @@ class BinpkgFetcher(CompositeTask):
                 finally:
                     if copier.isAlive():
                         copier.cancel()
-
+                if copier.returncode == os.EX_OK:
+                    fetcher.sync_timestamp()
             else:
                 fetcher.start()
                 try:
--- a/lib/portage/tests/emerge/meson.build
+++ b/lib/portage/tests/emerge/meson.build
@@ -1,6 +1,7 @@
 py.install_sources(
     [
         'test_actions.py',
+        'test_binpkg_fetch.py',
         'test_config_protect.py',
         'test_emerge_blocker_file_collision.py',
         'test_emerge_slot_abi.py',
--- a/lib/portage/tests/emerge/test_binpkg_fetch.py
+++ b/lib/portage/tests/emerge/test_binpkg_fetch.py
@@ -1,0 +1,226 @@
+# Copyright 2024 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+import shutil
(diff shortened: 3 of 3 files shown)


thesamesam:
cc @zmedico, just posting so you can use it / we can rebase it later / w/e

lib/portage/tests/update/test_update_dbentry.py
```diff
@@ -520,3 +527,175 @@ def testUpdateDbentryDbapiTestCaseWithSignature(self):
...
+                        os.path.join(
+                            bindb.bintree.pkgdir, "dev-libs", "A", "A-1-1.gpkg.tar"
+                        )
+                    )
+                    or True,
+                )
+            ),
+            # This should succeed if we've correctly saved it as A-1-1.gpkg.tar, not
+            # A-1-2.gpkg.tar, and then also try to unpack the right filename.
+            emerge_cmd
```
thesamesam, line 593: We can make this more finely-grained so we store the exit code, then check what the downloaded binpkg filename is, and then fail earlier on that, or fail if the earlier-stored exit code was bad too. That way, we get two distinct failures with and without https://github.com/gentoo/portage/pull/1226.

lib/portage/tests/update/test_update_dbentry.py
```diff
@@ -1,23 +1,30 @@
 # Copyright 2012-2023 Gentoo Authors
```
thesamesam, line 1: I'll move the new test into a new file.

zmedico:
As noted in [bug 921208 comment #9](https://bugs.gentoo.org/921208#c9) the issue was that the FileCopier added in 1db44d18578a7aee58449cb97e1991cb06c915c3 does not preserve mtime, so this is needed:
```diff
--- a/lib/_emerge/BinpkgFetcher.py
+++ b/lib/_emerge/BinpkgFetcher.py
@@ -106,7 +106,8 @@ class BinpkgFetcher(CompositeTask):
                 finally:
                     if copier.isAlive():
                         copier.cancel()
-
+                if copier.returncode == os.EX_OK:
+                    fetcher.sync_timestamp()
             else:
                 fetcher.start()
                 try:
```

zmedico (approved):
Thanks!

PR #1228, opened 2024-01-04
binarytree: Handle inject failures

zmedico:
Capture stdout and stderr for logging purposes during binarytree inject calls (like https://github.com/gentoo/portage/pull/1221), and use the inject return value to report success or failure.

For failures prior to pkg_pretend, use an eerror/elog message to indicate that the binary package is not usable. Move corresponding elog_process call to a finally block so that it is called for all pkg_pretend failures.

Bug: https://bugs.gentoo.org/921327


thesamesam (approved):

gkubaryk:
tested in coordination with @thesamesam - I already use portage-9999, and this patch applies cleanly in /etc/portage/patches and solves both test cases

PR #1229, opened 2024-01-04
find_smallest_cycle: Prefer installed nodes

zmedico:
Like 9823f70c6e4ef3cdd6abb4d9fc599ce02a138125, prefer installed nodes in order to try and avoid merging something too early.

Bug: https://bugs.gentoo.org/921333


zmedico:
Closing this since it didn't help. It seems that finer-grained scoring is needed as in discussed in https://bugs.gentoo.org/754777.

PR #1232, opened 2024-01-11
Test python 3.13

zmedico:


thesamesam (approved):
wonder how many of the warnings are new

PR #1234, opened 2024-01-13
_emerge: depgraph: lookup unrecognized atoms in updates

thesamesam:
This isn't suitable for merging right now, just posting so I can ask for some help.

Bug: https://bugs.gentoo.org/910572

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -5286,6 +5286,7 @@ class depgraph:
             for atom in sorted(arg.pset.getAtoms()):
                 self._spinner_update()
                 dep = Dependency(atom=atom, onlydeps=onlydeps, root=myroot, parent=arg)
+                fixedup_pkgmove = False
                 try:
                     pprovided = pprovideddict.get(atom.cp)
                     if pprovided and portage.match_from_list(atom, pprovided):
@@ -5318,14 +5319,9 @@ class depgraph:
                         )
 
                         if pkg:
-                            atom = Atom(atom.replace(atom.cp, pkg.cp))
-
-                            if not self._add_pkg(pkg, dep):
-                                writemsg(
-                                    f"\n\n!!! Problem resolving dependencies for {arg.arg=}, {arg.atom=}\n",
-                                    noiselevel=-1,
-                                )
-                                return False, myfavorites
+                            fixedup_pkgmove = True
+                            atom = Atom(f"{atom.replace(atom.cp, pkg.cp)}")
+                            arg = pkg.slot_atom
 
                     if debug:
                         writemsg_level(
@@ -5333,11 +5329,10 @@ class depgraph:
                             noiselevel=-1,
                             level=logging.DEBUG,
                         )
+
                     pkg, existing_node = self._select_package(
                         myroot, atom, onlydeps=onlydeps
                     )
-                    if not arg:
-                        arg = AtomArg(pkg.slot_atom)
 
                     # Is the package installed (at any version)?
                     if pkg and "update_if_installed" in self._dynamic_config.myparams:
@@ -5409,6 +5404,7 @@ class depgraph:
                     if (
                         pkg.installed
                         and "selective" not in self._dynamic_config.myparams
+                        and not fixedup_pkgmove
                         and not self._frozen_config.excluded_pkgs.findAtomForPackage(
                             pkg, modified_use=self._pkg_use_enabled(pkg)
                         )
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -5286,7 +5286,6 @@ class depgraph:
             for atom in sorted(arg.pset.getAtoms()):
                 self._spinner_update()
                 dep = Dependency(atom=atom, onlydeps=onlydeps, root=myroot, parent=arg)
-                fixedup_pkgmove = False
                 try:
                     pprovided = pprovideddict.get(atom.cp)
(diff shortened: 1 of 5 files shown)


lib/_emerge/depgraph.py
```diff
@@ -5024,10 +5025,40 @@ def _select_files(self, myfiles):
                 #   2) It takes away freedom from the resolver to choose other
                 #      possible expansions when necessary.
                 if "/" in x.split(":")[0]:
+                    atomarg_atom = Atom(x, allow_repo=True)
+
+                    # TODO: Doing the lookup here unfairly penalises the success case
+                    if atomarg_atom.package and not atomarg_atom.cp.startswith("null/"):
+                        found_any_match = any(
```
thesamesam, line 5032: @zmedico Do you have an idea for how to best do this? In particular, I don't want to penalise the success case (so "dev-build/make" having to incur a lookup).

zmedico: Maybe we can do this expansion inside _dep_expand, and possibly only do it after category expansion has failed?

thesamesam: OK, pushed a fixup (not squashed it). I tried a few things, including maybe treating this like the old PROVIDE case for virtuals, but this seemed to be the most elegant.

What do you reckon?

lib/_emerge/depgraph.py
```diff
@@ -5255,6 +5260,25 @@ def _resolve(self, myfavorites):
...
+                                pkgmoves_only=True,
+                            ),
+                            None,
+                        )
+
+                        if pkg:
+                            arg = AtomArg(pkg.slot_atom)
+                            atom = pkg.slot_atom
+                            self._add_pkg(pkg, dep)
+                            self._create_graph()
```
zmedico, line 5279: If _add_pkg or _create_graph returns 0 instead of 1 then you need to return. Normally _create_graph is delayed until later. after all the roots have been established.

lib/_emerge/depgraph.py
```diff
@@ -6863,14 +6914,39 @@ def _iter_match_pkgs_atom(self, root_config, pkg_type, atom, onlydeps=False):
                         matched_something = True
                         yield pkg
 
+        if pkgmoves_only or not matched_something:
+            # Try lookup the supplied atom in known updates instead to ease
+            # use in scripts (bug #910572).
+            updates = grab_global_updates(root_config.trees["porttree"].dbapi)
```
zmedico, line 6920: Should cache this in the dynamic config.

lib/_emerge/depgraph.py
zmedico: Seems like _select_package would be better than _iter_match_pkgs_any.

zmedico: Maybe just use _iter_match_pkgs_any to translate the catagory/package-name, and let _select_package handle the rest.

thesamesam: Tried to do that now - but worried I've misunderstood what you mean..

thesamesam: OH, thank you!

thesamesam: OK, in [`2a0f564` (#1234)](https://github.com/gentoo/portage/pull/1234/commits/2a0f564ddbe30ea2a1343589d5fea7ee5617ac9c), it works, but I'm calling `_add_pkg` myself rather than relying on `_select_package` and then the `_add_pkg` later on in the function.

In [`2cbc189` (#1234)](https://github.com/gentoo/portage/pull/1234/commits/2cbc189d871f47ba223c9096df313ab64996d554), I try to use the existing logic more, but that doesn't work.

thesamesam: `_select_package` ends up replacing it with an `installed` Package:
```
pkg=<Package ('binary', '/tmp/tmpruf_hub6/', 'dev-build/make-4.4', 'merge', 'None', '225', '0', '1705479700')>
   binary: dev-build/make-4.4::test_repo
installed: dev-build/make-4.4::test_repo
pkg=<Package ('installed', '/tmp/tmpruf_hub6/', 'dev-build/make-4.4', 'nomerge', 'installed')>
```

zmedico: So, _select_package operates on data from _set_args, so you need to call `_set_args` at some point with updated args. This loop isn't a good place to call `_set_args`, since it's supposed to be called before any packages are added to the graph (before _select_package is ever called), and called with all args at once rather than a single arg. Anyway, the tests pass with this diff:
```diff
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -5321,8 +5321,12 @@ class depgraph:
 
                         if pkg:
                             atom = Atom(atom.replace(atom.cp, pkg.cp))
+                            arg.atom = dep.atom = atom
+                            arg.pset.replace([atom])
+                            self._set_args([arg])
 
-                            if not self._add_pkg(pkg, dep):
+                            # goto _select_package below
+                            if False and not self._add_pkg(pkg, dep):
                                 writemsg(
                                     f"\n\n!!! Problem resolving dependencies for {arg.arg=}, {arg.atom=}\n",
```

zmedico: For example, you might modify `args` inside `_select_files` in the `for x in myfiles:` loop that creates the `args` list, or just after that loop. Then your transformations will already be present when `args` is passed to `_set_args`.

thesamesam: Thanks Zac, I'm going to try get back to this tonight.

PR #1235, opened 2024-01-15
installed_dynlibs: Resolve *.so symlinks

zmedico:
Resolve *.so symlinks to check if they point to regular files inside the top directory. If a symlink points outside the top directory then try to follow the corresponding file inside the top directory if it exists, and otherwise stop following.

Bug: https://bugs.gentoo.org/921170


floppym (approved):
Seems reasonable.

thesamesam (approved):

PR #1236, opened 2024-01-15
emerge: backtrack consistently regardless of --fetchonly

zmedico:
Make the _accept_blocker_conflicts method always return True when backtracking is enabled, so that backtracking results will be identical regardless of options that _accept_blocker_conflicts treats specially. This way, conflicts will only be accepted when backtracking is disabled or all backtracking tries have been exhausted. Make --nodeps imply --backtrack=0, since backtracking is only useful with dependencies.

Make _eliminate_rebuilds return early if there are slot conflicts, since the merge list is invalid anyway, and its possible that state inconsistencies could trigger unexpected exceptions as in bug 922038. Make the KeyError from bug 922038 a warning, and include relevant information to help trace the inconsistency back to the _eliminate_rebuilds method or some other source like  _solve_non_slot_operator_slot_conflicts.

Bug: https://bugs.gentoo.org/922038


thesamesam (approved):

lib/_emerge/depgraph.py
```diff
@@ -3777,11 +3787,45 @@ def _eliminate_rebuilds(self):
...
                 for parent, parent_atom in self._dynamic_config._parent_atoms[pkg]:
-                    priorities = self._dynamic_config.digraph.nodes[pkg][1][parent][:]
+                    try:
+                        priorities = self._dynamic_config.digraph.nodes[pkg][1][parent][
+                            :
+                        ]
+                    except KeyError:
+                        optional_msg = " ({} previously removed from graph)"
+                        warnings.warn(
+                            f"_eliminate_rebuilds inconsistency: parent priorities missing for {parent} -> {pkg} edge"(
```
thesamesam, line 3797: When running tests for https://github.com/gentoo/portage/pull/1234 (`lib/portage/tests/update/test_move_resolve.py`), I noticed:
```
lib/_emerge/depgraph.py:3797
  /home/sam/git/portage/lib/_emerge/depgraph.py:3797: SyntaxWarning: 'str' object is not callable; perhaps you missed a comma?
    f"_eliminate_rebuilds inconsistency: parent priorities missing for {parent} -> {pkg} edge"(
```

PR #1237, opened 2024-01-16
Handle SignatureException during package moves

zmedico:
Ignore package moves for packages that raise SignatureException, just as they are ignored for packages that have a valid signature.

Bug: https://bugs.gentoo.org/922142


thesamesam (approved):
Thanks!

PR #1239, opened 2024-01-26
bin: prefix eqawarn with 'QA Notice'

asarubbo:
Bug: https://bugs.gentoo.org/728046


zmedico (approved):
LGTM.

I used grep to check which files contain `eqwarn` and not `QA Notice:`, and I see you fixed all of those, plus 05prefix and 90world-writable which did not apply `QA Notice:` consistently.

thesamesam (approved):

PR #1245, opened 2024-01-29
estrip: Support debug info deduplication with sys-devel/dwz

zhuyifei1999:
This this gates behind FEATURES=dedupdebug, and packages can opt-out with RESTRICT=dedupdebug, though I'm not sure why a package would need to opt this out, unless dwz could break something I'm not aware of...

Bug: https://bugs.gentoo.org/906368

### DIFF
--- a/bin/estrip
+++ b/bin/estrip
@@ -15,8 +15,8 @@ exp_tf() {
 		eval ${var}_${flag}=$(tf has ${flag} ${!var})
 	done
 }
-exp_tf FEATURES compressdebug installsources nostrip splitdebug xattr
-exp_tf PORTAGE_RESTRICT binchecks installsources splitdebug strip
+exp_tf FEATURES compressdebug dedupdebug installsources nostrip splitdebug xattr
+exp_tf PORTAGE_RESTRICT binchecks dedupdebug installsources splitdebug strip
 
 if ! ___eapi_has_prefix_variables; then
 	EPREFIX= ED=${D}
@@ -201,6 +201,10 @@ fi
 [[ ${debugedit} ]] && debugedit_found=true || debugedit_found=false
 debugedit_warned=false
 
+dwz=$(type -P dwz)
+[[ ${dwz} ]] && dwz_found=true || dwz_found=false
+dwz_warned=false
+
 __multijob_init
 
 # Setup ${T} filesystem layout that we care about.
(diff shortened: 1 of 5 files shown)


zhuyifei1999:
Idk if the name / implementation is good or not, free free to bikeshed.

And if we need multi-file dedup, I can try to add that too.

ArsenArsen:
this lgtm so far (seems to be the same as the patch i 'reviewed' on bz), wrt. multi-file mode, that seems quite desirable (but i wouldn't hold the pr over it).  it seems to me that, for dwz multi-file mode, we'd need to do some more advanced processing rather than just installing to dwz/whatever in order for debug tools to pick up on it.  perhaps we should ask some folks about prior art?

thesamesam (approved):
This looks good to me, I'm going to daily-drive it for a few days.

This also reminds me of split dwarf, but the last time I looked into that, I found we were blocked on https://sourceware.org/bugzilla/show_bug.cgi?id=31085, although maybe there's another way of doing it that wouldn't be terrible.

zhuyifei1999:
> perhaps we should ask some folks about prior art?

Who would I ask about this? 

Though, I haven't looked into the details on how dwz works to see how similar its multi-file mode is similar to dwo files gcc makes.

ArsenArsen:
dwz has a discussion ML

looking into `dwo`s is likely worth it too, but I don't think it's related to the problem at hand, unless I'm missing something

ArsenArsen:
so, I spoke with mjw and he shared how fedora uses dwz multi-file mode.  in short, they generate a file per package and combine all the debug info into it, then wire it up so that debuginfod and debuggers know how to handle it.  this logic is in https://sourceware.org/cgit/debugedit/tree/scripts/find-debuginfo.in AFAIU 

this doesn't seem too hard to implement as part of this PR, as long as dedupdebug happens after splitdebug and other debuginfo-related passes

zhuyifei1999:
> so, I spoke with mjw and he shared how fedora uses dwz multi-file mode. in short, they generate a file per package and combine all the debug info into it, then wire it up so that debuginfod and debuggers know how to handle it. this logic is in https://sourceware.org/cgit/debugedit/tree/scripts/find-debuginfo.in AFAIU

Oh cool I'll take a look.

> this doesn't seem too hard to implement as part of this PR, as long as dedupdebug happens after splitdebug and other debuginfo-related passes

When I tried this the problem seems to be that dwz seems unable to process compressdebug-ed debug info. I can take another look to how to workaround this.

> (to be clear, I wouldn't hold the PR over it - if you wish to merge the current code, that's okay too. sam noticed significant gains already in single file mode, and I imagine very good results with multi-file mode)

I don't mind either way, but let me try multi-file and see how complex it'd be.

thesamesam:
I'll merge it as-is and we can do a followup later, as it seems to work great for me as-is. Thank you!

PR #1249, opened 2024-02-03
process.spawn: Use multiprocessing.Process for returnproc

zmedico:
Use multiprocessing.Process for returnproc, so that fork will stop being used when python makes "spawn" the default multiprocessing start method.

Continue to use _start_fork when returnproc is not enabled, for backward compatibility. Ultimately, it can be removed at the same time as the returnpid parameter.

### DIFF
--- a/lib/portage/process.py
+++ b/lib/portage/process.py
@@ -19,7 +19,7 @@ import warnings
 
 from dataclasses import dataclass
 from functools import lru_cache
-from typing import Any, Optional, Callable
+from typing import Any, Optional, Callable, Union
 
 from portage import os
 from portage import _encodings
@@ -28,6 +28,7 @@ import portage
 
 portage.proxy.lazyimport.lazyimport(
     globals(),
+    "portage.util._async.ForkProcess:ForkProcess",
     "portage.util._eventloop.global_event_loop:global_event_loop",
     "portage.util.futures:asyncio",
     "portage.util:dump_traceback,writemsg,writemsg_level",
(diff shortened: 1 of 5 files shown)


thesamesam (approved):

lib/portage/process.py
```diff
@@ -1370,6 +1375,48 @@ def _start_fork(
...
+    should ensure that any relevant file descriptors are
+    non-inheritable and therefore automatically closed on exec.
+
+    Note that for the "spawn" multiprocessing start method,
+    fd_pipes is sent to the child process asynchronously in a
+    background thread. Therefore, the caller cannot safely
+    close the corresponding file descriptors until after the
+    background thread has finished sending fd_pipes. The
+    current simplest practice is to close fd_pipes file
+    descriptors after waiting for the process to exit.
```
zmedico, line 1399: I ran the tests with multiprocessing start method spawn, and it looks like some file descriptors are being closed too early, so I need to hunt those down and see what can be done. Maybe an option is needed to have selected file descriptors in fd_pipes closed immediately after they've been sent to the child process.
```
Exception in callback SpawnProcess._main_exit(<Task finishe... descriptor')>)
handle: <Handle SpawnProcess._main_exit(<Task finishe... descriptor')>)>
  File "/usr/lib/python3.12/asyncio/events.py", line 84, in _run
    self._context.run(self._callback, *self._args)
  File "portage/lib/_emerge/SpawnProcess.py", line 198, in _main_exit
    main_task.result()
  File "portage/lib/portage/util/_async/ForkProcess.py", line 112, in _main
    await self.scheduler.run_in_executor(
  File "/usr/lib/python3.12/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "portage/lib/portage/util/_async/ForkProcess.py", line 103, in _send_fd_pipes
    multiprocessing.reduction.send_handle(
  File "/usr/lib/python3.12/multiprocessing/reduction.py", line 184, in send_handle
    sendfds(s, [handle])
  File "/usr/lib/python3.12/multiprocessing/reduction.py", line 149, in sendfds
    sock.sendmsg([msg], [(socket.SOL_SOCKET, socket.SCM_RIGHTS, fds)])
OSError: [Errno 9] Bad file descriptor
```

zmedico: Maybe the solution is to have ForkProcess created duplicates that it is free to close when it finishes sending them.

zmedico: Yeah, the duplicate fd idea turned out really well, allowing me to revert 69a7c68b45074bec28d6092573b01bb54ac39419.

thesamesam:
btw it might be worth us adding a single impl (3.12 maybe) to CI w/ a hack to use spawn

zmedico:
> btw it might be worth us adding a single impl (3.12 maybe) to CI w/ a hack to use spawn

Yeah, that's a necessity. I suppose we could use an environment variable to trigger `multiprocessing.set_start_method("spawn", force=True)` at the beginning of all portage programs. Then there's this little script I use to launch pytest:
```python
import multiprocessing
import sys
import pytest

if __name__ == "__main__":
    multiprocessing.set_start_method("spawn", force=True)
    sys.exit(pytest.console_main())
```

PR #1250, opened 2024-02-03
doebuild: Add returnproc parameter and deprecate returnpid

zmedico:
Raise NotImplementedError if returnproc is enabled for anything other than the "depend" phase, since corresponding returnpid support has long been deprecated.


thesamesam (approved):

zmedico:
I've moved the additional commits to https://github.com/gentoo/portage/pull/1251 as a draft due to test failures with `multiprocessing.set_start_method("spawn", force=True)`.

PR #1251, opened 2024-02-04
Bug 916566 returnproc migration

zmedico:
Now includes a 3.12-dev ci job patched to set the multiprocessing start method to spawn, which took 7m 11s compared to 4m 16s for the 3.12-dev fork ci job.

### DIFF
--- a/lib/portage/package/ebuild/doebuild.py
+++ b/lib/portage/package/ebuild/doebuild.py
@@ -19,6 +19,7 @@ import sys
 import tempfile
 from textwrap import wrap
 import time
+from typing import Union
 import warnings
 import zlib
 
@@ -246,14 +247,21 @@ def _doebuild_spawn(phase, settings, actionmap=None, **kwargs):
 
 
 def _spawn_phase(
-    phase, settings, actionmap=None, returnpid=False, logfile=None, **kwargs
+    phase,
+    settings,
+    actionmap=None,
+    returnpid=False,
+    returnproc=False,
+    logfile=None,
+    **kwargs,
 ):
-    if returnpid:
+    if returnproc or returnpid:
         return _doebuild_spawn(
             phase,
             settings,
             actionmap=actionmap,
             returnpid=returnpid,
+            returnproc=returnproc,
             logfile=logfile,
             **kwargs,
         )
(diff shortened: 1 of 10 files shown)


lib/portage/package/ebuild/doebuild.py
```diff
@@ -867,17 +882,25 @@ def doebuild(
...
             "portage.doebuild() called "
             "with returnpid parameter enabled. This usage will "
             "not be supported in the future.",
-            DeprecationWarning,
+            UserWarning,
             stacklevel=2,
         )
+    elif returnpid:
+        warnings.warn(
+            "The portage.doebuild() returnpid paramenter is deprecated and replaced by returnproc",
```
thesamesam, line 900:
```suggestion
            "The portage.doebuild() returnpid parameter is deprecated and replaced by returnproc",
```

lib/portage/util/futures/executor/fork.py
```diff
@@ -40,7 +40,9 @@ def submit(self, fn, *args, **kwargs):
         """
         future = self._loop.create_future()
         proc = AsyncFunction(
-            target=functools.partial(self._guarded_fn_call, fn, args, kwargs)
+            target=functools.partial(self._guarded_fn_call, fn, args, kwargs),
+            # Directly inherit stdio streanms and run in the foreground with no log.
```
thesamesam, line 44:
```suggestion
            # Directly inherit stdio streams and run in the foreground with no log.
```

thesamesam (approved):

zmedico:
> > Now includes a 3.12-dev ci job patched to set the multiprocessing start method to spawn, which took 7m 11s compared to 4m 16s for the 3.12-dev fork ci job.
> 
> I'm really curious about this part -- spawn should be quite a bit faster where `posix_spawn` is eligible at least. There's [python/cpython#114467](https://github.com/python/cpython/issues/114467) and [python/cpython#113117](https://github.com/python/cpython/issues/113117) which are two cases it's not used where it could be.

The issue is that there is a lot of back-and-forth between the parent and child process during the child process setup. The parent needs to pickle any referenced objects and send them. ForkProcess needs to send fd_pipes just before the target function is called.

It's slow for unit tests where a disproportionate number of processes are created and destroyed, but for normal operation the difference should be pretty negligible.

PR #1253, opened 2024-02-05
ForkProcess: Prevent redundant pipe and set_term_size recursion

zmedico:
When process.spawn is updated to call ForkProcess
for bug 916566, it needs to avoid recursion via
set_term_size.

Bug: https://bugs.gentoo.org/923750

NOTE: This patch is also in https://github.com/gentoo/portage/pull/1251.

### DIFF
--- a/lib/_emerge/SpawnProcess.py
+++ b/lib/_emerge/SpawnProcess.py
@@ -41,7 +41,7 @@ class SpawnProcess(SubProcess):
     )
 
     __slots__ = (
-        ("args", "log_filter_file")
+        ("args", "create_pipe", "log_filter_file")
         + _spawn_kwarg_names
         + (
             "_main_task",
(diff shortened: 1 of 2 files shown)


lib/_emerge/SpawnProcess.py
```diff
@@ -60,15 +60,27 @@ def _start(self):
         else:
             self.fd_pipes = self.fd_pipes.copy()
         fd_pipes = self.fd_pipes
+        previous_stdio_redir = tuple(fd for fd in (0, 1, 2) if fd in fd_pipes)
```
zmedico, line 63: This previous_stdio_redir variable is sort of an ugly heuristic to detect when we're called via `process.spawn`. I think I will just go ahead an add an explicit boolean parameter for _start_proc to provide. I guess we can call it `create_pipe` (a previous patch added a `create_pty` parameter, but we really should not create either a pipe or a pty when called via `process.spawn`.

zmedico: Added `create_pipe` parameter.

thesamesam (approved):

PR #1254, opened 2024-02-05
check_locale: Call asynchronously in EbuildPhase

zmedico:
Since os.fork() is unsafe in threaded processes, use multiprocessing.Process instead. This way the fork will be eliminated when the default multiprocessing start method changes to "spawn".

Change config.environ() check_locale calls to async_check_locale calls in the EbuildPhase _async_start method in order to eliminate synchronous waiting for child processes in the main event loop thread.

Bug: https://bugs.gentoo.org/923841

### DIFF
--- a/lib/_emerge/EbuildPhase.py
+++ b/lib/_emerge/EbuildPhase.py
@@ -1,4 +1,4 @@
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2024 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 import functools
@@ -24,6 +24,7 @@ from portage.package.ebuild.prepare_build_dirs import (
     _prepare_fake_distdir,
     _prepare_fake_filesdir,
 )
+from portage.eapi import _get_eapi_attrs
 from portage.util import writemsg, ensure_dirs
 from portage.util._async.AsyncTaskFuture import AsyncTaskFuture
 from portage.util._async.BuildLogger import BuildLogger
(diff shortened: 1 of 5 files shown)


lib/portage/util/locale.py
```diff
@@ -116,29 +139,21 @@ def check_locale(silent=False, env=None):
...
+    proc = multiprocessing.Process(
+        target=_set_and_check_locale,
+        args=(silent, env, None if env is None else portage._native_string(mylocale)),
+    )
+    proc.start()
+    proc = portage.process.MultiprocessingProcess(proc)
+    await proc.wait()
+
     pyret = None
-    if os.WIFEXITED(ret):
```
thesamesam, line 140: At some point, we should check over our instances of WIFEXITED and other shenanigans to see if we can port to ` os.waitstatus_to_exitcode(status)`.

lib/portage/util/locale.py
```diff
@@ -116,29 +139,21 @@ def check_locale(silent=False, env=None):
...
-                os._exit(2)
-            else:
-                os._exit(0 if ret else 1)
-        except Exception:
-            traceback.print_exc()
-            os._exit(2)
-
-    pid2, ret = os.waitpid(pid, 0)
-    assert pid == pid2
+    # TODO: Make async version of check_locale and call it from
```
thesamesam, line 142: This is stale now, right?

lib/_emerge/EbuildPhase.py
```diff
@@ -94,6 +96,31 @@ def _start(self):
         self._start_task(AsyncTaskFuture(future=future), self._async_start_exit)
 
     async def _async_start(self):
+
+        eapi_attrs = _get_eapi_attrs(self.settings["EAPI"])
+        if eapi_attrs.posixish_locale:
+            split_LC_ALL(self.settings)
```
zmedico, line 102: I've realized that this split_LC_ALL call will delete LC_ALL, and the config.reset() method will not restore the old LC_ALL unless it is in the config's backupenv. So, I need to think about how to fix this.

lib/portage/util/locale.py
zmedico: Yeah, I'll remove it.

lib/portage/util/locale.py
zmedico: Yeah, I didn't realize it was that prevalent but now I see there 7 remaining.

lib/_emerge/EbuildPhase.py
zmedico: I've solved this by making split_LC_ALL add an empty LC_ALL setting instead of deleting it. And when eapi_attrs.posixish_locale is True, config.environ() will assert correct LC_ALL state, and delete the empty LC_ALL setting.

thesamesam (approved):

PR #1255, opened 2024-02-06
actions: Add muliprocessing start-method to matrix

zmedico:
Add multiprocessing start-method to matrix so that we are prepared for when python changes to default to spawn. Exclude all python versions except 3.12-dev for now.

This adds a conditional ci step that patches the sources just for the spawn start-method. It patches all bin/* scripts with python shebangs, and also patches the test command in meson.build.

Bug: https://bugs.gentoo.org/914876

@thesamesam this does something like what you suggested in https://github.com/gentoo/portage/pull/1249#issuecomment-1925460360. This implementation is kind of ugly, but maybe good enough for the first iteration. I've also added this to https://github.com/gentoo/portage/pull/1251.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,9 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
+        start-method:
+          - 'fork'
+          - 'spawn'
         python-version:
           - '3.9'
           - '3.10'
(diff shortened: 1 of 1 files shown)


thesamesam (approved):
It's gnarly but disappointingly, there's no way to control it from the command line (was really hoping for something like the new var for CPU cores). Thank you!

zmedico:
I temporarily added a `multiprocessing.get_start_method() == "fork"` to one of the tests, and inadvertently found out that somehow `multiprocessing.set_start_method("spawn", force=True)` in the pytest command is practically ignored when the pytest-xdist plugin is used. We still get a lot of useful coverage anyway from the patched bin/* python scripts, but it will be nice if we can fix the pytest-xdist plugin somehow.

PR #1256, opened 2024-02-08
emaint: Print 'usage: ' only once in help message

imwints:
`ArgumentParser` prepends 'usage: ' to the usage message already.


thesamesam:
Thanks! Merged w/ your PR text added to the commit message.

PR #1258, opened 2024-02-10
GPG: Proactively stop to avoid "GPG keepalive failed" error in pypy ci jobs

zmedico:
GPG: Proactively stop to avoid "GPG keepalive failed" error in pypy ci jobs

This seems to help mitigate pypy ci job hangs like those fba76a545f2 triggered.

Use threading.Event for thread safety during GPG stop, and use the wait method to improve responsiveness for stop requests.

Bug: https://bugs.gentoo.org/924192

### DIFF
--- a/bin/quickpkg
+++ b/bin/quickpkg
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2024 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 import argparse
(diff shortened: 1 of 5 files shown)


bin/quickpkg
```diff
@@ -341,10 +341,6 @@ def quickpkg_main(options, args, eout):
         portage.settings.features.remove("xattr")
         portage.settings.lock()
 
-    if portage.settings.get("BINPKG_GPG_SIGNING_KEY", None):
```
thesamesam, line 344: I wasn't sure if this was OK because of `unmerge-backup` and `downgrade-backup` but it looks like they just call the entrypoint normally.

thesamesam (approved):

PR #1259, opened 2024-02-11
_overlap_dnf: deduplicate any-of blocks

zmedico:
Duplicate any-of blocks are eliminated since DNF expansion of duplicates is nonsensical.

Bug: https://bugs.gentoo.org/891137


thesamesam (approved):
Thanks!

PR #1260, opened 2024-02-11
spawn_wrapper: Make pre_exec function picklable

zmedico:
Local functions are unpicklable, which triggered this error
with the multiprocessing "spawn" start method:
```
AttributeError: Can't pickle local object 'spawn_wrapper.__call__.<locals>._pre_exec'
```
Bug: https://bugs.gentoo.org/924273

### DIFF
--- a/lib/portage/_selinux.py
+++ b/lib/portage/_selinux.py
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2024 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 # Don't use the unicode-wrapped os and shutil modules here since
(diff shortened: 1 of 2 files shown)


thesamesam (approved):
Maybe explicitly say in the commit message that it's the local/nested function which makes it unpickleable? Thanks!

zmedico:
> Maybe explicitly say in the commit message that it's the local/nested function which makes it unpickleable? Thanks!

I installed libselinux[python] to verify that `partial(setexec, self._con)` is picklable in the fixed version. In the commit message I've now included the exception that would have occurred during pickling:
```
AttributeError: Can't pickle local object 'spawn_wrapper.__call__.<locals>._pre_exec'
```

PR #1261, opened 2024-02-11
process._exec: Use _start_fork for os.fork() error handling

zmedico:
Use _start_fork for os.fork() error handling, ensuring that if exec fails then the child process will display a traceback before it exits via os._exit to suppress any finally blocks from parent's call stack (bug 345289).

Bug: https://bugs.gentoo.org/924313


thesamesam (approved):
Add Bug: tag for 345289? (I know I'm a bit pedantic with this but it's useful when doing archeology to have the reverse mapping).

(Also, this was much easier to review with -w, need to remember to do that for some of these rather than trying w/ gh, oops)

zmedico:
> Add Bug: tag for 345289? (I know I'm a bit pedantic with this but it's useful when doing archeology to have the reverse mapping).

Done, thanks!

> (Also, this was much easier to review with -w, need to remember to do that for some of these rather than trying w/ gh, oops)

Nice!

PR #1262, opened 2024-02-11
EbuildBuild: Execute EbuildFetchonly in fork

zmedico:
Execute EbuildFetchonly in a fork in case it needs to fetch a mirror layout file, since that would trigger "RuntimeError: This event loop is already running" in this context.

Bug: https://bugs.gentoo.org/924287


lib/portage/package/ebuild/fetch.py
```diff
@@ -1562,6 +1562,7 @@ def fetch(
                 tried_locations.add(loc)
                 if listonly:
                     writemsg_stdout(loc + " ", noiselevel=-1)
+                    fetched = 2
```
thesamesam, line 1565: This needs to become an enum but not something to worry about now.

thesamesam (approved):

PR #1264, opened 2024-02-12
sync: Add method to download zip archives

legionus:
Add a simple method for synchronizing repository as a snapshot in a zip archive. The implementation does not require external utilities to download and unpack archive. This makes the method very cheap.

The main usecase being considered is obtaining snapshots of github repositories, but many other web interfaces for git also support receiving snapshots in zip format.

For example, to get a snapshot of the master branch:

  https://github.com/gentoo/portage/archive/refs/heads/master.zip
  https://gitweb.gentoo.org/proj/portage.git/snapshot/portage-master.zip

or a link to a snapshot of the tag:

  https://github.com/gentoo/portage/archive/refs/tags/portage-3.0.61.zip

### DIFF
--- a/lib/portage/sync/modules/zipfile/__init__.py
+++ b/lib/portage/sync/modules/zipfile/__init__.py
@@ -21,10 +21,11 @@ module_spec = {
             "sourcefile": "zipfile",
             "class": "ZipFile",
             "description": doc,
-            "functions": ["sync"],
+            "functions": ["sync", "retrieve_head"],
             "func_desc": {
                 "sync": "Performs an archived http download of the "
                 + "repository, then unpacks it.",
+                "retrieve_head": "Returns the checksum of the unpacked archive.",
             },
             "validate_config": CheckSyncConfig,
             "module_specific_options": (),
--- a/lib/portage/sync/modules/zipfile/__init__.py
+++ b/lib/portage/sync/modules/zipfile/__init__.py
@@ -1,0 +1,33 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+# Copyright (C) 2024  Alexey Gladkov <gladkov.alexey@gmail.com>
+
+doc = """Zipfile plug-in module for portage.
+Performs a http download of a portage snapshot and unpacks it to the repo
+location."""
+__doc__ = doc[:]
+
+
+import os
+
+from portage.sync.config_checks import CheckSyncConfig
+
+
+module_spec = {
+    "name": "zipfile",
+    "description": doc,
+    "provides": {
+        "zipfile-module": {
+            "name": "zipfile",
+            "sourcefile": "zipfile",
+            "class": "ZipFile",
+            "description": doc,
+            "functions": ["sync"],
+            "func_desc": {
+                "sync": "Performs an archived http download of the "
+                + "repository, then unpacks it.",
+            },
+            "validate_config": CheckSyncConfig,
+            "module_specific_options": (),
+        },
+    },
+}
--- a/lib/portage/sync/modules/zipfile/zipfile.py
+++ b/lib/portage/sync/modules/zipfile/zipfile.py
@@ -1,0 +1,82 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+# Copyright (C) 2024  Alexey Gladkov <gladkov.alexey@gmail.com>
+
+import os
+import os.path
+import logging
+import zipfile
+import shutil
+import tempfile
+import urllib.request
+
+import portage
+from portage.util import writemsg_level
+from portage.sync.syncbase import SyncBase
+
+
+class ZipFile(SyncBase):
+    """ZipFile sync module"""
+
+    short_desc = "Perform sync operations on GitHub repositories"
+
+    @staticmethod
+    def name():
+        return "ZipFile"
+
+    def __init__(self):
+        SyncBase.__init__(self, "emerge", ">=sys-apps/portage-2.3")
+
+    def sync(self, **kwargs):
+        """Sync the repository"""
+        if kwargs:
+            self._kwargs(kwargs)
+
+        # initial checkout
+        zip_uri = self.repo.sync_uri
+
+        with urllib.request.urlopen(zip_uri) as response:
+            with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
+                shutil.copyfileobj(response, tmp_file)
+            zip_file = tmp_file.name
+
+        if not zipfile.is_zipfile(zip_file):
+            msg = "!!! file is not a zip archive."
+            self.logger(self.xterm_titles, msg)
+            writemsg_level(msg + "\n", noiselevel=-1, level=logging.ERROR)
+
+            os.unlink(zip_file)
+
+            return (1, False)
+
+        # Drop previous tree
+        shutil.rmtree(self.repo.location)
+
+        with zipfile.ZipFile(zip_file) as archive:
+            strip_comp = 0
+
+            for f in archive.namelist():
+                f = os.path.normpath(f)
+                if os.path.basename(f) == "profiles":
+                    strip_comp = f.count("/")
+                    break
+
+            for n in archive.infolist():
+                p = os.path.normpath(n.filename)
+
+                if os.path.isabs(p):
+                    continue
+
+                parts = p.split("/")
+                dstpath = os.path.join(self.repo.location, *parts[strip_comp:])
+
+                if n.is_dir():
+                    os.makedirs(dstpath, mode=0o755, exist_ok=True)
+                    continue
+
+                with archive.open(n) as srcfile:
+                    with open(dstpath, "wb") as dstfile:
+                        shutil.copyfileobj(srcfile, dstfile)
+
+        os.unlink(zip_file)
+
+        return (os.EX_OK, True)
--- a/lib/portage/sync/modules/zipfile/zipfile.py
+++ b/lib/portage/sync/modules/zipfile/zipfile.py
@@ -35,6 +35,16 @@ class ZipFile(SyncBase):
             return (os.EX_OK, info["etag"][0])
         return (1, False)
 
+    def _do_cmp(self, f1, f2):
+        bufsize = 8 * 1024
+        while True:
+            b1 = f1.read(bufsize)
+            b2 = f2.read(bufsize)
+            if b1 != b2:
+                return False
+            if not b1:
+                return True
+
     def sync(self, **kwargs):
         """Sync the repository"""
         if kwargs:
@@ -76,7 +86,15 @@ class ZipFile(SyncBase):
             return (1, False)
 
         # Drop previous tree
-        shutil.rmtree(self.repo.location)
+        tempdir = tempfile.mkdtemp(prefix=".temp", dir=self.repo.location)
+        tmpname = os.path.basename(tempdir)
+
+        for name in os.listdir(self.repo.location):
+            if name != tmpname:
+                os.rename(
+                    os.path.join(self.repo.location, name),
+                    os.path.join(tempdir, name),
+                )
 
         with zipfile.ZipFile(zip_file) as archive:
             strip_comp = 0
@@ -101,8 +119,20 @@ class ZipFile(SyncBase):
                     continue
 
                 with archive.open(n) as srcfile:
+                    prvpath = os.path.join(tempdir, *parts[strip_comp:])
+
+                    if os.path.exists(prvpath):
+                        with open(prvpath, "rb") as prvfile:
+                            if self._do_cmp(prvfile, srcfile):
+                                os.rename(prvpath, dstpath)
+                                continue
+                        srcfile.seek(0)
+
                     with open(dstpath, "wb") as dstfile:
                         shutil.copyfileobj(srcfile, dstfile)
+
+        # Drop previous tree
+        shutil.rmtree(tempdir)
 
         with open(os.path.join(self.repo.location, ".info"), "w") as infofile:
             if etag:
--- a/lib/portage/sync/modules/zipfile/zipfile.py
+++ b/lib/portage/sync/modules/zipfile/zipfile.py
@@ -10,7 +10,7 @@ import tempfile
 import urllib.request
 
 import portage
-from portage.util import writemsg_level
+from portage.util import writemsg_level, writemsg_stdout
 from portage.sync.syncbase import SyncBase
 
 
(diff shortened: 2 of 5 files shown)


lib/portage/sync/modules/zipfile/zipfile.py
```diff
@@ -0,0 +1,119 @@
...
+            with open(os.path.join(self.repo.location, ".desc"), "w") as descfile:
+                descfile.write(f"{algo}:{hexdigest}\n")
+
+        os.unlink(zip_file)
+
+        return (os.EX_OK, True)
+
+    def update(self):
+        """
+        Internal function to update an existing SVN repository
```
thesamesam, line 112: This isn't SVN:
```suggestion
        Internal function to update an existing SVN repository
```

lib/portage/sync/modules/zipfile/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+        "zipfile-module": {
+            "name": "zipfile",
+            "sourcefile": "zipfile",
+            "class": "ZipFile",
+            "description": doc,
+            "functions": ["sync", "new", "exists", "retrieve_head"],
+            "func_desc": {
+                "sync": "Performs an archived http download of the "
+                + "repository, then unpacks it.  Optionally it performs a "
+                + "gpg verification of the downloaded file(s)",
```
thesamesam, line 28: I can't spot where any gpg verification happens here.

lib/portage/sync/modules/zipfile/zipfile.py
```diff
@@ -0,0 +1,119 @@
...
+
+                if n.is_dir():
+                    os.makedirs(dstpath, mode=0o755, exist_ok=True)
+                    continue
+
+                with archive.open(n) as srcfile:
+                    with open(dstpath, "wb") as dstfile:
+                        shutil.copyfileobj(srcfile, dstfile)
+
+        with open(zip_file, "rb") as archive:
```
thesamesam, line 98: What's the purpose of this?

thesamesam (commented):
Could you give a bit more context if possible (motivation and such)? I didn't realise it was interesting to try fetch snapshots of ebuild repositories.

I'd also say the gzip module might be better as AFAIK git forges always have .tar.gz too. The idea sounds fine to me though.

legionus:
> Could you give a bit more context if possible (motivation and such)? I didn't realise it was interesting to try fetch snapshots of ebuild repositories.

The context of the problem is this: I was installing several new systems using gentoo stage3. I also have my specific overlay on github which is used for installation. But it turned out that gentoo stage3 does not contain either git or unzip. I have to constantly deliver git before adding overlay. The git package is not a very small program and it has dependencies on perl modules for example.

I decided to see what I could do about it and found out that my problem could be solved without external utilities at all. And this works for most git hosting.

I understand that the lack of git in stage3 is not a portage problem, but I thought it would be useful to have a sync methood that didn't require anything other than python.

> I'd also say the gzip module might be better as AFAIK git forges always have .tar.gz too. The idea sounds fine to me though.

I'm not against tar.gz either. I was guided by the fact that github has a link to zip in the interface. For brunch, he can also make a tarball, but the link is only to the “download zip”. This is the only reason :)

lib/portage/sync/modules/zipfile/zipfile.py
legionus: Oops! I made a spoiler which module I used as a basis. My mad. I will fix this.

lib/portage/sync/modules/zipfile/zipfile.py
legionus: If the question is about `copyfileobj`, then `zipfile.extract()` can be passed to the base directory relative to which the file will be unpacked. This doesn't work for us here. Plus, I wanted to control where exactly the file is created and with what permissions. I want to ignore the executable bit on the file if it is in the archive.

floppym: I think the question was about the ".desc" file.

legionus: @floppym  I added desc to remember the checksum of the file we are downloading.
Should I remove it ?

floppym: I don't see any obvious purpose for it.

Seems like there's quite a bit of unnecessary code here actually. I think you could just drop the `exists` and `retrieve_head` methods and let SyncBase provide defaults.

floppym: Also, given the the behavior of `new()` is identical to `update()`, you could inherit from SyncBase instead of NewBase and do the work in the `sync()` method.

legionus:
@floppym I changed the base class to SyncBase, left only the sync method, and threw out the ".desc" file.

It came to my mind the ".desc" may be needed if you implement the ETag header check. But for now there is no such thing, it really isn’t needed yet.

legionus:
I have implemented ETag processing. I may add this support as a separate commit to this PR, or it may be done sometime later.

lib/portage/sync/modules/zipfile/zipfile.py
```diff
@@ -0,0 +1,104 @@
...
+                with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
+                    shutil.copyfileobj(response, tmp_file)
+
+                zip_file = tmp_file.name
+                etag = response.headers.get("etag")
+
+        except urllib.error.HTTPError as resp:
+            if resp.code == 304:
+                writemsg_stdout(">>> The repository has not changed.\n", noiselevel=-1)
+                return (os.EX_OK, True)
```
zmedico, line 51: Seems the return behavior is not well documented here, but anyway the RsyncSync module would return `(os.EX_OK, False)` in an analogous situation in order to indicate that nothing as changed and caches (in case people use the sqlite cache module) do not need to be updated.

lib/portage/sync/modules/zipfile/zipfile.py
```diff
@@ -0,0 +1,104 @@
...
+        shutil.rmtree(self.repo.location)
+
+        with zipfile.ZipFile(zip_file) as archive:
+            strip_comp = 0
+
+            for f in archive.namelist():
+                f = os.path.normpath(f)
+                if os.path.basename(f) == "profiles":
+                    strip_comp = f.count("/")
+                    break
```
zmedico, line 79: It's unfortunate to need a heuristic like this to calculate strip_comp.

We could use module_specific_options to add an option to specify a number of components to strip.

zmedico: On second though maybe the heuristic isn't bad.

lib/portage/sync/modules/zipfile/zipfile.py
legionus: @zmedico Changed the return code to `False` if the repository has not changed.

zmedico: Thanks!

lib/portage/sync/modules/zipfile/zipfile.py
```diff
@@ -0,0 +1,104 @@
...
+            msg = "!!! file is not a zip archive."
+            self.logger(self.xterm_titles, msg)
+            writemsg_level(msg + "\n", noiselevel=-1, level=logging.ERROR)
+
+            os.unlink(zip_file)
+
+            return (1, False)
+
+        # Drop previous tree
+        shutil.rmtree(self.repo.location)
```
zmedico, line 70: Like https://github.com/gentoo/portage/pull/991, we could possibly recycle files that have not changed like this:
```diff
--- a/lib/portage/sync/modules/zipfile/zipfile.py
+++ b/lib/portage/sync/modules/zipfile/zipfile.py
@@ -5,2 +5,3 @@ import os
 import os.path
+import filecmp
 import logging
@@ -68,4 +69,10 @@ class ZipFile(SyncBase):
 
-        # Drop previous tree
-        shutil.rmtree(self.repo.location)
+        old = tempfile.mkdtemp(prefix=".old-revision", dir=self.repo.location)
+        oldbasename = os.path.basename(old)
+        for oldfile in os.listdir(self.repo.location):
+            if oldfile != oldbasename:
+                os.rename(
+                    os.path.join(self.repo.location, oldfile),
+                    os.path.join(old, oldfile),
+                )
 
@@ -96,2 +103,15 @@ class ZipFile(SyncBase):
                         shutil.copyfileobj(srcfile, dstfile)
+                        oldfile = os.path.join(
+                            old, os.path.join(old, *parts[strip_comp:])
+                        )
+                        try:
+                            files_equal = filecmp.cmp(dstfile, oldfile, shallow=False)
+                        except Exception:
+                            pass
+                        else:
+                            if files_equal:
+                                os.rename(oldfile, dstfile)
+
+        # Drop previous tree
+        shutil.rmtree(old)
```

lib/portage/sync/modules/zipfile/zipfile.py
```diff
@@ -0,0 +1,104 @@
...
+                    os.makedirs(dstpath, mode=0o755, exist_ok=True)
+                    continue
+
+                with archive.open(n) as srcfile:
+                    with open(dstpath, "wb") as dstfile:
+                        shutil.copyfileobj(srcfile, dstfile)
+
+        with open(os.path.join(self.repo.location, ".info"), "w") as infofile:
+            if etag:
+                infofile.write(f"etag {etag}\n")
```
zmedico, line 100: We can implement the retrieve_head method to retrieve the etag like this:
```python
    def retrieve_head(self, **kwargs):
        """Get information about the head commit"""
        if kwargs:
            self._kwargs(kwargs)
        info = portage.grabdict(os.path.join(self.repo.location, ".info"))
        if "etag" in info:
            return (os.EX_OK, info["etag"][0])
        return (1, False)
```

legionus: @zmedico Good catch! I'll add this as a separate commit.

lib/portage/sync/modules/zipfile/zipfile.py
legionus: @zmedico The idea of not deleting the location is good, but the rest will not work.

```python
shutil.copyfileobj(srcfile, dstfile)
```
after this line mtime of `dstfile` will be different and `filecmp.cmp()` will always return `False`.  If we really want to do such a check, then we need to get the checksum of the file.

legionus: Speaking of `self.repo.location`, you can make the following change:

```diff
diff --git a/lib/portage/sync/modules/zipfile/zipfile.py b/lib/portage/sync/modules/zipfile/zipfile.py
index 3cd210a64..24de2b0c7 100644
--- a/lib/portage/sync/modules/zipfile/zipfile.py
+++ b/lib/portage/sync/modules/zipfile/zipfile.py
@@ -76,7 +76,14 @@ class ZipFile(SyncBase):
             return (1, False)
 
         # Drop previous tree
-        shutil.rmtree(self.repo.location)
+        for name in os.listdir(self.repo.location):
+            if name == ".info":
+                continue
+            f = os.path.join(self.repo.location, name)
+            if os.path.isdir(f):
+                shutil.rmtree(f)
+            else:
+                os.unlink(f)
 
         with zipfile.ZipFile(zip_file) as archive:
             strip_comp = 0
```

legionus: > +                            files_equal = filecmp.cmp(dstfile, oldfile, shallow=False)

Ops. I missed that `shallow=False` is actually a content comparison.

zmedico: It would be better to do a byte comparison in memory to avoid writing dstfile to disk when possible, but it will get replaced by oldfile so quickly in this case that you would hope the blocks never go flushed to disk.

legionus: @zmedico I've added an implementation of this, but I implemented it in my own way. The `filecmp.cmp()` always wants to do stat on arguments, which makes the function inapplicable for ZipExtFile. We may independently verify the content is identical.

legionus: > It would be better to do a byte comparison in memory to avoid writing dstfile to disk when possible, but it will get replaced by oldfile so quickly in this case that you would hope the blocks never get flushed to disk.

@zmedico That's what I did. My current implementation will recreate directories. We don’t have to do this, but then we will need to save the list of files from the zip archive.

zmedico: Thanks this looks great! Let's not bother with directories unless you can come up with a really simple implementation.

legionus: The implementation will not be difficult, but it will require saving all the paths from the archive in memory. In my case (in the case of local overlay) this is not much, but in the case of the entire portage it will be quite a lot.

I would prefer to leave such changes to follow up PRs and merge it as it is.

lib/portage/sync/modules/zipfile/zipfile.py
zmedico: I went ahead and wrote a little test case that calls `emerge --sync` and checks the etag afterwards. It would be nice to include in the PR so we have some test coverage from the start.

```python
# Copyright 2024 Gentoo Authors

import http.server
import os
import shutil
import socketserver
import subprocess
import tempfile
import textwrap
import threading
from functools import partial

import portage
from portage.tests.resolver.ResolverPlayground import ResolverPlayground


class test_sync_zipfile_case(TestCase):
    def test_sync_zipfile(self):
        cpv = "dev-libs/A-0"
        ebuilds = {
            cpv: {"EAPI": "8"},
        }
        etag = "foo"

        server = None
        playground = None
        tmpdir = tempfile.mkdtemp()
        try:

            class Handler(http.server.SimpleHTTPRequestHandler):
                def end_headers(self):
                    self.send_header("etag", etag)
                    super().end_headers()

            server = socketserver.TCPServer(
                ("127.0.0.1", 0),
                partial(Handler, directory=tmpdir),
            )
            threading.Thread(target=server.serve_forever, daemon=True).start()

            playground = ResolverPlayground(
                ebuilds=ebuilds,
            )
            settings = playground.settings

            env = settings.environ()

            repos_conf = textwrap.dedent(
                """
                [test_repo]
                location = %(location)s
                sync-type = zipfile
                sync-uri = %(sync-uri)s
                auto-sync = true
            """
            )

            repo_location = f"{playground.eprefix}/var/repositories/test_repo"

            env["PORTAGE_REPOSITORIES"] = repos_conf % {
                "location": repo_location,
                "sync-uri": "http://{}:{}/test_repo.zip".format(*server.server_address),
            }

            shutil.make_archive(os.path.join(tmpdir, "test_repo"), "zip", repo_location)

            ebuild = playground.trees[playground.eroot]["porttree"].dbapi.findname(cpv)
            self.assertTrue(os.path.exists(ebuild))
            shutil.rmtree(repo_location)
            self.assertFalse(os.path.exists(ebuild))

            result = subprocess.run(
                [
                    "emerge",
                    "--sync",
                ],
                env=env,
                stdout=subprocess.PIPE,
                stderr=subprocess.STDOUT,
            )
            output = result.stdout.decode(errors="replace")
            try:
                self.assertEqual(result.returncode, os.EX_OK)
            except Exception:
                print(output)
                raise

            repo = settings.repositories["test_repo"]
            sync_mod = portage.sync.module_controller.get_class("zipfile")
            status, repo_revision = sync_mod().retrieve_head(options={"repo": repo})
            self.assertEqual(status, os.EX_OK)
            self.assertEqual(repo_revision, etag)
        finally:
            if server is not None:
                server.shutdown()
            shutil.rmtree(tmpdir)
            playground.cleanup()
```

legionus: @zmedico I'm not very familiar with the portage codebase, but I added your test as a separate commit.

zmedico: Great, thanks!

zmedico (approved):

thesamesam (approved):
I've only given it a cursory look since an earlier revision but I'm happy for it to go in once docs are updated.

Could you update the man pages, and also possibly add something to the wiki as well once this is merged?

legionus:
> I've only given it a cursory look since an earlier revision but I'm happy for it to go in once docs are updated.
> 
> Could you update the man pages, and also possibly add something to the wiki as well once this is merged?

I will try to add something to wiki when this PR will be merged.

Can someone merge it ?

legionus:
@floppym It looks like this PR is just waiting for your review.
Could you make it and put it apporve?

floppym (approved):

thesamesam:
Wiki can be done later, but what about man pages? Just a small example in `portage(5)` should suffice (there's already others there).

legionus:
> Wiki can be done later, but what about man pages? Just a small example in `portage(5)` should suffice (there's already others there).

@thesamesam I added the example to `portge(5)` as a separate commit.

thesamesam (approved):
Thanks! I'll merge when at computer (or someone else is free to ofc)

thesamesam:
Thank you both for the contribution and all for the reviews!

legionus:
Sorry for the off-topic. I don't understand how I can change the page https://wiki.gentoo.org/wiki/Project:Portage/Sync . I have an account with a verified email, but I can only open a discussion.

dol-sen:
Add the diff to discussion, a moderator or project dev will need to apply it.

PR #1267, opened 2024-02-13
EbuildPhase/EbuildMetadataPhase: async check_locale

zmedico:
Change config.environ() check_locale calls to async_check_locale calls in the EbuildPhase/EbuildMetadataPhase _async_start method in order to eliminate synchronous waiting for child processes in the main event loop thread.

Note that this series of changes causes access to the portdbapi doebuild_settings attribute to no longer be serialized via the EbuildMetadataPhase _start method. As a result, exclusive access to config instances needs to be guaranteed in some other way to avoid triggering problems (see bug 924319), such as by maintaining a config pool or by serializing access via an asyncio.Lock instance.

Bug: https://bugs.gentoo.org/923841
Bug: https://bugs.gentoo.org/924319

### DIFF
--- a/lib/_emerge/EbuildMetadataPhase.py
+++ b/lib/_emerge/EbuildMetadataPhase.py
@@ -46,6 +46,12 @@ class EbuildMetadataPhase(SubProcess):
     _files_dict = slot_dict_class(_file_names, prefix="")
 
     def _start(self):
+        asyncio.ensure_future(
+            self._async_start(), loop=self.scheduler
+        ).add_done_callback(self._async_start_done)
+        self._registered = True
+
+    async def _async_start(self):
         ebuild_path = self.ebuild_hash.location
 
         with open(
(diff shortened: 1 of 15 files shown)


lib/_emerge/EbuildMetadataPhase.py
```diff
@@ -139,6 +159,21 @@ def _start(self):
 
         self._proc = retval
 
+    def _async_start_done(self, future):
+        future.cancelled() or future.result()
+        if self._was_cancelled():
+            pass
```
thesamesam, line 165: Am I missing why we write it like this?

Why not just:
`if not self._was_cancelled() and future.cancelled():`

thesamesam (approved):

lib/_emerge/EbuildMetadataPhase.py
zmedico: Your version is equivalent so I updated it to use that.

PR #1271, opened 2024-02-14
BuildLogger: Fix portage.locks._open_fds memory leak

zmedico:
The _file_close_wrapper __getattribute__ method needs to be overridden to expose its close method, otherwise the underlying file's close method is called and the closed file object remains as a memory leak in the global portage.locks._open_fds dict. For reference, see similar classes like portage.util.atomic_ofstream which overrides methods in the same way.

Bug: https://bugs.gentoo.org/919072
Fixes:  https://github.com/gentoo/portage/pull/1199  df212738bbb2 ("BuildLogger: Close self._stdin after fork")

NOTE: This leak seems to be not much of a leak because the file descriptor numbers tend to recycle and kick the previous instances out of _open_fds.

### DIFF
--- a/lib/portage/util/_async/BuildLogger.py
+++ b/lib/portage/util/_async/BuildLogger.py
@@ -1,4 +1,4 @@
-# Copyright 2020-2023 Gentoo Authors
+# Copyright 2020-2024 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 import functools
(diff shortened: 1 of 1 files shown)


arkamar (approved):
Uf, took me some time to experiment and understand how exactly this works :) Now it works as expected. I also tested it according to this scenario https://bugs.gentoo.org/919072#c5 -> no deadlock.

PR #1272, opened 2024-02-14
emerge: Skip installed packages with emptytree in depgraph selection

gaboroszkar:
Running emerge with emptytree tries to find the best match for every atom it needs to install. Sometimes the best matches would be already installed packages (with `operation=nomerge`), but these packages would be silently skipped with full emptytree installation. This change makes sure that emerge attempts to install every package. If the package has unmet requirements, emerge will complain.

Bug: https://bugs.gentoo.org/651018
Signed-off-by: Gábor Oszkár Dénes <gaboroszkar@protonmail.com>

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -7629,6 +7629,19 @@ class depgraph:
                         continue
 
                     if (
+                        empty
+                        and pkg.installed
+                        and not self._frozen_config.excluded_pkgs.findAtomForPackage(
+                            pkg, modified_use=self._pkg_use_enabled(pkg)
+                        )
+                    ):
+                        # With --emptytree option we assume no packages
+                        # are installed, so we do not select them.
+                        # But we allow installed packages to satisfy dependency requirements
+                        # if they're explicitly excluded, so we allow them to be selected.
+                        continue
+
+                    if (
                         not pkg.installed
                         and self._frozen_config.excluded_pkgs.findAtomForPackage(
                             pkg, modified_use=self._pkg_use_enabled(pkg)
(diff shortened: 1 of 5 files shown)


lib/_emerge/depgraph.py
```diff
@@ -7622,6 +7622,10 @@ def _wrapped_select_pkg_highest_available_imp(
                     if pkg in self._dynamic_config._runtime_pkg_mask:
                         # The package has been masked by the backtracking logic
                         continue
+                    if empty and pkg.installed:
```
thesamesam, line 7625: Could you see if you could construct a testcase to add?

zmedico: Should test to verify that --exclude allows an installed package.

gaboroszkar: Thanks @zmedico , I made a small adjustment. I tested manually whenever `--exclude` is there, the package is not reinstalled. With my original modification emerge just fails, because the excluded package is masked.

gaboroszkar: I will add tests soon.

gaboroszkar: Tests are done. Could you rereview?

zmedico (approved):
We could add an `--exclude-unsatisfied` option to trigger the old  behavior in case somebody wanted that, but I suppose we could just wait and see if anyone complains about the new behavior.

gaboroszkar:
> We could add an --exclude-unsatisfied
Maybe, but in my opinion this would be an overly complicated user interface.

In any case, the user can add `--exclude` for the unsatisfied dependencies, and the emerge should be successful.
I've just added a small testcase for this in `test_emptytree_reinstall_unsatisfiability.py`: `dev-libs/C` is an unsatisfied dependency, but with `--exclude dev-libs/C`, emerge is successful.

thesamesam (approved):
Thank you! I'll merge later if zac hasn't by then

PR #1270, opened 2024-02-14
multiprocessing: Use a private context for set_start_method

zmedico:
Wrap the multiprocessing module to return a private multiprocessing context instead. It's safe to call set_start_method on this and it will only affect portage internals, which makes it compatible with the pytest-xdist plugin.

Bug: https://bugs.gentoo.org/924416

Seeing pytest-xdist worker crashes less frequently now:
https://github.com/gentoo/portage/actions/runs/7897425289/job/21552979356
```
2024-02-14T06:42:44.7157130Z =================================== FAILURES ===================================
2024-02-14T06:42:44.7157559Z ______________ lib/portage/tests/process/test_spawn_returnproc.py ______________
2024-02-14T06:42:44.7157935Z [gw1] linux -- Python 3.12.2 /opt/hostedtoolcache/Python/3.12.2/x64/bin/python
2024-02-14T06:42:44.7158702Z worker 'gw1' crashed while running 'lib/portage/tests/process/test_spawn_returnproc.py::SpawnReturnProcTestCase::testSpawnReturnProcWait'
```
https://github.com/gentoo/portage/actions/runs/7897425289/job/21553620257?pr=1270
```
2024-02-14T07:14:15.4605629Z =================================== FAILURES ===================================
2024-02-14T07:14:15.4605883Z ______________ lib/portage/tests/locks/test_asynchronous_lock.py _______________
2024-02-14T07:14:15.4606199Z [gw1] linux -- Python 3.12.2 /opt/hostedtoolcache/Python/3.12.2/x64/bin/python
2024-02-14T07:14:15.4606935Z worker 'gw1' crashed while running 'lib/portage/tests/locks/test_asynchronous_lock.py::AsynchronousLockTestCase::testAsynchronousLockWaitKill'
```
Maybe https://github.com/pytest-dev/pytest-rerunfailures/pull/158 could be a solution at this point, since it handles `pytest-xdist -n` worker crashes.

NOTE: Temporarily added the commits from https://github.com/gentoo/portage/pull/1267 to see if they trigger more pytest-xdist runner crashes.

PR #1273, opened 2024-02-14
actions: Use pytest-rerunfailures for pytest-xdist worker crash

zmedico:
Since pytest-xdist workers crash intermittently for the multiprocessing spawn start method, use pytest-rerunfailures to detect and handle this case. Only use pytest-rerunfailures for the spawn start-method since that is the only case where we've observed intermittent pytest-xdist worker crashes, and use `--only-rerun 'worker .* crashed while running'` to ensure that rerun only triggers for worker crashes.

Bug: https://bugs.gentoo.org/924416

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
           python -m site
           python -m pip install --upgrade pip
           # setuptools needed for 3.12+ because of https://github.com/mesonbuild/meson/issues/7702.
-          python -m pip install pytest pytest-xdist setuptools
+          python -m pip install pytest pytest-rerunfailures pytest-xdist setuptools
 
           # symlink /bin/true to /usr/bin/getuto (or do we want to grab the script from github?)
           sudo ln -s /bin/true /usr/bin/getuto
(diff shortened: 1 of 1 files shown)


thesamesam:
bleh. I'm not hard-against it, just worried we're papering over the problem with this. But I suppose it's better than flaky CI which may-or-may not fail for unrelated changes...

I haven't looked at what the actual crash is - do you have any ideas for it yet?

zmedico:
> bleh. I'm not hard-against it, just worried we're papering over the problem with this. But I suppose it's better than flaky CI which may-or-may not fail for unrelated changes...

It's a pytest-xdist problem though, since it doesn't trigger without it, and pytest-rerunfailures is designed to handle it since https://github.com/pytest-dev/pytest-rerunfailures/pull/158.

> I haven't looked at what the actual crash is - do you have any ideas for it yet? (Are we to blame?)

I haven't been able to trigger it locally, so that makes it extra mysterious. Since using a private multiprocessing context for the spawn start method didn't help in https://github.com/gentoo/portage/pull/1270, I think we can safely say it's not our fault. It might be https://github.com/pytest-dev/execnet/issues/96 since that's an "intermittent" issue.

The `--only-rerun 'worker .* crashed while running'` bit makes the workaround not so bad, since rerun is only triggered when absolutely necessary.

thesamesam:
AH, thank you, sorry, I thought it was an interaction between one of our setup fixtures and multiprocessing, just exposed by pytest-xdist.

thesamesam (approved):

PR #1274, opened 2024-02-15
Support PROPERTIES=test_userpriv not to drop perms for tests

mgorny:
Support PROPERTIES=test_userpriv and a corresponding ALLOW_TEST=userpriv to disable FEATURES=userpriv when running the test phase.  This can be used e.g. in dev-python/reflink that needs to be able to mount filesystem on a loopback device for testing.

Bug: https://bugs.gentoo.org/924585

### DIFF
--- a/man/ebuild.5
+++ b/man/ebuild.5
@@ -808,6 +808,11 @@ is installed.
 The package manager may run tests that require an internet connection, even if
 the ebuild has
 .IR RESTRICT=test .
+.TP
+.I test_privileged
+The package manager may run tests that require superuser permissions, even if
+the ebuild has
+.IR RESTRICT=test .
 .RE
 .PD 1
 .TP
--- a/bin/phase-functions.sh
+++ b/bin/phase-functions.sh
@@ -503,7 +503,8 @@ __dyn_test() {
 	fi
 
 	if has test ${PORTAGE_RESTRICT} && ! has all ${ALLOW_TEST} &&
-			! { has test_network ${PORTAGE_PROPERTIES} && has network ${ALLOW_TEST}; }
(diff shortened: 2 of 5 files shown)


mgorny:
That said, I'm not sure if `test_userpriv` is the best name. On one hand, it's consistent with `RESTRICT=userpriv` — but it does the exact opposite of `FEATURES=userpriv`.

zmedico:
> That said, I'm not sure if `test_userpriv` is the best name. On one hand, it's consistent with `RESTRICT=userpriv` — but it does the exact opposite of `FEATURES=userpriv`.

To be consistent with `test_network`, we could make it `test_privileged` or `test_as-root`. As it is, we have `test_network` rather than `test_network-sandbox`.

floppym:
Somewhat late question: why did we decide to use underscores for word separation here?

zmedico:
> Somewhat late question: why did we decide to use underscores for word separation here?

I suggested it for consistency with USE_EXPAND back in https://bugs.gentoo.org/553696#c12.

man/ebuild.5
```diff
@@ -807,6 +807,10 @@ is installed.
 .I test_network
 The package manager may run tests that require an internet connection, even if
 the ebuild has
```
zmedico, line 809: The end of this sentence was stolen by test_privileged below.

thesamesam (commented):
I'm OK with the idea but I haven't reviewed the implementation at all. It might be nice to include a warning in the man page that the tests should be carefully assessed first.

man/ebuild.5
mgorny: Fixed, sorry.

mgorny:
Ping.

zmedico (approved):

PR #1276, opened 2024-02-23
AsyncioEventLoop: Call process.run_exitfuncs() before close

zmedico:
For the event loop running in the main thread, call process.run_exitfuncs() before close with the event loop running so that anything attached can clean itself up (like the socks5 ProxyManager for bug 925240). This is necessary because process.spawn uses the event loop to implement the new returnproc parameter related to bug 916566.

Bug: https://bugs.gentoo.org/925240

### DIFF
--- a/lib/portage/util/socks5.py
+++ b/lib/portage/util/socks5.py
@@ -8,6 +8,13 @@ import os
 import socket
 from typing import Union
 
+import portage
+
+portage.proxy.lazyimport.lazyimport(
+    globals(),
+    "portage.util._eventloop.global_event_loop:global_event_loop",
+)
+
 import portage.data
 from portage import _python_interpreter
 from portage.data import portage_gid, portage_uid, userpriv_groups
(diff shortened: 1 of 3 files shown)


thesamesam (approved):

lib/portage/util/socks5.py
```diff
@@ -74,10 +81,15 @@ def stop(self) -> Union[None, asyncio.Future]:
         if self._proc is not None:
             self._proc.terminate()
             if loop is None:
-                asyncio.run(self._proc.wait())
+                # In this case spawn internals would have used
+                # portage's global loop when attaching a waiter to
+                # self._proc, so we are obligated to use that.
+                global_event_loop().run_until_complete(self._proc.wait())
```
zmedico, line 87: We can actually use the real `asyncio.get_event_loop()` in cases like this without triggering a DeprecationWarning, since portage's _safe_loop will have set the real asyncio global event loop like this:
```python
            except RuntimeError:
                _loop = _real_asyncio.new_event_loop()
                _real_asyncio.set_event_loop(_loop)
```

PR #1277, opened 2024-02-23
Poking at py3.9 CI failure

thesamesam:


zmedico:
Seems flawed in lib/portage/tests/emerge/conftest.py because it doesn't create the `dev-libs/B` package required for the `emerge -e` command to succeed with the new behavior:
```python
        # The next emerge has been added to split this test from the rest:
        make_package = Emerge("--buildpkg", "dev-libs/A")
        getbinpkgonly = Emerge(
            "-e",
            "--getbinpkgonly",
            "dev-libs/A",
            preparation=lambda: os.rename(pkgdir, binhost_dir),
        )
```

thesamesam:
Thanks!

PR #1279, opened 2024-02-23
process.spawn: Fix logic for missing libc.unshare on musl

zmedico:
Fix `unshare_*` variables to be False when the libc is missing, libc.unshare is missing, or libc.unshare fails.

Also, if socket.sethostname is missing then _exec2 needs libc for the network-sandbox sethostname call which is wrapped by a blanket Exception handler.

Fixes: 419cce79f908 ("process._exec: Use _start_fork for os.fork() error handling")
Bug: https://bugs.gentoo.org/925311

### DIFF
--- a/lib/portage/process.py
+++ b/lib/portage/process.py
@@ -956,114 +956,119 @@ def _exec(
     signal.signal(signal.SIGQUIT, signal.SIG_DFL)
 
     # Unshare (while still uid==0)
+    have_unshare = False
+    libc = None
     if unshare_net or unshare_ipc or unshare_mount or unshare_pid:
         filename = find_library("c")
         if filename is not None:
             libc = LoadLibrary(filename)
             if libc is not None:
-                # unshare() may not be supported by libc
-                if not hasattr(libc, "unshare"):
-                    unshare_net = False
-                    unshare_ipc = False
-                    unshare_mount = False
-                    unshare_pid = False
-                else:
-                    # Since a failed unshare call could corrupt process
-                    # state, first validate that the call can succeed.
-                    # The parent process should call _unshare_validate
-                    # before it forks, so that all child processes can
-                    # reuse _unshare_validate results that have been
-                    # cached by the parent process.
-                    errno_value = _unshare_validate(unshare_flags)
-                    if errno_value == 0 and libc.unshare(unshare_flags) != 0:
-                        errno_value = ctypes.get_errno()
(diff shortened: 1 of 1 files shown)


lib/portage/process.py
```diff
@@ -956,114 +956,112 @@ def _exec(
...
-                                portage._python_interpreter,
-                                os.path.join(portage._bin_path, "pid-ns-init"),
-                                str(main_child_pid),
-                            ]
-
-                            os.execve(binary, myargs, env)
+                unshare_net = False
+                unshare_ipc = False
+                unshare_mount = False
+                unshare_pid = False
```
floppym, line 969: This is still missing some cases.
```suggestion
    # Unshare (while still uid==0)
    have_unshare = False
    libc = None
    if unshare_net or unshare_ipc or unshare_mount or unshare_pid:
        filename = find_library("c")
        if filename is not None:
            libc = LoadLibrary(filename)
            if libc:
               have_unshare = hasattr(libc, "unshare")

    if not have_unshare:
        # unshare() may not be supported by libc
        unshare_net = False
        unshare_ipc = False
        unshare_mount = False
        unshare_pid = False
```

floppym:
I think this needs some more work to ensure we don't pass unshare_xxx=True to _exec2 in the fall-through case from the giant if block.

floppym:
In other words, if the unshare stuff fails in _exec(), we should not be calling mount, sethostname, or _configure_loopback_interface() in _exec2().

lib/portage/process.py
zmedico: Updated to use your have_unshare variable. Thanks!

lib/portage/process.py
```diff
@@ -956,114 +956,119 @@ def _exec(
...
+                    " ".join(involved_features),
+                ),
+                noiselevel=-1,
+            )
+
+            unshare_net = False
+            unshare_ipc = False
+            unshare_mount = False
+            unshare_pid = False
+
```
zmedico, line 1009: > In other words, if the unshare stuff fails in _exec(), we should not be calling mount, sethostname, or _configure_loopback_interface() in _exec2().

Updated to set `unshare_*` False if unshare fails here.

floppym (approved):
That looks better.

thesamesam (approved):
lgtm but let's separately do that find_library fallback for musl

PR #1280, opened 2024-02-24
Fix python 3.9 CI jobs since 92ff02b9189f

zmedico:
Use emerge -e to fix this error we've seen since
https://github.com/gentoo/portage/pull/1272 92ff02b9189f for python 3.9 CI jobs:

https://github.com/gentoo/portage/actions/runs/8014796128/job/21893963019
```
test_portage_baseline[binhost emerge-gpkg] - AssertionError: 'emerge' failed with args '('-e', '--getbinpkgonly', 'dev-libs/A')'
```
```
emerge: there are no binary packages to satisfy "dev-libs/B[flag]". (dependency required by "dev-libs/A-1::test_repo" [binary]) (dependency required by "dev-libs/A" [argument])
```
Fixes: 92ff02b9189f ("emerge: Skip installed packages with emptytree in depgraph selection")
Bug: https://bugs.gentoo.org/651018

Also see https://github.com/gentoo/portage/pull/1277.

### DIFF
--- a/lib/portage/tests/emerge/conftest.py
+++ b/lib/portage/tests/emerge/conftest.py
@@ -805,7 +805,7 @@ def _generate_all_baseline_commands(playground, binhost):
         test_commands["binhost emerge"] = Noop()
     else:
         # The next emerge has been added to split this test from the rest:
-        make_package = Emerge("--buildpkg", "dev-libs/A")
+        make_package = Emerge("-e", "--buildpkg", "dev-libs/A")
         getbinpkgonly = Emerge(
             "-e",
             "--getbinpkgonly",


thesamesam:
Thanks Zac. I haven't looked at it yet but do we get why this differs only for py3.9? cc @gaboroszkar.

gaboroszkar:
Still not sure what is happening exactly. As far as I saw, the problem is that some modules are not loaded, but not sure whether this is the only issue.

I believe this would be a good fix too: https://github.com/gentoo/portage/compare/master...gaboroszkar:portage:test-baseline-python39

I was able to replicate the issue by running `git clean -dfx`, and then running `pytest "lib/portage/tests/emerge/test_baseline.py`. After that I run `pytest lib/portage/tests/bin/test_doins.py::DoIns::testDoIns`, the baseline test seem to be fixed.

gaboroszkar:
I'm not sure still what's exactly the problem. But I was only able to reproduce one kind of error. Not sure how much this is related.

When I do `git clean -dfx`, and after that `pytest "lib/portage/tests/emerge/test_baseline.py::test_portage_baseline[xpak-emerge -1 dev-libs/A -v dev-libs/B]"`, the test fails. When I run `pytest lib/portage/tests/bin/test_doins.py::DoIns::testDoIns`, the baseline test seem to work. It complains about things not in the python cache. I added a small fix, which is still not the proper fix, but the tests run: https://github.com/gentoo/portage/compare/master...gaboroszkar:portage:test-baseline-python39

Not sure if this is a problem only on my environment, and how much this is related to this issue.

PR #1281, opened 2024-02-24
test_baseline: Improve robustness with cleanup

gaboroszkar:
The baseline tests need to cleanup the ResolverPlayground after each testcase, with each different parametrization. This is ensured by making the scope of the playground fixture the function instead of the module. With module the cleanup only happens before/after the switch from/to xpak and gpkg.

Signed-off-by: Gábor Oszkár Dénes <gaboroszkar@protonmail.com>

### DIFF
--- a/lib/portage/tests/emerge/conftest.py
+++ b/lib/portage/tests/emerge/conftest.py
@@ -406,7 +406,7 @@ def async_loop():
     yield asyncio._wrap_loop()
 
 
-@pytest.fixture(params=SUPPORTED_GENTOO_BINPKG_FORMATS, scope="module")
+@pytest.fixture(params=SUPPORTED_GENTOO_BINPKG_FORMATS, scope="function")
 def playground(request, tmp_path_factory):
     """Fixture that provides instances of ``ResolverPlayground``
     each one with one supported value for ``BINPKG_FORMAT``."""


gaboroszkar:
See also https://github.com/gentoo/portage/pull/1272.

gaboroszkar:
@zmedico @thesamesam 

As far as I understand the followings happened. The `PlaygroundResolver` cleanup did not run after previous gpkg runs, and so the last parametrization of this test `[gpkg-binhost emerge]` (or `[binhost emerge-gpkg]` as far as I understand they're the same) could run succesfully.

You can see the logs, that (as far as I saw, didn't look at all the logs) when the tests succeed, they do run some other gpkg tests before `[gpkg-binhost emerge]` on the same worker. When this test failed, it was the first run on that worker.

With this small fix, it will be ensured, that whenever someone (this time it was me) makes a change which would invalidate these tests, they will be caught early.

Another small issue I had: whenever I ran the test on a clean slate (`git clean -dfx`), the tests fail, because they need the `_xattr` and `movefile` portage utilities to be initialized before.

lib/portage/tests/emerge/test_baseline.py
```diff
@@ -38,6 +38,7 @@
 from portage.tests import cnf_etc_path
 from portage.util import ensure_dirs
 from portage.util.futures import asyncio
+from portage.util import movefile
```
zmedico, line 41: Our pylint config won't allow an unused import like this. How did the error that you're working around manifest itself?

gaboroszkar: Still not sure what would be the correct solution, unfortunately don't know enough about Portage. I understand this can't be merged in this form.

The errors I got is the following by running `git clean -dfx; pytest "lib/portage/tests/emerge/test_baseline.py::test_portage_baseline[gpkg-binhost emerge]"`.
```
>>> Install dev-libs/B-1 into /tmp/pytest-of-gabor/pytest-199/eprefix0/var/tmp/portage/portage/dev-libs/B-1/image/                                                                    
 * installing something...                                                                                                                                                            
 * ACCESS DENIED:  open_wr:            <...>/portage/lib/portage/util/__pycache__/movefile.cpython-311.pyc.140285464041136                                             
 * ACCESS DENIED:  open_wr:            <...>/portage/lib/portage/util/__pycache__/_xattr.cpython-311.pyc.140285462691888                                               
 * ACCESS DENIED:  open_wr:            <...>/portage/lib/portage/util/__pycache__/movefile.cpython-311.pyc.140557727815200                                             
 * ACCESS DENIED:  open_wr:            <...>/portage/lib/portage/util/__pycache__/_xattr.cpython-311.pyc.140557727825712                                               
 * ACCESS DENIED:  open_wr:            <...>/portage/lib/portage/util/__pycache__/movefile.cpython-311.pyc.140324497282448                                             
 * ACCESS DENIED:  open_wr:            <...>/portage/lib/portage/util/__pycache__/_xattr.cpython-311.pyc.140324497292960                                               
 * called src_install for dev-libs/B-1                                                                                      
```

(`<...>` is edited by me there, it's the absolute path to the directory `portage` is in.)

gaboroszkar: Note: this is an unrelated error you fixed.

zmedico: I can trigger a sandbox violation locally if I run `find lib -name '*.py[co]' -delete` before the tests, but I don't see anything related to `movefile`.

gaboroszkar: I don't understand unfortunately. Maybe it's not that important if it's only on my machine, but I can't verify that, don't understand the cause.

What do you suggest? Remove this `import`, and just merge this without this? I'm fine with that.

zmedico: The sandbox violation can trigger for more than just movefile, so we need to either allow write in sandbox or else pre-compile the pyc/pyo files.

gaboroszkar: This seems like a complicated problem, out of scope. I'll just remove the `import`.

lib/portage/tests/emerge/conftest.py
```diff
@@ -406,7 +406,7 @@ def async_loop():
     yield asyncio._wrap_loop()
 
 
-@pytest.fixture(params=SUPPORTED_GENTOO_BINPKG_FORMATS, scope="module")
+@pytest.fixture(params=SUPPORTED_GENTOO_BINPKG_FORMATS, scope="function")
 def playground(request, tmp_path_factory):
```
zmedico, line 410: When I ran the lib/portage/tests/emerge tests on my laptop with this change, the time only went up from 3:59 to 4:08, and the number of instantiated playground fixtures went up from 6 to 24. Seems worthwhile to improve cleanup.

zmedico (approved):

lib/portage/tests/emerge/conftest.py
gaboroszkar: Thanks, I didn't even check the run time.

thesamesam (approved):
Thanks!

lib/portage/tests/emerge/test_baseline.py
zmedico: This is a script I have in my portage checkout  to compile pyc/pyo files (probably inspired by [python-utils-r1.eclass](https://github.com/gentoo/gentoo/blob/master/eclass/python-utils-r1.eclass) usage of compileall in its python_optimize function):
```bash
#!/usr/bin/env bash

[[ $UID != 0 ]] && \
	find lib -uid 0 -name "*.py[co]" -print0 | xargs -r -0 rm -f
python -c 'import compileall ; compileall.compile_dir("./lib/", quiet=True)'
python -O -c 'import compileall ; compileall.compile_dir("./lib/", quiet=True)'
```

PR #1287, opened 2024-02-25
Scheduler: Support parallel-install with merge-wait

zmedico:
For system packages, always serialize install regardless parallel-install, in order to mitigate failures triggered by fragile states as in bug 256616. For other packages, continue to populate self._task_queues.merge, which will serialize install unless parallel-install is enabled.

Fixes: 825db01b91a3 ("Add merge-wait FEATURES setting enabled by default")
Bug: https://bugs.gentoo.org/256616
Bug: https://bugs.gentoo.org/925213

### DIFF
--- a/lib/_emerge/Scheduler.py
+++ b/lib/_emerge/Scheduler.py
@@ -1519,17 +1519,20 @@ class Scheduler(PollScheduler):
             self._deallocate_config(build.settings)
         elif build.returncode == os.EX_OK:
             self.curval += 1
-            merge = PackageMerge(merge=build, scheduler=self._sched_iface)
+            merge = PackageMerge(
+                is_system_pkg=(build.pkg in self._deep_system_deps),
+                merge=build,
+                scheduler=self._sched_iface,
+            )
             self._running_tasks[id(merge)] = merge
             # By default, merge-wait only allows merge when no builds are executing.
             # As a special exception, dependencies on system packages are frequently
             # unspecified and will therefore force merge-wait.
-            is_system_pkg = build.pkg in self._deep_system_deps
             if not build.build_opts.buildpkgonly and (
-                "merge-wait" in build.settings.features or is_system_pkg
+                "merge-wait" in build.settings.features or merge.is_system_pkg
             ):
                 self._merge_wait_queue.append(merge)
-                if is_system_pkg:
+                if merge.is_system_pkg:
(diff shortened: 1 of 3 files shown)


lib/_emerge/Scheduler.py
```diff
@@ -1804,13 +1807,29 @@ def _schedule_tasks(self):
...
-                state_change += 1
+                while self._merge_wait_queue:
+                    # If we added non-system packages to the merge queue in a
+                    # previous iteration of this loop, then for system packages
+                    # we need to come back later when the merge queue is empty.
+                    if (
+                        self._task_queues.merge
+                        and self._merge_wait_queue[0].is_system_pkg
+                    ):
+                        break
```
zmedico, line 1818: We could possibly search through the queue for non-system packages to merge here, but maybe this is good as-is for our initial implementation since this way it keeps the merge order in the same order that the builds finished.

thesamesam (approved):

man/make.conf.5
```diff
@@ -652,9 +652,10 @@ terminal to view parallel-fetch progress.
 .TP
 .B parallel\-install
 Use finer\-grained locks when installing packages, allowing for greater
-parallelism. Note that \fIparallel\-install\fR currently has no effect
-unless \fImerge\-wait\fR is disabled. For additional parallelism,
-disable \fIebuild\-locks\fR.
+parallelism. For additional parallelism disable \fIebuild\-locks\fR.
+Also disable \fImerge\-wait\fR for additional parallelism if desired,
+but this is not necessary because \fIparallel\-install\fR cooperates
```
thesamesam, line 657: Maybe tweak the phrasing here? I think users get confused by the different options and how to achieve maximum parallelism (and the risks of each).

lib/_emerge/Scheduler.py
zmedico: Yeah, added this:
```python
# TODO: Maybe promote non-system packages to the front of the
# queue and process them within the current loop, though that
# causes merge order to differ from the order builds finish.
```

man/make.conf.5
zmedico: Yeah, changed it to:
> Also disable merge-wait for additional parallelism if desired, but that increases the possibility of random build failures. When parallel-install is used together with merge-wait, parallel installation occurs in batches when there are no builds running.

PR #1294, opened 2024-03-02
Replace linux_reflink extension module

floppym:
Python 3.8 added support for copy_file_range in the os module, so we can just call that directly.

Also, we can use the FICLONE ioctl for fast file clones on supported filesystems (btrfs).

If we think this needs some performance testing, I would appreciate any help with that.


lib/portage/util/file_copy.py
```diff
@@ -0,0 +1,115 @@
...
+                return True
+            except OSError:
+                pass
+
+        if hasattr(os, "copy_file_range"):
+            try:
+                _do_copy_file_range(srcfd, dstfd)
+                return True
+            except OSError:
+                pass
```
zmedico, line 88: With this ignorance of OSError we won't catch the zfs bugs anymore.  :smirk:

floppym:
Closing in favor of #1296.

floppym:
Re-opening this since it has some commentary.

lib/portage/util/file_copy.py
zmedico: If we do a something like this then we can still catch zfs bugs:
```python
logging.warning("_do_copy_file_range failed", exc_info=sys.exc_info())
```

floppym: If there are specific errno values you think we should handle differently, happy to make changes here.

zmedico: If we do a something like this then we can still catch zfs bugs:
```python
logging.warning("_do_copy_file_range failed", exc_info=sys.exc_info())
```

zmedico: It will also give us some visibility to better understand _do_copy_file_range failures.

floppym: I added warnings for copy_file_range and sendfile. Let me know what you think.

zmedico: That looks great, thanks!

lib/portage/util/file_copy.py
```diff
@@ -0,0 +1,124 @@
...
+
+        try_cfr = hasattr(os, "copy_file_range")
+
+        for offset, count in _get_chunks(srcfd):
+            if count == 0:
+                os.ftruncate(dstfd, offset)
+            else:
+                if try_cfr:
+                    try:
+                        _do_copy_file_range(srcfd, dstfd, offset, count)
```
thesamesam, line 91: I was wondering why we're doing this / if c_f_r handles the holes part for us, but indeed, the man page says:
> If fd_in is a sparse file, then copy_file_range() may expand any
>       holes existing in the requested range.  Users may benefit from
>       calling copy_file_range() in a loop, and using the [lseek(2)](https://man7.org/linux/man-pages/man2/lseek.2.html)
>       SEEK_DATA and SEEK_HOLE operations to find the locations of data
>       segments.

thesamesam (commented):
Some initial comments. Thanks a lot for doing this.

I haven't yet tested it or worked through the calls/compared with docs yet though (I assume Zac is planning on doing the same too).

lib/portage/util/file_copy.py
```diff
@@ -0,0 +1,115 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+import errno
+import fcntl
+import os
+import platform
+import shutil
+
+
+FICLONE = 0x40049409
```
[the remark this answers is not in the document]
zmedico: Can use this:
```python
FICLONE = getattr(fcntl, "FICLONE", 0x40049409)
```

lib/portage/util/file_copy.py
```diff
@@ -0,0 +1,115 @@
...
+        if platform.system() == "Linux":
+            try:
+                fcntl.ioctl(dstfd, FICLONE, srcfd)
+                return True
+            except OSError:
+                pass
+
+        if hasattr(os, "copy_file_range"):
+            try:
+                _do_copy_file_range(srcfd, dstfd)
```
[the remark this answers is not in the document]
floppym: The current logic "caches" the result for the current file by setting `try_cfr` to false.

Doing anything beyond that would get complex really fast.

floppym:
> I haven't yet tested it or worked through the calls/compared with docs yet though (I assume Zac is planning on doing the same too).

I'm currently dog fooding it with an @world update of ~250 packages on my main system. No glaring issues noticed yet.

lib/portage/util/file_copy.py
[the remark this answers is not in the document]
floppym: Replaced with Zac's example.

lib/portage/tests/util/file_copy/test_copyfile.py
```diff
@@ -42,25 +42,27 @@ def testCopyFileSparse(self):
...
+            dest_stat = os.stat(dest_path)
+
+            self.assertEqual(src_stat.st_size, dest_stat.st_size)
 
             # If sparse blocks were preserved, then both files should
             # consume the same number of blocks.
-            self.assertEqual(os.stat(src_path).st_blocks, os.stat(dest_path).st_blocks)
+            # This is expected to fail when sparse copy is not implemented.
+            if src_stat.st_blocks != dest_stat.st_blocks:
+                pytest.xfail(reason="sparse copy is not implemented")
```
zmedico, line 66: We could possibly do things like use an `fcntl.ioctl` wrapper to check for a successful FICLONE ioctl call, and then adjust assertions accordingly, like can we assert that FICLONE results in a sparse copy?

zmedico: I was thinking of something roughly like this:
```diff
diff --git a/lib/portage/tests/util/file_copy/test_copyfile.py b/lib/portage/tests/util/file_copy/test_copyfile.py
index e4979160f3..a3bb658ee3 100644
--- a/lib/portage/tests/util/file_copy/test_copyfile.py
+++ b/lib/portage/tests/util/file_copy/test_copyfile.py
@@ -2,6 +2,9 @@
 # Distributed under the terms of the GNU General Public License v2
 
+import contextlib
+import fcntl
 import shutil
 import tempfile
+from unittest.mock import patch
 
 import pytest
@@ -10,5 +13,5 @@ from portage import os
 from portage.tests import TestCase
 from portage.checksum import perform_md5
-from portage.util.file_copy import copyfile
+from portage.util.file_copy import copyfile, FICLONE
 
 
@@ -33,4 +36,5 @@ class CopyFileTestCase(TestCase):
 class CopyFileSparseTestCase(TestCase):
     def testCopyFileSparse(self):
+        stack = contextlib.ExitStack()
         tempdir = tempfile.mkdtemp()
         try:
@@ -50,5 +54,13 @@ class CopyFileSparseTestCase(TestCase):
                 # the end of the file.
                 f.truncate(2**18)
+            orig_ioctl = fcntl.ioctl
 
+            def mock_ioctl(fd, request, arg=0, mutate_flag=True):
+                result = orig_ioctl(fd, request, arg, mutate_flag)
+                if request == FICLONE:
+                    mock_ioctl.ficlone_success = True
+                return result
+
+            stack.enter_context(patch("fcntl.ioctl", new=mock_ioctl))
             copyfile(src_path, dest_path)
 
@@ -64,5 +76,8 @@ class CopyFileSparseTestCase(TestCase):
             # This is expected to fail when sparse copy is not implemented.
             if src_stat.st_blocks != dest_stat.st_blocks:
+                if getattr(mock_ioctl, "ficlone_success", None):
+                    pytest.fail("sparse copy failed with FICLONE")
                 pytest.xfail(reason="sparse copy is not implemented")
         finally:
             shutil.rmtree(tempdir)
+            stack.close()
```

zmedico: Yeah, a _fastcopy mock to detect OSError sounds pretty reasonable.

floppym: Pushed a _fastcopy mock.

lib/portage/tests/util/file_copy/test_copyfile.py
```diff
@@ -42,25 +43,36 @@ def testCopyFileSparse(self):
...
-                f.seek(2**17, 1)
+                # the end of the file.
+                f.truncate(2**18)
 
-            copyfile(src_path, dest_path)
+            fastcopy_success = False
+
+            def mock_fastcopy(src, dst):
+                _fastcopy(src, dst)
+                fastcopy_success = True
```
dwfreed, line 58:
```suggestion
                nonlocal fastcopy_success
                _fastcopy(src, dst)
                fastcopy_success = True
```

In python closures, when you want to write to a variable in the outer scope, you need to first declare it `nonlocal`

dwfreed (changes requested):
Just saw this one issue, otherwise seems correct to me.

lib/portage/tests/util/file_copy/test_copyfile.py
floppym: I thought I tested this, but I think you are right. Updated.

floppym: Retested and confirmed that nonlocal is indeed necessary here.

lib/portage/util/file_copy.py
```diff
@@ -0,0 +1,135 @@
...
+                    break
+
+                # SEEK_DATA failed with EINVAL, return the whole file
+                yield (0, offset_end)
+                break
+            else:
+                offset_hole = os.lseek(src, offset_data, SEEK_HOLE)
+                yield (offset_data, offset_hole - offset_data)
+    except OSError:
+        logger.warning("_get_chunks failed unexpectedly", exc_info=sys.exc_info())
```
dwfreed, line 51:
```suggestion
    except OSError:
        logger.warning("_get_chunks failed unexpectedly", exc_info=sys.exc_info())
        raise
```

This needs to re-raise, or you won't copy the whole file.  This re-raise will trigger the `shutil.copyfile` fallback.

floppym: Good catch.

thesamesam (approved):

PR #1295, opened 2024-03-02
elog/mod_custom: Spawn processes in background

zmedico:
Since elog_process is typically called while the event loop is running, hold references to spawned processes and wait for them asynchronously, ultimately waiting for them if necessary when the AsyncioEventLoop _close_main method calls _async_finalize via portage.process.run_coroutine_exitfuncs().

ConfigProtectTestCase is useful for exercising this code, and this little make.globals patch can be used to test failure during finalize with this error message:

    !!! PORTAGE_ELOG_COMMAND failed with exitcode 1

```diff
--- a/cnf/make.globals
+++ b/cnf/make.globals
@@ -144 +144,2 @@ PORTAGE_ELOG_CLASSES="log warn error"
-PORTAGE_ELOG_SYSTEM="save_summary:log,warn,error,qa echo"
+PORTAGE_ELOG_SYSTEM="save_summary:log,warn,error,qa echo custom"
+PORTAGE_ELOG_COMMAND="/bin/false"
```
Bug: https://bugs.gentoo.org/925907

### DIFF
--- a/lib/portage/elog/mod_custom.py
+++ b/lib/portage/elog/mod_custom.py
@@ -1,10 +1,33 @@
 # elog/mod_custom.py - elog dispatch module
-# Copyright 2006-2020 Gentoo Authors
+# Copyright 2006-2024 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+import types
+
+import portage
 import portage.elog.mod_save
 import portage.exception
 import portage.process
+from portage.util.futures import asyncio
+
+# Since elog_process is typically called while the event loop is
+# running, hold references to spawned processes and wait for them
+# asynchronously, ultimately waiting for them if necessary when
+# the AsyncioEventLoop _close_main method calls _async_finalize
+# via portage.process.run_coroutine_exitfuncs().
+_proc_refs = None
+
+
+def _get_procs() -> list[tuple[portage.process.MultiprocessingProcess, asyncio.Future]]:
+    """
+    Return list of (proc, asyncio.ensure_future(proc.wait())) which is not
+    inherited from the parent after fork.
+    """
+    global _proc_refs
+    if _proc_refs is None or _proc_refs.pid != portage.getpid():
+        _proc_refs = types.SimpleNamespace(pid=portage.getpid(), procs=[])
+        portage.process.atexit_register(_async_finalize)
+    return _proc_refs.procs
 
 
 def process(mysettings, key, logentries, fulltext):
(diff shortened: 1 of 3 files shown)


lib/portage/elog/mod_custom.py
```diff
@@ -18,8 +39,49 @@ def process(mysettings, key, logentries, fulltext):
...
+    procs = _get_procs()
+    while procs:
+        proc, waiter = procs.pop()
+        if (await waiter) != 0:
+            raise portage.exception.PortageException(
+                f"!!! PORTAGE_ELOG_COMMAND failed with exitcode {waiter.result()}"
+            )
+
+
+def finalize():
```
floppym, line 69: Sorry for my ignorance, but is `finalize()` actually used anywhere?

zmedico: It's called in the atexit hook that's registered in the elog_process function here:
```python
                if hasattr(m, "finalize") and not m.finalize in _elog_atexit_handlers:
                    _elog_atexit_handlers.append(m.finalize)
                    atexit_register(m.finalize)
```
In practice the atexit hook is invoked a little early though, just before the event loop is closed, when we call `portage.process.run_exitfuncs()` at the end of the AsyncioEventLoop _close_main method:
```python
    async def _close_main(self):
        await mod_custom._async_finalize()

        # Even though this has an exit hook, invoke it here so that
        # we can properly wait for it and avoid messages like this:
        # [ERROR] Task was destroyed but it is pending!
        if socks5.proxy.is_running():
            await socks5.proxy.stop()

        portage.process.run_exitfuncs()
```
For both mod_custom and the socks5 proxy, the atexit hooks refer to processes that _close_main cleans up. We could refactor _close_main  to wait for both mod_custom and the socks5 proxy simultaneously using asyncio.gather.

floppym: Thanks for explaining.

zmedico: This made me realize that `portage.process.run_exitfuncs()` call actually triggers recursive calls to the AsyncioEventLoop close method via an [atexit hook](https://github.com/gentoo/portage/blob/dcbcac809213537afaa6b4f9822146a2e984f773/lib/portage/util/futures/_asyncio/__init__.py#L302). It's fine though because AsyncioEventLoop uses a _closing attribute to detect recursive calls and drop them.

lib/portage/util/_eventloop/asyncio_event_loop.py
```diff
@@ -75,12 +80,17 @@ def _close(self):
             self._closing = False
 
     async def _close_main(self):
+        tasks = []
+        tasks.append(asyncio.ensure_future(mod_custom._async_finalize()))
```
thesamesam, line 84: It feels a bit like an abstraction break but then again, the whole point of custom is it could do anything, so it's probably fine.

zmedico: We can add general purpose async atexit hooks, which we can use for both mod_custom and socks5 proxy.

zmedico: I've added a general purpose `portage.process.run_coroutine_exitfuncs()` function to handle it:
```diff
diff --git a/lib/portage/elog/mod_custom.py b/lib/portage/elog/mod_custom.py
index 1a74acdc95..0392c7afa7 100644
--- a/lib/portage/elog/mod_custom.py
+++ b/lib/portage/elog/mod_custom.py
@@ -26,4 +26,5 @@ def _get_procs() -> list[tuple[portage.process.MultiprocessingProcess, asyncio.F
     if _proc_refs is None or _proc_refs.pid != portage.getpid():
         _proc_refs = types.SimpleNamespace(pid=portage.getpid(), procs=[])
+        portage.process.atexit_register(_async_finalize)
     return _proc_refs.procs
 
diff --git a/lib/portage/process.py b/lib/portage/process.py
index b652e32942..0fedfea4ab 100644
--- a/lib/portage/process.py
+++ b/lib/portage/process.py
@@ -227,4 +227,26 @@ def run_exitfuncs():
 
 
+async def run_coroutine_exitfuncs():
+    """
+    This is the same as run_exitfuncs but it uses asyncio.iscoroutinefunction
+    to check which functions to run. It is called by the AsyncioEventLoop
+    _close_main method just before the loop is closed.
+    """
+    tasks = []
+    for index, (func, targs, kargs) in reversed(list(enumerate(_exithandlers))):
+        if asyncio.iscoroutinefunction(func):
+            del _exithandlers[index]
+            tasks.append(asyncio.ensure_future(func(*targs, **kargs)))
+    exc_info = None
+    for task in tasks:
+        try:
+            await task
+        except Exception:
+            traceback.print_exc()
+            exc_info = sys.exc_info()
+    if exc_info is not None:
+        raise exc_info[1].with_traceback(exc_info[2])
+
+
 atexit.register(run_exitfuncs)
 
diff --git a/lib/portage/util/_eventloop/asyncio_event_loop.py b/lib/portage/util/_eventloop/asyncio_event_loop.py
index 50cc0ef22a..2b34b28028 100644
--- a/lib/portage/util/_eventloop/asyncio_event_loop.py
+++ b/lib/portage/util/_eventloop/asyncio_event_loop.py
@@ -82,5 +82,5 @@ class AsyncioEventLoop(_AbstractEventLoop):
     async def _close_main(self):
         tasks = []
-        tasks.append(asyncio.ensure_future(mod_custom._async_finalize()))
+        tasks.append(asyncio.ensure_future(portage.process.run_coroutine_exitfuncs()))
 
         # Even though this has an exit hook, invoke it here so that
```

thesamesam (approved):
Thanks!

lib/portage/util/_eventloop/asyncio_event_loop.py
zmedico: Now the socks5 stuff cleans up nicely in https://github.com/gentoo/portage/pull/1299.

PR #1297, opened 2024-03-02
_validate_deps: Discard configdict["pkg"]["USE"]

zmedico:
Since configdict["pkg"]["USE"] may contain package.use settings from config.setcpv, it is inappropriate to use here (bug 675748), so discard it. This is only an issue because configdict["pkg"] is a sub-optimal place to extract metadata from. This issue does not necessarily indicate a flaw in the Package constructor, since passing in precalculated USE can be valid for things like autounmask USE changes.

Bug: https://bugs.gentoo.org/675748

### DIFF
--- a/lib/portage/tests/ebuild/test_doebuild_fd_pipes.py
+++ b/lib/portage/tests/ebuild/test_doebuild_fd_pipes.py
@@ -51,8 +51,21 @@ class DoebuildFdPipesTestCase(TestCase):
         ebuilds = {
             "app-misct/foo-1": {
                 "EAPI": "8",
+                "IUSE": "+foo +bar",
+                "REQUIRED_USE": "|| ( foo bar )",
                 "MISC_CONTENT": ebuild_body,
             }
+        }
+
+        # Populate configdict["pkg"]["USE"] with something arbitrary in order
+        # to try and trigger bug 675748 in doebuild _validate_deps.
+        arbitrary_package_use = "baz"
+
+        user_config = {
+            # In order to trigger bug 675748, package.env must be non-empty,
+            # but the referenced env file can be empty.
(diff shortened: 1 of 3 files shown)


thesamesam (approved):
Any chance of a test case, or is it a pain because of ResolverPlayground? Does it work if we do it with global USE? I guess not? If not, it's fine obviously and I wouldn't worry about it.

lib/portage/tests/ebuild/test_doebuild_fd_pipes.py
```diff
@@ -106,6 +119,10 @@ def testDoebuild(self):
             )
             settings.setcpv(pkg)
 
+            # Demonstrate that settings.configdict["pkg"]["USE"] contains our arbitrary
+            # package.use setting in order to trigger bug 675748.
+            self.assertEqual(settings.configdict["pkg"]["USE"], arbitrary_package_use)
+
```
zmedico, line 125: > Any chance of a test case, or is it a pain because of ResolverPlayground?

I was able to add some package.env and package.use settings in testDoebuild that create a state like the one which triggered bug 925625.

> Does it work if we do it with global USE? I guess not? If not, it's fine obviously and I wouldn't worry about it.

By global USE do you mean like configdict["env"]["USE"] which would be like the setting from emerge --info? That value should never have any effect on _validate_deps.

zmedico: I also realized that we could do this to avoid a redundant setcpv call in _validate_deps:
```python
    # A USE calculation from setcpv should always be available here because
    # mysettings.mycpv is not None, so use it to prevent redundant setcpv calls.
    metadata["USE"] = mysettings["PORTAGE_USE"]
```

PR #1299, opened 2024-03-03
socks5: Use run_coroutine_exitfuncs()

zmedico:
Since commit c3ebdbb42e72 from https://github.com/gentoo/portage/pull/1295 the atexit_register(proxy.stop) call in the get_socks5_proxy function can accept a coroutine function to execute in run_coroutine_exitfuncs(), so convert the ProxyManager stop method to a coroutine and use run_coroutine_exitfuncs().

Bug: https://bugs.gentoo.org/925240


thesamesam (approved):
Thank you!

PR #1301, opened 2024-03-06
util: set a timeout for urlopen calls

floppym:
A hung urlopen call can cause emerge to produce no output when fetching binhost data.

Bug: https://bugs.gentoo.org/926221

### DIFF
--- a/lib/portage/util/_urlopen.py
+++ b/lib/portage/util/_urlopen.py
@@ -26,10 +26,10 @@ def have_pep_476():
     return hasattr(__import__("ssl"), "_create_unverified_context")
 
 
-def urlopen(url, if_modified_since=None, headers={}, proxies=None):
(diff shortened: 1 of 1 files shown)


lib/portage/util/_urlopen.py
```diff
@@ -29,7 +29,7 @@ def have_pep_476():
 def urlopen(url, if_modified_since=None, headers={}, proxies=None):
```
zmedico, line 29: How about if we add a default `timeout=10` parameter up here?

floppym: Done.

zmedico (commented):
This seems like a reasonable fix and the socket.timeout or TimeoutError inherits from OSError so binarytree should handle it appropriately.

PR #1302, opened 2024-03-06
phase-functions: prematurely delete WORKDIR if FEATURES=merge-wait

Flowdalic:
Using the merge-wait feature together with many parallel emerge jobs potentially leads to a high disk space usage due to the created portage build directories waiting for their packages to be merged into the live filesystem prior portage will clean them. This can easily lead to out-of-space errors.

Prematurely deleting WORKDIR at the end of src_install() helps reducing the temporarily used disk space. All that portage need to merge a package into the live filesystem are the files in the image/ directory (D). (At least in theory)

### DIFF
--- a/bin/phase-functions.sh
+++ b/bin/phase-functions.sh
@@ -723,6 +723,13 @@ __dyn_install() {
 		>> DEBUGBUILD
 	fi
 	trap - SIGINT SIGQUIT
+
+	# Prematurely delete WORKDIR in case merge-wait is enabled to
+	# decrease the space used by portage build directories until the
+	# packages are merged and cleaned.
+	if has merge-wait ${FEATURES} && ! has keepwork ${FEATURES}; then
+		rm -rf "${WORKDIR}"
+	fi
 }
 
 __dyn_help() {


bin/phase-functions.sh
```diff
@@ -722,6 +722,14 @@ __dyn_install() {
...
 	fi
+
+	# Prematurely delete WORKDIR in case merge-wait is enabled, to
+	# decrease the space used by portage build directories until the
+	# packages are merged and cleaned.
+	if has merge-wait ${FEATURES} && ! has keepwork ${FEATURES}; then
+		rm -rf "${WORKDIR}"
+	fi
+
 	trap - SIGINT SIGQUIT
```
zmedico, line 733: We should do it after this `trap - SIGINT SIGQUIT` since the __abort_install critical section is over.

Flowdalic: Fair point, fixed.

zmedico (approved):

thesamesam (approved):

floppym:
Hey, so this PR breaks [install-qa-check.d/90config-impl-decl](https://github.com/gentoo/portage/blob/8eb2502bf264d54dfbf180222f162f7a593b947d/bin/install-qa-check.d/90config-impl-decl#L69):
```
>>> Completed installing sys-libs/libxcrypt-4.4.36-r3 into /x/portage/sys-libs/libxcrypt-4.4.36-r3/image

 * Final size of build directory: 13072 KiB (12.7 MiB)
 * Final size of installed tree:   1448 KiB ( 1.4 MiB)

find: ‘/x/portage/sys-libs/libxcrypt-4.4.36-r3/work’: No such file or directory
strip: x86_64-pc-linux-gnu-strip --strip-unneeded -N __gentoo_check_ldflags__ -R .comment -R .GCC.command.line -R .note.gnu.gold-version
```

zmedico:
We can remove it just after the hooks instead.

Flowdalic:
thanks @zmedico

thesamesam:
We should also skip it if we're doing `ebuild ... clean install` or so. I often do that while testing ebuilds and then may go look in the work dir after. There's no value in using merge-wait w/ `ebuild`.

zmedico:
> We should also skip it if we're doing `ebuild ... clean install` or so. I often do that while testing ebuilds and then may go look in the work dir after. There's no value in using merge-wait w/ `ebuild`.

https://github.com/gentoo/portage/pull/1307 has a fix.

PR #1303, opened 2024-03-08
meson.build: Install dist-info METADATA

zmedico:
Install dist-info METADATA for pip to resolve dependencies:
```console
# pip freeze | grep portage
portage==3.0.63
```
Suggested-by: Eli Schwartz <eschwartz93@gmail.com>
Bug: https://bugs.gentoo.org/920330

### DIFF
--- a/cnf/METADATA
+++ b/cnf/METADATA
@@ -1,0 +1,3 @@
+Metadata-Version: 2.1
+Name: portage
+Version: @VERSION@
--- a/meson.build
+++ b/meson.build
@@ -97,6 +97,20 @@ if native_extensions
     subdir('src')
 endif
 
+if system_wide
+	METADATA = configure_file(
+		input : 'cnf/METADATA',
+		output : 'METADATA',
+		configuration : conf_data
+	)
+	install_data(
+		[
+			METADATA
+		],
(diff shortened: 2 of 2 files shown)


chewi (approved):
Thanks for keeping it simple in the end. :slightly_smiling_face:

thesamesam (approved):

mgorny:
Please give me 24 hours or so to attempt to rewrite `sys-apps/portage` ebuild to use plain PEP517 build instead of the awful pile of hacks it's doing right now.

chewi:
> Please give me 24 hours or so to attempt to rewrite `sys-apps/portage` ebuild to use plain PEP517 build instead of the awful pile of hacks it's doing right now.

I honestly don't believe my work was an awful pile of hacks. Portage does not abuse Meson's Python support, and Meson installs the files to the same locations that PEP517 would. You're effectively saying that Meson's Python support is worthless unless used together with meson-python. Feel free to try something different, but I think it's good that Portage can currently be installed system-wide using the regular Meson commands.

mgorny:
Nevermind. I didn't realize that mesonpy is so strict about wheel specification, and doesn't allow installing to other paths like setuptools does. This is hopeless.

mgorny (approved):

PR #1304, opened 2024-03-09
__dyn_install: Record REPO_REVISIONS in build-info

zmedico:
Record REPO_REVISIONS as a json object that maps repo name to revision for an ebuild's source repository and any repositories that eclasses were inherited from:
```console
$ cat /var/tmp/portage/sys-apps/portage-3.0.63/build-info/REPO_REVISIONS
{"gentoo": "34875e30e73e33d3597d1101cdf97dc22729b268"}
```
Ultimately the intention is to expose this information in binhost metadata so that clients can select consistent revisions of source repositories.

Bug: https://bugs.gentoo.org/924772


thesamesam (approved):
I think this is going to be generally useful for reproducible build efforts / determinism / auditing.

PR #1305, opened 2024-03-09
EbuildPhase/EbuildBinpkg: Ensure PKGDIR subdirectory permissions

zmedico:
Bug: https://bugs.gentoo.org/712222

### DIFF
--- a/lib/portage/dbapi/bintree.py
+++ b/lib/portage/dbapi/bintree.py
@@ -72,6 +72,7 @@ import traceback
 import warnings
 from gzip import GzipFile
 from itertools import chain
+from pathlib import PurePath
 from urllib.parse import urlparse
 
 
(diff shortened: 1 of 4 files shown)


lib/portage/dbapi/bintree.py
```diff
@@ -839,11 +839,18 @@ def _ensure_dir(self, path):
...
         pkgdir_gid = pkgdir_st.st_gid
         pkgdir_grp_mode = 0o2070 & pkgdir_st.st_mode
-        try:
-            ensure_dirs(path, gid=pkgdir_gid, mode=pkgdir_grp_mode, mask=0)
-        except PortageException:
-            if not os.path.isdir(path):
-                raise
+
+        components = []
+        for component in path[len(self.pkgdir) + 1 :].split(os.sep):
```
thesamesam, line 844: You might be able to do this more elegantly with pathlib.

zmedico: Yeah, relative_to and parts worked nicely here, thanks!
```python
for component in pathlib.PurePosixPath(path).relative_to(self.pkgdir).parts:
```

thesamesam (approved):

lib/portage/dbapi/bintree.py
thesamesam: Nice!

PR #1306, opened 2024-03-11
bintree: Add REPO_REVISIONS to package index header

zmedico:
As a means for binhost clients to select source repo revisions which are consistent with binhosts, inject REPO_REVISIONS from a package into the index header, using a history of synced revisions to guarantee forward progress. This queries the relevant repos to check if any new revisions have appeared in the absence of a proper sync operation.

The history of synced revisions is provided by a new get_repo_revision_history function and corresponding /var/lib/portage/repo_revisions file, with history limit currently capped at 25 revisions. If a change is detected and the current process has permission to update the repo_revisions file, then the file will be updated with any newly detected revisions. This functions detects revisions which are not yet visible to the current process due to the sync-rcu option.

The emaint revisions --purgerepos and --purgeallrepos options allow revisions for some or all repos to be easily purged from the history. For example, the emerge-webrsync script uses this emaint commmand to purge the revision history of the gentoo repo when the emerge-webrsync --revert option is used to roll back to a previous snapshot:

	emaint revisions --purgerepos="${repo_name}"

Bug: https://bugs.gentoo.org/924772


lib/portage/sync/revision_history.py
```diff
@@ -0,0 +1,121 @@
...
+    This functions detects revisions which are not yet visible to the
+    current process due to the sync-rcu option.
+
+    @param eroot: EROOT to query
+    @type eroot: string
+    @param repos: list of RepoConfig instances to check for new revisions
+    @type repos: list
+    @rtype: dict
+    @return: mapping of repo name to list of revisions in descending
+             order by time
```
zmedico, line 35: Ultimately, when binhosts clients need to checkout a specific source repo revision, we don't want that to lead to an unordered revision list. I think in that case we might just purge newer revisions from the history, and set the selected source repo revision as the latest revision in the history.

zmedico: Nothing prevents users from doing things to create unordered revision lists. It doesn't seem like an issue for plain rsync, but emerge-webrsync is capable of selecting dated snapshots, and nothing stops git sync users from checking out random revisions that can then get recorded in the history.

zmedico: When emerge-webrsync installs a dated snapshot, we can easily wipe previous history and just add the snapshot's revision to the history.

Checkouts of random git revisions are much harder to handle.

zmedico: Maybe we can prevent random git revisions from entering the history by making the GitSync retrieve_head method return an error it's not on a branch that's in sync with a remote tracking branch. Maybe we can use this same mechanism on binhost clients to prevent specifically checked out revisions from entering the history.

zmedico: When a binhost client needs to checkout a specific revision, it can insert that revision into the appropriate location in its revision history, since will have access to git history at that point.

zmedico: I've fixed it to ignore revisions from volatile repos, which seems like it will cover the random git revisions problem.

I still want to handle emerge-webrsync snapshots.

zmedico: Fixed `emerge-webrsync --revert` to do this:
```bash
emaint revisions --purgerepos="${repo_name}"
```

thesamesam:
Sorry I haven't had a chance to look yet, the next few days are still quite busy but I will try by/on the weekend.

lib/portage/dbapi/bintree.py
```diff
@@ -1916,6 +1929,56 @@ def _read_metadata(self, filename, st, keys=None, binpkg_format=None):
...
+                header_rev_index = None
+            try:
+                repo_revision_index = rev_list.index(repo_revision)
+            except ValueError:
+                repo_revision_index = None
+            if repo_revision_index is not None and (
+                header_rev_index is None or repo_revision_index < header_rev_index
+            ):
+                # There is forward progress when repo_revision is more recent
+                # than header_rev or header_rev was not found in the history.
```
zmedico, line 1971: The repo_revisions history file exists solely so that we can perform this comparison in order to ensure forward progress.

As an alternative that avoids the need for a repo revision history, we could simply expose the latest synced revision here regardless of the revision that the package was built with. In most cases I would expect the package would be built against the latest revision anyway (unless there's a cron job that syncs periodically in the background).

zmedico: @akhuettel, for gentoo binhost builds, is there a cron job that sync repos in the background or is sync triggered prior to builds? Either will work if we track the history of repo revisions, but I'm just checking since tracking repo revision history just so we can validate revisions exposed in the binhost Packages header could be overkill.

thesamesam:
Sorry, I'm not ignoring this, I really will try to get to it soon, I just don't want to give a half-assed review given it involves a file format and changing those isn't easy.

zmedico:
> Sorry, I'm not ignoring this, I really will try to get to it soon, I just don't want to give a half-assed review given it involves a file format and changing those isn't easy.

Something I noticed when with working with Amazon Linux 2023 is that they have a dnf releasever mechanism which prevents new binary packages from being installed too early (it's the opposite of the problem reported in [bug 924772](https://bugs.gentoo.org/924772)), which is something that REPO_REVISIONS can't help with:

https://docs.aws.amazon.com/linux/al2023/ug/deterministic-upgrades-usage.html

lib/portage/emaint/modules/revisions/revisions.py
```diff
@@ -0,0 +1,95 @@
...
+    short_desc = "Purge repo_revisions history file."
+
+    @staticmethod
+    def name():
+        return "revisions"
+
+    def __init__(self, settings=None):
+        """Class init function
+
+        @param settings: optional portge.config instance to get EROOT from.
```
thesamesam, line 20:
```suggestion
        @param settings: optional portage.config instance to get EROOT from.
```

thesamesam (approved):
Thank you! (And thank you for waiting.)

lib/portage/dbapi/bintree.py
akhuettel: @zmedico The sync is triggered and completed before the build starts (both for the dedicated binhosts and for the catalyst builds).
The dedicated binhosts use a separate (from the host) repo dir for that, and the catalyst builds have their squashfs snapshot.

PR #1309, opened 2024-03-18
sync: don't use ipv6 for rsync when it's disabled

0x501D:
socket.has_ipv6 gives a false result:

```sh
$ sysctl net.ipv6.conf.all.disable_ipv6=1
net.ipv6.conf.all.disable_ipv6 = 1
$ python
Python 3.11.8 (main, Feb 24 2024, 17:10:38) [GCC 13.2.1 20240210] on linux Type "help", "copyright", "credits" or "license" for more information.
>>> import socket
>>> socket.has_ipv6
True
```

This patch uses the built-in _has_ipv6() function, which returns the correct result.

Bug: https://bugs.gentoo.org/927241

### DIFF
--- a/lib/portage/process.py
+++ b/lib/portage/process.py
@@ -631,8 +631,8 @@ def spawn(
         fd_pipes[1] = pw
         fd_pipes[2] = pw
 
-    # Cache _has_ipv6() result for use in child processes.
-    _has_ipv6()
+    # Cache has_ipv6() result for use in child processes.
+    has_ipv6()
 
     # This caches the libc library lookup and _unshare_validator results
     # in the current process, so that results are cached for use in
@@ -751,7 +751,7 @@ def spawn(
 __has_ipv6 = None
 
 
-def _has_ipv6():
+def has_ipv6():
     """
     Test that both userland and kernel support IPv6, by attempting
     to create a socket and listen on any unused port of the IPv6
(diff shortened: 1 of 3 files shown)


zmedico:
Thanks @0x501D this looks great! Can you add a sign off to comply with our [Certificate of Origin](https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin)?

0x501D:
> Thanks @0x501D this looks great! Can you add a sign off to comply with our [Certificate of Origin](https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin)?

Sign off added.

0x501D:
Linter errors was fixed.

floppym:
> This patch uses the built-in _has_ipv6() function, which returns the correct result.

Nit: there is no "built-in" _has_ipv6() function. I would refer to `portage.process._has_ipv6() instead.

0x501D:
> > This patch uses the built-in _has_ipv6() function, which returns the correct result.
> 
> Nit: there is no "built-in" _has_ipv6() function. I would refer to `portage.process._has_ipv6()` instead.

Thanks. Commit message was fixed.

floppym (approved):

zmedico (approved):

floppym:
I suppose we should remove the underscore from the front of the function name if we are going to call it from another module. It's no longer a "private" function.

0x501D:
> I suppose we should remove the underscore from the front of the function name if we are going to call it from another module. It's no longer a "private" function.

Done. Added in separate commit.

floppym:
Perfect, thank you.

PR #1310, opened 2024-03-20
depgraph: Do not trigger rebuilds for some options

zmedico:
Do not trigger rebuilds for options which do not merge packages like --buildpkgonly, --fetchonly, and --fetch-all-uri.

Bug: https://bugs.gentoo.org/927301

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -109,6 +109,8 @@ if TYPE_CHECKING:
     import _emerge.stdout_spinner.stdout_spinner
 
 
+_NO_MERGE_OPTS = ("--buildpkgonly", "--fetchonly", "--fetch-all-uri")
+
 # Exposes a depgraph interface to dep_check.
 _dep_check_graph_interface = collections.namedtuple(
     "_dep_check_graph_interface",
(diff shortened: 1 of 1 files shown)


mgorny:
Still happens with the patch applied on top of 3.0.63.

mgorny:
> Still happens with the patch applied on top of 3.0.63.

Though the conditions seem to work. I guess some other code needs to be disabled as well.

lib/_emerge/depgraph.py
```diff
@@ -1942,7 +1944,9 @@ def _process_slot_conflicts(self):
         for conflict in self._dynamic_config._package_tracker.slot_conflicts():
             self._process_slot_conflict(conflict)
 
-        if self._dynamic_config._allow_backtracking:
+        if self._dynamic_config._allow_backtracking and not any(
+            opt in self._frozen_config.myopts for opt in _NO_MERGE_OPTS
+        ):
             self._slot_operator_trigger_reinstalls()
```
zmedico, line 1950: It seems like --buildpkgonly needs to be in a class of its own here, since --fetch* should behave mostly like regular build commands in order to avoid prevent discrepancies in behavior (like the slot conflicts in https://bugs.gentoo.org/161422 thought those would exist without --fetchonly you get the idea).

PR #1311, opened 2024-03-21
save-build-env: Strip out MAKEOPTS

ismell:
MAKEOPTS contains a host specific value. If the same package is built on a host with different core counts, the binpkgs will have different hashes. This change drops MAKEOPTS after src_install so that it's not included in final environment.gz.

Bug: https://bugs.gentoo.org/914441

I'm pulling this patch into chromiumos here: https://chromium-review.googlesource.com/c/chromiumos/bazel/+/5356724

### DIFF
--- a/bin/save-ebuild-env.sh
+++ b/bin/save-ebuild-env.sh
@@ -16,6 +16,11 @@ __save_ebuild_env() {
 		unset S __E_DESTTREE __E_INSDESTTREE __E_DOCDESTTREE __E_EXEDESTTREE \
 			PORTAGE_DOCOMPRESS_SIZE_LIMIT PORTAGE_DOCOMPRESS \
 			PORTAGE_DOCOMPRESS_SKIP PORTAGE_DOSTRIP PORTAGE_DOSTRIP_SKIP
+		# This value contains build host specific configuration. We want
+		# binpkgs generated on different sized hosts to be identical,
+		# so strip this value from the binpkg. It's also not needed for
+		# installing / removing a package.
+		unset MAKEOPTS
 		if [[ -n ${PYTHONPATH} &&
 			${PYTHONPATH%%:*} -ef ${PORTAGE_PYM_PATH} ]] ; then
 			if [[ ${PYTHONPATH} == *:* ]] ; then


zmedico (approved):
Seems reasonable. Thanks!

thesamesam:
I suppose this might help with https://bugs.gentoo.org/736794 too.

bin/save-ebuild-env.sh
```diff
@@ -16,6 +16,11 @@ __save_ebuild_env() {
 		unset S __E_DESTTREE __E_INSDESTTREE __E_DOCDESTTREE __E_EXEDESTTREE \
 			PORTAGE_DOCOMPRESS_SIZE_LIMIT PORTAGE_DOCOMPRESS \
 			PORTAGE_DOCOMPRESS_SKIP PORTAGE_DOSTRIP PORTAGE_DOSTRIP_SKIP
+		# This value contains build host specific configuration. We want
+		# binpkgs generated on different sized hosts to be identical,
+		# so strip this value from the binpkg. It's also not needed for
+		# installing / removing a package.
+		unset MAKEOPTS
```
zmedico, line 23: How about if we also add MAKEOPTS to environ_whitelist in lib/portage/package/ebuild/_config/special_env_vars.py so that we'll always have the current user setting in the ebuild environment? That way any MAKEOPTS usage in binary package ebuild phases will be covered (like https://bugs.gentoo.org/736794).

ismell: SGTM. I do have a question. When does the `has --exclude-init-phases` block run? I assumed when we were creating a binpkg. Does it make a different if we move the unset out of this block? i.e., put it down the the `unset COLORTERM DISPLAY` ,etc?

I also think we need to exclude `NINJAOPTS`. See https://chromium-review.googlesource.com/c/chromiumos/bazel/+/5445316

zmedico: The `--exclude-init-phases` part is used after src_install and also when persisting state after pkg_* phases.

It's better if we keep the unset MAKEOPTS under the `--exclude-init-phases` block in case the variable is modified within the ebuild environment, since variable mutations within the ebuild environment should generally persist unless there is a very good reason not to.

ismell: Thanks for the clarification. That's what I had originally understood.

ismell:
The change has been updated to also exclude NINJAOPTS, EPOCHREALTIME, EPOCHSECONDS, and SRANDOM.

ismell:
Anything blocking this?

zmedico (approved):
> Anything blocking this?

I think it's fine. @sam what do you think?

PR #1312, opened 2024-03-21
Reapply "config: Don't directly modify FEATURES"

ismell:
This change addresses the feedback from https://github.com/gentoo/portage/pull/1098#issuecomment-1761638611

As pointed out by @zmedico, It turns out we need to keep the open coded filter because otherwise we create a persistent copy.


zmedico (approved):
Looks good, thanks!

PR #1319, opened 2024-05-10
emerge: Make --root-deps install build deps to ROOT plus / for all EAPIs

chewi:
Rather than instead of / for EAPI 6 and below. This changes the behaviour, but it is arguably not a breaking change, as no installations are being dropped. The previous behaviour was highly likely to break anyway.

I'm not 100% sure about the changes in `_complete_graph` and `_validate_blockers`, but I think the idea here was to skip processing when you were discarding dependencies for /. We're not doing that for bare `--root-deps` any more, only `rdeps` specifically.


lib/_emerge/depgraph.py
```diff
@@ -8390,7 +8398,7 @@ def _complete_graph(self, required_sets=None):
         for root in self._frozen_config.roots:
             if root != self._frozen_config.target_root and (
                 "remove" in self._dynamic_config.myparams
-                or self._frozen_config.myopts.get("--root-deps") is not None
+                or self._frozen_config.myopts.get("--root-deps") != "rdeps"
```
zmedico, line 8401: Looking back at f241cdb50b0f8a448216a1e5155dba8d0abddf56, I'm not sure what I fixed. Maybe we could try removing --root-deps from this expression entirely (basically revert f241cdb50b0f8a448216a1e5155dba8d0abddf56).

lib/_emerge/depgraph.py
```diff
@@ -8625,7 +8633,7 @@ def _validate_blockers(self):
             dep_keys = Package._runtime_keys
             for myroot in self._frozen_config.trees:
                 if (
-                    self._frozen_config.myopts.get("--root-deps") is not None
+                    self._frozen_config.myopts.get("--root-deps") != "rdeps"
                     and myroot != self._frozen_config.target_root
```
zmedico, line 8637: Looking back at e9740bcd93e01633695dee32425d1bab38e607c7, I'm not sure about the optimization logic. Once again, maybe we could try removing --root-deps from this expression entirely (basically revert e9740bcd93e01633695dee32425d1bab38e607c7).

It seems a9de7a29a25c876caab34b288cf87629514c7ac2 which was committed around the same time was also reverted at some point.

chewi: I don't think keeping this does any harm, and I'd rather not break anything. Same for the above. I do intend to remove the `rdeps` option when we drop EAPI 6, although I was discussing the need for such an option that works with EAPI 7+ with someone today. `rdeps` is a terrible name, so I'd probably call it something else. I'm also considering other ways to solve their problem.

zmedico: Okay that seems fair to preserve backward compat. Should these conditions use `==` rather than `!=`, in order to affect the `--root-deps=rdeps` case, like this?
```diff
-               or self._frozen_config.myopts.get("--root-deps") is not None
+               or self._frozen_config.myopts.get("--root-deps") == "rdeps"
```

chewi: Ah, good point. :upside_down_face:

PR #1323, opened 2024-05-17
install-qa-checks.d: suppress some gnulib implicit configure declarations

eli-schwartz:
These happen in tons of GNU packages because of using gnulib, which pulls in macros that check for some functionality and spit out an implicit function declaration error if they aren't supported, which is *expected*.

### DIFF
--- a/bin/install-qa-check.d/90config-impl-decl
+++ b/bin/install-qa-check.d/90config-impl-decl
@@ -49,6 +49,15 @@ add_default_skips() {
 		res_ndestroy
 		statacl
 	)
+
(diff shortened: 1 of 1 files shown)


bin/install-qa-check.d/90config-impl-decl
```diff
@@ -49,6 +49,15 @@ add_default_skips() {
 		res_ndestroy
 		statacl
 	)
+
+	# Available in c23, these gnulib checks are expected to fail
```
thesamesam, line 53: Move the c23 comment to above the 3 instead?

PR #1324, opened 2024-05-21
Improve legibility of scheduler code

Flowdalic:


lib/_emerge/Scheduler.py
```diff
@@ -1953,26 +1953,21 @@ def _schedule_tasks_imp(self):
...
-                return bool(state_change)
+                return False
 
             if (
-                self._choose_pkg_return_early
-                or self._merge_wait_scheduled
+                self._merge_wait_scheduled
                 or (self._jobs and self._unsatisfied_system_deps)
                 or not self._can_add_job()
                 or self._job_delay()
```
zmedico, line 1964: A reason to include self._choose_pkg_return_early in this condition is to skip unnecessary calls to self._can_add_job() and self._job_delay() when self._choose_pkg() will return None anyway. It saves some cpu time spent scheduling.

lib/_emerge/Scheduler.py
```diff
@@ -1953,26 +1953,21 @@ def _schedule_tasks_imp(self):
         @return: True if state changed, False otherwise.
         """
 
-        state_change = 0
-
         while True:
             if not self._keep_scheduling():
-                return bool(state_change)
+                return False
```
zmedico, line 1958: If it added a package in the previous iteration of the loop, then that means the state changed, and it's supposed to return True so that the calling _schedule_tasks loop will continue to reconcile the state until there are no changes.

Flowdalic: I see, that is a good reason to keep it. I've dropped the according commit.

lib/_emerge/Scheduler.py
Flowdalic: I see, that is a good reason to keep it. I've dropped the according commit.

lib/_emerge/Scheduler.py
Flowdalic: I feel bad that I missed that. Modified the commit to simply convert state_change to boolean (which I like better, but it is probably subjective).

Flowdalic:
@zmedico Thank you very much for your review. I've updated the PR.

PR #1327, opened 2024-05-21
install-qa-checks.d: suppress some gnulib implicit decls on musl

eli-schwartz:
These happen in tons of GNU packages because of using gnulib's regex.m4 specifically, which pulls in a macro that checks for some functionality and spit out many implicit function declaration errors if regex.h isn't GNU's specifically.

The compile tests do fail either way, it's just very dirty in the logs.

Bug: https://bugs.gentoo.org/906027

Followup to 8256473c6a9fa93e7cf81c46fa920cd522507c21 / https://github.com/gentoo/portage/pull/1323

### DIFF
--- a/bin/install-qa-check.d/90config-impl-decl
+++ b/bin/install-qa-check.d/90config-impl-decl
@@ -58,6 +58,18 @@ add_default_skips() {
 		# also gnulib, but checks both linux/non-linux headers
 		MIN
 	)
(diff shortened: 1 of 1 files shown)


floppym (approved):
I have gotten a few bug reports from ago about this personally, so this would be a welcome change.

thesamesam (approved):

PR #1329, opened 2024-05-25
binrepos.conf: Support "frozen" attribute

zmedico:
In order to allow consistent and reproducible dependency calculations during mixed source and binary updates, add a "frozen" binrepos.conf attribute which will freeze binrepo index updates and cause messsages to indicate that the repo is frozen rather than up-to-date:

    Local copy of remote index is frozen and will be used.

This should only be set temporarily in order to guarantee consistent and reproducible dependency calculations for mixed binary and source updates.

Bug: https://bugs.gentoo.org/932739

### DIFF
--- a/lib/portage/binrepo/config.py
+++ b/lib/portage/binrepo/config.py
@@ -1,4 +1,4 @@
-# Copyright 2020 Gentoo Authors
+# Copyright 2020-2024 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 from collections import OrderedDict
@@ -12,6 +12,7 @@ from portage.util.configparser import SafeConfigParser, ConfigParserError, read_
 
 class BinRepoConfig:
     __slots__ = (
+        "frozen",
         "name",
         "name_fallback",
         "fetchcommand",
@@ -19,6 +20,7 @@ class BinRepoConfig:
         "resumecommand",
         "sync_uri",
     )
+    _bool_opts = ("frozen",)
 
     def __init__(self, opts):
         """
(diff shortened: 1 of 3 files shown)


thesamesam (commented):
Can you give me a scenario where you'd use it? I can see it in theory but I can't yet see a time when I'd want to set it.

Or is the idea that we'll have some tooling for the other feature set it temporarily?

zmedico:
> Can you give me a scenario where you'd use it? I can see it in theory but I can't yet see a time when I'd want to set it.
> 
> Or is the idea that we'll have some tooling for the other feature set it temporarily?

When performing mixed source / binary updates, there's a desire to have the binary package in sync with the source tree, which [REPO_REVISIONS](https://github.com/gentoo/portage/pull/1306) will make possible. However, once you have your source repository in sync with you binrepos, it's highly desirable for them to stay in sync until you've completed your current batch of mixed source / binary updates. Otherwise, the binrepos state could start to diverge, leading to inconsistent and dependency calculations which are not reproducible.

thesamesam:
Ah! Got it, thanks. It's something which has bothered me for ages (how you end up mutating state just by even e.g. `emerge -p`) but I never sat down and thought about. LGTM.

zmedico:
> Ah! Got it, thanks. It's something which has bothered me for ages (how you end up mutating state just by even e.g. `emerge -p`) but I never sat down and thought about. LGTM.

I've recently switched to mixed source / binary updates, when I saw how [Amazon Linux used dnf releasever](https://github.com/gentoo/portage/pull/1306#issuecomment-2130434252) to hold back binary updates, I realized that we would need to hold back binary updates ourselves. For us, a frozen binrepo serves as a convenient alternative to a releasever (and [REPO_REVISIONS](https://github.com/gentoo/portage/pull/1306) will link a frozen binrepo to corresponding source repo state).

PR #1332, opened 2024-05-26
tests: add testcase for app-arch/tar merge order with binpkgs

thesamesam:
tests: add testcase for app-arch/tar merge order with binpkgs

In the bug, dilfridge reports releng@ is hitting the following:
```
$ emerge -epvk world | grep -E "(app-arch/tar|sys-apps/acl)"
[ebuild  N     ] sys-apps/acl-2.3.2-r1::gentoo  USE="nls -static-libs" 363 KiB
[binary   R    ] app-arch/tar-1.35-1::gentoo  USE="acl* nls* xattr* -minimal (-selinux) -verify-sig" 0 KiB
[...]
```

Test for bug #922629 where binary app-arch/tar[acl] was merged
before its dependency sys-apps/acl (with virtual/acl merged but
unsatisfied).

It (appears to be) a bad interaction with @system containing app-alternatives/tar
plus a circular dependency on app-arch/tar. The USE change is also important,
as e.g. dropping "sys-apps/attr nls" from package.use makes things okay.

XFAIL'd for now.

Bug: https://bugs.gentoo.org/922629

### DIFF
--- a/lib/portage/tests/resolver/meson.build
+++ b/lib/portage/tests/resolver/meson.build
@@ -81,6 +81,7 @@ py.install_sources(
         'test_slot_operator_update_probe_parent_downgrade.py',
         'test_solve_non_slot_operator_slot_conflicts.py',
         'test_targetroot.py',
+        'test_tar_merge_order.py',
         'test_unmerge_order.py',
         'test_unnecessary_slot_upgrade.py',
         'test_update.py',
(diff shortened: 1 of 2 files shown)


zmedico (approved):
Let's go ahead and revert 49e01d041c74680a81860b819daff812d83df02f.

Later we can try to optimize it again but without breaking this test case.

thesamesam:
Thanks!

PR #1334, opened 2024-05-27
atomic_ofstream: Use mkstemp rather than getpid (pid namespace safety)

zmedico:
Bug: https://bugs.gentoo.org/851015


thesamesam (approved):
Thanks!

PR #1338, opened 2024-06-02
tar_safe_extract: Use tarfile.fully_trusted_filter

zmedico:
This suppresses a DeprecationWarning triggered because the tarfile.data_filter will become the new default in python3.14. The fully_trusted filter should be suitable here because tar_safe_extract already performs security validation on tar members prior to extraction.

Bug: https://bugs.gentoo.org/933433

@syu-nya

### DIFF
--- a/lib/portage/gpkg.py
+++ b/lib/portage/gpkg.py
@@ -628,6 +628,15 @@ class tar_safe_extract:
         if self.closed:
             raise OSError("Tar file is closed.")
         temp_dir = tempfile.TemporaryDirectory(dir=dest_dir)
+        # The below tar member security checks can be refactored as a filter function
+        # that raises an exception. Use tarfile.fully_trusted_filter for now, which
+        # is simply an identity function:
+        # def fully_trusted_filter(member, dest_path):
+        #     return member
+        try:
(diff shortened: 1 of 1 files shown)


lib/portage/gpkg.py
```diff
@@ -628,6 +628,10 @@ def extractall(self, dest_dir: str):
         if self.closed:
             raise OSError("Tar file is closed.")
         temp_dir = tempfile.TemporaryDirectory(dir=dest_dir)
+        try:
+            self.tar.extraction_filter = tarfile.fully_trusted_filter
+        except AttributeError:
+            pass
```
zmedico, line 634: We could refactor the security checks into a filter function that raises exceptions for violations. The `fully_trusted_filter` is simply an identity function like this:
```python
def fully_trusted_filter(member, dest_path):
    return member
```

thesamesam: Yeah, that might be a nice idea and worth a TODO comment. Not a blocker though obviously.

thesamesam (approved):

lib/portage/gpkg.py
zmedico: Comment added, thanks!

PR #1339, opened 2024-06-03
Add binpkgs info to mtimedb resume data

zmedico:
In order to fix emerge --resume and --keep-going to make appropriate binary package selections, store a list of binpkgs in the resume data. By adding the data as a new key which older versions of portage will ignore, the extension is backward compatible.

Without this fix, emerge --resume and --keep-going make poor package selection choices which do not account for the --binpkg-respect-use option.

Bug: https://bugs.gentoo.org/933442


thesamesam (approved):
Thanks! I haven't verified that older Portage ignores it, but it looks like it should.

PR #1342, opened 2024-06-13
bin/ebuild.sh: disable globskipdots in Bash 5.2

thesamesam:
See 69cac73ba0a7bcf2e2cff88c60d389895a550623.

globskipdots is a new option in bash-5.2 that is also default-on in that release. The default value is not gated by BASH_COMPAT (see bug #907061), hence we need to disable it for older Bashes to avoid behaviour changes in ebuilds and eclasses.

Bug: https://bugs.gentoo.org/907061

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -14,6 +14,12 @@ Features:
   as well as / for all EAPIs rather than instead of / for EAPI 6 and below
   (bug #435066).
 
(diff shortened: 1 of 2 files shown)


thesamesam:
Not tested yet.

PR #1343, opened 2024-06-13
output.py: support title setting via curses/terminfo

gerstner-hub:
To support terminal title setting for a broader range of terminals, implement curses/terminfo support for setting terminal titles, instead of relying only on the hard coded list of supported $TERM values.

### DIFF
--- a/lib/portage/output.py
+++ b/lib/portage/output.py
@@ -268,20 +268,65 @@ def nc_len(mystr):
 _legal_terms_re = re.compile(
     r"^(xterm|xterm-color|Eterm|aterm|rxvt|screen|kterm|rxvt-unicode|gnome|interix|tmux|st-256color|alacritty|konsole|foot)"
 )
-_disable_xtermTitle = None
-_max_xtermTitle_len = 253
+_xterm_title_supported = None
+_max_xterm_title_len = 253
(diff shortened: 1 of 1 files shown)


lib/portage/output.py
```diff
@@ -268,26 +268,67 @@ def nc_len(mystr):
...
+
+    _xterm_title_supported = False
+
+    if not sys.__stderr__.isatty() or "TERM" not in os.environ:
+        return
+
+    try:
+        # by default check if we can dynamically query the proper title
+        # setting sequence via terminfo
+        curses = init_curses(sys.stderr)
```
zmedico, line 295: We should be consistent in our usage of `sys.__stderr__` vs. `sys.stderr`, and `sys.__stderr__` is preferable because sometimes we override `sys.stderr` with a temporary `io.StringIO` instance.

zmedico: Maybe it's better if the fd argument defaults to `sys.__stderr__` in init_curses, so we'll only have to explicitly pass it in for an unusual case.

gerstner-hub: Right, I've pushed a commit on top to address this.

zmedico (approved):
LGTM, thanks!

PR #1344, opened 2024-06-13
Allow GIL to be disabled in whirlpool C extension

zmedico:
In 3.13 python extensions need to [declare support for GIL features](https://github.com/python/cpython/issues/116322), for example if they don't declare Py_MOD_GIL_NOT_USED then it will cause the GIL to be enabled even when python was launched in free-threaded mode.

This requires "multi-phase initialization" because `Py_mod_create` is incompatible with `m_slots`. There's a `PyUnstable_Module_SetGIL()` function that can be used with single-phase init, but it's an unstable API, so it's better to use multi-phase init. There's no need to use `PyModule_GetState()` because the whirlpool module has no mutable state.

Bug: https://bugs.gentoo.org/934220

### DIFF
--- a/src/portage_util__whirlpool.c
+++ b/src/portage_util__whirlpool.c
@@ -1112,30 +1112,42 @@ static PyTypeObject WhirlpoolType = {
     .tp_methods = Whirlpool_methods,
 };
 
+static int
+whirlpool_exec(PyObject *mod)
+{
+    if (PyType_Ready(&WhirlpoolType) < 0)
+        return -1;
+
+    Py_INCREF(&WhirlpoolType);
+    if (PyModule_AddObject(mod, "Whirlpool", (PyObject *) &WhirlpoolType) < 0) {
+        Py_DECREF(&WhirlpoolType);
+        return -1;
+    }
+
+    return 0;
+}
+
+static PyModuleDef_Slot _whirlpoolmodule_slots[] = {
+    {Py_mod_exec, whirlpool_exec},
+#ifdef Py_MOD_PER_INTERPRETER_GIL_SUPPORTED
+    {Py_mod_multiple_interpreters, Py_MOD_PER_INTERPRETER_GIL_SUPPORTED},
+#endif
+#ifdef Py_MOD_GIL_NOT_USED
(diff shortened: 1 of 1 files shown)


thesamesam:
Thanks for looking at this already. I haven't dared try -gil yet...

src/portage_util__whirlpool.c
```diff
@@ -1112,30 +1112,42 @@ static PyTypeObject WhirlpoolType = {
...
 };
 
+static int
+whirlpool_exec(PyObject *mod)
+{
+    if (PyType_Ready(&WhirlpoolType) < 0)
+        return -1;
+
+    Py_INCREF(&WhirlpoolType);
+    if (PyModule_AddObject(mod, "Whirlpool", (PyObject *) &WhirlpoolType) < 0) {
```
zmedico, line 1122: This passes tests but I'm not sure if it's really correct. I've seen other modules like https://github.com/python/cpython/blob/main/Modules/termios.c store all of their state in a struct.

zmedico: The docs here indicate that state should be specific to the module object (e.g. using PyModule_GetState()):

https://docs.python.org/3.13/c-api/module.html#multi-phase-initialization

zmedico: I think this is okay because the "state" is WhirlpoolType an it's immutable.

floppym:
Having just given myself a crash course on "multi-phase initialization", this looks ok to me.

zmedico:
> Having just given myself a crash course on "multi-phase initialization", this looks ok to me.

I've found that there's a `PyUnstable_Module_SetGIL()` function we can use with single-phase init modules, but it's an unstable API so it seems better to migrate to multiphase init:

https://docs.python.org/3.13/c-api/module.html#c.PyUnstable_Module_SetGIL

zmedico:
> Thanks for looking at this already. I haven't dared try -gil yet...

I'm running emerge with `PYTHON_GIL=0 EPYTHON=python3.13` now, and I haven't noticed any issues. I suspect all of our code is safe, so I don't expect to see an issue unless it is in python itself.

thesamesam (approved):

PR #1350, opened 2024-06-19
vartree, movefile: Warn when rewriting a symlink

ulm:
See PMS section 13.4.1 (Rewriting):
Any absolute symlink whose link starts with D must be rewritten with the leading D removed. The package manager should issue a notice when doing this.

Bug: https://bugs.gentoo.org/934514

### DIFF
--- a/lib/portage/dbapi/vartree.py
+++ b/lib/portage/dbapi/vartree.py
@@ -5563,6 +5563,16 @@ class dblink:
                 myabsto = myabsto.lstrip(sep)
                 if self.settings and self.settings["D"]:
                     if myto.startswith(self.settings["D"]):
+                        self._eqawarn(
+                            "preinst",
+                            [
+                                _(
+                                    "QA Notice: Absolute symlink %s points to %s inside the image directory.\n"
+                                    "Removing the leading %s from its path."
+                                )
+                                % (mydest, myto, self.settings["D"])
+                            ],
+                        )
                         myto = myto[len(self.settings["D"]) - 1 :]
                 # myrealto contains the path of the real file to which this symlink points.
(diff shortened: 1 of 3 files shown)


lib/portage/dbapi/vartree.py
```diff
@@ -5563,6 +5563,16 @@ def mergeme(
                 myabsto = myabsto.lstrip(sep)
                 if self.settings and self.settings["D"]:
                     if myto.startswith(self.settings["D"]):
+                        self._eqawarn(
+                            "preinst",
+                            [
+                                _(
+                                    "QA Notice: Symlink %s points to %s inside the image directory.\n"
```
Flowdalic, line 5570: Could we maybe clarify this a little bit? Somehow most (all?) symlinks point to somewhere inside the image directory. The problem are symlinks using parts of the image directory's path. So maybe

```suggestion
                                    "QA Notice: Symlink %s uses path of image directory to point to %s.\n"
```

lib/portage/util/movefile.py
```diff
@@ -210,6 +210,11 @@ def movefile(
         try:
             target = os.readlink(src)
             if mysettings and "D" in mysettings and target.startswith(mysettings["D"]):
+                writemsg(
+                    f"!!! {_('Symlink points inside image directory.')}\n",
```
Flowdalic, line 214:
```suggestion
                    f"!!! {_('Symlink based on path of image directory.')}\n",
```

lib/portage/dbapi/vartree.py
ulm: I've replaced "symlink" by "absolute symlink" which I think should clarify it.

floppym (approved):

thesamesam (approved):

PR #1351, opened 2024-06-19
Add emerge jobs tmpdir blocks and files threshold options

zmedico:
This supersedes https://github.com/gentoo/portage/pull/1345 and distinguishes between both used blocks and used files (inodes).

--jobs-tmpdir-blocks-threshold[=RATIO]

  Specifies the maximum ratio of used blocks allowed (a
  floating-point number between 0.0 and 1.0) in PORTAGE_TMPDIR
  when starting a new job. With no argument, removes a previous
  blocks ratio threshold. For example, use a ratio of 0.50 to stop
  starting new jobs when the blocks usage in PORTAGE_TMPDIR
  exceeds 50%. This option conflicts with FEATURES="keepwork".
  WARNING: Since the job scheduler is unable to predict the future
  consumption of jobs that it has scheduled, users are advised to
  set a threshold that provides a significant amount of headroom,
  in order to decrease the probability that jobs will fail due
  to ENOSPC errors.

--jobs-tmpdir-files-threshold[=RATIO]

  Specifies the maximum ratio of used files (inodes) allowed (a
  floating-point number between 0.0 and 1.0) in PORTAGE_TMPDIR
  when starting a new job. With no argument, removes a previous
  files ratio threshold. For example, use a ratio of 0.50 to stop
  starting new jobs when the files usage in PORTAGE_TMPDIR
  exceeds 50%. This option conflicts with FEATURES="keepwork".
  WARNING: Since the job scheduler is unable to predict the future
  consumption of jobs that it has scheduled, users are advised to
  set a threshold that provides a significant amount of headroom,
  in order to decrease the probability that jobs will fail due
  to ENOSPC errors.

Bug: https://bugs.gentoo.org/934382

### DIFF
--- a/lib/_emerge/Scheduler.py
+++ b/lib/_emerge/Scheduler.py
@@ -26,6 +26,7 @@ from portage._sets import SETPREFIX
 from portage._sets.base import InternalPackageSet
 from portage.util import ensure_dirs, writemsg, writemsg_level
 from portage.util.futures import asyncio
+from portage.util.path import first_existing
 from portage.util.SlotObject import SlotObject
 from portage.util._async.SchedulerInterface import SchedulerInterface
 from portage.package.ebuild.digestcheck import digestcheck
@@ -64,7 +65,7 @@ FAILURE = 1
 
 
 class Scheduler(PollScheduler):
-    # max time between loadavg checks (seconds)
+    # max time between loadavg and tmpdir statvfs checks (seconds)
     _loadavg_latency = 30
 
     # max time between display status updates (seconds)
@@ -229,6 +230,10 @@ class Scheduler(PollScheduler):
         if max_jobs is None:
             max_jobs = 1
         self._set_max_jobs(max_jobs)
+        self._jobs_tmpdir_blocks_threshold = myopts.get(
+            "--jobs-tmpdir-blocks-threshold"
+        )
+        self._jobs_tmpdir_files_threshold = myopts.get("--jobs-tmpdir-files-threshold")
         self._running_root = trees[trees._running_eroot]["root_config"]
         self.edebug = 0
         if settings.get("PORTAGE_DEBUG", "") == "1":
@@ -1573,7 +1578,11 @@ class Scheduler(PollScheduler):
         self._main_exit = self._event_loop.create_future()
 
         if (
-            self._max_load is not None
+            (
+                self._max_load is not None
+                or self._jobs_tmpdir_blocks_threshold is not None
+                or self._jobs_tmpdir_files_threshold is not None
+            )
             and self._loadavg_latency is not None
             and (self._max_jobs is True or self._max_jobs > 1)
         ):
@@ -1791,6 +1800,53 @@ class Scheduler(PollScheduler):
 
     def _running_job_count(self):
         return self._jobs
+
+    def _can_add_job(self):
+        if not super()._can_add_job():
+            return False
+
+        running_job_count = self._running_job_count()
+        if running_job_count == 0 and not self._merge_wait_queue:
+            # Ensure there is forward progress if there are no running
+            # jobs and no jobs in the _merge_wait_queue.
+            return True
+
+        if (
+            self._jobs_tmpdir_blocks_threshold is not None
+            or self._jobs_tmpdir_files_threshold is not None
+        ) and hasattr(os, "statvfs"):
+            tmpdirs = set()
+            for root in self.trees:
+                settings = self.trees[root]["root_config"].settings
+                if settings["PORTAGE_TMPDIR"] in tmpdirs:
+                    continue
+                tmpdirs.add(settings["PORTAGE_TMPDIR"])
+                tmpdir = first_existing(
+                    os.path.join(settings["PORTAGE_TMPDIR"], "portage")
+                )
+                try:
+                    vfs_stat = os.statvfs(tmpdir)
+                except OSError as e:
+                    writemsg_level(
+                        f"!!! statvfs('{tmpdir}'): {e}\n",
+                        noiselevel=-1,
+                        level=logging.ERROR,
+                    )
+                else:
+                    if (
+                        self._jobs_tmpdir_blocks_threshold is not None
+                        and (vfs_stat.f_blocks - vfs_stat.f_bavail) / vfs_stat.f_blocks
+                    ) >= self._jobs_tmpdir_blocks_threshold:
+                        return False
+                    if (
+                        self._jobs_tmpdir_files_threshold is not None
+                        and vfs_stat.f_files > 0
+                        and ((vfs_stat.f_files - vfs_stat.f_favail) / vfs_stat.f_files)
+                        >= self._jobs_tmpdir_files_threshold
+                    ):
+                        return False
+
+        return True
 
     def _schedule_tasks(self):
         while True:
--- a/lib/_emerge/main.py
+++ b/lib/_emerge/main.py
@@ -1,4 +1,4 @@
-# Copyright 1999-2023 Gentoo Authors
+# Copyright 1999-2024 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 import argparse
@@ -165,6 +165,8 @@ def insert_optional_args(args):
         "--getbinpkgonly": y_or_n,
         "--ignore-world": y_or_n,
         "--jobs": valid_integers,
+        "--jobs-tmpdir-blocks-threshold": valid_floats,
+        "--jobs-tmpdir-files-threshold": valid_floats,
         "--keep-going": y_or_n,
         "--load-average": valid_floats,
         "--onlydeps-with-ideps": y_or_n,
@@ -521,6 +523,14 @@ def parse_opts(tmpcmdline, silent=False):
         "--jobs": {
             "shortopt": "-j",
             "help": "Specifies the number of packages to build " + "simultaneously.",
+            "action": "store",
+        },
+        "--jobs-tmpdir-blocks-threshold": {
+            "help": "Specifies maximum used blocks ratio when starting a new job.",
+            "action": "store",
+        },
+        "--jobs-tmpdir-files-threshold": {
+            "help": "Specifies maximum used files ratio when starting a new job.",
             "action": "store",
         },
         "--keep-going": {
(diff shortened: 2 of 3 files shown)


lib/_emerge/main.py
```diff
@@ -1304,6 +1350,27 @@ def emerge_main(args: Optional[list[str]] = None):
     emerge_config.action, emerge_config.opts, emerge_config.args = parse_opts(
         tmpcmdline
     )
+    if (
+        "--jobs-tmpdir-blocks-threshold" in emerge_config.opts
+        and "keep-work" in emerge_config.running_config.settings.features
+    ):
+        writemsg_level(
+            "--jobs-tmpdir-blocks-threshold conflicts with FEATURES=keep-work\n",
```
thesamesam, line 1358:
```suggestion
            "--jobs-tmpdir-blocks-threshold conflicts with FEATURES=keepwork\n",
```
(and elsewhere)

thesamesam (approved):
Needs (functional) typo fixing but otherwise looks good, thanks! Hopefully @akhuettel can test it though to see if it works well for him.

lib/_emerge/main.py
zmedico: Fixed, thanks!

lib/_emerge/Scheduler.py
```diff
@@ -1792,6 +1801,41 @@ def _is_work_scheduled(self):
...
+    def _can_add_job(self):
+        if not super()._can_add_job():
+            return False
+
+        if (
+            self._jobs_tmpdir_blocks_threshold is not None
+            or self._jobs_tmpdir_files_threshold is not None
+        ) and hasattr(os, "statvfs"):
+            tmpdir = first_existing(
+                os.path.join(self.settings["PORTAGE_TMPDIR"], "portage")
```
zmedico, line 1813: It's possible to have a separate PORTAGE_TMPDIR per PORTAGE_CONFIGROOT, so we could possibly handle multiple ROOTs here. Obviously PORTAGE_TMPDIR overrides via package.env are also completely neglected here.

zmedico: Added a loop to cover multiple ROOTs, with deduplication of identical PORTAGE_TMPDIR values.

Flowdalic (commented):
I still think that an absolute threshold is more sensible than a relative threshold. And considering the fact that it would be good to minimize the number of configuration options of Portage, we may want to start with an absolute threshold. A relative one can always be added later on, if the need arises.

It is also probably sufficient that users are only able to specify the absolute value in GiB, so we do not need to parse bytes in a human readable format.

Therefore, I believe instead of
`--jobs-tmdir-blocks-threshold`
we should start with
`--jobs-tmpdir-require-free-gb`
(or so)

Again, just my personal opinion. While I believe what I suggest is a small change resulting in a relevant improvement, this does not constitute in any shape or form me blocking this PR in its current form.

zmedico:
> I still think that an absolute threshold is more sensible than a relative threshold. And considering the fact that it would be good to minimize the number of configuration options of Portage, we may want to start with an absolute threshold. A relative one can always be added later on, if the need arises.
> 
> It is also probably sufficient that users are only able to specify the absolute value in GiB, so we do not need to parse bytes in a human readable format.
> 
> Therefore, I believe instead of `--jobs-tmdir-blocks-threshold` we should start with `--jobs-tmpdir-require-free-gb` (or so)

I'm afraid this `--jobs-tmpdir-require-free-gb` option might give users a false sense of security, thinking the the last scheduled job will have at least this amount of space reserved.

When emerge starts multiple jobs around the same time, the corresponding space usage does not manifest itself until later in time, after the scheduling decisions were already made. We can try to mitigate this issue by adding a long delay in between starting jobs, but this strategy could slow things down unnecessarily.

One way to deal with this kind of issue is to set a threshold that requires plenty of head room, like a ratio of 0.5 or so. An option like `--jobs-tmpdir-require-free-gb` makes less sense when we consider that space usage can be difficult to predict.

Flowdalic:
> I'm afraid this `--jobs-tmpdir-require-free-gb` option might give users a false sense of security, thinking the last scheduled job will have at least this amount of space reserved.

That is, broadly speaking, also true for `--jobs-tmdir-blocks-threshold 0.85`, which could give the impression that  the last scheduled jobs will have 15% of space reserved.

Unrelated if the setting denotes an absolute or relative value, one way to mitigate this is to make this a setting scale with the number of emerge jobs. For example, assume `--jobs 3`, then we could simply multiply required free space by three.<sup>1</sup>

And this is a example why an absolute threshold is preferable over a relative one. Assume an analysis of ::gentoo showed that the worst-case `PORTAGE_TMPDIR` consumption per package is 6 GiB. Then we could simply set the default of `--jobs-tmpdir-require-free-gb` to 6. But you can't hardcode such an value in portage which `--jobs-tmdir-blocks-threshold`. You can only dynamically calculate it one you know `PORTAGE_TMPDIR`.

The gist is, a good threshold is derived from our packages, not from the size of the filesystem that holds `PORTAGE_TMPDIR`. Therefore, a setting that specifies an absolute value is preferable over one that specifies a relative one.

Again, not blocking this PR. However, I strongly believe a relative threshold is not sensible for the reasons explained above. And, if it helps to avoid a relative threshold, then I am happy to work on `--jobs-tmpdir-require-free-gb`.

1: Instead of linearly scaling the free-space requirements with the number of emerge jobs, we could also use some more elaborate scheme using a decay function if we deem this sensible (which I believe it is).

lib/_emerge/Scheduler.py
```diff
@@ -1792,6 +1801,47 @@ def _is_work_scheduled(self):
     def _running_job_count(self):
         return self._jobs
 
+    def _can_add_job(self):
+        if not super()._can_add_job():
+            return False
+
```
Flowdalic, line 1807: I think this requires a check that returns true if there are no running jobs, as otherwise we will never add a job but just stall indefinitely.

zmedico: I agree that it's a useful property to always allow at least one job, however there could be an edge case where this allows a job to start at an inappropriate time, like just before it's supposed to start processing packages from _merge_wait_queue which only happens when the running job count is zero.

I think it will probably be fine because _schedule_tasks processes _merge_wait_queue before it calls _schedule_tasks_imp to start new jobs, but we need to be careful and should add a comment about this scheduling order requirement in _schedule_tasks.

zmedico:
> > I'm afraid this `--jobs-tmpdir-require-free-gb` option might give users a false sense of security, thinking the last scheduled job will have at least this amount of space reserved.
> 
> That is, broadly speaking, also true for `--jobs-tmdir-blocks-threshold 0.85`, which could give the impression that the last scheduled jobs will have 15% of space reserved.

Yes, so now I'd like to fix the documentation to explain the difficulties in predicting space usage, and suggest a safer ratio that provides more headroom.

> Again, not blocking this PR. However, I strongly believe a relative threshold is not sensible for the reasons explained above. And, if it helps to avoid a relative threshold, then I am happy to work on `--jobs-tmpdir-require-free-gb`.
> 
> 1: Instead of linearly scaling the free-space requirements with the number of emerge jobs, we could also use some more elaborate scheme using a decay function if we deem this sensible (which I believe it is).

Yes, please do work on `--jobs-tmpdir-require-free-gb`. However, it seems clear that it will introduce some significant complexities (decay functions and so on), and I think a relative threshold is a reasonable way to keep things simple for the first iteration of this feature.

lib/_emerge/Scheduler.py
Flowdalic: > I agree that it's a useful property

I'ts not only a useful property, something like this is required to prevent portage from hanging if the tmpdir space requirement is not fulfilled. Try it for yourself

```
$ git checkout 0f2a0b2b5a8456821747a698cd98fb15228fd37f # This PRs commit
$ sudo ./bin/emerge --jobs-tmpdir-blocks-threshold 0.01 --jobs 3 -v app-misc/hello app-text/pdfjam 
# hangs
```

I currently use

https://github.com/gentoo/portage/pull/1353/files#diff-e07339ce877e9437fcd80afb3da63e264d5ae26a50dd27d6bab58a6019ec1b73R1815-R1819

however, we could alternatively bail with an error if the tmpdir resource requirements are not meet and there _merge_wait_queue is empty.

zmedico: Okay, I've copied your code:
```python
        running_job_count = self._running_job_count()
        if running_job_count == 0 and not self._merge_wait_queue:
            # Ensure there is forward progress if there are no running
            # jobs and no jobs in the _merge_wait_queue.
            return True
```

Flowdalic:
> However, it seems clear that it will introduce some significant complexities (decay functions and so on)

I don't think that this is how the situation presents. There is no additional complexity in my PR besides the [`scale_to_jobs(num)`](https://github.com/gentoo/portage/pull/1353/files#diff-e07339ce877e9437fcd80afb3da63e264d5ae26a50dd27d6bab58a6019ec1b73R1845) function. Which isn't very complex and attempts to mitigate an issue that exists also in this PR. In other words, it is not specific to an absolute value. We could also drop it, if it is deemed to much complexity for "just" a mitigation.

> and I think a relative threshold is a reasonable way [...] for the first iteration of this feature.

Unfortunately, I'd rather we do not. The my main motivation of authoring what I present in https://github.com/gentoo/portage/pull/1353 is that we do not end up with
- `--jobs-tmpdir-blocks-threshold`
- `--jobs-tmpdir-files-threshold`
- `--jobs-tmpdir-require-free-gb`
- `--jobs-tmpdir-require-free-kilo-inodes`

portage/emerge already has many command line arguments. Therefore I hope we can decide for either the absolute option or the relative one.

Fortunately, both approaches are now implemented, and I've provided the advantages of an absolute value over an relative one in my previous posts, so people can make an informed decision what they feel is best.

Again, not blocking this PR. I just personally believe that we should not go with  `--jobs-tmpdir-blocks-threshold` (and `--jobs-tmpdir-files-threshold`).

zmedico:
> Unfortunately, I'd rather we do not. The my main motivation of authoring what I present in #1353 is that we do not end up with
> 
> * `--jobs-tmpdir-blocks-threshold`
> * `--jobs-tmpdir-files-threshold`
> * `--jobs-tmpdir-require-free-gb`
> * `--jobs-tmpdir-require-free-kilo-inodes`
> 
> portage/emerge already has many command line arguments. Therefore I hope we can decide for either the absolute option or the relative one.

We're in one of those situations where a plugin system is needed in order to satisfy everyone's needs. We can use a single `--jobs-tmpdir` option that can be specified multiple times in order to load multiple plugins, and then prefix the arguments with plugin names like `--jobs-tmpdir=require-free-gb:16 --jobs-tmpdir=require-free-kilo-inodes=32`.

Flowdalic:
> We're in one of those situations where a plugin system is needed in order to satisfy everyone's needs.

We should avoid introducing additional complexity, especially since portage has already plenty of it.

Therefore, if I couldn't establish to you, or anybody else, that the relative-threshold approach of PR isn't ideal and that the absolute-threshold approach I suggest has unique advantages, just merge this PR and not mine. Carrying the semantics of both PRs is unnecessary and just adds even more knobs to portage. If it is decided to go with this PR (not sure how such decisions are made, portage dev team vote?), then please consider adjusting this PR so that emerge also emits a one-time warning if the threshold is it (see my PR). I am also not sure why this PR marks the threshold conflicting with the keepwork feature.

lib/_emerge/main.py
```diff
@@ -1304,6 +1350,27 @@ def emerge_main(args: Optional[list[str]] = None):
     emerge_config.action, emerge_config.opts, emerge_config.args = parse_opts(
         tmpcmdline
     )
+    if (
+        "--jobs-tmpdir-blocks-threshold" in emerge_config.opts
```
thesamesam, line 1354: I suppose we should check for merge-wait too?

lib/_emerge/main.py
```diff
@@ -1033,6 +1043,42 @@ def parse_opts(tmpcmdline, silent=False):
 
         myoptions.jobs = jobs
 
+    if myoptions.jobs_tmpdir_blocks_threshold == "True":
```
thesamesam, line 1046: Just coming back to this: all of this really makes me think we need to revisit the option handling :(

But it's of course fine for now

thesamesam (commented):
I'm already hesitant to add the 2 new options but I feel like they should make most people happy.

lib/_emerge/main.py
zmedico: `True` here indicates a missing argument, and this can mean different things for different options. For example, for --jobs a missing argument means unlimited jobs, and so it's handled like this instead:
```python
        if myoptions.jobs == "True":
            jobs = True
```

lib/_emerge/main.py
zmedico: It's valid to use this option with or without merge-wait because either way the user could desire to  use disk pressure to limit starting of new jobs.

PR #1352, opened 2024-06-20
eapi.py: Use attrs instead of hardcoding EAPIs in functions

ulm:
Adding new attrs as needed. Their name is the same as the corresponding PMS feature label (if one exists).


ulm:
@thesamesam has asked me to file a PR for this.

Not entirely sure what the default should be in the eapi is None case. I've used EAPI 6 for them, because that seems what is done elsewhere.

ulm:
ping

thesamesam (approved):

ulm:
Merged.

PR #1353, opened 2024-06-23
Add emerge options to require free space/inodes in tmpdir

Flowdalic:
Alternative to #1351. Seems to work as intended, but more testing is always welcomed.

This adds
  --jobs-tmpdir-require-free-gb=GB
  --jobs-tmpdir-require-free-kilo-inodes=INODES
as emerge emerge options.

When used with --jobs, makes portage/emerge check that PORTAGE_TMPDIR has sufficient free resources before a new job is started.

Thanks goes out to Zac Medico, as this was inspired by https://github.com/gentoo/portage/pull/1351, with the following differences:

- options are absolute values, not relatives ones
- defaults for both options are specified
- option values are scaled, using a decaying function, considering the number or runnings jobs
- emit a warning once a threshold is reached

Bugs: https://bugs.gentoo.org/934382

This seems to work as intended, but more testing is always welcomed.

### DIFF
--- a/lib/_emerge/Scheduler.py
+++ b/lib/_emerge/Scheduler.py
@@ -26,6 +26,7 @@ from portage._sets import SETPREFIX
 from portage._sets.base import InternalPackageSet
 from portage.util import ensure_dirs, writemsg, writemsg_level
 from portage.util.futures import asyncio
+from portage.util.path import first_existing
 from portage.util.SlotObject import SlotObject
 from portage.util._async.SchedulerInterface import SchedulerInterface
 from portage.package.ebuild.digestcheck import digestcheck
@@ -64,7 +65,7 @@ FAILURE = 1
 
 
 class Scheduler(PollScheduler):
-    # max time between loadavg checks (seconds)
+    # max time between loadavg and tmpdir statvfs checks (seconds)
     _loadavg_latency = 30
 
     # max time between display status updates (seconds)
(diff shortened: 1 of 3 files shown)


lib/_emerge/Scheduler.py
```diff
@@ -1792,6 +1805,96 @@ def _is_work_scheduled(self):
...
+                        return res
+
+                    if (
+                        self._jobs_tmpdir_require_free_gb
+                        and self._jobs_tmpdir_require_free_gb != 0
+                    ):
+                        required_free_bytes = (
+                            self._jobs_tmpdir_require_free_gb * 1024 * 1024 * 1024
+                        )
+                        required_free_bytes = scale_to_jobs(required_free_bytes)
```
zmedico, line 1861: How can we be sure that this scaled required_free_bytes value stays within bounds of the file-system size?

Flowdalic: This is currently not ensured, although we could cap `required_free_bytes` and `required_free_inodes)

However, I am not sure if this is even required. Such a case is not fatal, "just" portage's emerge job parallelism is limited.

zmedico: The scaling is a significant adjustment to the user's input, so we should document it in the man page.

lib/_emerge/Scheduler.py
```diff
@@ -230,6 +231,18 @@ def __init__(
...
+            # dev-lang/rust-1.77.1: ~16 GiB
+            # www-client/chromium-126.0.6478.57: ~18 GiB
+            self._jobs_tmpdir_require_free_gb = 18
+        self._jobs_tmpdir_free_kilo_inodes = myopts.get(
+            "--jobs-tmpdir-require-free-kilo-inodes"
+        )
+        if not self._jobs_tmpdir_free_kilo_inodes:
+            # dev-lang/rust-1.77.1: ~ 450k inodes
+            # www-client/chromium-126.0.6478.57: ~1011K
+            self._jobs_tmpdir_free_kilo_inodes = 1100
```
zmedico, line 245: It looks like you should possibly distinguish between 0 and None so that the user can specify 0 to explicitly disable the reature.

Flowdalic: The `--jobs-tmpdir-require-free-kilo-inodes 0` case is already handled vai https://github.com/gentoo/portage/pull/1353/files#diff-e07339ce877e9437fcd80afb3da63e264d5ae26a50dd27d6bab58a6019ec1b73R1878

lib/_emerge/Scheduler.py
Flowdalic: Documented the scaling behavior in the man page.

PR #1356, opened 2024-07-03
avoid using 'l' as variable

maekke:
as detected by ruff. try to use a more meaningful variable name instead.

### DIFF
--- a/lib/_emerge/Binpkg.py
+++ b/lib/_emerge/Binpkg.py
@@ -123,15 +123,15 @@ class Binpkg(CompositeTask):
                 fetch_log = os.path.join(
                     _emerge.emergelog._emerge_log_dir, "emerge-fetch.log"
                 )
-                msg = (
+                msgs = (
                     "Fetching in the background:",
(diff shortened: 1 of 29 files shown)


floppym:
This feels pretty pointless to me.

ajakk:
Do you propose that some combination of ruff+black be used for Portage? I note that you have some black fixes in this PR, do we know why we're not seeing complaints for those in CI?

eli-schwartz:
The black fixes in this PR aren't actually black fixes for portage. They are the classic case of "my first commit introduced some linter errors, so I added a second commit to fix the first commit instead of using `git commit --amend`".

eli-schwartz:
Overall I agree this seems pointless. Using `i` as an iteration variable in a loop isn't a bad thing, even if you spell it as an "L" instead.

There are varying opinions on this, obviously. I commented about the general concept on the ruff issue tracker, in fact. :D https://github.com/astral-sh/ruff/issues/4866#issuecomment-1579651071

PR #1357, opened 2024-07-04
Do not export A

Flowdalic:
Instead of passing A as part of the process environment, we pass it via a file. Since A is usually the greatest contributor to the process environment, removing it from the process environment significantly helps running into MAX_ARG_STRLEN when spawning a new child process.

This means that A is now unexported in the ebuild. However, A is mostly used as part of the default_src_unpack function. And there A does not need to be exported.

Proof-of-concept for https://bugs.gentoo.org/721088

Bug: https://bugs.gentoo.org/721088

### DIFF
--- a/bin/ebuild.sh
+++ b/bin/ebuild.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2024 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 # Prevent aliases from causing portage to act inappropriately.
@@ -9,6 +9,11 @@ unalias -a
 # Make sure this isn't exported to scripts we execute.
 unset BASH_COMPAT
 declare -F ___in_portage_iuse >/dev/null && export -n -f ___in_portage_iuse
+
+if [[ -v PORTAGE_EBUILD_EXTRA_SOURCE ]]; then
+	source "${PORTAGE_EBUILD_EXTRA_SOURCE}" || exit 1
+	unset PORTAGE_EBUILD_EXTRA_SOURCE
+fi
 
 source "${PORTAGE_BIN_PATH}/isolated-functions.sh" || exit 1
 
--- a/lib/portage/package/ebuild/doebuild.py
+++ b/lib/portage/package/ebuild/doebuild.py
@@ -82,6 +82,7 @@ from portage.dep import (
 )
 from portage.dep.libc import find_libc_deps
 from portage.eapi import (
+    eapi_exports_A,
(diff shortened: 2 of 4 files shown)


mgorny:
Why do you need to pass it via a file? It shouldn't be used by any external process, so it should be sufficient to leave it as a non-exported bash variable.

Flowdalic:
> It shouldn't be used by any external process, so it should be sufficient to leave it as a non-exported bash variable.

I assumed that this is not possible. The `A` variable can't be part of the `env` dict that is passed to `spawn()`, as otherwise it would be part of the process environment and therefore subject to `MAX_ARG_STRLEN`.

I'd love to pass the variable directly, without a file. But I don't see how this is possible. What am I missing?

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2131,9 +2131,35 @@ def spawn(
...
+    )
+
+    if dont_export_a:
+        orig_env = mysettings.environ()
+        # Copy since we are potentially removing keys from the dict.
+        env = orig_env.copy()
+        ebuildExtraSource = os.path.join(env["T"], "portage-ebuild-extra-source")
+        with open(ebuildExtraSource, mode="w") as f:
+            for name, value in orig_env.items():
+                if name != "A":
```
Flowdalic, line 2151: I realize that this is not optimal, but that's a leftover from a previous implementation where everything was passed via file and not via `env`. This should be changed once this PR leaves the PoC stage.

mgorny:
Perhaps we could write it straight into `environment` that gets sourced by bash-side of Portage anyway.

Flowdalic:
> Perhaps we could write it straight into `environment` that gets sourced by bash-side of Portage anyway.

I've looked into this and I believe the resulting code would be more convoluted that we we have right now.

bin/ebuild.sh
```diff
@@ -10,6 +10,10 @@ unalias -a
 unset BASH_COMPAT
 declare -F ___in_portage_iuse >/dev/null && export -n -f ___in_portage_iuse
 
+if [[ -v PORTAGE_EBUILD_EXTRA_SOURCE ]]; then
+	source "${PORTAGE_EBUILD_EXTRA_SOURCE}" || exit 1
+fi
+
```
zmedico, line 16: Since this variable doesn't need to stay in the environment, we could unset it here, or later in bin/save-ebuild-env.sh.

Flowdalic: Sure, sounds sensible.

Flowdalic:
Closing, since the (better) alternative https://github.com/gentoo/portage/pull/1407 was merged.

PR #1361, opened 2024-07-22
_EbuildFetcherProcess: Handle SIGTERM

zmedico:
Fix _EbuildFetcherProcess to handle SIGTERM, so that FETCHCOMMAND processes will not be left running in the background:

* Convert the fetch function to an async_fetch coroutine function so that it can use asyncio.CancelledError handlers to terminate running processes.

* Use multiprocessing.active_children() to detect and terminate any processes that asyncio.CancelledError handlers did not have an opportunity to terminate because the exception arrived too soon after fork/spawn.

* Add unit test to verify that a child process is correctly killed when EbuildFetcher is cancelled, with short timeout in case it takes some time for the process to disappear.

Bug: https://bugs.gentoo.org/936273

### DIFF
--- a/lib/_emerge/EbuildFetcher.py
+++ b/lib/_emerge/EbuildFetcher.py
@@ -4,6 +4,8 @@
 import copy
 import functools
 import io
+import multiprocessing
+import signal
 import sys
 
 import portage
@@ -17,11 +19,12 @@ from portage.package.ebuild.fetch import (
     _check_distfile,
     _drop_privs_userfetch,
     _want_userfetch,
-    fetch,
+    async_fetch,
 )
 from portage.util._async.AsyncTaskFuture import AsyncTaskFuture
 from portage.util._async.ForkProcess import ForkProcess
 from portage.util._pty import _create_pty_or_pipe
+from portage.util.futures import asyncio
 from _emerge.CompositeTask import CompositeTask
 
 
@@ -34,6 +37,7 @@ class EbuildFetcher(CompositeTask):
         "logfile",
         "pkg",
         "prefetch",
+        "pre_exec",
         "_fetcher_proc",
     )
 
@@ -253,6 +257,7 @@ class _EbuildFetcherProcess(ForkProcess):
             self._get_manifest(),
             self._uri_map,
             self.fetchonly,
+            self.pre_exec,
         )
         ForkProcess._start(self)
 
@@ -263,7 +268,10 @@ class _EbuildFetcherProcess(ForkProcess):
         self._settings = None
 
     @staticmethod
-    def _target(settings, manifest, uri_map, fetchonly):
+    def _target(settings, manifest, uri_map, fetchonly, pre_exec):
+        if pre_exec is not None:
+            pre_exec()
+
         # Force consistent color output, in case we are capturing fetch
         # output through a normal pipe due to unavailability of ptys.
         portage.output.havecolor = settings.get("NOCOLOR") not in ("yes", "true")
@@ -273,17 +281,53 @@ class _EbuildFetcherProcess(ForkProcess):
         if _want_userfetch(settings):
             _drop_privs_userfetch(settings)
 
-        rval = 1
         allow_missing = manifest.allow_missing or "digest" in settings.features
-        if fetch(
-            uri_map,
-            settings,
-            fetchonly=fetchonly,
-            digests=copy.deepcopy(manifest.getTypeDigests("DIST")),
-            allow_missing_digests=allow_missing,
-        ):
-            rval = os.EX_OK
-        return rval
+
+        async def main():
+            loop = asyncio.get_event_loop()
+            task = asyncio.ensure_future(
+                async_fetch(
+                    uri_map,
+                    settings,
+                    fetchonly=fetchonly,
+                    digests=copy.deepcopy(manifest.getTypeDigests("DIST")),
+                    allow_missing_digests=allow_missing,
+                )
+            )
+
+            def sigterm_handler(signum, _frame):
+                loop.call_soon_threadsafe(task.cancel)
+                signal.signal(signal.SIGTERM, signal.SIG_IGN)
+
+            signal.signal(signal.SIGTERM, sigterm_handler)
+            try:
+                await task
+            except asyncio.CancelledError:
+                # If asyncio.CancelledError arrives too soon after fork/spawn
+                # then handers will not have an opportunity to terminate
+                # the corresponding process, so clean up after this race.
+                for proc in multiprocessing.active_children():
+                    proc.terminate()
+
+                # Use a non-zero timeout only for the first join because
+                # later joins are delayed by the first join.
+                timeout = 0.25
+                for proc in multiprocessing.active_children():
+                    proc.join(timeout)
+                    timeout = 0
+
+                for proc in multiprocessing.active_children():
+                    proc.kill()
+                    # Wait upon the process in order to ensure that its
+                    # pid will trigger ProcessLookupError for tests.
+                    proc.join()
+
+                signal.signal(signal.SIGTERM, signal.SIG_DFL)
+                os.kill(os.getpid(), signal.SIGTERM)
+
+            return os.EX_OK if task.result() else 1
+
+        return asyncio.run(main())
 
     def _get_ebuild_path(self):
         if self.ebuild_path is not None:
--- a/lib/portage/package/ebuild/fetch.py
+++ b/lib/portage/package/ebuild/fetch.py
@@ -1,4 +1,4 @@
-# Copyright 2010-2021 Gentoo Authors
+# Copyright 2010-2024 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 __all__ = ["fetch"]
@@ -73,6 +73,7 @@ from portage.util import (
     writemsg_level,
     writemsg_stdout,
 )
+from portage.util.futures import asyncio
 from portage.process import spawn
 
 _download_suffix = ".__download__"
@@ -111,7 +112,7 @@ def _drop_privs_userfetch(settings):
     """
     spawn_kwargs = dict(_userpriv_spawn_kwargs)
     try:
-        _ensure_distdir(settings, settings["DISTDIR"])
+        asyncio.run(_ensure_distdir(settings, settings["DISTDIR"]))
     except PortageException:
         if not os.path.isdir(settings["DISTDIR"]):
             raise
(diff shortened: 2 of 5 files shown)


lib/portage/tests/ebuild/test_fetch.py
```diff
@@ -36,6 +38,72 @@
...
+        pid = pr.recv()
+        pr.close()
+
+        # Poll the process table until the pid has disappeared,
+        # and fail if a short timeout expires.
+        tries = 10
+        while tries:
+            tries -= 1
+            try:
+                self.assertRaises(ProcessLookupError, os.kill, pid, 0)
```
zmedico, line 76: This passes on my laptop, but not in CI. I wonder if the difference is something about pid 1 needing to reap zombies because the parent didn't wait on the process after it sent SIGTERM.

zmedico: Something is different about the wget behavior, and and apparently we have to follow up with a SIGKILL if SIGTERM does not cause immediate exit.

lib/_emerge/EbuildFetcher.py
```diff
@@ -273,17 +281,43 @@ def _target(settings, manifest, uri_map, fetchonly):
...
+            except asyncio.CancelledError:
+                # If asyncio.CancelledError arrives too soon after fork/spawn
+                # then handers will not have an opportunity to termintate
+                # the corresponding process, so clean up after this race.
+                for proc in multiprocessing.active_children():
+                    proc.terminate()
+                    proc.join(0)
+                    if proc.exitcode is None:
+                        proc.kill()
+                        proc.join(0)
```
Flowdalic, line 314: That `proc.join(0)` right after kill seems unnecessary. I think it should be either dropped, or invoked without any argument.

lib/_emerge/EbuildFetcher.py
```diff
@@ -273,17 +281,43 @@ def _target(settings, manifest, uri_map, fetchonly):
...
+                signal.signal(signal.SIGTERM, signal.SIG_IGN)
+
+            signal.signal(signal.SIGTERM, sigterm_handler)
+            try:
+                await task
+            except asyncio.CancelledError:
+                # If asyncio.CancelledError arrives too soon after fork/spawn
+                # then handers will not have an opportunity to termintate
+                # the corresponding process, so clean up after this race.
+                for proc in multiprocessing.active_children():
```
Flowdalic, line 309: Why not loop multiple times over `active_children()`. Especially first sending SIGTERM to all children, and *then* joining them, gives the childs and the OS time to actually terminate the process correctly.

First time calls terminate, second time calls `join()`, where the first invocation of `join` has set time small timeout value, like 250ms. All others use 0 as explicit timeout value. And with the third and last iteration over `active_children()`, we kill all stragglers.

So this single loop becomes:

```python
for proc in multiprocessing.active_children():
    proc.terminate()

timeout = 0.25
for proc in multiprocessing.active_children():
     proc.join(timeout)
     timeout = 0

for proc in multiprocessing.active_children():
    proc.kill()
```

Flowdalic (commented):
Hadn't time to look over the complete changeset, but the logic terminating the child processes immediately caught my eye.

lib/_emerge/EbuildFetcher.py
zmedico: The intention is to try and ensure that the pid is removed from the process table as soon as possible, since the unit test fails otherwise, and this can only happen if the process waited upon (either by its parent or a pid 1 reaper).

We should be able to omit the argument, since SIGKILL should be completely effective here.

lib/_emerge/EbuildFetcher.py
zmedico: Your version looks great except I'd add `proc.join()` after kill in order to ensure that the pid is reaped ASAP for reasons mentioned in https://github.com/gentoo/portage/pull/1361/files#r1688755132.

lib/_emerge/EbuildFetcher.py
zmedico: Argument removed.

lib/_emerge/EbuildFetcher.py
zmedico: Updated, thanks!

Flowdalic: Thanks, the added code comments make then intention now clear.

lib/portage/package/ebuild/fetch.py
```diff
@@ -201,6 +226,19 @@ def _userpriv_test_write_file(settings, file_path):
...
+        "userfetch" in settings.features
+        and os.getuid() == 0
+        and portage_gid
+        and portage_uid
+        and hasattr(os, "setgroups")
+    ):
+        rval = os.access(os.path.dirname(file_path), os.W_OK)
+        _userpriv_test_write_file_cache[file_path] = rval
+        return rval
+
```
zmedico, line 241: I've triggered an `RuntimeError: This event loop is already running` error here by running `sudo ebuild <ebuild> manifest clean install`:
```
  File "/usr/lib/python-exec/python3.12/ebuild", line 444, in <module>
    main()
  File "/usr/lib/python-exec/python3.12/ebuild", line 405, in main
    a = portage.doebuild(
        ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/portage/package/ebuild/doebuild.py", line 1387, in doebuild
    return not digestgen(mysettings=mysettings, myportdb=mydbapi)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/portage/package/ebuild/digestgen.py", line 161, in digestgen
    if not fetch({myfile: uris}, mysettings):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/portage/package/ebuild/fetch.py", line 835, in fetch
    return asyncio.run(
           ^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/portage/util/futures/_asyncio/__init__.py", line 115, in run
    return _safe_loop().run_until_complete(coro)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/portage/util/_eventloop/asyncio_event_loop.py", line 173, in _run_until_complete
    return self._loop.run_until_complete(future)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/portage/package/ebuild/fetch.py", line 1199, in async_fetch
    _ensure_distdir(mysettings, mysettings["DISTDIR"])
  File "/usr/lib/python3.12/site-packages/portage/package/ebuild/fetch.py", line 281, in _ensure_distdir
    if _userpriv_test_write_file(settings, write_test_file):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/portage/package/ebuild/fetch.py", line 248, in _userpriv_test_write_file
    returncode = _spawn_fetch(settings, args)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/portage/package/ebuild/fetch.py", line 173, in _spawn_fetch
    rval = spawn_func(args, env=env, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/portage/process.py", line 730, in spawn
    retval = loop.run_until_complete(pid.wait())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/portage/util/_eventloop/asyncio_event_loop.py", line 173, in _run_until_complete
    return self._loop.run_until_complete(future)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/base_events.py", line 663, in run_until_complete
    self._check_running()
  File "/usr/lib/python3.12/asyncio/base_events.py", line 622, in _check_running
    raise RuntimeError('This event loop is already running')
RuntimeError: This event loop is already running
```

zmedico: As an approximation, I've used `os.stat` to test for portage group write access.

This makes an assumption about the implementation details for DISTDIR write access with `FEATURES=userfetch`, since the approximation using `os.stat`  works with normal Unix permissions but ignores ACLs.

lib/portage/package/ebuild/fetch.py
```diff
@@ -201,6 +226,28 @@ def _userpriv_test_write_file(settings, file_path):
...
+    if asyncio.get_event_loop().is_running():
+        # Approximate the permission test in order to avoid this error:
+        # RuntimeError: This event loop is already running
+        # FIXME: This approximation only supports Unix permissions and ignores ACLs.
+        st = os.stat(os.path.dirname(file_path))
+        rval = st.st_gid == int(portage_gid) and stat.S_IMODE(st.st_mode) & 0o020
+        _userpriv_test_write_file_cache[file_path] = rval
+        return rval
+
```
zmedico, line 250: Since this approximation doesn't support ACLs, we could possibly use it just as an optimization for the case where the `st.st_gid == int(portage_gid) and stat.S_IMODE(st.st_mode) & 0o020` expression holds (ACLs no used), and if the result of this expression is `False` then proceed with the spawn in order to account for ACLs.

lib/portage/package/ebuild/fetch.py
zmedico: We need to change the _drop_privs_userfetch function to an async coroutine function, since it calls _userpriv_test_write_file via _ensure_distdir (so all three functions need to be coroutine functions).

lib/portage/package/ebuild/fetch.py
zmedico: The function is now async, so I've removed the `loop.is_running()` check and converted this Unix permission code to a conditional optimization:
```python
    # Optimize away the spawn if we can detect a portage group write
    # permission bit, but if this optimization fails then continue with
    # the spawn for ACL support.
    st = os.stat(os.path.dirname(file_path))
    if st.st_gid == int(portage_gid) and stat.S_IMODE(st.st_mode) & 0o020:
        _userpriv_test_write_file_cache[file_path] = True
        return True
```

lib/portage/package/ebuild/fetch.py
zmedico: Since the `_drop_privs_userfetch` function is always called before the event loop is running, I simply adapted it like this:
```diff
--- a/lib/portage/package/ebuild/fetch.py
+++ b/lib/portage/package/ebuild/fetch.py
@@ -114,3 +114,3 @@ def _drop_privs_userfetch(settings):
     try:
-        _ensure_distdir(settings, settings["DISTDIR"])
+        asyncio.run(_ensure_distdir(settings, settings["DISTDIR"]))
     except PortageException:
```

lib/portage/package/ebuild/fetch.py
zmedico: It turns out that the approximation / optimization to check for the portage group write bit is not needed, since we can simply use an `os.access` check in any case that the `_want_userfetch` function returns False. So, we have just this:
```python
    # Optimize away the spawn when privileges do not need to be dropped.
    if not _want_userfetch(settings):
        rval = os.access(os.path.dirname(file_path), os.W_OK)
        _userpriv_test_write_file_cache[file_path] = rval
        return rval
```

lib/portage/tests/ebuild/test_fetch.py
```diff
@@ -36,6 +38,85 @@
 
 
 class EbuildFetchTestCase(TestCase):
+
+    async def _test_interrupt(self, loop, server, async_fetch, pkg, ebuild_path):
```
thesamesam, line 42: Thanks for the test case. This is all so hairy that it's good to have as much as we can.

thesamesam (approved):

PR #1363, opened 2024-08-03
Make portage not reliant on _ctypes

mid-kid:
This is useful for bootstrapping purposes, as _ctypes requires a dynamic
linker.


floppym:
The commit summary is not very good. Please say something like this:

Make portage.util.compression_probe work when ctypes is unavailable

floppym (approved):

PR #1364, opened 2024-08-05
Pass through MAKEFLAGS and exclude from environment.bz2

zmedico:
Allow the MAKEFLAGS environment variable to pass through, in case a centralized GNU Make POSIX Jobserver is available. In order to prevent persistence of this variable in environment.bz2, exclude it when the __save_ebuild_env --exclude-init-phases argument is given.

Ultimately we may want to add support for portage to parse MAKEFLAGS and use it to allocate job tokens in various circumstances. For example, emerge could allocate a job token for each job started for emerge --jobs. This would remove a job token from the pool that is available to nested make calls, but is reasonable because nested make calls will execute jobs serially when no jobserver tokens remain.

Bug: https://bugs.gentoo.org/692576

### DIFF
--- a/lib/portage/package/ebuild/doebuild.py
+++ b/lib/portage/package/ebuild/doebuild.py
@@ -599,11 +599,12 @@ def doebuild_environment(
                     )
                     mysettings.features.remove(feature)
 
-        if "MAKEOPTS" not in mysettings:
+        # MAKEOPTS conflicts with MAKEFLAGS, so skip this if MAKEFLAGS exists.
+        if "MAKEOPTS" not in mysettings and "MAKEFLAGS" not in mysettings:
             nproc = get_cpu_count()
             if nproc:
                 mysettings["MAKEOPTS"] = "-j%d" % (nproc)
-            if "GNUMAKEFLAGS" not in mysettings and "MAKEFLAGS" not in mysettings:
+            if "GNUMAKEFLAGS" not in mysettings:
                 mysettings["GNUMAKEFLAGS"] = (
                     f"--load-average {nproc} --output-sync=line"
                 )
--- a/man/make.conf.5
+++ b/man/make.conf.5
@@ -1,4 +1,4 @@
-.TH "MAKE.CONF" "5" "Jan 2024" "Portage @VERSION@" "Portage"
+.TH "MAKE.CONF" "5" "Aug 2024" "Portage @VERSION@" "Portage"
 .SH "NAME"
 make.conf \- custom settings for Portage
 .SH "SYNOPSIS"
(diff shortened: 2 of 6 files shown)


mgorny:
Wouldn't this pass option like `-k` though, too?

zmedico:
> Wouldn't this pass option like `-k` though, too?

Yes, in fact -k is considered a "simple" option because it has no argument, and make represents these simple options as a group of letters at the beginning of MAKEFLAGS like `MAKEFLAGS="k -j4 -l4 --jobserver-auth=fifo:/tmp/GMfifo30205"`.

If there are no simple options then MAKEFLAGS begins with a space.

lib/portage/package/ebuild/doebuild.py
```diff
@@ -598,7 +598,8 @@ def doebuild_environment(
                     )
                     mysettings.features.remove(feature)
 
-        if "MAKEOPTS" not in mysettings:
+        # MAKEOPTS conflicts with MAKEFLAGS, so skip this if MAKEFLAGS exists.
+        if "MAKEOPTS" not in mysettings and "MAKEFLAGS" not in mysettings:
```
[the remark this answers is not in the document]
zmedico: Yeah, that second `"MAKEFLAGS" not in mysettings` became redundant so I've removed it now. Thanks!

thesamesam (approved):

mgorny:
> > Wouldn't this pass option like `-k` though, too?
> 
> Yes, in fact -k is considered a "simple" option because it has no argument, and make represents these simple options as a group of letters at the beginning of MAKEFLAGS like `MAKEFLAGS="k -j4 -l4 --jobserver-auth=fifo:/tmp/GMfifo30205"`.
> 
> If there are no simple options then MAKEFLAGS begins with a space.

I mean, passing arbitrary options sounds dangerous and undesirable. If someone currently runs `emerge` via `make -k` as part of some bigger script, I don't think we should pass that one to make invocations within package builds.

man/make.conf.5
```diff
@@ -948,6 +948,14 @@ Setting this and other *FLAGS variables arbitrarily may cause compile or
...
 enabled for these flags may be closed as INVALID.
 .TP
+.B MAKEFLAGS
+Use this variable instead of \fBMAKEOPTS\fR if you want to inject a
+centralized job server for make. In this case \fBMAKEOPTS\fR should be
+unset or else it can cause the make jobserver mode to reset. Include
+a \fB--jobserver-auth=fifo:PATH\fR flag to specify the path of the
+centralized jobserver fifo, which needs to be readable and writable by
+the portage group when userpriv is enabled.
+.TP
```
zmedico, line 958: If we parse MAKEFLAGS then we can automatically add a sandbox exemption for the jobserver fifo. Otherwise users will have to do this themselves or put the fifo in an unsandboxed location like $PORTAGE_TMPDIR (they might have to do that for `sesandbox` too).

zmedico: > > > Wouldn't this pass option like `-k` though, too?
> > 
> > 
> > Yes, in fact -k is considered a "simple" option because it has no argument, and make represents these simple options as a group of letters at the beginning of MAKEFLAGS like `MAKEFLAGS="k -j4 -l4 --jobserver-auth=fifo:/tmp/GMfifo30205"`.
> > If there are no simple options then MAKEFLAGS begins with a space.
> 
> I mean, passing arbitrary options sounds dangerous and undesirable. If someone currently runs `emerge` via `make -k` as part of some bigger script, I don't think we should pass that one to make invocations within package builds.

If we parse MAKEFLAGS in order to add a sandbox exemption for the fifo, I suppose we might filter out the simple options while we're there.

Flowdalic (approved):

PR #1366, opened 2024-08-06
run_exitfuncs: Support loop close via hook

zmedico:
Handle the case where the loop has not been explicitly closed before exit. In this case, run_exitfuncs previously tried to call coroutine functions as though they were normal functions, which obvously would not behave correctly.

Solve this problem by storing the coroutine functions in a separate _coroutine_exithandlers list that belongs exclusively to the run_coroutine_exitfuncs function, so that it is safe to close the loop and call run_coroutine_exitfuncs from inside a run_exitfuncs hook. A _thread_weakrefs_atexit hook already exists that will close weakly referenced loops. The _thread_weakrefs_atexit hook is now fixed to release its lock when it closes a loop, since the same lock may need to be re-acquired when run_coroutine_exitfuncs runs.

The included test case demonstrates that run_exitfuncs will run via an atexit hook and correctly terminate the socks5 proxy in a standalone program using the portage API (like eclean).

Due to a deadlock that will occur if an _exit_function atexit hook from the multiprocessing module executes before run_exitfuncs, a portage.process._atexit_register_run_exitfuncs() function needs to be called in order to re-order the hooks after the first process has been started via the multiprocessing module. The natural place to call this is in the ForkProcess class, using a global variable to trigger the call just once.

Fixes: https://github.com/gentoo/portage/pull/1295 c3ebdbb42e72 ("elog/mod_custom: Spawn processes in background")
Bug: https://bugs.gentoo.org/937384

### DIFF
--- a/lib/portage/process.py
+++ b/lib/portage/process.py
@@ -3,6 +3,7 @@
 # Distributed under the terms of the GNU General Public License v2
 
 
+import asyncio as _asyncio
 import atexit
 import errno
 import fcntl
@@ -193,6 +194,7 @@ def spawn_fakeroot(mycommand, fakeroot_state=None, opt_name=None, **keywords):
 
 
 _exithandlers = []
+_coroutine_exithandlers = []
 
 
 def atexit_register(func, *args, **kargs):
@@ -200,7 +202,12 @@ def atexit_register(func, *args, **kargs):
     what is registered.  For example, when portage restarts itself via
     os.execv, the atexit module does not work so we have to do it
     manually by calling the run_exitfuncs() function in this module."""
-    _exithandlers.append((func, args, kargs))
+    # The internal asyncio wrapper module would trigger a circular import
+    # if used here.
+    if _asyncio.iscoroutinefunction(func):
+        _coroutine_exithandlers.append((func, args, kargs))
+    else:
+        _exithandlers.append((func, args, kargs))
 
 
 def run_exitfuncs():
@@ -232,12 +239,16 @@ async def run_coroutine_exitfuncs():
     This is the same as run_exitfuncs but it uses asyncio.iscoroutinefunction
     to check which functions to run. It is called by the AsyncioEventLoop
     _close_main method just before the loop is closed.
+
+    If the loop is explicitly closed before exit, then that will cause
+    run_coroutine_exitfuncs to run before run_exitfuncs. Otherwise, a
+    run_exitfuncs hook will close it, causing run_coroutine_exitfuncs to be
+    called via run_exitfuncs.
     """
     tasks = []
-    for index, (func, targs, kargs) in reversed(list(enumerate(_exithandlers))):
-        if asyncio.iscoroutinefunction(func):
-            del _exithandlers[index]
-            tasks.append(asyncio.ensure_future(func(*targs, **kargs)))
+    while _coroutine_exithandlers:
+        func, targs, kargs = _coroutine_exithandlers.pop()
+        tasks.append(asyncio.ensure_future(func(*targs, **kargs)))
     tracebacks = []
     exc_info = None
     for task in tasks:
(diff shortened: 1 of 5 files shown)


lib/portage/process.py
```diff
@@ -232,12 +239,16 @@ async def run_coroutine_exitfuncs():
...
+    called via run_exitfuncs.
     """
     tasks = []
-    for index, (func, targs, kargs) in reversed(list(enumerate(_exithandlers))):
-        if asyncio.iscoroutinefunction(func):
-            del _exithandlers[index]
-            tasks.append(asyncio.ensure_future(func(*targs, **kargs)))
+    while _coroutine_exithandlers:
+        func, targs, kargs = _coroutine_exithandlers.pop()
+        tasks.append(asyncio.ensure_future(func(*targs, **kargs)))
```
zmedico, line 251: I've created a test case that starts a socks5 proxy from a subprocess and then calls `run_exitfuncs()` to test the cleanup. This has revealed a deadlock in `_thread_weakrefs_atexit()` that was resolved by making the lock reentrant.

zmedico: I changed it to release the lock while closing the loop, and then the reentrant lock was not needed to avoid a deadlock.

lib/portage/tests/util/test_socks5.py
```diff
@@ -269,3 +270,40 @@ def _testSocks5ServerLoopCloseSubprocess():
...
+        settings = {
+            "PORTAGE_TMPDIR": tempdir,
+            "PORTAGE_BIN_PATH": PORTAGE_BIN_PATH,
+        }
+
+        socks5.get_socks5_proxy(settings)
+        loop.run_until_complete(socks5.proxy.ready())
+        pid = socks5.proxy._proc.pid
+        portage.process.run_exitfuncs()
+        return pid
```
zmedico, line 309: Ideally `run_exitfuncs()` would be called automatically by an atexit hook, but this test case currently fails without an explicit `run_exitfuncs()`  call, and it's not just this test case because a standalone script demonstrates a similar problem.

zmedico: The `atexit.register(run_exitfuncs)` call occurs in portage.process module's global scope, but it's ineffective. If I call `atexit.register(portage.process.run_exitfuncs)` at the end of my test script then it works.

zmedico: This issue is somehow related to the `get_socks5_proxy` call, because `atexit.register(portage.process.run_exitfuncs)` is only effective when called after `get_socks5_proxy`.

zmedico: Despite this flaw, the MR is still an improvement over the status quo.

I guess I'll use [this hack](https://stackoverflow.com/questions/16042463/how-can-i-get-the-list-of-registered-atexit-functions-in-python3/63029332#63029332) to examine the registered atexit handlers as a workaround for https://github.com/python/cpython/issues/61388.

zmedico: Apparently it's an issue with an `_exit_function` from the multiprocessing module that triggers this, since `run_exitfuncs` only runs if it's added a second time after the `_exit_function`:

[<function shutdown at 0x7a027b9be5c0>, <function run_exitfuncs at 0x7a027afc5760>, <function _exit_function at 0x7a027a7432e0>]

This explains the interaction with `get_socks5_proxy`.

zmedico: Also, atexit only works in subprocesses after fork since https://github.com/python/cpython/issues/83856 was fixed in python3.13 with no plan for backport.

zmedico: We can use a global variable to trigger `atexit.register(portage.process.run_exitfuncs)` after the first subprocess is started by ForkProcess, in order to prevent the multiprocessing `_exit_function` deadlock.

thesamesam (approved):

PR #1367, opened 2024-08-09
Extend QA_PREBUILT handling to QA_SONAME_NO_SYMLINK

antecrescent:
https://bugs.gentoo.org/924953 was caused by `QA_SONAME_NO_SYMLINK` not inheriting the regex-transformed paths from `QA_PREBUILT` despite the ebuild.5 manpage implying this behavior based on the ordering of the QA variables.

The [tested ebuild](https://gitweb.gentoo.org/repo/proj/guru.git/commit/?id=0643c00df81ff8a08cff0d2f7e929ae341404212) contains:
```sh
QA_PREBUILT="
	usr/lib/*
	usr/lib64/*
"
```
which according to the `man 5 ebuild` should be enough to silence the following warning.
```
 * QA Notice: Missing soname symlink(s):
 * 
 * 	usr/lib64/amdvlk64.so.1 -> amdvlk64.so
 *
```

These commits make `QA_SONAME_NO_SYMLINK` inherit regex-transformed paths from `QA_PREBUILT` and update the QA variables, which are affected by `QA_PREBUILT`, in the ebuild.5 manpage.

### DIFF
--- a/bin/phase-functions.sh
+++ b/bin/phase-functions.sh
@@ -555,7 +555,8 @@ __dyn_install() {
 
 		# These ones support regular expressions, so translate
 		# fnmatch patterns to regular expressions
-		for x in QA_DT_NEEDED QA_FLAGS_IGNORED QA_PRESTRIPPED QA_SONAME ; do
+		for x in QA_DT_NEEDED QA_FLAGS_IGNORED QA_PRESTRIPPED \
+			QA_SONAME QA_SONAME_NO_SYMLINK; do
 			if [[ $(declare -p ${x} 2>/dev/null) = declare\ -a* ]] ; then
 				eval "${x}=(\"\${${x}[@]}\" ${QA_PREBUILT//\*/.*})"
 			else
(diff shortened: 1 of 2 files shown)


floppym:
I don't think QA_PREBUILT should imply QA_SONAME_NO_SYMLINK. It is possible that broken pre-built packages exist, and it is technical possible to resolve the issue using `dosym` in the ebuild if appropriate.

floppym:
Eh, then again I guess QA_PREBUILT is really just a big "I don't care if it's broken" flag, so maybe this is ok.

zmedico (approved):
I see, you've fixed the code to fit the `listed below` part of the QA_PREBUILT documentation.

antecrescent:
I have no strong feelings about whether QA_PREBUILT should or should not imply QA_SONAME_NO_SYMLINK. What's important to me is that the man page's `listed below` part of the QA_PREBUILT documentation is in sync with how Portage actually behaves.

> I don't think QA_PREBUILT should imply QA_SONAME_NO_SYMLINK. It is possible that broken pre-built packages exist, and it is technical possible to resolve the issue using `dosym` in the ebuild if appropriate.

I do see your point here. I'm happy to drop my first commit and fix the second one if that's what we agree upon :)

floppym (approved):

PR #1368, opened 2024-08-11
Support coroutine exitfuncs for non-main loops

zmedico:
Since an API consumer can cause loops to be instantiated for non-main threads, support coroutine exitfuncs for each loop. The included Socks5ServerAtExitThreadedTestCase calls get_socks5_proxy from a non-main thread, and demonstrates that coroutine exitfuncs for the resulting non-main loop will reliably stop the socks5 proxy via atexit hook.

The _thread_weakrefs_atexit function will now make a temporary adjustment to _thread_weakrefs.loops so that a loop is associated with the current thread when it is closing. Also, the _get_running_loop function will now store weak references to all _AsyncioEventLoop instances it creates, since each has a _coroutine_exithandlers attribute that can be modified by atexit_register calls.

Bug: https://bugs.gentoo.org/937740

### DIFF
--- a/lib/portage/tests/util/test_socks5.py
+++ b/lib/portage/tests/util/test_socks5.py
@@ -194,17 +194,17 @@ class Socks5ServerTestCase(TestCase):
         asyncio.run(self._test_socks5_proxy())
 
     async def _test_socks5_proxy(self):
-        loop = asyncio.get_running_loop()
+        loop = global_event_loop()
 
         host = "127.0.0.1"
         content = b"Hello World!"
         path = "/index.html"
         proxy = None
         tempdir = tempfile.mkdtemp()
-        previous_exithandlers = portage.process._coroutine_exithandlers
+        previous_exithandlers = loop._coroutine_exithandlers
 
         try:
-            portage.process._coroutine_exithandlers = []
+            loop._coroutine_exithandlers = []
             with AsyncHTTPServer(host, {path: content}, loop) as server:
                 settings = {
                     "PORTAGE_TMPDIR": tempdir,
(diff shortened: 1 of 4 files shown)


floppym:
Sorry, but I don't have enough of a grasp on the async code in Portage to review this in any meaningful way.

lib/portage/tests/util/test_socks5.py
```diff
@@ -295,24 +297,36 @@ def testSocks5ServerAtExit(self):
...
+    loop = global_event_loop()
+    socks5.get_socks5_proxy(settings)
+    loop.run_until_complete(socks5.proxy.ready())
+    print(socks5.proxy._proc.pid, flush=True)
+
+if __name__ == "__main__":
+    if threaded:
+        t = threading.Thread(target=main)
+        t.start()
+        t.join()
```
zmedico, line 324: Note that functions like `get_socks5_proxy` doe not currently have internal locking for thread safety, but we can safely neglect that here since this test case only has a single thread interacting with the portage API at a given time.

We can consider adding locking in the future, but for now our main goal should be to simply allow the portage APIs to be used in a single worker thread when the main thread is busy running a UI.

zmedico: Even though we use weak references to loops inside our asycio wrapper module, we can be sure that the loop won't be garbage collected before the socks5 proxy is stopped, since the socks5 proxy references the loop indirectly via ForkProcess and [_GCProtector](https://github.com/gentoo/portage/pull/1284).

zmedico:
> Sorry, but I don't have enough of a grasp on the async code in Portage to review this in any meaningful way.

Thanks anyway. By improving async support in atexit hooks, this MR adds low-level support that has been missing since we adopted async code as the basis for the portage.process module in https://github.com/gentoo/portage/pull/1249.

thesamesam (approved):

PR #1369, opened 2024-08-12
dispatch-conf: ignore SHELL in spawn_shell

floppym:
There is no need to use SHELL here, and this can actually cause problems when SHELL is set to "nologin" or "false".

Bug: https://bugs.gentoo.org/910560

### DIFF
--- a/bin/dispatch-conf
+++ b/bin/dispatch-conf
@@ -574,26 +574,18 @@ def clear_screen():
     os.system("clear 2>/dev/null")
 
 
-shell = os.environ.get("SHELL")
-if not shell or not os.access(shell, os.EX_OK):
-    shell = find_binary("sh")
-
-
 def spawn_shell(cmd):
-    if shell:
-        sys.__stdout__.flush()
-        sys.__stderr__.flush()
-        spawn(
-            [shell, "-c", cmd],
-            env=os.environ,
-            fd_pipes={
-                0: portage._get_stdin().fileno(),
-                1: sys.__stdout__.fileno(),
-                2: sys.__stderr__.fileno(),
-            },
-        )
-    else:
-        os.system(cmd)
+    sys.__stdout__.flush()
+    sys.__stderr__.flush()
+    spawn(
+        ["sh", "-c", cmd],
+        env=os.environ,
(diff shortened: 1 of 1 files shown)


zmedico (approved):

floppym:
It seems that prefix  [patches glibc](https://github.com/gentoo/gentoo/blob/2f9c42cb12b476a78214e2f7d5bb3f984ce05de6/profiles/features/prefix/standalone/profile.bashrc#L17-L35) to change `/bin/sh` to `${EPREFIX}/bin/sh` when calling `system(3)`.

We will need to deal with that somehow. Maybe we should look for `sh` in PATH intead.

floppym:
Updated to try the following:

1. ${BROOT}/bin/sh
2. PATH
3. os.system

bin/dispatch-conf
```diff
@@ -574,8 +574,9 @@ def clear_screen():
     os.system("clear 2>/dev/null")
 
 
-shell = os.environ.get("SHELL")
-if not shell or not os.access(shell, os.EX_OK):
+broot = portage.settings["BROOT"] or "/"
```
zmedico, line 577: It's messy for a cli tool to use BROOT outside of the ebuild env like this.

PATH is a fairly standard thing to use here, and we use it ourselves elsewhere in prefix support.

floppym: How would you feel about using `portage.const.EPREFIX` instead?

I'm basically trying to replicate what libc would do in os.system(). If we don't think that's important I can drop this.

floppym: I went ahead and updated it to just use PATH instead.

zmedico: Yeah, the glibc patch is reasonable, but here PATH is a reasonable alternative.

zmedico (approved):

thesamesam (approved):
LGTM but give @chewi or @grobian a chance to comment.

chewi (approved):
I agree, PATH is best here.

grobian:
sh from PATH should be OK

grobian (approved):

PR #1371, opened 2024-08-14
_EbuildFetcherProcess: Suppress CancelledError

zmedico:
Suppress CancelledError when attempting to cache the result in the _async_uri_map method. The cancelled result is returned for the caller to handle.

Bug: https://bugs.gentoo.org/937888

### DIFF
--- a/lib/_emerge/EbuildFetcher.py
+++ b/lib/_emerge/EbuildFetcher.py
@@ -373,7 +373,7 @@ class _EbuildFetcherProcess(ForkProcess):
         def cache_result(result):
             try:
                 self._uri_map = result.result()
-            except Exception:
+            except (CancelledError, Exception):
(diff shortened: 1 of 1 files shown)


thesamesam (approved):

zmedico:
Merged in 74e29110d86a55ac74fdde3e1b79506b3fb695c2.

lib/_emerge/EbuildFetcher.py
```diff
@@ -373,7 +373,7 @@ def _async_uri_map(self):
         def cache_result(result):
             try:
                 self._uri_map = result.result()
-            except Exception:
+            except (CancelledError, Exception):
```
zmedico, line 376: Fixed to to use asyncio.CancelledError in e4f0fa4a316aba2cf2b1a47c9d2974710a1d8bd3.

PR #1374, opened 2024-08-27
sync: add a 5 second timeout to gemato's openpgp refresh

eli-schwartz:
If it takes more than 5 seconds to download a miniscule file from https://gentoo.org/.well-known/openpgpkey/ then something has gone dreadfully wrong. Most commonly, what went wrong is that the user has broken ipv6 connectivity and requests simply hangs forever (no joke -- it doesn't implement Happy Eyeballs and closed the multiple bug reports as "please use stackoverflow to ask questions about how to use requests, we are not a support forum").

Pass a timeout so that we eventually give up and try ipv4 instead. This is a crude hack and the proper solution is to make gemato handle this better, but until gemato is fixed to use a better download library we do the best we can.

Bug: https://github.com/psf/requests/issues/6788
Bug: https://github.com/projg2/gemato/issues/35
Bug: https://bugs.gentoo.org/779766

### DIFF
--- a/lib/portage/sync/syncbase.py
+++ b/lib/portage/sync/syncbase.py
@@ -330,6 +330,7 @@ class SyncBase:
 
     def _get_openpgp_env(self, openpgp_key_path=None, debug=False):
         if gemato is not None:
+            timeout = 15
             # Override global proxy setting with one provided in emerge configuration
             if "http_proxy" in self.spawn_kwargs["env"]:
                 proxy = self.spawn_kwargs["env"]["http_proxy"]
@@ -337,14 +338,15 @@ class SyncBase:
                 proxy = self.spawn_kwargs["env"]["https_proxy"]
             else:
                 proxy = None
+                timeout = 5
 
             if openpgp_key_path:
                 openpgp_env = gemato.openpgp.OpenPGPEnvironment(
-                    proxy=proxy, debug=debug
+                    proxy=proxy, debug=debug, timeout=timeout
                 )
             else:
                 openpgp_env = gemato.openpgp.OpenPGPSystemEnvironment(
-                    proxy=proxy, debug=debug
+                    proxy=proxy, debug=debug, timeout=timeout
                 )
 
             return openpgp_env


lib/portage/sync/syncbase.py
```diff
@@ -340,11 +340,11 @@ def _get_openpgp_env(self, openpgp_key_path=None, debug=False):
 
             if openpgp_key_path:
                 openpgp_env = gemato.openpgp.OpenPGPEnvironment(
-                    proxy=proxy, debug=debug
```
thesamesam, line 343: Let's do 10-15, perhaps? I'm willing to compromise on 5 and something higher if a proxy is set.

I'm thinking of the Tor case where e.g. no circuits have been built yet.

eli-schwartz: Fair enough, timeout is 15 if a proxy is set and 5 otherwise.

lib/portage/sync/syncbase.py
```diff
@@ -330,21 +330,23 @@ def noisy_refresh_keys():
...
                 proxy = self.spawn_kwargs["env"]["https_proxy"]
             else:
                 proxy = None
+                timeout = 5
 
             if openpgp_key_path:
                 openpgp_env = gemato.openpgp.OpenPGPEnvironment(
-                    proxy=proxy, debug=debug
+                    proxy=proxy, debug=debug, timeout=timeout
                 )
```
zmedico, line 346: Does this timeout mean that callers should be adjusted to handle `requests.Timeout` exceptions?

We could possibly catch it here and the re-raise it as `portage.exeception.TimeoutException` or something.

eli-schwartz: Hmm, possibly. It doesn't look like gemato handles requests.Timeout nor uses the base exception classes.

eli-schwartz: How do you want to handle this given that portage doesn't directly depend on requests?

zmedico: We can substitute TimeoutError if requests is missing like this:
```python
try:
	from requests.exceptions import Timeout
execept ImportError:
	Timeout = TimeoutError
```

mgorny:
Does domain resolution count towards that timeout? Because if it does, then 5 seconds is very low.

selckin:
Even something as high as 30s if you are worried about it would be a big improvement, now it just hangs forever.
since gentoo.org resolves to 6 ip adresses, this would still potentially hang for 6*30s then, but i at least usually start a sync in a terminal in the background and swap to something else to continue working, coming back to it 10min later and it still hanging is quite annoying, and if it fails the terminal app can notify me

waltercool:
+1 to this. Works on my system. I get a lot this never-ending sync problem lately. This PR works to me

PR #1379, opened 2024-09-09
unpack: Don't display "Unpacking ..." for skipped files, match file extensions case-sensitively

ulm:
As discussed in `#gentoo-pms`.

PMS currently says that "Any unrecognised file format shall be skipped silently." This wording was added with the draft of what later became EAPI 4: https://gitweb.gentoo.org/proj/pms.git/commit/?id=634c32f231e1bc94d64588e2b2edf0ad1ca60f1f

The commit message doesn't give any rationale for "silently". It may well be that the wording is a remnant of the rejected `unpack --if-compressed` item. See the [discussion in the 2009-04-23 council meeting](https://projects.gentoo.org/council/meeting-logs/20090423.txt) (starting at 21:35).

Patch for PMS (presumably, replacing "silently" with "without raising an error") will follow.

### DIFF
--- a/bin/phase-helpers.sh
+++ b/bin/phase-helpers.sh
@@ -334,7 +334,7 @@ unpack() {
 		y=${y##*.}
 		y_insensitive=$(LC_ALL=C tr "[:upper:]" "[:lower:]" <<< "${y}")
 
-		# wrt PMS 11.3.3.13 Misc Commands
+		# wrt PMS 12.3.15 Misc Commands
 		if [[ ${x} != */* ]]; then
 			# filename without path of any kind
 			srcdir=${DISTDIR}/
--- a/bin/phase-helpers.sh
+++ b/bin/phase-helpers.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Copyright 1999-2023 Gentoo Authors
+# Copyright 1999-2024 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 if ___eapi_has_DESTTREE_INSDESTTREE; then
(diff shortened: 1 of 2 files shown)


thesamesam:
Please, if possible, include the PR description within the commit message(s) where appropriate, so that the relevant info is encoded just in the git repo.

bin/phase-helpers.sh
```diff
@@ -360,14 +360,33 @@ unpack() {
...
+			7z)      ___eapi_unpack_supports_7z  && suffix_known=1 ;;
+			rar)     ___eapi_unpack_supports_rar && suffix_known=1 ;;
+			lha|lzh) ___eapi_unpack_supports_lha && suffix_known=1 ;;
+			xz)      ___eapi_unpack_supports_xz  && suffix_known=1 ;;
+			txz)     ___eapi_unpack_supports_txz && suffix_known=1 ;;
+		esac
+
+		if ___eapi_unpack_is_case_sensitive \
+				&& [[ ${suffix} != "${suffix_insensitive}" ]] \
+				&& ! has ${suffix} ZIP Z 7Z RAR LHA LHa; then
```
mgorny, line 375:
```suggestion
				&& ! has "${suffix}" ZIP Z 7Z RAR LHA LHa "${suffix_insensitive}"; then
```

should be the same, right? Also note that technically `suffix` can contain spaces.

Or perhaps we could even make this simpler and do something like:

```
___eapi_unpack_is_case_sensitive || suffix=${suffix,,}
```

ulm: Good catch, fixed.

floppym (approved):

thesamesam (approved):

ulm:
Merged.

PR #1382, opened 2024-09-20
vartree: use stat symbols instead of magic numbers

henning-schild:
This is a purely cosmetic change to improve readability a bit.


thesamesam (approved):
I'm sorry for missing this before now!

PR #1383, opened 2024-09-21
RFC: Drop _emerge/getloadavg.py

Flowdalic:
CC @zmedico You added the `getloadavg()` wrapper to portage with https://github.com/gentoo/portage/commit/8607a39dc9d9. However, the commit message lacks a rationale why such a wrapper became necessary. This PR assumes that the reasons for the wrapper are now no longer existent. Could you clarify if this is true, otherwise we should consider adding a code comment to `_emerge/getloadavg.py` explaining the reason for the wrapper.


zmedico:
I've found the original commit, and we should be able to remove the
/proc/loadaverage usage if all of your supported libc implementations have
getloadaverage(). It was probably lacking in uclibc at the time of this
commit:

https://github.com/gentoo/portage/commit/8607a39dc9d979713f9434839ed1aef2602276a1

grobian:
Hi @zmedico os.getloadavg() returns data everywhere I tried.  Should it be unavailable, I think we should just ignore it, and disable or bark or something when load average limits are used.  It's not like loadaverage means the same on every platform anyway.

chewi:
Sounds good to me, although I don't do much with alternative libcs.

zmedico (approved):

thesamesam (approved):

thesamesam:
On reflection, please update the commit message to..
1) include a summary of https://github.com/gentoo/portage/pull/1383#issuecomment-2375471855 (and perhaps a link),
2) describe why we're ignoring it if it fails (because we know uclibc failed in the past and we don't want to fail on such cases, just silently chug on)

Flowdalic:
I've updated the commit message

> 2\. describe why we're ignoring it if it fails

this behavior would change with this PR applied. In the past, we simply ignored a missing `os.getloadavg()` and tried some fallback. After this PR, portage would fail if there is no `os.getloadavg()` provided by Python's runtime library.

So we where ignoring "getloadavg() failures" in the past, but after this PR, we would fail.

thesamesam:
I'd kind of prefer it if we caught it not existing at all (not just OSError) but if others are OK with failing hard, I guess it's OK with me.

thesamesam:
The issue is more with ports to new platforms IMO.

PR #1387, opened 2024-10-16
depgraph: Ignore blockers when computing virtual deps visibility

ismell:
If a `virtual/` package declares a blocker to a package that doesn't
exist, then `_virt_deps_visible` incorrectly marks the virtual as
"not visible". Blockers should be ignored when performing the deps
visibility test since blockers are handled later in the pipeline.

### DIFF
--- a/lib/portage/tests/resolver/test_virtual_slot.py
+++ b/lib/portage/tests/resolver/test_virtual_slot.py
@@ -71,7 +71,13 @@ class VirtualSlotResolverTestCase(TestCase):
             },
             "virtual/jdk-1.7.0": {
                 "SLOT": "1.7",
-                "RDEPEND": "|| ( =dev-java/icedtea-7* =dev-java/oracle-jdk-bin-1.7.0* )",
+                "RDEPEND": """
+                    || (
+                        =dev-java/icedtea-7*
+                        =dev-java/oracle-jdk-bin-1.7.0*
+                    )
+                    !virtual/jdk-does-not-exist
+                """,
             },
         }
 
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -6022,6 +6022,8 @@ class depgraph:
 
         for atoms in rdepend.values():
             for atom in atoms:
+                if atom.blocker:
+                    continue
                 if ignore_use:
                     atom = atom.without_use
                 pkg, existing = self._select_package(pkg.root, atom)


thesamesam:
Any chance of a testcase?

ismell:
Sure! Can you point me to something similar that I can reference?

ismell:
Just FYI, here is some more detailed information: https://issuetracker.google.com/issues/373581679#comment13

Is it valid to put the Google bug tracker in the commit?

thesamesam:
Thanks on both counts! Yes, please do in fact. I just use `Bug:` for these but `Link:` or `Reference:` is acceptable as well.

For the testcase, check out any of the `ResolverPlayground` tests. I think a testcase with:
* a virtual containing a blocker;
* an installed version of a package satisfying the virtual;
* a pending upgrade for said package;
* checking the merge list on upgrade to make sure the new version of that package is queued
would work.

ismell:
Is there documentation on running the tests, or should I just copy/paste from the [workflow config](https://github.com/gentoo/portage/actions/runs/11372409632/workflow?pr=1387#L85)?

thesamesam:
Bog standard `pytest` should work (you can give args for specific tests if you want), `meson test` (which has a bit less flexibility for this), or you can use `tox` (it's in the root of the repo), or copy what the workflow does.

zmedico:
Here is a minimal test case:
```diff
--- a/lib/portage/tests/resolver/test_virtual_slot.py
+++ b/lib/portage/tests/resolver/test_virtual_slot.py
@@ -71,7 +71,7 @@ class VirtualSlotResolverTestCase(TestCase):
             },
             "virtual/jdk-1.7.0": {
                 "SLOT": "1.7",
-                "RDEPEND": "|| ( =dev-java/icedtea-7* =dev-java/oracle-jdk-bin-1.7.0* )",
+                "RDEPEND": "|| ( =dev-java/icedtea-7* =dev-java/oracle-jdk-bin-1.7.0* ) !virtual/jdk-cip",
             },
         }

```

ismell:
Oh, awesome! I've been trying to recreate the conditions in a test case, but with no luck. I can add it into my patch on Monday. Thanks!

ismell:
Thanks Zac!

lib/portage/tests/resolver/test_virtual_slot.py
```diff
@@ -71,7 +71,13 @@ def testVirtualSlotUpdate(self):
             },
             "virtual/jdk-1.7.0": {
                 "SLOT": "1.7",
```
ismell, line 73: Bah,
So I figured out why I was having problems trying to write a test case in our 2.3.75 branch:
https://chromium-review.googlesource.com/c/chromiumos/third_party/portage_tool/+/2936978

We  mark EAPI 0 as deprecated.

Since none of these test packages have their EAPI specified, they default to 0. That means that these packages have the `EAPI.deprecated` mask which in turn makes the `visible` property false.

Changing `ResolverPlayground._create_ebuilds` to default to EAPI 5 in our fork makes the test pass.

zmedico (approved):

thesamesam (approved):

ismell:
Just FYI, I'm tracking a new similar bug here: https://issuetracker.google.com/issues/376518885
I haven't confirmed it it's reproducible using upstream portage yet, so it might just be due to our archaic version of portage.

PR #1389, opened 2024-10-21
Handle python3.14 ChildWatcher changes

zmedico:
Bug: https://bugs.gentoo.org/941955

### DIFF
--- a/lib/portage/util/_eventloop/asyncio_event_loop.py
+++ b/lib/portage/util/_eventloop/asyncio_event_loop.py
@@ -6,12 +6,16 @@ import signal
 
 import asyncio as _real_asyncio
 from asyncio.events import AbstractEventLoop as _AbstractEventLoop
-from asyncio.unix_events import ThreadedChildWatcher
 
 try:
-    from asyncio.unix_events import PidfdChildWatcher
+    from asyncio.unix_events import _ThreadedChildWatcher as ThreadedChildWatcher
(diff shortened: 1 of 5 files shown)


lib/portage/util/_eventloop/asyncio_event_loop.py
```diff
@@ -6,12 +6,22 @@
...
-    from asyncio.unix_events import PidfdChildWatcher
+    from asyncio.unix_events import _ThreadedChildWatcher as ThreadedChildWatcher
 except ImportError:
-    PidfdChildWatcher = None
+    try:
+        from asyncio.unix_events import ThreadedChildWatcher
+    except ImportError:
+        ThreadedChildWatcher = None
+
+try:
```
thesamesam, line 18: Our minimum Python is 3.9, I think, so do we need this conditional? Looks OK otherwise.

thesamesam: I think the same might apply to https://github.com/gentoo/portage/commit/7218c9cea6185c82fbc1719489a2d8e814d05ff1.

zmedico: Yes on both accounts, so updated to remove the unnecessary conditionals. Thanks!

thesamesam (approved):
Thanks!

PR #1393, opened 2024-11-01
gpkg: do not consider symlinks targets for size estimation

zmedico:
Symlinks size is already accounted for, so there is no need to account for the pointed to file.
Moreover, previous code failed to handle permission error when using ROOT= and having absolute symlinks pointing to running root.

Bug: https://bugs.gentoo.org/942512

### DIFF
--- a/lib/portage/gpkg.py
+++ b/lib/portage/gpkg.py
@@ -1960,14 +1960,10 @@ class gpkg:
 
                 image_max_link_length = max(image_max_link_length, path_link_length)
 
-                try:
-                    file_size = os.path.getsize(f)
-                except FileNotFoundError:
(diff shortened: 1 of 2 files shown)


lib/portage/gpkg.py
```diff
@@ -1960,14 +1960,10 @@ def _check_pre_image_files(self, root_dir, image_prefix="image"):
...
-                try:
-                    file_size = os.path.getsize(f)
-                except FileNotFoundError:
-                    # Ignore file not found if symlink to non-existing file
-                    if os.path.islink(f):
-                        continue
-                    else:
-                        raise
+                if os.path.islink(f):
+                    continue
```
zmedico, line 1964: There's an earlier islink call we can use the result from here.

PR #1395, opened 2024-11-04
Fix deprecated datetime.datetime.utcnow() usage

zmedico:
DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).

Beware that this may break interaction with gemato timestamps if they are not UTC aware:
```python
>>> datetime.datetime.now(datetime.UTC) - datetime.datetime.utcnow()
  File "<stdin>", line 1, in <module>
TypeError: can't subtract offset-naive and offset-aware datetimes
```

Also at least python3.11 is required for the datetime.UTC alias for datetime.timezone.utc.

### DIFF
--- a/lib/portage/sync/modules/rsync/rsync.py
+++ b/lib/portage/sync/modules/rsync/rsync.py
@@ -437,7 +437,11 @@ class RsyncSync(NewBase):
                             raise RuntimeError("Timestamp not found in Manifest")
                         if (
                             self.max_age != 0
-                            and (datetime.datetime.utcnow() - ts.ts).days > self.max_age
+                            and (
+                                datetime.datetime.now(datetime.timezone.utc)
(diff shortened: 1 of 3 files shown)


lib/portage/sync/modules/rsync/rsync.py
```diff
@@ -437,7 +437,8 @@ def update(self):
                             raise RuntimeError("Timestamp not found in Manifest")
                         if (
                             self.max_age != 0
-                            and (datetime.datetime.utcnow() - ts.ts).days > self.max_age
+                            and (datetime.datetime.now(datetime.UTC) - ts.ts).days
+                            > self.max_age
```
zmedico, line 441: The ts.ts from gemato may need to be fixed to be UTC aware here.

PR #1399, opened 2024-11-11
update set for "rust-rebuild" to adapt to changes in gentoo

eli-schwartz:
The virtual package is deprecated and we use an eclass to depend on an any-of group for the relevant rust compilers. This set no longer does anything without an update.

Start checking BDEPEND for the specific rust compilers. Also keep checking for the virtual, so that "rust-rebuild" reports on packages that have not, in fact, been rebuilt yet.


zmedico (approved):

thesamesam:
Please tag https://bugs.gentoo.org/906044.

floppym (approved):

eli-schwartz:
Updated the commit metadata.

zmedico (approved):

PR #1401, opened 2024-12-06
webrsync: Support sync-webrsync-verify-signature = false

zmedico:
Bug: https://bugs.gentoo.org/945861


thesamesam (approved):
I think in theory, this could've (maybe) worked before with the semi-parsing of config that emerge-webrsync tries/tried to do, but it certainly doesn't now, and that old way isn't the right direction.

So definitely lgtm -- thanks!

PR #1402, opened 2024-12-07
Scheduler: Handle SIGWINCH for JobStatusDisplay

zmedico:
Do not use curses in the get_term_size function, since curses caches the terminal width and does not account for resize.

Bug: https://bugs.gentoo.org/945382

### DIFF
--- a/lib/_emerge/JobStatusDisplay.py
+++ b/lib/_emerge/JobStatusDisplay.py
@@ -65,11 +65,9 @@ class JobStatusDisplay:
             if not isinstance(v, str):
                 self._term_codes[k] = v.decode(encoding, "replace")
 
-        if self._isatty:
-            width = portage.output.get_term_size()[1]
-        else:
-            width = self.max_display_width
-        self._set_width(width)
+        if not self._isatty:
+            self._set_width(self.max_display_width)
+        self.sigwinch()
 
     def _set_width(self, width):
         if width == getattr(self, "width", None):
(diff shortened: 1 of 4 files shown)


lib/portage/output.py
```diff
@@ -522,16 +522,19 @@ def get_term_size(fd=None):
         fd = sys.stdout
     if not hasattr(fd, "isatty") or not fd.isatty():
         return (0, 0)
-    try:
-        import curses
 
+    # Do not use curses by default because it returns stale values after terminal resize.
+    if cached:
```
zmedico, line 527: This leaves behind dead code, but maybe it will be fixed in curses at some point in the future. Meanwhile, it serves as documentation of why not to use curses here.

Flowdalic: Interesting. I found https://stackoverflow.com/a/53019662/194894, which suggests that we should use

```
curses.initscr().getmaxy()
```

instead of `curses.tigetnum()`.

lib/_emerge/JobStatusDisplay.py
```diff
@@ -79,6 +75,13 @@ def _set_width(self, width):
         object.__setattr__(self, "width", width)
         object.__setattr__(self, "_jobs_column_width", width - 32)
 
+    def sigwinch(self):
+        if self._isatty:
```
Flowdalic, line 79: I guess we could also simply do nothing, i.e., early exit,  if `not self._isatty()`.

```python
def sigwinch(self):
  if not self._isatty:
    return
  width = portage.output.get_term_size()[1]
  self._set_width(width)
```

If that is possible, then it would be my personal preference. But nothing I would insist on.

lib/portage/output.py
zmedico: Hmm, I tested this, and it seems `curses.initscr()` has some unwanted side-effects.

lib/_emerge/JobStatusDisplay.py
zmedico: updated to use your version.

lib/portage/output.py
zmedico: I've removed to cached parameter and just comments about not using curses here.

Flowdalic (approved):

PR #1406, opened 2024-12-27
Flush the merge-wait queue on SIGUSR2

Flowdalic:
If portage receives SIGUSR2, then flush the merge-wait queue on the next possible occasion (typically after all running "install" jobs are finished).

### DIFF
--- a/lib/_emerge/Scheduler.py
+++ b/lib/_emerge/Scheduler.py
@@ -210,6 +210,8 @@ class Scheduler(PollScheduler):
         # empty.
         self._merge_wait_scheduled = []
 
+        self._flush_merge_wait_queue = False
+
         # Holds system packages and their deep runtime dependencies. Before
         # being merged, these packages go to merge_wait_queue, to be merged
         # when no other packages are building.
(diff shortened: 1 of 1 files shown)


zmedico (approved):

lib/_emerge/Scheduler.py
```diff
@@ -1193,6 +1199,10 @@ def sighandler(signum, frame):
                     signal.signal(signal.SIGCONT, earlier_sigcont_handler)
                 else:
                     signal.signal(signal.SIGCONT, signal.SIG_DFL)
+                if earlier_sigusr2_handler is not None:
```
thesamesam, line 1202: No need to do this now, but I wonder if we could do bette rwith this with a context manager.

lib/_emerge/Scheduler.py
```diff
@@ -1809,11 +1819,16 @@ def _schedule_tasks(self):
             # special packages and we want to ensure that
             # parallel-install does not cause more than one of
             # them to install at the same time.
-            if (
-                self._merge_wait_queue
-                and not self._jobs
-                and not self._task_queues.merge
+            if self._merge_wait_queue and (
```
thesamesam, line 1822: This might be nice as a variable instead.

thesamesam (approved):

lib/_emerge/Scheduler.py
Flowdalic: Not sure if I understand. "This" being `_merge_wait_queue` and you want something like `merge_wait_queue_not_empty`?

lib/_emerge/Scheduler.py
Flowdalic: This would require creating a custom context manager class for signals, right?

lib/_emerge/Scheduler.py
thesamesam: I mean the whole expression is pretty icky over multiple lines (although this is partly Black's fault).

PR #1407, opened 2024-12-30
Do not export PMS variables in EAPI 9

Flowdalic:
Instead of passing PMS variables as part of the process environment, we pass them via a file that is later sourced by the ebuild's bash.

Since, for example, A is usually the greatest contributor to the process environment, removing it from the process environment significantly avoids running into MAX_ARG_STRLEN when spawning a new child process.

This means that A and other PMS variables are no longer exported in the ebuild and hence unavaiable to child processes. However, A is mostly used as part of the default_src_unpack function and there A does not need to be exported.

This started as a change that only unexported A, but was later extended to most PMS variables as suggested by ulm in https://bugs.gentoo.org/721088#c23.

Thanks to Zac Medico for helpful input on this change.

Closes: https://bugs.gentoo.org/721088


lib/portage/const.py
```diff
@@ -181,6 +181,7 @@
         "digest",
         "distcc",
         "distlocks",
+        "dont-export-pms-vars",
```
zmedico, line 184: Could we make this `export-pms-vars` and enable it by default?

Flowdalic: Unfortunately, I don't think so, as there is no way to disable a feature. Or does `FEATURES="-export-pms-vars"` work (havn't checked)?

The FEATURE=dont-export-pms-vars is there to test the behavior with current EAPIs. That is, it is opt-in, instead of opt-out. In theory, we could also remove the dont-export-pms-vars FEATURE once we are reasonably convinced that it does not break anything.

thesamesam: Yes, FEATURES="-x" works.

Flowdalic: Ok, not sure if I missed it in make.conf's man page, but if it isn't there, then it should potentially be documented.

So what would
```python
 if "dont-export-pms-vars" in mysettings.features or not eapi_exports_pms_vars(eapi):
```

become with FEATURES="-export-pms-vars"? Simply

```python
 if "-export-pms-vars" in mysettings.features or not eapi_exports_pms_vars(eapi):
```
?

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2133,9 +2185,37 @@ def spawn(
...
+        if not os.path.isdir(t):
+            os.makedirs(t)
+
+        ebuildExtraSource = os.path.join(t, ".portage-ebuild-extra-source")
+        with open(ebuildExtraSource, mode="w") as f:
+            for var_name in unexported_env_vars:
+                var_value = orig_env.get(var_name)
+                if var_value is None:
+                    continue
+                f.write(f"{var_name}='{var_value}'\n")
```
Flowdalic, line 2209: Should we

```suggestion
                quoted_var_value = shlex.quote(var_value)
                f.write(f"{var_name}={quoted_var_value}\n")
```
?

lib/portage/const.py
thesamesam: It would be `"export-pms-vars" not in mysettings.features`.

zmedico (approved):

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2133,9 +2185,38 @@ def spawn(
...
+    if unexported_env_vars:
+        orig_env = mysettings.environ()
+        # Copy since we are potentially removing keys from the dict.
+        env = orig_env.copy()
+
+        t = env["T"]
+        if not os.path.isdir(t):
+            os.makedirs(t)
+
+        ebuildExtraSource = os.path.join(t, ".portage-ebuild-extra-source")
```
zmedico, line 2203: Note that the T directory is currently not ideal for the "depend" phase, since the parent build directory is not locked and cleaned up afterwards. Therefore, a file under T will not be cleaned up for the "depend" phase.

Flowdalic: Any recommendation of an alternative approach? Should and could we run the code creating the .portage-ebuild-extra-source file only of an ebuild phase function is going to be invoked afterwards?

Flowdalic: We could maybe

```python
        t = env["T"]
        if os.path.isdir(t):
            ebuild_extra_source_path = os.path.join(t, ".portage-ebuild-extra-source")
        else:
            ebuild_extra_source_path = tempfile.NamedTemporaryFile(
                suffix=f"portage-ebuild-extra-source",
                delete=True,
                delete_on_close=False,
            ).name
```

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2133,9 +2185,43 @@ def spawn(
...
+
+        t = env["T"]
+        if os.path.isdir(t):
+            ebuild_extra_source_path = os.path.join(t, ".portage-ebuild-extra-source")
+        else:
+            ebuild_extra_source_path = tempfile.NamedTemporaryFile(
+                prefix=f"portage-ebuild-extra-source",
+                delete=True,
+                delete_on_close=False,
+            ).name
```
zmedico, line 2207: I believe this will leave an orphaned temporary file. We could save the path as `mysettings["PORTAGE_EBUILD_EXTRA_SOURCE"]` and clean it up after the process exits in EbuildMetadataPhase.

zmedico: Maybe something like this:
```diff
diff --git a/lib/_emerge/EbuildMetadataPhase.py b/lib/_emerge/EbuildMetadataPhase.py
index 54177840c7..32ea846676 100644
--- a/lib/_emerge/EbuildMetadataPhase.py
+++ b/lib/_emerge/EbuildMetadataPhase.py
@@ -38,2 +38,3 @@ class EbuildMetadataPhase(SubProcess):
         "deallocate_config",
+        "portage_ebuild_extra_source",
         "write_auxdb",
@@ -167,2 +168,3 @@ class EbuildMetadataPhase(SubProcess):
 
+        self.portage_ebuild_extra_source = self.settings.get("PORTAGE_EBUILD_EXTRA_SOURCE")
         if self.deallocate_config is not None and not self.deallocate_config.done():
@@ -193,2 +195,4 @@ class EbuildMetadataPhase(SubProcess):
         SubProcess._unregister(self)
+        if self.portage_ebuild_extra_source:
+            os.unlink(self.portage_ebuild_extra_source)
```

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2133,9 +2185,48 @@ def spawn(
...
+    if "export-pms-vars" not in mysettings.features or not eapi_exports_pms_vars(eapi):
+        unexported_env_vars = _unexported_pms_vars
+
+    if unexported_env_vars:
+        orig_env = mysettings.environ()
+        # Copy since we are potentially removing keys from the dict.
+        env = orig_env.copy()
+
+        t = env["T"]
+        if os.path.isdir(t):
```
Flowdalic, line 2200: @zmedico is there a way to determine if we are in a phase that has the parent directory of T locked? I fear that `os.path.isdir(t)` is not robust, as it could` also simply return true if the directory happens to exist

Flowdalic: Maybe we should

```python
if phase not in _global_pid_phases and os.path.isdir(t)
```

not sure if the `and os.path.isdir(t)` is even required then

Flowdalic: Yep, that seems to work and do what we want

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2133,9 +2185,57 @@ def spawn(
...
+        orig_env = mysettings.environ()
+        # Copy since we are potentially removing keys from the dict.
+        env = orig_env.copy()
+
+        t = env["T"]
+        # The os.path.isdir() is probably not required, as T should
+        # exists if we are processing a phase that is in
+        # _phase_func_map. But better safe than sorry.
+        if mysettings.get("EBUILD_PHASE") in _phase_func_map.keys() and os.path.isdir(
+            t
```
zmedico, line 2206: We could probably make this just `mysettings.get("EBUILD_PHASE") != "depend"`.

bin/isolated-functions.sh
```diff
@@ -8,6 +8,15 @@ if ___eapi_has_version_functions; then
 	source "${PORTAGE_BIN_PATH}/eapi7-ver-funcs.sh" || exit 1
 fi
 
+if [[ -v PORTAGE_EBUILD_EXTRA_SOURCE ]]; then
+	source "${PORTAGE_EBUILD_EXTRA_SOURCE}" || exit 1
+	# We delierbately do not unset PORTABE_EBUILD_EXTRA_SOURCE, so
```
zmedico, line 13: s/delierbately/deliberately/

lib/portage/package/ebuild/doebuild.py
Flowdalic: I considered `mysettings.get("EBUILD_PHASE") != "depend"`. The drawback is that, as soon as portage adds another phase with similar constraints like "depend", the code would need to be adjusted or otherwise the if-condition would erroneously hold. Whereas, if `mysettings.get("EBUILD_PHASE") in _phase_func_map.keys()` is true, then we know with the current implementation, that `T` will exists (since it is mandated for all PMS ebuild phase functions to exist) and this is also likely to be true for potential future PMS ebuild phase functions (simply because T is useful all the time).

That said, this was just explaining the rationale why I did it this way. I ultimately leave the decision to you, because you have far more experience with portage. Let me know if it should stay as it currently is, or if you want me to change it to `mysettings.get("EBUILD_PHASE") != "depend"`.

Flowdalic:
Now is a good time for others to test this patch and the feature. Set `FEATURES="${FEATURES} -export-pms-vars"` in `make.conf` to enable the behavior in all existing EAPIs. Make sure to use sys-apps/portage-9999, since sys-apps/portage was the only ebuild I have encountered so far that is affected by this change, and the -9999 ebuild of portage takes it into consideration.

lib/portage/package/ebuild/doebuild.py
zmedico: I think it's nicer to use `mysettings.get("EBUILD_PHASE") != "depend"` since we only implement the temp file cleanup for the `depend` phase. If we add another phase with similar constraints then it's easy enough to add a special case for it here.

Flowdalic: > since we only implement the temp file cleanup for the `depend` phase.

Then we seem to have a general problem. I've added a debug print to emit the phase names we are seeing and got:
```
EBUILD_PHASE: clean
EBUILD_PHASE: cleanrm
EBUILD_PHASE: compile
EBUILD_PHASE: configure
EBUILD_PHASE: install
EBUILD_PHASE: instprep
EBUILD_PHASE: None
EBUILD_PHASE: package
EBUILD_PHASE: postinst
EBUILD_PHASE: postrm
EBUILD_PHASE: preinst
EBUILD_PHASE: prepare
EBUILD_PHASE: prerm
EBUILD_PHASE: setup
EBUILD_PHASE: test
EBUILD_PHASE: unpack
```

We don't even see a phase named "depend" there. Is that to be expeced?
From those phases, these are the ones that are neither "depend" or in `_phase_func_map.keys()`:
```
clean
cleanrm
instprep
None
package
```

Note that we are also see an invocation where EBUILD_PHASE is `None`. Is that to be expected? I also see that the current approach leaves `/tmp/portage-ebuild-extra-source-*` files behind. Which would be in line with what we are seeing. I wonder if we shouldn't simply cleanup the file in the `finally` block at the end of the function, instead via `EbuildMetadataPhase`.

Flowdalic: A simple solution would be to only unexport the PMS variables if we are spawning in an EBUILD_PHASE in `_phase_func_map.keys()`, because then we can place the `PORTAGE_EBUILD_EXTRA_SOURCE` file in`T`. I've pushed this now.

The drawback would be, that the PMS variables are still exported when the ebuild is invoked without the intention to call a specific ebuild phase function. But it makes the implementation of the feature much cleaner, as we can bind the life cycle of the `PORTAGE_EBUILD_EXTRA_SOURCE` to `T`'s life cycle.

zmedico: You will only observe the `depend` phase when there is missing metadata cache for a particular ebuild. Once the cache is populated, you won't observe it again until there is an ebuild or eclass modification.

zmedico: > If we also want to unexport the PMS vars in the clean, cleanrm, instprep, None, and package phases, then we need to come up with a good scheme to clean up the PORTAGE_EBUILD_EXTRA_SOURCE file at the end of those phases.

The problem is that we need to unexport them for all phases in order to prevent E2BIG errors.

Flowdalic: > The problem is that we need to unexport them for all phases in order to prevent E2BIG errors.

Right, not sure how I could miss that.

But if I understand the situation correctly, we have a life cycle issue then. It is trivial to manage the life cycle of the `PORTAGE_EBUILD_EXTRA_SOURCE` file in all phases that map to PMS ebuild phases. It also possible to manage the life cycle in the 'depend' phase using `EbuildMetadataPhase`. But how to we manage the life cycle of  the `PORTAGE_EBUILD_EXTRA_SOURCE` in the remaining phases? Those phases include clean, cleanrm, instprep, None, and package.

In which phases, besides the PMS ebuild phases, is `"${PORTAGE_TMPDIR}/portage/${CATEGORY}/${PF}/` locked and `T` exists? It appears to be also in the `None` phase. But, for example, not in the 'clean' phase.

Flowdalic: I came up with something that should avoid E2BIG. We simply side-step the life cycle issues for the problematic phases by still exporting the PMS variables there, *but* strip the problematic ones (currently 'A'): https://github.com/gentoo/portage/pull/1407/files#diff-b95b5cc27d828361746142e69c03078573680277416d7fe5a60b1f61bed38072R2237-R2241

I still hope we find a solution to the life cycle issue, though.

zmedico: The None phase is MiscFunctionsProcess I think, so we can add cleanup similar to EbuildMetadataPhase in there. For clean and cleanrm it should suffice to strip problematic vars.

zmedico: Actually I see you assume T is available for None phase which I think is fine.

zmedico (approved):

lib/portage/package/ebuild/doebuild.py
Flowdalic: I am currently considering managing the livecycle with a temporary per-emerge/ebuild process directory that is cleaned up via an atexit handler: https://github.com/gentoo/portage/pull/1407/files#diff-b95b5cc27d828361746142e69c03078573680277416d7fe5a60b1f61bed38072R2212-R2217

Flowdalic: It just occurred to me that we could also simply

```bash
trap "rm \"${PORTAGE_EBUILD_EXTRA_SOURCE}\"" EXIT
```

I am probably going to add that.

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2133,9 +2187,65 @@ def spawn(
...
+            global _emerge_tmpdir
+            if _emerge_tmpdir is None:
+                _emerge_tmpdir = tempfile.mkdtemp(
+                    prefix=f"portage-tmpdir-{portage.getpid()}-"
+                )
+                portage.process.atexit_register(shutil.rmtree, _emerge_tmpdir)
+            ebuild_extra_source_fd, ebuild_extra_source_path = tempfile.mkstemp(
+                prefix=f"portage-ebuild-extra-source-{phase}-",
+                dir=_emerge_tmpdir,
+            )
```
zmedico, line 2221: We may need to chown the file so that it is accessible by the portage group.

Flowdalic: Making it world readable, as done I t he following line, seems to be sufficient. At least, I did not see any related failures during my testing so far.

zmedico: Oh right, world readable is enough.

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2133,9 +2187,65 @@ def spawn(
...
+            t = env["T"]
+            ebuild_extra_source_path = os.path.join(
+                t, f".portage-ebuild-extra-source-{phase}"
+            )
+        else:
+            global _emerge_tmpdir
+            if _emerge_tmpdir is None:
+                _emerge_tmpdir = tempfile.mkdtemp(
+                    prefix=f"portage-tmpdir-{portage.getpid()}-"
+                )
```
zmedico, line 2216: This parent directory needs to be made world executable.

Flowdalic: This doesn't seem to be required. The phases that trigger this case seem to be able to access it. Which makes me wonder if, due the latest changes, the chmod() above is still required.

However, I can see an argument to make it always world readable. That would also helpwith debugging. In contrast to the argument to not introduce unnecessary code.

So we probably should either chmod the file and the directory, or do none of that.

I am currently leaning towards making it world readable, but first let me run some experiments.

zmedico: Also, the atexit hook is not ideal for things like egencache that will create a large number of temporary files. It would be much nicer to have EbuildMetadataPhase clean up the temporary files as soon as their lifecycle is complete.

Flowdalic: Since ebuild.sh removes the up the file, the at exit hook just cleans up the (empty) directory.

I just shove all files under a directory to avoid cluttering the tmp directory.

Flowdalic: The `_emerge_tmpdir` dir is now world-readable.

zmedico: Removing the file in ebuild.sh will work, but the parent directory needs to be writable by the portage user.

Flowdalic: I don't see any current issues with the directory not being writable by the portage user. I assume all phases that hit the case are running as root. To make sure, I've added a `|| die` to the rm in the latest version of the patch.

Are you sure that there is an issue here?

zmedico: The depend phase should not run as root if userpriv is enabled. You can run egencache to trigger it.

Flowdalic: Alright, I will look into it. Thanks.

I wonder if we should set the sticky bit as well, as otherwise unprivileged phases could delete files from privileged ones.

zmedico: I gave it a spin, and the trap never executes because the depend phase never calls __ebuild_main.

Flowdalic: Got it, thanks!

So we need to consider three cases: A) ebuild phase with T available. B) privileged ebuild phase with T unavailable and C) phase=depend

For C, I have resurrected the life cycle managed via `EbuildMetadataPhase`. Please have a look at the latest revision of the patch.

And thanks for your assistance and patience. That was a tremendous help.

bin/phase-functions.sh
```diff
@@ -962,6 +963,18 @@ __ebuild_main() {
...
+		# Cleanup PORTAGE_EBUILD_EXTRA_SOURCE after ebuild.sh
+		# (__ebuild_main()) finishes if PORTAGE_EBUILD_EXTRA_SOURCE is
+		# not under T.
+		__portage_ebuild_exit() {
+			rm "${PORTAGE_EBUILD_EXTRA_SOURCE}" ||
+				die "failed to remove PORTAGE_EBUILD_EXTRA_SOURCE file (${PORTAGE_EBUILD_EXTRA_SOURCE})"
+		}
+		trap __portage_ebuild_exit EXIT
+	fi
+
```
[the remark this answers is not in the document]
zmedico: Actually, maybe it can be removed entirely, since it should never trigger due to the `${PORTAGE_EBUILD_EXTRA_SOURCE} != ${T}/*` condition.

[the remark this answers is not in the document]
Flowdalic: Adding
```bash
echo "Installing EXIT trap in EBUILD_PHASE=${EBUILD_PHASE}" | tee -a /tmp/phase
```
before installing the EXIT trap shows:
```
Installing EXIT trap in EBUILD_PHASE=clean
Installing EXIT trap in EBUILD_PHASE=cleanrm
Installing EXIT trap in EBUILD_PHASE=clean
```

so it is triggered.

However adding/adjusting
```python
            if phase == "depend":  # case C
                # The file will be deleted by EbuildMetadataPhase._unregister()
                mysettings["PORTAGE_EBUILD_EXTRA_SOURCE"] = ebuild_extra_source_path
            else:
                print(f"Should invoke EXIT trap in phase={phase}")
```
shows
```
Should invoke EXIT trap in phase=clean
Should invoke EXIT trap in phase=instprep
Should invoke EXIT trap in phase=cleanrm
Should invoke EXIT trap in phase=clean
```

Here, the `instprep` phase is additionally shown.  The `instprep` phase seems to have `T`, hence I added it to `is_ebuild_phase_with_t`.

Since `None` is also in `is_ebuild_phase_with_t`, we already handle `MiscFunctionsProcess`, by simply placing `PORTAGE_EBUILD_EXTRA_SOURCE` under `T`, where it will eventually be deleted.

PR #1408, opened 2024-12-31
Drop qa warnings

ulm:
As scheduled. :smile:

### DIFF
--- a/bin/ebuild-helpers/fperms
+++ b/bin/ebuild-helpers/fperms
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Copyright 1999-2023 Gentoo Authors
+# Copyright 1999-2024 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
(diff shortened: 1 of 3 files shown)


bin/ebuild-helpers/fperms
```diff
@@ -20,12 +20,6 @@ for arg; do
 		args+=( "${arg}" )
 	else
 		args+=( "${ED%/}/${arg#/}" )
-		# remove the QA warning after 2024-12-31
-		if [[ ${arg:0:1} != / ]]; then
-			eqawarn "QA Notice: ${0##*/}: Path '${arg}' does not start with '/'."
```
thesamesam, line 25: Should we make this a die?

ulm: No, why? The path is relative to `${ED}` and therefore need not start with a slash. See https://bugs.gentoo.org/905223 for a longer explanation.

thesamesam: Sorry, I'd completely forgot that context, my bad.

thesamesam (approved):

PR #1413, opened 2025-01-11
Remove support for 4-slot-abi EAPI

ulm:
Deprecated in commit db799d6ffe3a. Features are available in EAPI 5 and later.


zmedico (approved):
Thanks1

PR #1414, opened 2025-01-11
make.globals: Set PORTAGE_USERNAME and PORTAGE_GRPNAME variables

duxsco:
"man 5 make.conf" states that "PORTAGE_USERNAME" and "PORTAGE_GRPNAME" default to "portage":
- https://github.com/gentoo/portage/blob/597229aff02810764223a4cdd1e8056142d70ed0/man/make.conf.5#L1257-L1261
- https://github.com/gentoo/portage/blob/597229aff02810764223a4cdd1e8056142d70ed0/man/make.conf.5#L1109-L1113

But, the two variables are not set in the files pointed out by "man 5 make.conf":
https://github.com/gentoo/portage/blob/597229aff02810764223a4cdd1e8056142d70ed0/man/make.conf.5#L21-L31

This commits solves the discrepancy.

Bug: https://bugs.gentoo.org/941977

### DIFF
--- a/cnf/make.globals
+++ b/cnf/make.globals
@@ -55,6 +55,10 @@ BINPKG_GPG_VERIFY_BASE_COMMAND="/usr/bin/gpg --verify --batch --no-tty --no-auto
 # The binary package default GPG home directory for verify
 BINPKG_GPG_VERIFY_GPG_HOME="/etc/portage/gnupg"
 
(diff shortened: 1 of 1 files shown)


duxsco:
Testing on my laptop for now.

zmedico (approved):

PR #1424, opened 2025-01-17
Support for new home-dir-template-copy FEATURE

john-r-graham:
If enabled, as part of the ebuild environment setup, copies the PORTAGE_USERNAME home directory into the build environment HOME directory. Useful predominantly for ebuilds that use the git-r3 eclass, to give a place to put SSH keys and user- specific git configuration files used during the build.

See bug #947822 for rationale and additional details.

### DIFF
--- a/lib/portage/package/ebuild/prepare_build_dirs.py
+++ b/lib/portage/package/ebuild/prepare_build_dirs.py
@@ -135,7 +135,7 @@ def prepare_build_dirs(myroot=None, settings=None, cleanup=False):
         # a temporary PORTAGE_TMPDIR setting (for fetchonly).
         _prepare_features_dirs(mysettings)
         # Support for home-dir-template-copy FEATURE:
-        if not cleanup and "home-dir-template-copy" in settings.features:
+        if "home-dir-template-copy" in settings.features:
             portage_username = mysettings.get("PORTAGE_USERNAME", "portage")
             home_template_dir = pwd.getpwnam(portage_username).pw_dir
             build_env_home_dir = mysettings["HOME"]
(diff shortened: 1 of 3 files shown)


lib/portage/package/ebuild/prepare_build_dirs.py
```diff
@@ -133,6 +134,27 @@ def makedirs(dir_path):
         # Avoid spurious permissions adjustments when fetching with
         # a temporary PORTAGE_TMPDIR setting (for fetchonly).
         _prepare_features_dirs(mysettings)
+        # Support for home-dir-template-copy FEATURE:
+        if not cleanup and "home-dir-template-copy" in settings.features:
```
zmedico, line 138: I'm not sure that the `cleanup` flag should disable this feature.

john-r-graham: It's possible that I don't understand the meaning of the cleanup flag well enough. Let me go back to the drawing board for a bit. Should I move the pull request to "work in progress"?

zmedico: Sure, move it to "draft" if you're going back to the drawing board for a bit.

john-r-graham: You were right. The cleanup flag makes sure that remnants from a previous phase that might interfere with the next phase are deleted, but there still must be a full, correct build directory tree left behind, so the home directory template copy must be done regardless of the state of the cleanup flag.

 I stand by my decision not to do the copy in the info, fetch, and pretend phases. It wouldn't be _harmful,_ but it's a small efficiency.

Ready for review.

zmedico (approved):

PR #1426, opened 2025-02-02
Ban 'assert' and introduce 'pipestatus' instead in EAPI 9

Flowdalic:
Starting with EAPI 9, we decided to phase out 'assert', because, for example, its name is confusing. Instead 'assert' is going to be replaced with 'pipestatus'.

This also introduces the portage internal __pipestatus helper that can be used in bash code for all EAPIs.

Bug: https://bugs.gentoo.org/566342

### DIFF
--- a/bin/eapi9-pipestatus.sh
+++ b/bin/eapi9-pipestatus.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+# Copyright 2024-2025 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+# @FUNCTION: __pipestatus
+# @USAGE: [-v]
+# @RETURN: last non-zero element of PIPESTATUS, or zero if all are zero
+# @DESCRIPTION:
+# Check the PIPESTATUS array, i.e. the exit status of the command(s)
+# in the most recently executed foreground pipeline.  If called with
+# option -v, also output the PIPESTATUS array.
+__pipestatus() {
+	local status=( "${PIPESTATUS[@]}" )
+	local s ret=0 verbose=""
+
+	[[ ${1} == -v ]] && { verbose=1; shift; }
+	[[ $# -ne 0 ]] && die "usage: ${FUNCNAME} [-v]"
+
+	for s in "${status[@]}"; do
+		[[ ${s} -ne 0 ]] && ret=${s}
+	done
+
+	[[ ${verbose} ]] && echo "${status[@]}"
+
+	return "${ret}"
+}
--- a/bin/isolated-functions.sh
+++ b/bin/isolated-functions.sh
@@ -12,6 +12,13 @@ fi
 # It _must_ preceed all the calls to die and assert.
 shopt -s expand_aliases
 
+source "${PORTAGE_BIN_PATH}/eapi9-pipestatus.sh" || exit 1
+if ___eapi_has_pipestatus; then
+	pipestatus() {
+		__pipestatus "$@"
+	}
+fi
+
 if ___eapi_has_assert; then
 	assert() {
 		local x pipestatus=${PIPESTATUS[*]}
--- a/bin/isolated-functions.sh
+++ b/bin/isolated-functions.sh
@@ -12,12 +12,18 @@ fi
 # It _must_ preceed all the calls to die and assert.
 shopt -s expand_aliases
 
-assert() {
-	local x pipestatus=${PIPESTATUS[*]}
-	for x in ${pipestatus} ; do
-		[[ ${x} -eq 0 ]] || die "$@"
-	done
-}
+if ___eapi_has_assert; then
+	assert() {
+		local x pipestatus=${PIPESTATUS[*]}
+		for x in ${pipestatus} ; do
+			[[ ${x} -eq 0 ]] || die "$@"
+		done
+	}
+else
+	assert() {
+		die "assert is banned since EAPI 9 (current EAPI: ${EAPI}), use pipestatus instead"
+	}
+fi
 
 __assert_sigpipe_ok() {
 	# When extracting a tar file like this:
--- a/bin/save-ebuild-env.sh
+++ b/bin/save-ebuild-env.sh
@@ -50,7 +50,7 @@ __save_ebuild_env() {
 	done
 	unset x
 
-	unset -f assert __assert_sigpipe_ok \
+	unset -f assert __assert_sigpipe_ok __pipestatus \
 		__dump_trace die \
 		__quiet_mode __vecho __elog_base eqawarn elog \
 		einfo einfon ewarn eerror ebegin __eend eend KV_major \
@@ -91,6 +91,7 @@ __save_ebuild_env() {
 	___eapi_has_eapply && unset -f eapply eapply_user
 	___eapi_has_in_iuse && unset -f in_iuse
 	___eapi_has_version_functions && unset -f ver_cut ver_rs ver_test
+	___eapi_has_pipestatus && unset -f pipestatus
 
 	# Clear out the triple underscore namespace as it is reserved by the PM.
 	while IFS=' ' read -r _ _ REPLY; do
--- a/lib/portage/tests/bin/test_eapi9_pipestatus.py
+++ b/lib/portage/tests/bin/test_eapi9_pipestatus.py
@@ -1,0 +1,111 @@
+# Copyright 2018 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+import subprocess
+import tempfile
+import itertools
+
+from portage.const import PORTAGE_BIN_PATH
+from portage.tests import TestCase
+
+
+class TestEAPI9Pipestatus(TestCase):
+    test_script_prelude = """\
+pipestatus() {
+	__pipestatus "$@"
+}
+
+tps() {
+    local cmd=${1}
+    eval "${cmd}; pipestatus"
+    echo $?
+}
+
+tpsv() {
+    local cmd=${1}
+    local out ret
+    out=$(eval "${cmd}; pipestatus -v")
+    ret=$?
+    echo $ret $out
+}
+
+ret() {
+    return ${1}
+}
+
+die() {
+    if [[ $@ ]]; then
+        2>&1 echo "$@"
+    fi
+    exit 42
+}
+"""
+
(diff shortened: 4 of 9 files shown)


ulm:
There should be separate commits for `assert` and `pipestatus` as well as separate feature test functions.

bin/save-ebuild-env.sh
```diff
@@ -50,7 +50,7 @@ __save_ebuild_env() {
 	done
 	unset x
 
-	unset -f assert __assert_sigpipe_ok \
+	unset -f __pipestatus __assert_sigpipe_ok \
```
ulm, line 53: `assert` should be kept here because it still exists as a banned function.
```suggestion
	unset -f assert __pipestatus __assert_sigpipe_ok \
```

bin/save-ebuild-env.sh
```diff
@@ -92,6 +92,12 @@ __save_ebuild_env() {
 	___eapi_has_in_iuse && unset -f in_iuse
 	___eapi_has_version_functions && unset -f ver_cut ver_rs ver_test
 
+	if ___eapi_has_pipestatus; then
+		unset -f pipestatus
+	else
+		unset -f assert
+	fi
```
ulm, line 99: ... and it is not needed here:
```suggestion
	___eapi_has_pipestatus && unset -f pipestatus
```

bin/eapi9-pipestatus.sh
```diff
@@ -0,0 +1,30 @@
...
+	[[ $# -ne 0 ]] && die "usage: ${FUNCNAME} [-v]"
+
+	for s in "${status[@]}"; do
+		[[ ${s} -ne 0 ]] && ret=${s}
+	done
+
+	[[ ${verbose} ]] && echo "${status[@]}"
+
+	return "${ret}"
+}
```
ulm, line 19: AFAICS the body of this function is an exact copy from `eapi9-pipestatus.eclass`. It would have been nice to mention this in the commit message and give credit to its author. :smile:

ulm (changes requested):

bin/save-ebuild-env.sh
Flowdalic: Good catch. That was a leftover from a previous version of the patch.

bin/eapi9-pipestatus.sh
Flowdalic: Done

Flowdalic:
> There should be separate commits for `assert` and `pipestatus` as well as separate feature test functions.

I feel like banning `assert` and introducing `pipestatus` should be one atomic change. And I feel similar about the feature test functions. One could argue that the feature test function could be renamed to `__eapi_has_pipestatus_and_assert_banned` for clarity. How about that?

Unrelated to your comment above. How do you (and others) feel about

```
pipestatus() {
    __pipestatus $@
}
```

vs
```
alias pipestatus=__pipestatus
```

I considered using an alias for that, as it felt cleaner and clearly expresses the intent. But maybe there are downsides I am not aware of?

ulm:
> > There should be separate commits for `assert` and `pipestatus` as well as separate feature test functions.
> 
> I feel like banning `assert` and introducing `pipestatus` should be one atomic change. And I feel similar about the feature test functions. One could argue that the feature test function could be renamed to `__eapi_has_pipestatus_and_assert_banned` for clarity. How about that?

This is definitely not one atomic change. Read the [log of the 2024-12-08 council meeting](https://projects.gentoo.org/council/meeting-logs/20241208.txt) where several council members asked for the `assert` ban to be revisited before the final approval of EAPI 9. You commit would create a fait accompli preventing this. So please split it and also make them separate feature tests. Also, these are separate features in PMS.

> Unrelated to your comment above. How do you (and others) feel about
> 
> ```
> pipestatus() {
>     __pipestatus $@
> }
> ```
> 
> vs
> 
> ```
> alias pipestatus=__pipestatus
> ```
> 
> I considered using an alias for that, as it felt cleaner and clearly expresses the intent. But maybe there are downsides I am not aware of?

That would be the only alias in the Portage code and it feels brittle. As a matter of fact, all previously defined aliases (with `assert` being one of them) were removed or replaced by a function.

Anyway, the `pipestatus` function should use `"$@"` rather than `$@` (sorry that I missed that before):

```
pipestatus() {
    __pipestatus "$@"
}
```

ulm:
Two more comments:

1.  Why do you define the functions in a separate `eapi9-pipestatus.sh` file instead of `isolated-function.sh`? It seems unnecessary.
2. Please don't define `pipestatus` and then unset it again in EAPIs that don't have the function. `unset -f pipestatus` is just horrible.

Flowdalic:
> 1. Why do you define the functions in a separate `eapi9-pipestatus.sh` file instead of `isolated-function.sh`? It seems unnecessary.

Having the code isolated in a single file makes testing easier. It also follows the pattern of `eapi7-ver-funcs.sh`.

bin/isolated-functions.sh
```diff
@@ -12,12 +12,25 @@ fi
...
+	}
+fi
+
+if ___eapi_has_assert; then
+	assert() {
+		local x pipestatus=${PIPESTATUS[*]}
+		for x in ${pipestatus} ; do
+			[[ ${x} -eq 0 ]] || die "$@"
+		done
+	}
```
ulm, line 28: I wonder if this should be simplified and use `__pipestatus` internally?
```suggestion
	assert() {
		__pipestatus || die "$@"
	}
```

ulm:
> > 1. Why do you define the functions in a separate `eapi9-pipestatus.sh` file instead of `isolated-function.sh`? It seems unnecessary.
> 
> Having the code isolated in a single file makes testing easier. It also follows the pattern of `eapi7-ver-funcs.sh`.

Yes, but then follow its pattern entirely and source the file only when `___eapi_has_pipestatus` is true. IMHO it makes no sense to split out a single function to its own file which is then unconditionally sourced.

bin/save-ebuild-env.sh
```diff
@@ -91,6 +91,7 @@ __save_ebuild_env() {
 	___eapi_has_eapply && unset -f eapply eapply_user
 	___eapi_has_in_iuse && unset -f in_iuse
 	___eapi_has_version_functions && unset -f ver_cut ver_rs ver_test
+	___eapi_has_pipestatus && -f pipestatus
```
ulm, line 94:
```suggestion
	___eapi_has_pipestatus && unset -f pipestatus
```

Flowdalic:
> Yes, but then follow its pattern entirely and source the file only when `___eapi_has_pipestatus` is true. IMHO it makes no sense to split out a single function to its own file which is then unconditionally sourced.

Sourcing the file conditionally has disadvantages, as it's` __pipestatus` is used internally by portage as EAPI agnostic `assert`/`pipestatus`. If it would only be available in certain EAPIs, then we would need to guard every `portage-internal` use of assert/pipestatus with an EAPI condition.

And again, `__pipestatus()` resides in its own file to facilitate its unit testing.

ulm:
> Sourcing the file conditionally has disadvantages, as it's` __pipestatus` is used internally by portage as EAPI agnostic `assert`/`pipestatus`. If it would only be available in certain EAPIs, then we would need to guard every `portage-internal` use of assert/pipestatus with an EAPI condition.
> 
> And again, `__pipestatus()` resides in its own file to facilitate its unit testing.

Obviously we disagree here. So, I guess it is for the Portage maintainers to decide whether a function with 11 SLOC should live in its own file.

thesamesam:
Sorry, I'd completely forgot this PR existed, or I of course wouldn't have merged that one with overlapping changes.

Flowdalic:
> I agree it makes sense for it to be in its own file if there's no way you can think of to unittest it otherwise.

I fiddled around with it a little bit by sourcing `phase-function.sh` (where `pipestatus` is currently defined). This results in a lot of noise because `phase-functions.sh` really expect to be sourced by `ebuild` and I ultimately did not get it to work. But I only investigated an half hour of my time doing so. Maybe there is a way, but I don't see it and I am not inclined to spend more time on it.

Having the function under test in an extra module (which, in case of bash, the next best thing seems to be a file) just seems very sensible from a software engineering perspective, as it avoids all the side-effects of that `phase-functions.sh` brings with it.

lib/portage/tests/bin/test_eapi9_pipestatus.py
```diff
@@ -0,0 +1,111 @@
...
+            ("true | false | true", 1),
+            ("true |false | ret 5 | true", 5),
+        ]
+        test_cases = itertools.starmap(lambda a, b: (f"tps {a}", f"{b}"), test_cases)
+        self._test_pipestatus(test_cases)
+
+    def test_pipestatus_v(self):
+        test_cases = [
+            # (command, expected exit status, expected output)
+            ("true | true | true", 0, "0 0 O"),
```
ulm, line 84: Does this actually work? Looks like a typo.

ulm:
> I fiddled around with it a little bit by sourcing `phase-function.sh` (where `pipestatus` is currently defined). This results in a lot of noise because `phase-functions.sh` really expect to be sourced by `ebuild` and I ultimately did not get it to work. But I only investigated an half hour of my time doing so. Maybe there is a way, but I don't see it and I am not inclined to spend more time on it.

Not sure what I am missing here, but why would you have to source `phase-functions.sh`? AFAICS, assigning `EAPI=9` and then sourcing `eapi.sh` and `phase-helpers.sh` would work.

(And if in the long term we are to add more unit tests for the base code, we'll have to solve this problem anyway.)

lib/portage/tests/bin/test_eapi9_pipestatus.py
```diff
@@ -0,0 +1,111 @@
...
+            ("true", 0),
+            ("false", 1),
+            ("true | true", 0),
+            ("false | true", 1),
+            ("true | false", 1),
+            ("ret 2 | true", 2),
+            ("true | false | true", 1),
+            ("true |false | ret 5 | true", 5),
+        ]
+        test_cases = itertools.starmap(lambda a, b: (f"tps {a}", f"{b}"), test_cases)
```
ulm, line 78: Shouldn't the parameter in `tps {a}` be quoted? Otherwise the test script will contain lines like `tps true | true` which produces no output.

Makes me wonder how these test cases could have passed. Maybe we need a meta unit test for the unit test. :smile:

PR #1427, opened 2025-02-07
Show @lru_cache stats if PORTAGE_SHOW_LRU_CACHE_INFO env var is set

Flowdalic:
This introduces the PORTAGE_SHOW_LRU_CACHE_INFO environment variable.

Setting the PORTAGE_SHOW_LRU_CACHE_INFO environment variable allows to analyze the effectiveness and potential bottlenecks of the @lru_cache annotation used by portage. The cache information will be printed after dependency calculation and after emerge "finished".

Example output:
```
Portage @lru_cache information
use_reduce_cached: hit ratio: 71.25% (total: 48575, hits: 34611, misses: 13964) util: 100.00% (1024 / 1024)
encoded_length: hit ratio: 0.00% (total: 0, hits: 0, misses: 0) util: 0.00% (1024 / 0) 
vercmp: hit ratio: 69.50% (total: 1777, hits: 1235, misses: 542) util: 52.93% (1024 / 542) 
catpkgsplit: hit ratio: 66.90% (total: 191511, hits: 128129, misses: 63382) util: 100.00% (10240 / 10240) 
get_eapi_attrs: hit ratio: 100.00% (total: 267206, hits: 267200, misses: 6) util: 18.75% (32 / 6)
```

### DIFF
--- a/lib/portage/util/portage_lru_cache.py
+++ b/lib/portage/util/portage_lru_cache.py
@@ -1,0 +1,42 @@
+# Copyright 2025 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+import os
+import portage
+
+
+def show_lru_cache_info():
+    if not os.environ.get("PORTAGE_SHOW_LRU_CACHE_INFO"):
+        return
+
+    portage_lru_caches = {
+        portage.dep._use_reduce_cached: "use_reduce_cached",
+        portage.eapi._get_eapi_attrs: "get_eapi_attrs",
+        portage.process._encoded_length: "encoded_length",
+        portage.versions.catpkgsplit: "catpkgsplit",
+        portage.versions.vercmp: "vercmp",
+    }
+
+    print("Portage @lru_cache information")
+    for method, name in portage_lru_caches.items():
+        cache_info = method.cache_info()
+
+        hits = cache_info.hits
+        misses = cache_info.misses
+        maxsize = cache_info.maxsize
+        currsize = cache_info.currsize
+
+        total = hits + misses
+        if total:
+            hitratio = hits / total
+        else:
+            hitratio = 0
+
+        if maxsize:
+            utilization = currsize / maxsize
+        else:
+            utilization = 0
+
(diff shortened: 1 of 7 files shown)


lib/portage/util/portage_lru_cache.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2025 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+import os
+import portage
+
+
+def show_lru_cache_info():
+    if not os.environ.get("PORTAGE_SHOW_LRU_CACHE_INFO"):
+        return
```
zmedico, line 10: Maybe make this a FEATURES setting instead of environment variable?

Flowdalic: I considered that. But it feels more like a debug thing that can be enabled via an env var, than a user facing env var. That is, I feel like to should be more hidden from users.That said, I wouldn't oppose a FEATURE if there is consensus that it's the right thing to do.

Flowdalic: > I considered that. But it feels more like a debug thing, that can be enabled via an env var, than a user facing FEATURE setting. I feel like to should be more hidden from users.That said, I wouldn't oppose a FEATURE if there is consensus that it's the right thing to do.

eli-schwartz: I don't think this makes sense as a FEATURES. That implies that users are supposed to report whether it's set as part of e.g. attaching `emerge --info` to a bug report... it also implies that users are expected to set this in make.conf so it prints on every run, rather than running it as part of development profiling.

zmedico: Okay, I agree that FEATURES is not a good fit for this.

zmedico (approved):

thesamesam:
Can you add something into NEWS as well as the man page (noting that it may disappear in future, output is unstable, it's for development only)?

Flowdalic:
> Can you add something into NEWS as well as the man page (noting that it may disappear in future, output is unstable, it's for development only)?

Added NEWS item, man page section, and rebased on the lastest master.

PR #1429, opened 2025-03-10
selinux: fully disable selinux-awareness for FEATURES="-selinux"

WavyEbuilder:
It should be expected that FEATURES="-selinux" will completely disable selinux awareness; the system administrator should be able to recover from a situation where labeling is invalid using portage without having to fully disabling SELinux from being loaded by the kernel.

### DIFF
--- a/lib/portage/package/ebuild/config.py
+++ b/lib/portage/package/ebuild/config.py
@@ -3446,7 +3446,7 @@ class config:
     def selinux_enabled(self):
         if getattr(self, "_selinux_enabled", None) is None:
             self._selinux_enabled = 0
-            if "selinux" in self["USE"].split():
+            if "selinux" in self.features:
                 if selinux:
                     if selinux.is_selinux_enabled() == 1:
                         self._selinux_enabled = 1
--- a/man/make.conf.5
+++ b/man/make.conf.5
@@ -736,6 +736,11 @@ dependencies.
 .B sandbox
 Enable sandbox\-ing when running \fBemerge\fR(1) and \fBebuild\fR(1).
 .TP
+.B selinux
+Enable SELinux awareness.  Portage will install objects to the filesystem with
+a SELinux context calculated from the current loaded policy.
+Do not toggle this \fBFEATURE\fR yourself.
+.TP
 .B sesandbox
 Enable SELinux sandbox\-ing.  Do not toggle this \fBFEATURE\fR yourself.
 .TP


lib/portage/package/ebuild/config.py
```diff
@@ -3446,7 +3446,7 @@ def archlist(self):
     def selinux_enabled(self):
         if getattr(self, "_selinux_enabled", None) is None:
             self._selinux_enabled = 0
-            if "selinux" in self["USE"].split():
+            if "selinux" in self["USE"].split() and "selinux" in self.features:
```
floppym, line 3449: I don't really see why we are checking USE at all here. I think we should check FEATURES exclusively.

In trying to track this down, I found eb2b07ef27bc8f790e2ab280bbd19a012841f9cf which seems relevant.

WavyEbuilder: I don't really see a reason to check USE either here, I assumed there was a reason for it, but looking at that commit doesn't seem like it's necessary (or desired). Also worth noting that `selinux` is missing from `make.conf(5)` under the `FEATURES` section, so we should probably add that in. However as this is a fairly big change, cc'ing @perfinion and @0xC0ncord .

floppym (approved):
This looks good to me, but I will wait until we hear from the selinux folks or another Portage maintainer before merging.

perfinion (approved):
I think this way around is much better.

I think it came from that the visible part of profiles/features/selinux is that we turn on the USE-flag. any package that has selinux deps always has them behind teh selinux use-flag. eg chrony would have `selinux? ( sec-policy/selinux-chronyd )` which gets loaded into the system-wide policy. 

There are two different things that portage does regarding SELinux:

1. it relabels everything in `${D}` right before it merges into the live system so they have the right labels. every file on an selinux system always has a label regardless of if the package knows anything about selinux. Even if there is no app-specific policy or an selinux USE-flag you can't skip relabelling otherwise you'd be missing the basic `bin_t` and `ls` wouldnt work. 
2. The other thing that portage does is transition to a separate domain for the sandbox, `sesandbox` is just an extra layer on top of the regular `FEATURES=sandbox`  The only time i recall having to touch sesandbox was probably when the PID namespacing rules had not yet made it into policy (easily more than a decade ago tho).

If for some reason a systems policy or policycoreutils are completely broken, then the path to fix it all would be rebuild but dont relabel, then once everything is rebuilt you relabel the entire system. `FEATURES="-selinux" emerge -av @world` makes most sense to me for that. the `sesandbox` would also be disabled in that case and thats fine. I can think of cases where you want portage to label but the sandbox interferes, i cant really think of a case where you'd use the sesandbox but would skip relabelling.

thesamesam:
LGTM, but @WavyEbuilder, can you include some of perfinion's comment in your commit message too?

WavyEbuilder:
> LGTM, but @WavyEbuilder, can you include some of perfinion's comment in your commit message too?

Will do

PR #1435, opened 2025-04-30
Remove LD_PRELOAD environment variable mask

htv04:
This pull request removes LD_PRELOAD from the environment variables masked by Portage.

* Portage's sandbox appears to support prepending to an existing LD_PRELOAD: https://github.com/gentoo/sandbox/blob/7240cc9fb7cd6a9bc83b9ade5a385633f42e58d1/data/sandbox.bashrc#L4-L8
* When the sandbox is disabled, LD_PRELOAD doesn't appear to be used at all by Portage.

However, Portage currently masks LD_PRELOAD, nullifying this functionality. Having LD_PRELOAD as a Portage variable can be useful for various reasons, in my specific case using a custom allocator when supported by the build process.

I have tested this myself and it appears to work properly for the above use case, with and without the sandbox enabled (although the sandbox has its own custom allocator).

### DIFF
--- a/lib/portage/package/ebuild/_config/special_env_vars.py
+++ b/lib/portage/package/ebuild/_config/special_env_vars.py
@@ -112,6 +112,7 @@ environ_whitelist = frozenset(
         "FEATURES",
         "FILESDIR",
         "HOME",
+        "LD_PRELOAD",
         "MAKEFLAGS",
         "MERGE_TYPE",
         "NOCOLOR",


thesamesam:
The commit message should always match the PR description. The commit also needs `Signed-off-by`.

htv04:
> The commit message should always match the PR description. The commit also needs `Signed-off-by`.

@thesamesam Got it, thanks for the heads up. Commit fixed.

eli-schwartz:
I would also recommend reflowing the commit message, following markdown style. This is conventional for the rest of the repository already.

See e.g. https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

URLs, and triple-backticks code blocks, of course are an exception to the "72 characters" rule of thumb.

htv04:
> I would also recommend reflowing the commit message, following markdown style. This is conventional for the rest of the repository already.
> 
> See e.g. https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
> 
> URLs, and triple-backticks code blocks, of course are an exception to the "72 characters" rule of thumb.

I believe the commit message does already adhere to markdown, are you saying I should just have it wrap to 72 characters?

floppym:
I don't have any specific objection, but my gut tells me this is dangerous.

floppym:
I think this would cause LD_PRELOAD to be saved in the environment file, which also gets saved inside of binpkgs. That could be problematic.

I think some more thought needs to be put into this.

htv04:
> I don't have any specific objection, but my gut tells me this is dangerous.

@floppym Inherently `LD_PRELOAD` is a dangerous environment variable, so I have to agree there. But I don't think it's dangerous to not blacklist it in Portage's operations when desired. Especially when `LD_PRELOAD` already affects Portage to an extent until the ebuild environment is prepared (as seen by the configuring and install phases, in testing with an invalid library name).

However, `LD_PRELOAD` shouldn't be set by default in normal situations and therefore I think it's allowance here should be fine. If `LD_PRELOAD` is set globally, that already has other dangerous implications.

eli-schwartz:
> I believe the commit message does already adhere to Markdown, are you saying I should just have it wrap to 72 characters? Or use `-` instead of `*` for bullet points?

"reflow, using the same rules that markdown uses for reflowing".

I was simply clarifying that two paragraphs are defined by a blank line in between two blocks of text, not by the newline character. It is also quite "conventional" for people writing markdown to adhere to the same 72/80 character lines.

htv04:
> > I believe the commit message does already adhere to Markdown, are you saying I should just have it wrap to 72 characters? Or use `-` instead of `*` for bullet points?
> 
> "reflow, using the same rules that markdown uses for reflowing".
> 
> I was simply clarifying that two paragraphs are defined by a blank line in between two blocks of text, not by the newline character. It is also quite "conventional" for people writing markdown to adhere to the same 72/80 character lines.

I understand now. Will fix after the environment issue mentioned by @floppym is resolved.

eli-schwartz:
Feels like we just want this variable respected when set, for the duration of the `src_*` phases on the system that has set the variable, right? But it probably shouldn't propagate from `src_*` --> `pkg_*`.

I don't know that portage actually has a way to specify that sort of logical split...

floppym:
> Feels like we just want this variable respected when set, for the duration of the src_* phases on the system that has set the variable, **right?**

I think adding LD_PRELOAD to [environ_whitelist](https://github.com/gentoo/portage/blob/portage-3.0.68/lib/portage/package/ebuild/_config/special_env_vars.py#L83) would basically accomplish that: it would allow LD_PRELOAD to "leak in" from the calling environment for each phase instead of being sourced from the environment file.

htv04:
> > Feels like we just want this variable respected when set, for the duration of the src_* phases on the system that has set the variable, **right?**
> 
> I think adding LD_PRELOAD to [environ_whitelist](https://github.com/gentoo/portage/blob/portage-3.0.68/lib/portage/package/ebuild/_config/special_env_vars.py#L83) would basically accomplish that: it would allow LD_PRELOAD to "leak in" from the calling environment for each phase instead of being sourced from the environment file.

@floppym Interesting, so how is this different from preventing from being unset? Could LD_PRELOAD be "unset" and still leak through to the compile phase with this?

Also, this would work just for my specific use case, for whatever reason someone may want to use `LD_PRELOAD` to impact all of Portage's operations. Is there a way we could just avoid it being stored persistently in any way?

floppym:
I cannot think of any valid reason to preserve LD_PRELOAD in the environment file.

htv04:
> I cannot think of any valid reason to preserve LD_PRELOAD in the environment file.

What environment file? I might be misunderstanding here. The only functionality I'm trying to achieve with this PR is the allowance of `LD_PRELOAD` during the runtime of Portage's operations. Anything that preserves it, shouldn't, but as you said here...

> I think this would cause LD_PRELOAD to be saved in the environment file, which also gets saved inside of binpkgs. That could be problematic.

...it seems that not blacklisting it preserves it somehow. Is there a way it could *only* be used at runtime? I'm willing to implement it on my end if it isn't too much trouble.

eli-schwartz:
> What environment file? I might be misunderstanding here. The only functionality I'm trying to achieve with this PR is the allowance of `LD_PRELOAD` during the runtime of Portage's operations. Anything that preserves it, shouldn't, but as you said here...

https://dev.gentoo.org/~ulm/pms/head/pms.html#section-11.2

The file this PR modifies, exists solely to produce a file called "the environment file", which portage uses to implement the linked part of the package manager specification.

"The runtime of portage's operations" includes installing binary packages, and uninstalling packages sometime in the future, and running `emerge --config`...

floppym:
Right, what Eli said.

Portages uses a file called "environment" to store the state of variables between phases. This file exists as `${T}/environment` during execution, is saved in metadata.tar for binpkgs, and is saved under `/var/db/pkg/${CATEGORY}/${PR}/environment.bz2` for installed packages.

I am saying that I don't think we should preserve LD_PRELOAD in the environment file since it is likely we don't want to have it set when installing a binpkg on another system.

I would be happier with a PR to modify the environ_whitelist variable I pointed out.

htv04:
I see. Just fixed both the whitelist and the commit message formatting. Seems to work fine on my end and no issues seem to be present. Let me know if this implementation is okay.

htv04:
Just wanted to check in, everything okay with my latest commit?

floppym (approved):

PR #1436, opened 2025-05-06
estrip: salt build IDs with ${CATEGORY}/${PF}:${SLOT}

thesamesam:
Use debugedit to salt build IDs with ${CATEGORY}/${PF}:${SLOT} to avoid file collisions (same build ID generated as another package) where a bundled library is installed.

Bug: https://bugs.gentoo.org/549672
Bug: https://bugs.gentoo.org/953869

### DIFF
--- a/bin/estrip
+++ b/bin/estrip
@@ -338,9 +338,26 @@ save_elf_debug() {
 			# Symlink so we can read the name back.
 			__try_symlink "${dst}" "${inode_debug}"
 
-			# If we don't already have build-id from debugedit, look it up
+			# If we don't already have build-id from debugedit, look it up.
+			# This should only happen with FEATURES=-installsources, as
+			# it's done in save_elf_sources.
 			if [[ -z ${buildid} ]] ; then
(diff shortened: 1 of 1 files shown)


mgilbert-crain:
It looks like we only call debugedit in save_elf_sources if FEATURES=installsources is enabled.

Should we call debugedit in other scenarios?

floppym:
It looks like we only call debugedit in save_elf_sources if FEATURES=installsources is enabled.

Should we call debugedit in other scenarios?

thesamesam:
Good spot. We need to do it with splitdebug too (as that's when we'll have a build-id symlink created, see what we do in `save_elf_debug`) or we'll have collisions.

thesamesam:
Attempted that, not yet tested, but I don't really love how it looks. Ideas?

floppym:
I think this would probably work.

A slightly cleaner approach might be to always call debugedit in save_elf_sources, but vary the parameters depending on FEATURES=installsources.

PR #1440, opened 2025-05-31
Implement FEATURE `suppress-bindist-buildpkg`

Kangie:
This feature combines with `buildpkg` to enable the installation of packages with "RESTRICT=bindist" without the creation of a binpkg.

This should help binhost builders improve the library of pre-packaged software without running the risk of accidentally (bin)packaging a dependency that we can't legally distribute (or don't want to pay royalties on!)

**Current Behaviour**:
FEATURES=buildpkg + RESTRICT=bindist
	=> Package is installed, binpkg created
FEATURES=buildpkg + RESTRICT=bindist + ACCEPT_RESTRICT=-bindist
	=> Package is masked, cannot be selected

**New Behaviour**:
FEATURES=buildpkg + RESTRICT=bindist
	=> Package is installed, binpkg created
FEATURES=buildpkg suppress-bindist-buildpkg + RESTRICT=bindist
	=> Package is installed, binpkg creation stage skipped
FEATURES=* + RESTRICT=bindist + ACCEPT_RESTRICT=-bindist
	=> Any combination of features and the above are masked

Bug: https://bugs.gentoo.org/542480

### DIFF
--- a/lib/portage/tests/emerge/test_buildpkg.py
+++ b/lib/portage/tests/emerge/test_buildpkg.py
@@ -1,0 +1,482 @@
+# Copyright 2025 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+import subprocess
+import sys
+
+import portage
+from portage import _unicode_decode, os
+from portage.const import (
+    PORTAGE_PYM_PATH,
+    USER_CONFIG_PATH,
+    SUPPORTED_GENTOO_BINPKG_FORMATS,
+)
+from portage.process import find_binary
+from portage.tests import TestCase
+from portage.tests.resolver.ResolverPlayground import ResolverPlayground
+from portage.util import ensure_dirs
+from portage.output import colorize
+
+
+# We test the buildpkg functionality at a basic level with all binary package formats,
+# but avoid repeating the same tests for each format to avoid redundancy (just use gpkg / the fastest).
+class BuildpkgTestCase(TestCase):
+    """
+    Test suite for binary package creation using the 'buildpkg' feature in Portage.
+
+    This class focuses on verifying the correct behavior of binary package
+    creation in various scenarios, including:
+
+    - Basic binary package creation with the 'buildpkg' feature enabled.
+    - Handling of packages with RESTRICT=bindist.
+    - The 'suppress-bindist-buildpkg' feature, which prevents binary package
+        creation for packages with RESTRICT=bindist.
+    - Integration between 'suppress-bindist-buildpkg' and package installation.
+
+    The tests use a ResolverPlayground to simulate a Portage environment and
+    assert that binary packages are created or not created as expected based
+    on the configuration and package restrictions.
+    """
+
+    def testBasicBuildpkg(self):
+        """
+        Test basic binary package creation with FEATURES=buildpkg flag.
+        Verify that binary packages are actually created.
+        """
+        debug = False
+
+        from portage.tests.emerge.conftest import _INSTALL_SOMETHING
+
+        ebuilds = {
+            "dev-libs/test-basic-1::test": {
+                "EAPI": "8",
+                "SLOT": "0",
+                "KEYWORDS": "x86",
+                "MISC_CONTENT": _INSTALL_SOMETHING,
+            },
+        }
+
+        for binpkg_format in SUPPORTED_GENTOO_BINPKG_FORMATS:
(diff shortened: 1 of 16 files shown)


man/make.conf.5
```diff
@@ -796,6 +796,14 @@ Have portage react strongly to conditions that may conflict with system
 security provisions (for example textrels, executable stack).  Read about
 the \fIQA_STRICT_*\fR variables in \fBmake.conf\fR(5).
 .TP
+.B suppress\-bindist\-binpkg
```
thesamesam, line 799: Maybe add something about how it only does something if ACCEPT_RESTRICT excludes bindist? I know that's obvious, but it makes things clearer for users trying to understand how things fit together.

lib/portage/tests/resolver/test_accept_restrict_bindist.py
```diff
@@ -0,0 +1,300 @@
...
+                "RESTRICT": "test bindist strip",
+            },
+            # Package that depends on bindist-restricted package
+            "dev-libs/depends-on-bindist-1": {
+                "EAPI": "7",
+                "RDEPEND": "dev-libs/bindist-restricted",
+            },
+        }
+
+        # Test default behavior (ACCEPT_RESTRICT="*" - should accept all)
```
thesamesam, line 42: Thanks for the nice comments, I like seeing these for testcases especially.

lib/portage/tests/emerge/test_buildpkg.py
```diff
@@ -0,0 +1,483 @@
...
+                + (
+                    "--oneshot",
+                    "dev-libs/test-integration",
+                ),
+            )
+
+            success = self._run_commands(playground, test_commands, debug)
+            self.assertTrue(
+                success,
+                "Package with RESTRICT=bindist should be installable with suppress-bindist-buildpkg feature",
```
thesamesam: ... where we don't have ACCEPT_RESTRICT="... -bindist", right?

man/make.conf.5
```diff
@@ -796,6 +796,14 @@ Have portage react strongly to conditions that may conflict with system
...
+When enabled, this feature prevents the creation of binary packages for
+ebuilds that have the \fIbindist\fR restriction. This allows users to
+install such packages normally while ensuring that no binary packages
```
thesamesam, line 802: Perhaps explicitly state what happens when it's disabled (the default)?

lib/portage/tests/emerge/test_buildpkg.py
Kangie: Yes, right - that would result in the package being masked. I'll clear that up.

pyproject.toml
```diff
@@ -3,7 +3,7 @@ name = 'portage'
 dynamic = ['version']
 description = 'Portage is the package management and distribution system for Gentoo'
 readme = 'README.md'
-requires-python = '>=3.9'
+requires-python = '>=3.10'
```
eli-schwartz, line 6: If your only reason to do this is that the Gentoo repository doesn't support it anymore then just don't do it?

I'm generally quite opposed to bumping versions just for the heck of it, without actually benefitting from doing so. No, dropping it from a CI matrix doesn't count.

Consider also, that dropping python versions means non-Gentoo entities who also use portage, may have their workflow broken. e.g. stable LTS system builders, who should at least be given the opportunity to use any non-EOL version. Also uhhhh the elephant in the room being Google, who have interesting perspectives on which version of Python to support. :)

If you can't even muster up a *single* code change to make as a result of dropping 3.9 support then it indicates your whole claim of dropping support is entirely hollow.

...

Also, please propose completely unrelated distinctive changes in different PRs, for better reviewability.

eli-schwartz (changes requested):

pyproject.toml
Kangie: Well, since you got here before I had a chance to comment:

I've bumped the minimum requirement to py3.10 as this change fails with py3.9 (I used `match` in a test), which wasn't picked up until CI runs because I don't develop against EoL'd (or almost EoL'd) Python.

It wouldn't be hard to rework the code to work with older python, but I also don't see much point in making such a change if we're not actually sure that we want py3.9 anyway. Since we're reviewing changes, I opted to drop the older version (which was privately discussed earlier as a possibility). I'm equally happy to just use `if/else` if we _really_ want old python support.

PR #1442, opened 2025-06-09
EAPI 9

ulm:


thesamesam:
LGTM but maybe let's wait until after the release I'll be tagging soon (there's a lot in master I'd like to flush out).

bin/isolated-functions.sh
```diff
@@ -23,7 +23,11 @@ fi
 shopt -s expand_aliases
 
 assert() {
-	IFS='|' expression=${PIPESTATUS[*]} let '! expression' || die "$@"
+	local x pipestatus=( "${PIPESTATUS[@]}" )
+	___eapi_has_assert || die "'${FUNCNAME}' banned in EAPI ${EAPI}"
```
Flowdalic, line 27: Shouldn't we also adjust usages of `assert` by portage code when introducing this? Like I did in the alternative PR to this https://github.com/gentoo/portage/pull/1453, for example https://github.com/gentoo/portage/pull/1426/files#diff-4eacf848f4d2cbea61f0c0e5285e082233b162cd5d096566a20947f484961b4dL543-R543

Flowdalic:
I am sorry if I am raising this after it got merged, but I was on vacation for the last two weeks. And I wasn't aware that the PR with the title "EAPI 9" also includes `pipestatus` related changes, which I already worked on in https://github.com/gentoo/portage/pull/1426

The change introducing `pipestatus` in this PR does not adjust  the`ebuild(5)` man page regarding the availability of `pipestatus` and the unavailability of `assert` in newer EAPIs, it does not have a unit test, and does not fix the portage internal call sites of `asssert`. All those points are addressed in https://github.com/gentoo/portage/pull/1426

ulm:
Sorry, somehow I had forgotten that #1426 was still in the queue. I guess the best thing would be to rebase it onto current master.

Also, I'm not in favour of `pipestatus` living in its own file. It should be defined in the same place where `die` and `assert` are defined, and IMHO splitting the source into many tiny files for every function makes no sense. Why can't you simply source `isolated-functions.sh` in the unit test? Most ebuild helpers (e.g. `bin/ebuild-helpers/doins`) do this, and they're stand-alone shell scripts.

ulm:
> `pipestatus` is not declared in `isolated-fucntions.sh` and `phase-helpers.sh` is not independently sourceable (same is true for `ebuild.sh`).

But it could be? Or `__pipestatus` could, and `pipestatus` would call it.

PR #1443, opened 2025-06-15
binarytree: Fix local vs remote timestamp comparision

Flowdalic:
There seems to be a slight timestamp inconsistency that causes the "local_timestamp != remote_timestap" expression to always evaluate to true.

For example, on one of my systems, using
https://distfiles.gentoo.org/releases/amd64/binpackages/23.0/x86-64 as as binhost, I end up with

remote timestamp: 1749900593
local  timestamp: 1749907793

The local timestamp is newer than the remote one. However portage still performs expensive operations if "local_timestamp != remote_timestaps", causing a noticeable delay. Interestingly, it will still print

    Local copy of remote index is up-to-date and will be used.

Changing the condition from

    local_timestamp != remote_timestamp

to

    local_timestamp < remote_timestamp

makes portage nearly immediately print the "Local copy … is up-to-date" message, where it previously would take some seconds.


Flowdalic:
@zmedico could I bother you to have a look at this. It makes emerge significantly snappier for me. But I am not absolutely sure if it is correct.

thesamesam:
> I assume the timestamp inconsistency is due somewhere portage converting a UTC timestamp into the local timezone.

Nice catch!

I spotted this:
```python
def _timestamp_to_http(timestamp):
    dt = datetime.fromtimestamp(float(int(timestamp) + TIMESTAMP_TOLERANCE))
    stamp = mktime(dt.timetuple())
    return formatdate(timeval=stamp, localtime=False, usegmt=True)
```

The docs for https://docs.python.org/3/library/time.html#time.mktime say that its argument is in local time?

Flowdalic:
Of course those timestamps are strings and have to be converted to int before we perform the less than comparison.

PR #1445, opened 2025-06-19
improve portage.dep.Atom regex parsing, conformance fixes

Jannik2099:
this simplifies the Atom parser logic just a little, I hope.

Since we now enforce formatting, I also took the liberty to reorganize the regexes into something that you can actually read and comprehend. I put every capture group into it's own indent.

Lastly, I found a couple cases of non-conforming behaviour:
* Unicode identifiers have never been allowed in PMS (see also https://bugs.gentoo.org/435934 ).
* Python regex is secretly evil, `$` matches `\n$` - replace with `\Z`, which has the intended effect
* categories, packages and slots must not begin with a plus
* disallow ":/subslot" slots


thesamesam:
The commit message for [`f54b874` (#1445)](https://github.com/gentoo/portage/pull/1445/commits/f54b874b09e1ae7a820cf5e5924e5a9fc25d19c6) should include something like you had in the PR description:
> Unicode identifiers have never been allowed in PMS (see also [bugs.gentoo.org/435934](https://bugs.gentoo.org/435934) ). (and a Bug: tag for that)

The explicit other commentary from the PR desc should be propagated into the commit messages where appropriate too. Assume GH may disappear :)

lib/portage/dep/__init__.py
```diff
@@ -125,25 +125,24 @@ def _get_atom_re(eapi_attrs: portage.eapi._eapi_attrs) -> re.Pattern:
...
-        + _slot_loose
-        + ")?"
-        + _repo
-        + ")("
-        + _use
-        + ")?\\Z",
+        "^"
+        + (
+            "(?P<without_use>"
+            + (
```
eli-schwartz: It is mysterious to me, that you think this is more readable than, say,

```python
rf"""
    ^
        (?P<without_use>
            (?:
                (?P<op>{_op}{cpv_re})
                |
                (?P<star>={cpv_re}\*)
                |
                (?P<simple>{cp_re})
            )

        ({_slot_separator}{_slot_loose})?
        {_repo}
        )
    ({_use})?
    \Z"""

**Review by eli-schwartz — CHANGES_REQUESTED** [review]

File: lib/portage/dep/__init__.py (hunk 1 above)
Comment (Jannik2099) [reply lib/portage/dep/__init__.py]: because I just didn't think of using multi-line strings. Will fix tomorrow.

Comment (eli-schwartz) [reply lib/portage/dep/__init__.py]: Mysteriously, the code already used re.VERBOSE but never bothered to make use of it??? :D

File: lib/portage/dep/__init__.py [hunk 2]
```diff
@@ -143,7 +143,7 @@ def _get_atom_re(eapi_attrs: portage.eapi._eapi_attrs) -> re.Pattern:
         + _repo
         + ")("
         + _use
-        + ")?$",
+        + ")?\\Z",
```
mgorny: I'm fine with this change, but I think we should also look into using `.fullmatch()` wherever applicable.

lib/_emerge/is_valid_package_atom.py
```diff
@@ -7,7 +7,7 @@
 
 def insert_category_into_atom(atom, category):
     # Handle '*' character for "extended syntax" wildcard support.
-    alphanum = re.search(r"[\*\w]", atom, re.UNICODE)
+    alphanum = re.search(r"[\*\w]", atom, re.ASCII)
```
mgorny, line 10: Mixed feelings here. PMS yes, but it was generally considered fine for local repositories.

lib/portage/dep/__init__.py
```diff
@@ -1555,7 +1565,7 @@ def __init__(
             cp = m_group(base + 3)
             slot = m_group(atom_re.groups - 2)
             repo = m_group(atom_re.groups - 1)
-            use_str = m_group(atom_re.groups)
+            use_str = m_group("usedeps")
```
mgorny, line 1579: Any reason not to replace the remaining uses as well?

Jannik2099: Done so for `slot` and `repo`.

I've tried to untangle `cp` and `cpv` and put them into groups, but there's so much validation logic that relies on their group indices that I just gave up.
It's also made harder by the fact that you can't reuse capture group names, so for example the `op | star | simple` variant would require capture groups named `op_cp`, `star_cp`, `simple_cp` etc.

lib/_emerge/is_valid_package_atom.py
Jannik2099: As per the resolution in https://bugs.gentoo.org/435934 I think unicode is considered flat out unsupported?

mgorny: Oh, so that's why I had a vague recollection of seeing Unicode in ebuild names. Yeah, it's fine then.

PR #1447, opened 2025-06-22
Improve binhost status message, show timestamps when repository is up-to-date

Flowdalic:


Flowdalic:
> Maybe include some before/after in the commit messages?

Good point. I had it in an early version of this patchset, but seems that it got lost in a later rebase. Re-added again.

lib/portage/dbapi/bintree.py
```diff
@@ -1585,13 +1590,28 @@ def _populate_remote(self, getbinpkg_refresh=True, pretend=False):
...
+                    desc = "up-to-date"
+                    if remote_timestamp:
+                        extra_info = f" (local: {convUnixTs(local_timestamp)}, remote: {convUnixTs(remote_timestamp)})"
                 writemsg_stdout("\n")
                 writemsg_stdout(
                     colorize(
                         "GOOD",
-                        _("[%s] Local copy of remote index is %s and will be used.")
-                        % (binrepo_name, desc),
+                        _("[%s] Local copy of remote index is %s and will be used%s.")
```
eli-schwartz: This is very noisy IMO and I'd expect it to be controlled by e.g. a `--debug` flag.

lib/portage/dbapi/bintree.py
```diff
@@ -1594,8 +1600,8 @@ def _populate_remote(self, getbinpkg_refresh=True, pretend=False):
                 # This includes URLError which is raised for SSL
                 # certificate errors when PEP 476 is supported.
                 writemsg(
-                    _("\n\n!!! Error fetching binhost package" " info from '%s'\n")
-                    % _hide_url_passwd(base_url)
+                    _("\n\n!!! [%s] Error fetching binhost package" " info from '%s'\n")
```
eli-schwartz: This feels somehow noisy if we are already printing the URL that failed.

lib/portage/dbapi/bintree.py
Flowdalic: You are right, it is noisy. But tying this behavior to the `--debug` flags feels like a touch to much, i.e., I want to see this message in cases where I don't want to see the other `--debug` output. It feels like something that should be behind `-vv` (which portage currently has not).

Latest version of this PR ties this behavior behind `--verbose`.

lib/portage/dbapi/bintree.py
Flowdalic: I thought the same but added it never the less because of symmetry. However, I have no strong feelings, if there is consensus that we don't want it here then it can be removed.

lib/portage/dbapi/bintree.py
```diff
@@ -1640,7 +1673,7 @@ def _populate_remote(self, getbinpkg_refresh=True, pretend=False):
                         error_msg = str(
                             uerror.object, encoding="utf_8", errors="replace"
                         )
-                    writemsg(f"!!! {error_msg}\n\n")
+                    writemsg(f"!!! [${binrep_name}] {error_msg}\n\n")
```
thesamesam: Shouldn't this be:
```suggestion
                    writemsg(f"!!! [{binrepo_name}] {error_msg}\n\n")
```

lib/portage/dbapi/bintree.py
```diff
@@ -1628,8 +1658,11 @@ def _populate_remote(self, getbinpkg_refresh=True, pretend=False):
                     # This includes URLError which is raised for SSL
                     # certificate errors when PEP 476 is supported.
                     writemsg(
-                        _("\n\n!!! Error fetching binhost package" " info from '%s'\n")
-                        % _hide_url_passwd(base_url)
+                        _(
```
thesamesam, line 1661: OK but we should remove these gettext wrappers at some point, we're not using them, and I don't see us translating Portage any time soon.

lib/portage/dbapi/bintree.py
```diff
@@ -1596,22 +1607,41 @@ def _populate_remote(self, getbinpkg_refresh=True, pretend=False):
...
+                                ),
                                 noiselevel=-1,
                             )
                     break
                 except UseCachedCopyOfRemoteIndex:
                     changed = False
                     rmt_idx = pkgindex
                     if getbinpkg_refresh or repo.frozen:
-                        desc = "frozen" if repo.frozen else "up-to-date"
+                        extra_info = ""
```
thesamesam, line 1622: You don't need to handle this now, but we really need to try combat the excessive levels of indentation we have, it makes things so hard to follow.

lib/portage/dbapi/bintree.py
Flowdalic: I thought the same, but this was outside of the scope of this PR.

lib/portage/dbapi/bintree.py
Flowdalic: Doing some heavy work in `bintree.py` and while doing it causing another level of indentation, I totally feel you.

eli-schwartz: This is your daily reminder that that is because of black, not the indents.

thesamesam: Can you show me where your proposal was submitted?

thesamesam: (And maybe, but Portage has problems w/ not factoring out functions and having complicated control flow anyway.)

PR #1448, opened 2025-07-05
Hostname support in titles for all titles

Steering7253:
Move hostname support to the xtermTitle function, instead of (only two of) the places that call it. Notably, this adds the hostname to the initial static "emerge" display when syncing or calculating dependencies. It also removes duplicated code.

You could also pull my master branch which goes a step further and doesn't require people to export HOSTNAME from their shell.


thesamesam:
Mind switching the PR to use the patches from your master, and then add `Signed-off-by`? Thanks.

Steering7253:
Updated

PR #1450, opened 2025-08-19
file_copy: borrow list of non-fatal cfr errors from coreutils

floppym:
Bug: https://bugs.gentoo.org/961738


thesamesam (approved):
Let's try this but I'm OK with the other approach if we have to.

PR #1452, opened 2025-08-28
process: do not poll join() in MultiprocessingProcesss

Flowdalic:
Based on the initial analysis of Esteve Varela Colominas, the polling on join() with the fixed 100ms delay incurs a significant performance penalty, especially for short-lived processes. And since portage is prone to spawning many of those, the penalty adds up easily.

Instead of pooling proc.join() with a fixed 100ms delay, we now use the blocking variant of join() started in an executor which we await in the coroutine.

Before this change:
```
( cd lib; python3 -m timeit 'import portage.process; portage.process.spawn("true")' ) 
2 loops, best of 5: 104 msec per loop
```

After this change:

```
( cd lib; python3 -m timeit 'import portage.process; portage.process.spawn("true")' )
50 loops, best of 5: 4.45 msec per loop
```

Bug: https://bugs.gentoo.org/958635

### DIFF
--- a/lib/portage/process.py
+++ b/lib/portage/process.py
@@ -426,10 +426,6 @@ class MultiprocessingProcess(AbstractProcess):
     An object that wraps OS processes created by multiprocessing.Process.
     """
 
-    # Number of seconds between poll attempts for process exit status
-    # (after the sentinel has become ready).
-    _proc_join_interval = 0.1
-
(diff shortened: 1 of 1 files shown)


mid-kid:
That was... deceptively simple. I have no idea where you found this option 😅 
Thanks a lot for this!

zmedico (approved):
Looks good. Thanks!

PR #1454, opened 2025-09-09
bintree: Accelerate index fetch by requesting Packages.gz first

Flowdalic:
Fetching the compressed package index (Packages.gz) significantly reduces the time required to transfer the index. At the time of writing this, the size of the uncompressed package index of the official Gentoo binhost is ~16MiB, while the compressed variant's size is ~1.7MiB.

Portage got support writing the a compressed package index with 11c0619c63b5 ("Portage writes a compressed copy of 'Packages' index file."). Unfortunately, this did not include *fetching* the compressed package index.

This commit adjusts Portage logic to first try to fetch the compressed package index with a fallback to the uncompressed one.

The diff of this change looks wild, but essentially it introduced only four new lines of code:
- a new loop: for remote_pkgindex_file in ("Packages.gz", "Packages"):
  - which required indenting the existing code by one level
- a break statement at the end of the loop (if it was successful)
- wrapping the existing file object into GzipFile if required via if remote_pkgindex_file == "Packages.gz": f = GzipFile(fileobj=f, mode="rb")

Bug: https://bugs.gentoo.org/248089


thesamesam:
Could you rebase? CI on both of your PRs got stuck for some reason.

Flowdalic:
> Could you rebase? CI on both of your PRs got stuck for some reason.

rebased

PR #1459, opened 2025-09-11
Python 2 cleanup

Flowdalic:


lib/portage/xml/metadata.py
```diff
@@ -246,11 +246,7 @@ def _get_herd_email(self, herd):
...
             return None
 
-        try:
-            # Python 2.7 or >=3.2
-            iterate = self._herdstree.iter
-        except AttributeError:
-            iterate = self._herdstree.getiterator
+        iterate = self._herdstree.iter
 
         for node in iterate("herd"):
```
thesamesam, line 251: No need to do this now but I'll note we haven't had herds since GLEP 67 (10 years!)

lib/portage/elog/mod_save_summary.py
```diff
@@ -77,11 +77,6 @@ def process(mysettings, key, logentries, fulltext):
 
     time_fmt = "%Y-%m-%d %H:%M:%S %Z"
     time_str = time.strftime(time_fmt, time.localtime(time.time()))
-    # Avoid potential UnicodeDecodeError in Python 2, since strftime
-    # returns bytes in Python 2, and %Z may contain non-ascii chars.
-    time_str = _unicode_decode(
```
thesamesam, line 82: and of course we still have the general unicode layer crap which was just for Python 2 :/

PR #1462, opened 2025-09-12
Fixed a minor bug in writemsg() from lib/portage/util

john-r-graham:
When passing a regular text file descriptor to writemsg(), some bug-avoidance code was not identifying all of the file descriptor types that don't need Unicode encoding. The file descriptor was being tested for being an instance of io.StringIO instead of the base class for text streams (io.TextIOBase). As such, passing a file descriptor that was opened in the most natural way for a text file, for example

    fd = open(logfile, "w")

would cause an exception because Unicode encoding would be applied unnecessarily, creating a binary object on which fd.write() would raise an exception.

### DIFF
--- a/lib/portage/util/__init__.py
+++ b/lib/portage/util/__init__.py
@@ -112,7 +112,7 @@ def writemsg(mystr: str, noiselevel: int = 0, fd: Optional[TextIO] = None) -> No
         fd = sys.stderr
     if noiselevel <= noiselimit:
         # avoid potential UnicodeEncodeError
-        if isinstance(fd, io.StringIO):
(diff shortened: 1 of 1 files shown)


Flowdalic:
The change looks good to me. However, I was unable to test it. Therefore, could you
- improve the commit subject so that explains a bit what it's doing? Maybe "Use TextIoBase instead of StringIo to ...."
- Show us a minimal reproducer of code that previous threw an exception. And consider adding that reproducer to the commit message

john-r-graham:
The CI/Build test that failed _appears_ to have failed for reasons other than my patch.

john-r-graham:
Forced checks to repeat; all tests passed.

PR #1463, opened 2025-09-12
process: do not poll join() in MultiprocessingProcesss (v2)

Flowdalic:
Based on the initial analysis of Esteve Varela Colominas, the polling on join() with the fixed 100ms delay incurs a significant performance penalty, especially for short-lived processes. And since portage is prone to spawning many of those, the penalty adds up easily.

Instead of pooling proc.join() with a fixed 100ms delay, we now use the blocking variant of join() started in a thread which we await in the coroutine.

Before this change:

( cd lib; python3 -m timeit 'import portage.process; portage.process.spawn("true")' ) 2 loops, best of 5: 104 msec per loop

After this change:

( cd lib; python3 -m timeit 'import portage.process; portage.process.spawn("true")' ) 50 loops, best of 5: 4.45 msec per loop

Bug: https://bugs.gentoo.org/958635

### DIFF
--- a/lib/portage/process.py
+++ b/lib/portage/process.py
@@ -428,10 +428,6 @@ class MultiprocessingProcess(AbstractProcess):
     An object that wraps OS processes created by multiprocessing.Process.
     """
 
-    # Number of seconds between poll attempts for process exit status
-    # (after the sentinel has become ready).
-    _proc_join_interval = 0.1
-
     def __init__(self, proc: multiprocessing.Process):
         self._proc = proc
         self.pid = proc.pid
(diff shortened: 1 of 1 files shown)


lib/portage/process.py
```diff
@@ -482,13 +478,7 @@ async def _proc_join(self):
...
                 pass
 
-        # Now that proc.sentinel is ready, poll until process exit
-        # status has become available.
-        while True:
-            proc.join(0)
-            if proc.exitcode is not None:
-                break
-            await asyncio.sleep(self._proc_join_interval, loop=loop)
+        await _asyncio.to_thread(proc.join)
```
zmedico: I looked into it, and unfortunately `asyncio.to_thread` calls run_in_executor, so it suffers from the same problem described in https://github.com/gentoo/portage/pull/1452#issuecomment-3286298392.

zmedico: We could use a simple implementation of `to_thread` like this:
```python
def _thread_wrapper(loop, future, func, args, kwargs):
    try:
        result = func(*args, **kwargs)
    except Exception as e:
        loop.call_soon_threadsafe(future.set_exception, e)
    else:
        loop.call_soon_threadsafe(future.set_result, result)

async def to_thread(func, *args, **kwargs):
    loop = asyncio.get_running_loop()
    future = loop.create_future()
    t = threading.Thread(
        target=_thread_wrapper, args=(loop, future, func, args, kwargs)
    )
    t.start()
    return await future
```

zmedico: I tested the simple `to_thread` implementation, and Socks5ServerAtExitTestCase hangs with the multiprocessing fork start method, while it does not hang with the spawn start method. Maybe it's just not safe to start a thread during shutdown with the multiprocessing fork start method.

So, maybe it makes sense to fall back to the old _proc_join_interval approach during shutdown.

Flowdalic: Seems like we worked in parallel on the problem. :) I've just pushed what I came up with and am curious about the CI results. I think the fallback to old _proj_join_interval during shutdown does not fix the issue for when the `fork` start method is active. At least, that is what my experiments have shown. But please see if you can come up with something. It would be great to have the non-poll variant with the `fork` method. But maybe we can get at least the non-poll variant going with the `spawn` method (which should be the default from Python 3.14 on, right?).

zmedico: I've just re-merged `dev-lang/python-3.14.9999` and it seems that the current default is forkserver. However, forkserver raises the same RuntimeError as spawn, so we can use one of these conditions:
```python
multiprocessing.get_start_method() in ("forkserver", "spawn")
multiprocessing.get_start_method() != "fork"
```

PR #1466, opened 2025-09-14
bintree: Prefer 'Last-Modified' over 'timestamp' header

Flowdalic:
The Last-Modified header is a standard header and seems to be more in sync accross different mirrors than 'timestamp'. Prefer it over 'timestamp'.


thesamesam (approved):
OK with news update

Flowdalic:
> .. maybe split the typo fix vs the 'prefer' change, actually?

Both done

PR #1468, opened 2025-09-14
github ci: run CI also with Python 3.14 forkserver

Flowdalic:
Python 3.14 uses the forkserver multiprocessing start method per default. However, our current CI configuration only ran with the 'spawn' method when using Python 3.14.

This adjusts the CI configuration matrix to also run with Python 3.14 and the forkserver start method.


zmedico:
I've merged the changes from this PR in https://github.com/gentoo/portage/pull/1483 .

PR #1469, opened 2025-09-15
bintree: Improve UseCachedCopyOfRemoteIndex handling, report if TTL

Flowdalic:
Move the logic where parts of the user-exposed message is constructed from the except block to right before we raise the UseCachedCopyOfRemoteIndex exception. This means we can get rid of some logic in the except block, which tried to determine what the cause of the UseCachedCopyOfRemoteIndex exception was.

As benefit, we will no report if the use the cached copy if it is within its TTL.

### DIFF
--- a/lib/portage/dbapi/bintree.py
+++ b/lib/portage/dbapi/bintree.py
@@ -1666,7 +1666,7 @@ class binarytree:
                             colorize(
                                 "GOOD",
                                 _(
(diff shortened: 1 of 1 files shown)


Flowdalic:
Sorry, there where some commits/changes hanging in my local branch that aren't pushed. See
https://github.com/gentoo/portage/pull/1470
https://github.com/gentoo/portage/pull/1471

PR #1472, opened 2025-09-15
bintree: fix UnboundLocalError due to extra_info

Flowdalic:


Flowdalic:
We went with https://github.com/gentoo/portage/pull/1473

PR #1474, opened 2025-09-16
repos.conf: set sync-git-verify-commit-signature to 'true'

Flowdalic:
https://wiki.gentoo.org/wiki/Portage_with_Git recommends

    eselect repository enable gentoo

to enable git-based sync of ::gentoo. However, since Portage does not enable sync-git-verify-commit-signature per default, this ends up in an insecure configuration (when compared with an rsync based sync).

Fix this by enabling the setting per default for the ::gentoo repository in Portage.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -14,6 +14,9 @@ Features:
 * Prefer the 'Last-Modified' over the 'timestamp' HTTP header when
   querying binhosts.
 
+* Enable sync-git-verify-commit-signature for the 'gentoo' repository
+  per default.
+
 Bug fixes:
 
 * Fix timestamp comparison for Packages index by using UTC consistently
--- a/cnf/repos.conf
+++ b/cnf/repos.conf
@@ -17,3 +17,4 @@ sync-openpgp-key-refresh-retry-delay-exp-base = 2
 sync-openpgp-key-refresh-retry-delay-max = 60
 sync-openpgp-key-refresh-retry-delay-mult = 4
 sync-webrsync-verify-signature = yes
+sync-git-verify-commit-signature = true


thesamesam:
There's at least one bug tag to add here, but I'll note that we may need a news item for people using gentoo.git w/o verification (the raw repo) and mention they want to use our developer keyring?

Flowdalic:
> There's at least one bug tag to add here,

You mean https://bugs.gentoo.org/901385? Seems only tangential related. Or do you mean some other issue?


> but I'll note that we may need a news item for people using gentoo.git w/o verification (the raw repo)

Right, if we change portage's default, then we probably need that. (We wouldn't need that if this where solved via eselect-repository).

thesamesam:
I mean https://bugs.gentoo.org/959831 (and I'd include the eselect-repo bug on gh too), though I would tag https://bugs.gentoo.org/901385 too.

> Right, if we change portage's default, then we probably need that. (We wouldn't need that if this where solved via eselect-repository).

Yeah, it's a bit awkward, but on the other hand, I think we should do it either way in portage even if we later make eselect-repo improvements.

lib/portage/repository/config.py
```diff
@@ -269,9 +269,27 @@ def __init__(self, name, repo_opts, local_config=True):
...
+        if sync_openpgp_key_refresh == "yes":
+            sync_openpgp_key_refresh = "true"
+        elif sync_openpgp_key_refresh == "no":
+            sync_openpgp_key_refresh = "false"
+        elif sync_openpgp_key_refresh not in (
+            "true",
+            "false",
+            "wkd",
+            "keyserver",
+            "false-nowarn",
```
thesamesam, line 284: The only thing I'm not sure about is `nowarn` as it feels a bit ugly, but I don't have a better solution, so I can live with it if nobody has a better idea.

lib/portage/repository/config.py
```diff
@@ -269,9 +269,27 @@ def __init__(self, name, repo_opts, local_config=True):
...
+        ):
+            writemsg(
+                f"!!! Invalid sync-openpgpg-key-refresh setting for repo {name}: {sync_openpgp_key_refresh}\n",
```
thesamesam, line 287:
```suggestion
                f"!!! Invalid sync-openpgp-key-refresh setting for repo {name}: {sync_openpgp_key_refresh}\n",
```

lib/portage/sync/syncbase.py
```diff
@@ -282,11 +284,12 @@ def _refresh_keys(self, openpgp_env):
...
             return
 
-        out.ebegin("Refreshing keys via WKD")
-        if openpgp_env.refresh_keys_wkd():
-            out.eend(0)
-            return
-        out.eend(1)
+        if self.repo.sync_openpgp_key_refresh in ("true", "wkd"):
+            out.ebegin("Refreshing keys via WKD")
+            if openpgp_env.refresh_keys_wkd():
```
thesamesam, line 289: We bypass the retry logic here. That feels unexpected, or at least needs some acknowledgement in the commit message and a comment I think.

thesamesam (commented):
Some small comments. You're OK to handle the news item before the release containing this?

The news item shouldn't scare people too much and explain how to know if they're affected.

lib/portage/sync/syncbase.py
Flowdalic: > We bypass the retry logic here.

Are we? Maybe it's to early for me, but I don't see which retry logic is bypassed here.

thesamesam: No, it was the other way around: too early for _me_.

PR #1475, opened 2025-09-19
Add support for sccache

idealseal:
This depends on https://github.com/projg2/shadowman/pull/4 and a patched `sccache` ebuild.

It works exactly as the `ccache` implementation.

Docs are still missing. Maybe this can be combined with the `ccache` instructions?

### DIFF
--- a/bin/ebuild.sh
+++ b/bin/ebuild.sh
@@ -121,7 +121,7 @@ export PORTAGE_BZIP2_COMMAND=${PORTAGE_BZIP2_COMMAND:-bzip2}
 
 # These two functions wrap sourcing and calling respectively.  At present they
 # perform a qa check to make sure eclasses and ebuilds and profiles don't mess
-# with shell opts (shopts).  Ebuilds/eclasses changing shopts should reset them 
+# with shell opts (shopts).  Ebuilds/eclasses changing shopts should reset them
 # when they are done.
 
 __qa_source() {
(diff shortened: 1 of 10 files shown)


idealseal:
I'm thinking about how rust could be integrated: should we export `RUSTC_WRAPPER` and where would this happen? Should this happen in the eclasses?

thesamesam:
Can you explain what this is useful for, given the build cache issue (it keys on the full path)? Are you really using it with live ebuilds enough for it to pay off?

eli-schwartz:
I would personally be less concerned about "how often is it even used, and is it worth the hassle" and be more concerned about the exact opposite problem:

People will use it.

And it will be a bad thing that they use it. They will enable it because they think it sounds like a good thing and then they will suffer a pure loss due to the cache fullpath issue.

I'm strongly against merging sccache support in any form, unless it at least answers the question of how to make sure users have a fighting chance of understanding when it's a good idea vs when it's a bad idea. Mostly indifferent as long as that is handled.

One possibility of the top of my head is to make the FEATURES=sccache be ignored unless PROPERTIES=live is also set?

mgorny:
Isn't the path issue specific to Rust?

thesamesam (commented):
Can you update the man page(s) too? (Just search for ccache.)

The changes look good to me (not tested yet though). Do I need to do any setup?

idealseal:
0.17.0 is out, I'm trying to take another look at it, there is a client-side architecture which might just solve all my issues.

PR #1476, opened 2025-09-19
github ci: create gtar symlink

zmedico:
Fixes: e18f0aab473b3bfaa4507bc600dcb27b23125d19


eli-schwartz:
Thanks. I didn't have time to enhance my commit and @thesamesam wanted to merge it fast to unbreak users on portage-9999.ebuild.

I think @thesamesam wanted portage itself to fall back to `tar` if `gtar` isn't found on PATH. That will need a slightly more elaborate fix though...

PR #1497, opened 2025-11-01
Atom: drink up the string soup

Jannik2099:
First step to making `portage.dep.Atom` properly typed.

It still retains a `__str__` method, as that will always be useful. Some call sites will still do something like `str(atom).split(...)`, because they usually operate on variant types of `str | Atom | foo` and are completely untyped :(

Some properties of `Atom` should probably be dropped entirely, and some more be added. For memory efficiency, we may also want to synthesize some of them from the string representation as needed, instead of storing them.

Also, it seems that `depgraph.py` is more cursed than previously thought possible, and writes to `Atom._orig_atom`. I've just added it as a field for now...

Ultimately there's still a lot more `str(atom)` users than I'd like, hence this is just a draft.
There are a few remaining test failures that I am working on right now.

Note that there's also a selinux sandbox commit to fix some tests locally, that's ofc not related and I'll drop the commit later. Working with @WavyEbuilder to figure out why this only appears in tests.

### DIFF
--- a/bin/ebuild.sh
+++ b/bin/ebuild.sh
@@ -186,6 +186,12 @@ addread "${PORTAGE_TMPDIR}/portage"
 addwrite "${PORTAGE_TMPDIR}/portage"
 [[ -n ${PORTAGE_GPG_DIR} ]] && addpredict "${PORTAGE_GPG_DIR}"
 
+if [[ -f /sys/fs/selinux/enforce ]]; then
+	addpredict /proc/self/attr/fscreate
+	addpredict /proc/thread-self/attr/fscreate
+	addpredict /proc/self/attr/current
+fi
+
 # Avoid sandbox violations in temporary directories.
 if [[ -w ${T} ]] ; then
 	export TEMP=${T}
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -1753,8 +1753,8 @@ def action_deselect(settings, trees, opts, atoms):
         expanded_atoms = set(atoms)
 
         for atom in atoms:
-            if not atom.startswith(SETPREFIX):
-                if atom.cp.startswith("null/"):
+            if not str(atom).startswith(SETPREFIX):
+                if atom.category == "null":
                     # try to expand category from world set
                     null_cat, pn = portage.catsplit(atom.cp)
                     for world_atom in world_atoms:
@@ -1762,7 +1762,7 @@ def action_deselect(settings, trees, opts, atoms):
                         if pn == world_pn:
                             expanded_atoms.add(
                                 Atom(
-                                    atom.replace("null", cat, 1),
+                                    str(atom).replace("null", cat, 1),
                                     allow_repo=True,
                                     allow_wildcard=True,
                                 )
@@ -1775,13 +1775,13 @@ def action_deselect(settings, trees, opts, atoms):
         discard_atoms = set()
         for atom in world_set:
             for arg_atom in expanded_atoms:
-                if arg_atom.startswith(SETPREFIX):
-                    if atom.startswith(SETPREFIX) and arg_atom == atom:
+                if str(arg_atom).startswith(SETPREFIX):
+                    if str(atom).startswith(SETPREFIX) and arg_atom == atom:
                         discard_atoms.add(atom)
                         break
                 else:
                     if (
-                        not atom.startswith(SETPREFIX)
+                        not str(atom).startswith(SETPREFIX)
                         and arg_atom.intersects(atom)
                         and not (arg_atom.slot and not atom.slot)
                         and not (arg_atom.repo and not atom.repo)
@@ -1789,13 +1789,13 @@ def action_deselect(settings, trees, opts, atoms):
                         discard_atoms.add(atom)
                         break
         if discard_atoms:
-            for atom in sorted(discard_atoms):
+            for atom in sorted(discard_atoms, key=str):
                 if pretend:
                     action_desc = "Would remove"
                 else:
                     action_desc = "Removing"
 
-                if atom.startswith(SETPREFIX):
+                if str(atom).startswith(SETPREFIX):
                     filename = "world_sets"
                 else:
                     filename = "world"
@@ -2095,7 +2095,7 @@ def action_info(settings, trees, myopts, myfiles):
                 if not atom.blocker:
                     atoms.append((x, atom))
 
-    myvars = sorted(set(atoms))
+    myvars = sorted(set(atoms), key=lambda t: str(t[0]))
 
     cp_map = {}
     cp_max_len = 0
@@ -3956,11 +3956,11 @@ def run_action(emerge_config):
                     # look at the ebuilds, since EAPI 4 allows running pkg_info
                     # on non-installed packages
                     valid_atom = dep_expand(x, mydb=vardb)
-                    if valid_atom.cp.split("/")[0] == "null":
+                    if valid_atom.category == "null":
                         valid_atom = dep_expand(x, mydb=portdb)
 
                     if (
-                        valid_atom.cp.split("/")[0] == "null"
+                        valid_atom.category == "null"
                         and "--usepkg" in emerge_config.opts
                     ):
                         valid_atom = dep_expand(x, mydb=bindb)
--- a/lib/portage/dbapi/dep_expand.py
+++ b/lib/portage/dbapi/dep_expand.py
@@ -47,7 +47,7 @@ def dep_expand(mydep, mydb=None, use_cache=1, settings=None):
 
     if has_cat:
         # Optimize most common cases to avoid calling cpv_expand.
-        if not mydep.cp.startswith("virtual/"):
+        if mydep.category != "virtual":
             return mydep
         if not hasattr(mydb, "cp_list") or mydb.cp_list(mydep.cp):
             return mydep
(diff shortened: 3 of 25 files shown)


lib/portage/dep/__init__.py
```diff
@@ -2480,13 +2593,15 @@ def match_from_list(mydep, candidate_list):
     if not candidate_list:
         return []
 
+    if isinstance(mydep, Atom):
+        mydep = str(mydep)
+
```
zmedico: The `unevaluated_atom` attribute is lost here, which causes [the testSlotCollision failure](https://github.com/gentoo/portage/pull/1497#issuecomment-3476640634).

Jannik2099: Superb find, thanks!

Yeah, this is why I think there's too many `foo = str(atom)` conversions in this PR right now. Honestly a miracle that I only managed to break a single test?

Jannik2099: Hmm, not sure.

There's only one place where the `unevaluated_atom` attribute is getattr'd with a fallback, in `paren_enclose`. All other accesses to the attribute would throw if they were operating on a str, so this is the only site where dropping the attribute could go wrong, right?

Yet if I patch that function to do

```
            if unevaluated_atom:
                if not hasattr(x, "unevaluated_atom"):
                    x = Atom(x).unevaluated_atom
                else:
                    x = getattr(x, "unevaluated_atom", x)
```

it still fails.

zmedico: The specific code in match_from_list that will cause testSlotCollision to fail if you don't preserve unevaluated_atom is this:
```python
    if mydep.unevaluated_atom.use:
        candidate_list = mylist
        mylist = []
        for x in candidate_list:
            use = getattr(x, "use", None)
            if use is not None:
                if mydep.unevaluated_atom.use and not x.iuse.is_valid_flag(
                    mydep.unevaluated_atom.use.required
                ):
                    continue
```

lib/_emerge/actions.py
```diff
@@ -2095,7 +2095,7 @@ def action_info(settings, trees, myopts, myfiles):
                 if not atom.blocker:
                     atoms.append((x, atom))
 
-    myvars = sorted(set(atoms))
+    myvars = sorted(set(atoms), key=lambda t: str(t[0]))
```
zmedico, line 2098: It looks like we can omit this sort because myvars is only used to populate the cp_map variable.

lib/portage/dbapi/dep_expand.py
```diff
@@ -47,12 +47,12 @@ def dep_expand(mydep, mydb=None, use_cache=1, settings=None):
...
+        if mydep.category != "virtual":
             return mydep
         if not hasattr(mydb, "cp_list") or mydb.cp_list(mydep.cp):
             return mydep
         # Fallback to legacy cpv_expand for old-style PROVIDE virtuals.
         mydep = mydep.cp
 
     expanded = cpv_expand(mydep, mydb=mydb, use_cache=use_cache, settings=settings)
-    return Atom(orig_dep.replace(mydep, expanded, 1), allow_repo=True)
+    return Atom(str(orig_dep).replace(mydep, expanded, 1), allow_repo=True)
```
zmedico, line 58: Should pass `str(mydep)` to replace in case it's an Atom.

lib/portage/dep/dep_check.py
```diff
@@ -1091,7 +1089,7 @@ def dep_wordreduce(mydeplist, mysettings, mydbapi, mode, use_cache=1):
                     mydep = mydbapi.match(deplist[mypos], use_cache=use_cache)
                 if mydep is not None:
                     tmp = len(mydep) >= 1
-                    if deplist[mypos][0] == "!":
+                    if str(deplist[mypos]).startswith("!"):
```
zmedico, line 1092: They say startswith is marginally slower than slice comparison, and for the large number of these it might make sense to avoid startswith.

zmedico:
> I discovered that dep_getcpv, dep_getusedeps and get_operator have no uses within portage. Can these be removed, or are they considered public API?

They are public but deprecated, so let's just add a DeprecationWarning for now.

Actually, we could use UserWarning to make them more noisy.

lib/portage/dep/dep_check.py
zmedico: Actually, it's better if we use the Atom.blocker attribute here since that also avoids the `str` conversion.

lib/portage/dep/__init__.py
```diff
@@ -2355,15 +2438,12 @@ def dep_getkey(mydep):
     return mydep.cp
 
 
-def match_to_list(mypkg, mylist):
+def match_to_list(mypkg: Union[str, Atom], mylist: list) -> list:
```
zmedico, line 2441: mypkg should not be Atom, but _pkg_str or Package.

lib/portage/dep/__init__.py
```diff
@@ -2377,12 +2457,11 @@ def match_to_list(mypkg, mylist):
     return result
 
 
-def best_match_to_list(mypkg, mylist):
+def best_match_to_list(mypkg: Union[str, Atom], mylist: list) -> Optional[Atom]:
```
zmedico, line 2460: mypkg should not be Atom, but _pkg_str or Package.

lib/portage/dep/__init__.py
```diff
@@ -2480,12 +2559,11 @@ def match_from_list(mydep, candidate_list):
...
 
-    if "!" == mydep[:1]:
-        if "!" == mydep[1:2]:
-            mydep = mydep[2:]
-        else:
-            mydep = mydep[1:]
-    if not isinstance(mydep, Atom):
+    if isinstance(mydep, Atom):
+        if mydep.blocker:
+            mydep = Atom(str(mydep).lstrip("!"), allow_wildcard=True, allow_repo=True)
```
zmedico, line 2564: It might not really be necessary to do this, and would be better to avoid if possible because it discards the unevaluated_atom.

lib/portage/dep/__init__.py
```diff
@@ -2515,11 +2592,17 @@ def match_from_list(mydep, candidate_list):
...
     if mydep.extended_syntax:
         for x in candidate_list:
-            cp = getattr(x, "cp", None)
-            if cp is None:
+            if isinstance(x, Atom):
+                cp = x.cp
+            elif hasattr(x, "cp"):
+                # Package object
+                cp = x.cp
+            else:
```
zmedico, line 2600: Generally, only _pkg_str and Package are valid here, not Atom.

However, VariableSet in lib/portage/_sets/dbapi.py passes in a Atom here:
```python
if Atom(exclude_output).match(atom):
```
I'm not happy about this usage and would prefer that some alternative match function be created for this. Maybe this works because the atom matched a package that's stored in the ebuild variable:
```diff
--- a/lib/portage/_sets/dbapi.py
+++ b/lib/portage/_sets/dbapi.py
@@ -182,7 +182,7 @@ class VariableSet(EverythingSet):
             # 'exclude_output' here refers to filtering out any packages
             # matching the 'includes' criteria
             for exclude_output in self._excludes_output:
-                if Atom(exclude_output).match(atom):
+                if Atom(exclude_output).match(ebuild):
                     return False
 
             for include in self._includes:
```
But there's more Atom passed in further down in VariableSet._filter:
```python
                matches_any_exclude = any(
                    exclude_atom.match(x) for exclude_atom in exclude_atoms
                )
```
```python
                matches_any_include = any(
                    include_atom.match(x) for include_atom in include_atoms
                )
```

zmedico: I suppose we should just go ahead an support Atom here.

lib/portage/dep/__init__.py
```diff
@@ -2617,8 +2715,12 @@ def match_from_list(mydep, candidate_list):
 
     elif operator == "~":  # version, any revision, match
         for x in candidate_list:
-            xs = getattr(x, "cpv_split", None)
-            if xs is None:
+            if isinstance(x, _pkg_str):
+                xs = x.cpv_split
+            elif hasattr(x, "cpv"):
+                # Package object
+                xs = x.cpv.cpv_split
```
zmedico, line 2722: This could be Atom since it has a cpv attribute, right?

lib/portage/dep/__init__.py
```diff
@@ -2699,7 +2801,12 @@ def match_from_list(mydep, candidate_list):
         candidate_list = mylist
         mylist = []
         for x in candidate_list:
-            use = getattr(x, "use", None)
+            # Only package objects have 'use' and 'iuse' attributes
+            if not hasattr(x, "use"):
+                mylist.append(x)
+                continue
```
zmedico, line 2807: This could be Atom since it has a `use` attribute, right? We can't handle it like Package because it has no `iuse` attribute.

lib/portage/dep/__init__.py
Jannik2099: I had a hard time following the typing for these functions, thanks.
Ideally these would be decoupled so that the auxiliary functions in the Atom file don't get Package objects, seeing that Package imports Atom... I'll see if I can unmangle them.

bin/ebuild.sh
```diff
@@ -186,6 +186,12 @@ addread "${PORTAGE_TMPDIR}/portage"
 addwrite "${PORTAGE_TMPDIR}/portage"
 [[ -n ${PORTAGE_GPG_DIR} ]] && addpredict "${PORTAGE_GPG_DIR}"
 
+if [[ -f /sys/fs/selinux/enforce ]]; then
+	addpredict /proc/self/attr/fscreate
```
thesamesam, line 190: BTW, this is https://bugs.gentoo.org/976051. The reason you only saw it with tests is that it depends on the profiles right now. I plan to move it out.

mattst88:
Merged in #1594.

PR #1525, opened 2025-12-04
fetch: Only fetch from GENTOO_MIRRORS if ebuild is in ::gentoo

Flowdalic:
Portage hits GENTOO_MIRRORS even for ebuilds within ebuild repositories other than ::gentoo. However, GENTOO_MIRRORS will only hold distfiles from ::gentoo.

To avoid unnecessary network requests and load on the mirrors only use GENTOO_MIRRORS if the ebuild's repository is ::gentoo.

Bug: https://bugs.gentoo.org/967090

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,8 @@ Release notes take the form of the following optional categories:
 * Bug fixes
 * Cleanups
 
+* fetch: Only fetch from GENTOO_MIRRORS if ebuild is in ::gentoo (bug #967090).
+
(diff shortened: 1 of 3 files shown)


Soccera1:
> 50:50 on this. It would impact people who either copy ebuilds from `::gentoo`, or have their own ebuilds for matching software. Not to mention all the crap like crates.

My best idea would be to have an opt-in thing in the ebuild to not fetch from GENTOO_MIRRORS. Unsure how viable this is though.

PR #1526, opened 2025-12-05
add srcroot for abssymlink

clan:
so that symlink dereferencing works as if in a chroot(srcroot).

Closes: https://bugs.gentoo.org/954361


thesamesam:
Can you file a Portage bug and have the runit bug depend on that please?

thesamesam (commented):
I think this looks okay but I'd appreciate another check.

PR #1527, opened 2025-12-06
Binary package selection

jeth-ro:
A number of changes towards the end goal of more granular control over which packages are to be installed using binary packages instead of ebuilds.

1. Extend emerge command line interface with the following options:
`--usepkg-include <atoms>` defines a whitelist for binary packages (complements existing `--usepkg-exclude` option)
`--getbinpkg-exclude <atoms>` defines blacklist for remote binary packages
`--getbinpkg-include <atoms>` defines whitelist for remote binary packages
`--nobindeps` uses install target as the binary package whitelist
_No effect without -k or -g._

2. Extend repos.conf syntax with the following attributes:
`usepkg-exclude` configures a persistent and repository specific blacklist for binary packages
`usepkg-include` configures a persistent and repository specific whitelist for binary packages
_No effect without -k._

3. Extend binrepos.conf syntax with the following attributes:
`getbinpkg-exclude` configures a persistent and binhost specific blacklist for remote binary packages
`getbinpkg-include` configures a persistent and binhost specific whitelist for remote binary packages
_No effect without -g._

4. Populate an implicit `--getbinpkg-include` list from any packages with `FEATURES="getbinpkg"` set in package.env and also imply `-g` if said list is not empty (see [bug #463964](https://bugs.gentoo.org/463964)).
_Only works in absence of -g or FEATURES=getbinpkg (in make.conf or emerge process environment) both of which still have global effect._

6. Add visible indication in emerge output of when remote binary packages are being used by showing a `g` flag in the same place as `f` and `F` would appear for fetch restrict ebuilds.

Finally, have slipped in allowing the local cache of a binhost pkgindex to persist when said host is unreachable provided the `--pretend` option is in effect. This might not be desired, but I found this behaviour useful.

No tests yet, but will look at that if there is interest in actually merging this. For now have done my best to ensure existing tests still pass.

There remain a number of obvious problems, and no doubt others:
 - no warnings for no-ops like specifying `--usepkg-include` without `-k` or `--getbinpkg-include` without `-g`
 - no warnings for conflicting selections, e.g. same package for `--usepkg-include` and `--usepkg-exclude`, etc
 - the order of priority or override between *.conf and the command line is probably not correct (exclude always wins)
 - the package.env implementation is proof-of-concept at best, I suspect passing atoms to config.setcpv() is not sufficient
 - use of ::repo notation or operators is restricted on command line, but used in new repos.conf or binrepos.conf attributes will probably crash emerge

Will tackle these things if it looks like this will go anywhere, but felt I'd taken this far enough to seek feedback.


lib/_emerge/actions.py
```diff
@@ -3543,6 +3555,20 @@ def run_action(emerge_config):
     if "getbinpkg" in emerge_config.target_config.settings.features:
         emerge_config.opts["--getbinpkg"] = True
 
+    if "--getbinpkg" not in emerge_config.opts:
+        penvdict = emerge_config.target_config.settings._penvdict
+        pkgsettings = portage.config(clone=emerge_config.target_config.settings)
+        for cp in penvdict:
+            for atom, env in penvdict[cp].items():
+                pkgsettings.setcpv(atom)
```
jeth-ro: Suspect this is probably not correct in that other uses of `setcpv()` usually pass a `_pkg_str` or `Package` rather than `Atom`. Suggestions welcome!

jeth-ro:
I believe that `--getbinpkgonly` implies `--usepkgonly` (although the man page doesn't say so explicitly, it's in the region of `actions.py:3570`) and `--usepkgonly` is implemented by removing porttree from the dependency calculation entirely. So while this indeed forces an error if there's no available binary packages for something which is needed, this is presumably because emerge cannot find a solution (without porttree) that won't potentially break your system...?

In the case of `--update --getbinpkgonly` then vartree can presumably be used for unavailable binaries, i.e. simply don't update them rather than bailing. A quick rudimentary test indicates that this may be the current behaviour.

So I think I'm missing something as you say you are doing this already. When if I run `emerge -1gp <x>` and then `emerge -1gp <y>` where `<y>` is a list of the binary-only subset of the packages to be installed then the second command still pulls the remaining packages from ebuilds. Of course there's always `--nodep`, but I'd worry that adding features that need to _imply_ `--nodep` to work is going to lead nasty surprises for users.

eli-schwartz:
> So I think I'm missing something as you say you are doing this already. When if I run `emerge -1gp <x>` and then `emerge -1gp <y>` where `<y>` is a list of the binary-only subset of the packages to be installed then the second command still pulls the remaining packages from ebuilds. Of course there's always `--nodep`, but I'd worry that adding features that need to _imply_ `--nodep` to work is going to lead nasty surprises for users.

I think that you're overthinking this a lot.

I have 5 leaf packages in my world file. Two of them are available right now as binaries, and three are not available as binaries. All of them have gotten new version bumps. A theoretical `emerge -uDNg @world --ignore-missing-binpkgs` would result in only installing the two that have binaries and simply ignoring the other three (which I'll probably install tomorrow or something, after the server gets around to building binaries for them).

Recursive dependencies either don't exist or for the purpose of this discussion already have binaries if the leaf package does.

thesamesam:
i.e. it's in the realm of helping [bug 924772](https://bugs.gentoo.org/924772).

jeth-ro:
Yeah, I may be overthinking it :) I haven't yet run into #924772 myself, so had not appreciated that complexity. My usage of binpkgs is pretty light, thus the desire for a whitelist.

@thesamesam - assuming looking closer means code changes, here's a few things I'm pretty sure I've not take the ideal approach with. Stlil don't know the code very well...
 - Extending the `binarytree.populate()` interface feels messy, in particular the need to prepare `kwargs` in multiple places. Is there a better way to get the command line lists into `binarytree`?
 - Whether it's ok to update `depgraph._frozen_config` in the `_resolve()` function, unsure if "frozen" implies some degree of immutability? Did it here because there looks to be a lot of machinery to arrive at the desired Atoms which hasn't yet run at the time of `_frozen_config.__init__()`.
 - If portage/repository/config.py is an appropriate place to have added `_validate_usepkg_list()` or not. Could potentially reuse `_find_bad_atoms()` from _emerge/main.py, but was unsure if imports from _emerge into portage are desirable or not.

Might still be overthinking... Thanks both!

thesamesam:
@zmedico Would you mind taking a look at the above?

lib/_emerge/actions.py
zmedico: Since it seems that we only access pkgsettings.features, maybe we could add an optimized pkgsettings.features_contains(atom, feature) method to call here.

lib/portage/tests/resolver/ResolverPlayground.py
```diff
@@ -695,17 +752,10 @@ def _load_config(self):
...
-                    repo_name,
-                    "\n".join(f"{k} = {v}" for k, v in repo_config.items()),
-                )
-                for repo_name, repo_config in self._repositories.items()
-            ),
-        }
+        path = f"{self.eprefix}/usr/sbin:{self.eprefix}/usr/bin:{os.environ['PATH']}"
+        env = {"PATH": path}
+        with open(os.path.join(self.eprefix, USER_CONFIG_PATH, "repos.conf")) as f:
+            env["PORTAGE_REPOSITORIES"] = f.read()
```
jeth-ro, line 762: Did not think this should be necessary as `repos.conf` is now always written to the resolver playground's temporary environment, but many (not all) unrelated tests failed without this environment variable set.

lib/portage/tests/resolver/ResolverPlayground.py
```diff
@@ -69,11 +90,11 @@ class ResolverPlayground:
             "package.use.stable",
             "package.use.stable.force",
             "package.use.stable.mask",
+            "repos.conf",
             "soname.provided",
             "use.force",
             "use.mask",
             "use.stable",
-            "layout.conf",
```
jeth-ro, line 76: Removing duplicate, otherwise unrelated to this change.

lib/portage/tests/resolver/ResolverPlayground.py
```diff
@@ -411,8 +457,8 @@ def _create_binpkgs(self, binpkgs):
             else:
                 raise InvalidBinaryPackageFormat(binpkg_format)
 
-            bintree = binarytree(pkgdir=self.pkgdir, settings=self.settings)
-            bintree.populate(force_reindex=True)
+        bintree = binarytree(pkgdir=repo_dir, settings=self.settings)
+        bintree.populate(force_reindex=True)
```
jeth-ro, line 465: Did not think it necessary to call populate for every binpkg created, once at the end seems to work too?

lib/portage/tests/resolver/test_binpackage_selection.py
```diff
@@ -0,0 +1,1324 @@
...
+            test_cases,
+            binpkgs=binpkgs,
+            binrepos=binrepos,
+            ebuilds=ebuilds,
+            user_config=user_config,
+        )
+
+    @pytest.mark.xfail(
+        reason="pulls baz::other_repo instead of ::test_repo despite binrepo priority"
+    )
```
jeth-ro: ~~About this I am puzzled and will keep digging... I am 95% sure it is nothing to do with the feature under test and more something to do with the way the test is set up. Inverting the priorities in `binrepos.conf` produces the desired test outcome.~~

EDIT: pretty sure this was an artifact of test setup. xfail mark has been removed.

lib/_emerge/Package.py
```diff
@@ -116,6 +117,11 @@ def __init__(self, **kwargs):
                 raise
             db = self.root_config.trees["porttree"].dbapi
 
+        if self.type_name == "binary":
```
thesamesam: I'd personally prefer:
```
self.remote = False
if self.type_name == "binary":
	self.remote = db.bintree.isremote(self.cpv)
```

lib/portage/tests/resolver/ResolverPlayground.py
thesamesam: This should be split into it own commit and could go in now.

lib/portage/repository/config.py
```diff
@@ -89,6 +90,34 @@ def _gen_valid_repo(name):
     return name
 
 
+def _find_bad_atoms(atoms, less_strict=False):
```
thesamesam: Why not..
```suggestion
def _find_bad_atoms(atoms, strict=True):
```

EDIT: Ah, nevermind, it got moved.

jeth-ro:
> ... In particular, printing warnings from in _frozen_depgraph_config clashes with the spinner.

Above is fixed; validation of *.conf files now in `BinRepoConfig` and `RepoConfig`, validation of emerge command line options now (mostly) earlier in `action.py`. The business of `--getbinpkg-*` command line options overriding binrepos.conf is still implemented in `bintree.py` as the parsed configuration that these options override isn't really exposed.

I might still prefer to not have `binarytree` doing this last bit of config fix-up and to also avoid the need to change the `populate()` interface (see above comment 16-Dec), but don't see any minimally invasive way around these points. So will draw a line under it for now. Unlike the above spinner thing this isn't user visible as `binarytree.populate()` is called early enough for those warnings to appear up front and not interleaved with later emerge output.

lib/portage/repository/config.py
```diff
@@ -89,6 +91,34 @@ def _gen_valid_repo(name):
...
+    """
+    Declares all atoms as invalid that have an operator,
+    a use dependency, a blocker or a repo spec.
+    It accepts atoms with wildcards.
+    In less_strict mode it accepts operators and repo specs.
+    """
+    from _emerge.is_valid_package_atom import insert_category_into_atom
+
+    bad_atoms = []
+    for x in " ".join(atoms or []).split():
```
jeth-ro, line 104: Note minor change here in relocated function, can now do `_find_bad_atoms(None)` without an exception.

hololeap:
> I think it's an interesting idea though, but I also am not sure how much it matters given tools should be able to handle mangling binrepos.conf just fine too (what exactly is wrong with the ergonomics of doing that right now, and could such issues really not be handled by a tool)?

It's usable as-is.

> IMO we're better off exploring bigger changes like that once this is in (and I'd like to have that idea recorded in a bug for posterity). I'm glad you mentioned it now, but I think we could move to that later if it seems desirable.

I understand my idea would involve a lot of extra moving parts, and it's better to get the current PR pulled. I'll make a bgo bug with this proposal tomorrow.

jeth-ro:
This was it - https://bugs.gentoo.org/924772.

Am not going to pretend to have understood the full and detailed comment trail on that bug just yet, but it seemed to me that the initial need there (though indeed not the specific wording of the request) is to not ever be building poppler/boost/libreoffice/whatever locally, i.e. to exclude the ebuild entirely. Not strictly possible under the current state of this change, but I gather would just be `-*` under @hololeap's proposal. So arguably it is an improvement :)

> I remain sceptical of something FEATURES-centric. It means you have to be sure it'll work for any future FEATURES too (as well as any others we currently have)

As a user, twiddling features per-package does makes a certain kind of sense to me as a way to achieve this. Doing it via `FEATURES` and `package.env` seemed very awkward, but features itself as the knob to move seems ok. That said, I am for certain lacking any historical perspective on this... Also take your point about then having to deal with all other features to achieve it, which sounds big and may never get over the line. 

Have amended `portage(5)` on this branch, no other changes.

jeth-ro:
Understood :) I took some pains to make self contained and individually tested commits, which is part of why this has taken a while. So the *.conf stuff (latest two commits) can theoretically be left out or clawed back later without affecting the CLI changes.

Just in case it wasn't obvious - I have only marked the repos.conf and binrepos.conf attributes as "experimental" in `portage(5)`. The new command line options --usepkg-include, --getbinkg-include, --getbinpkg-exclude, and --nobindeps have no such qualification in `emerge(1)`.

lib/portage/dbapi/bintree.py
```diff
@@ -1765,6 +1839,17 @@ def _populate_remote_repo(
...
+                # Respect remote binary exclude and include lists if defined
+                in_getbinpkg_exclude = (
+                    have_getbinpkg_exclude and getbinpkg_exclude.containsCPV(cpv)
+                )
+                in_getbinpkg_include = (
+                    not have_getbinpkg_include or getbinpkg_include.containsCPV(cpv)
+                )
+                if in_getbinpkg_exclude or not in_getbinpkg_include:
+                    continue
+
```
zmedico, line 1852: We should optimize the containsCPV because it currently iterates over all atoms regardless of atom.cp.

zmedico: https://github.com/gentoo/portage/pull/1548

lib/_emerge/depgraph.py
```diff
@@ -5269,6 +5268,14 @@ def _select_files(self, myfiles):
...
         self._dynamic_config._initial_arg_list = args[:]
 
+        # set usepkg-include set to expanded args if --nobindeps is
+        # in effect, both usepkg-include and usepkg-exclude should
+        # be assumed empty at this point
+        if "--nobindeps" in self._frozen_config.myopts:
+            for arg in self._expand_set_args(args):
+                arg_cp = (a.cp for a in arg.pset.getAtoms())
+                self._frozen_config.usepkg_include.update(arg_cp)
+
```
zmedico, line 5278: This has me wondering why we're including package set members but not their dependencies.

Why can't we just make self._frozen_config.usepkg_include empty here?

zmedico: Oh nevermind I see the man page now.

mid-kid:
Thanks a lot for doing this! This was one of the great blockers for getbinpkg adoption to me, and being able to selectively allow/disallow certain packages from specific hosts is really powerful.

I just have one question: Does this feature support aggregating `getbinpkg-exclude` rules across multiple lines? I find putting a bunch of packages on a single line to be difficult to read, and I'd like to be able to comment why certain packages are excluded (e.g. patched, machine-specific settings or optimizations, prefer binpkg from different repo, etc). Ideally I'd also spread this over multiple files, like I'd do with package.env, but I can live with a file per repo.
If not, I'll play around with it for a bit and submit a feature request in the bug tracker. I think it would be very useful, at least in my workflow.

jeth-ro:
> Does this feature support aggregating getbinpkg-exclude rules across multiple lines?

No, I don't believe so sorry.

> I'd like to be able to comment why certain packages are excluded (e.g. patched, ...)

If "patched" means user patched, see https://github.com/gentoo/portage/pull/1574.

> Ideally I'd also spread this over multiple files, like I'd do with package.env,

Perhaps https://bugs.gentoo.org/969628 is of interest, and probably a better place to register interest in alternative config interfaces than here. I had a POC somewhere, it'll be out of date now.

mid-kid:
Thanks for the links! That's exactly what I was looking for. I'll subscribe to the bug ticket.

> > I'd like to be able to comment why certain packages are excluded (e.g. patched, ...)
> 
> If "patched" means user patched, see #1574.

Thanks for doing this, but for me it doesn't really solve the use-case, as some of my patches are scripted modifications in /etc/portage/env, and I'm planning on doing things with bashrc in a custom profile. Even matching on ebuild+eclass+bashrc+env checksum wouldn't be sufficient as at least one of them reads a non-portage file in /etc. I really don't mind specifying exclusions manually, to avoid having to devise extensive heurestics.

jeth-ro:
Do post the difficulties you have stated with binpkg selection via `repos.conf` and `binrepos.conf` to the linked bug. That interface is still documented as experimental, so I imagine it could change if people supply good reasons. But I doubt anyone is looking here for that feedback.

PR #1562, opened 2026-03-04
EbuildPhase: prefer C.UTF-8 for LC_COLLATE

thesamesam:
EbuildPhase: prefer C.UTF-8 for LC_COLLATE

Prefer C.UTF-8 for LC_COLLATE over C and set LC_COLLATE in the validation
loop if required.

Modern glibc (>=2.35) guarantees C.UTF-8 (though Gentoo had it patched
in before) and musl always has C.UTF-8.

This should still comply with the spec:
> The package manager must ensure that the LC_CTYPE and LC_COLLATE locale
> categories are equivalent to the POSIX locale, as far as characters in the ASCII
> range (U+0000 to U+007F) are concerned.

We would previously try for LC_CTYPE:
1) C.UTF-8
2) en_US.UTF-8
3) en_GB.UTF-8
4) C

but 3) and 4) aren't equivalent to C in the ASCII range for LC_COLLATE
which we want to add now, so change the iteration list to:
1) C.UTF-8
2) C
and we can maybe add the other two en_* if required at the end, but I
don't think those would satisfy the spec requirement.

Bug: https://bugs.gentoo.org/913903

### DIFF
--- a/lib/_emerge/EbuildPhase.py
+++ b/lib/_emerge/EbuildPhase.py
@@ -53,11 +53,11 @@ async def _setup_locale(settings):
     eapi_attrs = _get_eapi_attrs(settings["EAPI"])
     if eapi_attrs.posixish_locale:
         split_LC_ALL(settings)
-        settings["LC_COLLATE"] = "C"
         # check_locale() returns None when check can not be executed.
         if await async_check_locale(silent=True, env=settings.environ()) is False:
             # try another locale
-            for l in ("C.UTF-8", "en_US.UTF-8", "en_GB.UTF-8", "C"):
+            for l in ("C.UTF-8", "C"):
+                settings["LC_COLLATE"] = l
                 settings["LC_CTYPE"] = l
                 if await async_check_locale(silent=True, env=settings.environ()):
                     # TODO: output the following only once


lib/_emerge/EbuildPhase.py
```diff
@@ -56,11 +56,12 @@ async def _setup_locale(settings):
     eapi_attrs = _get_eapi_attrs(settings["EAPI"])
     if eapi_attrs.posixish_locale:
         split_LC_ALL(settings)
-        settings["LC_COLLATE"] = "C"
+        settings["LC_COLLATE"] = "C.UTF-8"
         # check_locale() returns None when check can not be executed.
         if await async_check_locale(silent=True, env=settings.environ()) is False:
             # try another locale
             for l in ("C.UTF-8", "en_US.UTF-8", "en_GB.UTF-8", "C"):
+                settings["LC_COLLATE"] = l
```
floppym, line 60: So this means if C.UTF-8 is unusable for some reason, we might have LC_COLLATE fall back to en_US.UTF-8 or en_GB.UTF-8. Is that intended and within spec?

thesamesam: The change extends it to `LC_COLLATE` rather than just `LC_CTYPE` (which I think is itself right), but I think it's wrong as-is actually?

If C.UTF-8 can't be used, then I think we should fallback to C (at least first), rather than the others:
> The package manager must ensure that the LC_CTYPE and LC_COLLATE locale categories are equivalent to the POSIX locale, as far as characters in the ASCII range (U+0000 to U+007F) are concerned. Only for EAPIs listed in such a manner in table [11.6](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-111040x6).

floppym: en_US.UTF-8 and en_GB.UTF-8 should be equivalent to C/C.UTF8 for LC_CTYPE in the ASCII range.

LC_COLLATE is another matter; I think there are some differences in sort order between C and en_US/en_GB.

floppym: So I think we agree here.

thesamesam: How about what I've pushed now? Note that there's also a `mylocale = "C"` in `async_check_locale` which we may want to change:
https://github.com/gentoo/portage/blob/24431d643340999918ef9d2ba8ebf82d4fa6d313/lib/portage/util/locale.py#L122-L129

floppym: Yeah, I guess limiting it to C.UTF-8 or C is the simple solution. Seems good to me.

thesamesam: OK, thanks. I'll check it in.

I'm not against doing something else but I didn't have any inspiration for doing it cleanly.

PR #1567, opened 2026-03-20
Add `CHECKSUMCOMMAND` variable for remote digest computation

mattst88:
When distfiles are stored on a remote system and accessed via NFS,
computing digests locally reads every byte across the network. Add
support for a `CHECKSUMCOMMAND` setting that delegates the hash
computation to an external command (e.g., via ssh to the file server),
while keeping the digest comparison local.

The command receives the same style of variable expansion as
`FETCHCOMMAND`, with the following variables:
- `${FILE}`             - full path to the file
- `${DISTFILE}`         - basename of the file
- `${HASHTYPE}`         - comma-separated list of hash types to compute
- `${DISTDIR}`          - the distfiles directory
- `${PORTAGE_SSH_OPTS}` - SSH options, if set

The command must print lines of "HASHTYPE hexdigest" followed by
"size filesize" on stdout. The `portage-checksum` helper script
implements this protocol using Portage's checksum functions.

Example make.conf configuration:
```
  CHECKSUMCOMMAND="ssh portage@fileserver \
      /usr/bin/portage-checksum \${HASHTYPE} \${DISTDIR}/\${DISTFILE}"
```

When `CHECKSUMCOMMAND` is not set, behavior is unchanged.

This is intended to be used in conjunction with `FETCHCOMMAND` and
`RESUMECOMMAND`.

Running `ebuild media-fonts/noto/noto-20260201.ebuild setup` with
`/var/cache/distfiles` on NFS over a WiFi network takes 5 minutes. With
`CHECKSUMCOMMAND`, it takes 5 seconds.


mattst88:
FWIW, I just tried cachefilesd and discovered that it doesn't work with ZFS (which my laptop uses).

PR #1570, opened 2026-04-08
netlink: clean up socket logic

floppym:
Call connect() to avoid the need to pass the kernel netlink address in all send calls.

Peek at the message header determine the full size of received messages.


thesamesam (approved):
Nice.

PR #1571, opened 2026-04-13
bin/isolated-functions.sh: add new function __normalize_owner used then by {ins,dir,exec,lib}opts

PPN-SD:
move __resolve_owner from fowners to isolated-function so it can be used by other helpers. Use die instead of die_helpers.

add a new function __normalize_owner using __resolve_owner to replace literal -o/--owner/-g/--group with numeric-uid-gid. Use it for {ins,dir,exec,lib}opts.

Not sure things are in the right place but it’s a start.

Bug: https://bugs.gentoo.org/971120


floppym:
I'll try to take a look at this.

PR #1574, opened 2026-04-19
Skip binpkgs when local user patches are supplied

jeth-ro:
Partial solution to https://bugs.gentoo.org/917047. Still draft as I intend to look also look at accommodating locally patched binpkgs, as per comments on the bug.

One open question might be how this should interact with `--usepkgonly`. Current implementation will revert to ignoring user patches, but some may prefer it to fail rather than effectively reverting user patches. Either behaviour is accessible via `--usepkg-exclude-patches=y/n`, just a question of which is the default. Falling back on the user-patched ebuild is obviously not an option with `--usepkgonly`.

Reporting a sensible reason for failure with `--usepkgonly --usepkg-exclude-patches=y` is also missing at this point.

### DIFF
--- a/bin/phase-helpers.sh
+++ b/bin/phase-helpers.sh
@@ -1139,6 +1139,7 @@ if ___eapi_has_eapply_user; then
 		local basename basedir columns tagfile hr d f
 		local -A patch_by
 		local -a dirents
+		local IFS
 
 		[[ ${EBUILD_PHASE} == prepare ]] || \
 			die "eapply_user() called during invalid phase: ${EBUILD_PHASE}"
@@ -1185,13 +1186,32 @@ if ___eapi_has_eapply_user; then
 		done
 
 		if (( ${#patch_by[@]} > 0 )); then
+			mkdir -p -- "${PORTAGE_BUILDDIR}"/build-info || die "eapply_user: could not create build-info directory"
+			local userpatch_digests="${PORTAGE_BUILDDIR}"/build-info/user_patch.digests
+			local userpatches_hash="${PORTAGE_BUILDDIR}"/build-info/USER_PATCHES
+
 			printf -v hr "%$(( columns - 3 ))s"
 			hr=${hr//?/=}
 			einfo "${PORTAGE_COLOR_INFO}${hr}${PORTAGE_COLOR_NORMAL}"
 			einfo "Applying user patches from ${basedir} ..."
+
+			local hash
+			local total=""
 			while IFS= read -rd '' basename; do
 				eapply -- "${patch_by[$basename]}"
+
+				read -r hash _ < <(sha256sum < "${patch_by[$basename]}") \
+				&& wait "$!" \
+				&& total+="${hash}" \
+				&& printf '%s\0' "${hash}" "${basename}" >> "${userpatch_digests}" \
+				|| die "eapply_user: could not generate individual user patch digests"
 			done < <(printf '%s\0' "${!patch_by[@]}" | LC_ALL=C sort -z)
+
+			read -r hash _ < <(printf '%s' "${total}" | sha256sum) \
+			&& wait "$!" \
+			&& printf '%s\n' "${hash}" > "${userpatches_hash}" \
+			|| die "eapply_user: could not generate hash over all user patches"
+
 			einfo "User patches applied."
 			einfo "${PORTAGE_COLOR_INFO}${hr}${PORTAGE_COLOR_NORMAL}"
 		fi
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -7741,6 +7741,10 @@ class depgraph:
                         if in_usepkg_exclude or not in_usepkg_include:
                             break
 
+                        # do not select binpkgs if user patches exist (see bug #917047)
+                        if pkg.cpv in pkgsettings._user_patches:
+                            break
+
                     # We can choose not to install a live package from using binary
                     # cache by disabling it with option --usepkg-exclude-live in the
                     # emerge call.
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -7642,6 +7642,9 @@ class depgraph:
         usepkg = "--usepkg" in self._frozen_config.myopts
         usepkgonly = "--usepkgonly" in self._frozen_config.myopts
         usepkg_exclude_live = "--usepkg-exclude-live" in self._frozen_config.myopts
+        usepkg_exclude_patches = self._frozen_config.myopts.get(
+            "--usepkg-exclude-patches", not usepkgonly
+        )
         empty = "empty" in self._dynamic_config.myparams
         selective = "selective" in self._dynamic_config.myparams
         reinstall = False
@@ -7742,7 +7745,10 @@ class depgraph:
                             break
 
                         # do not select binpkgs if user patches exist (see bug #917047)
-                        if pkg.cpv in pkgsettings._user_patches:
+                        if (
+                            usepkg_exclude_patches
+                            and pkg.cpv in pkgsettings._user_patches
+                        ):
                             break
 
                     # We can choose not to install a live package from using binary
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -7994,7 +7994,7 @@ class depgraph:
                         # do not select binpkgs if user patches exist (see bug #917047)
                         if (
                             binpkg_respect_user_patches
-                            and pkg in pkgsettings._user_patches
+                            and pkg.user_patches != pkgsettings.userPatchDigest(pkg)
                         ):
                             self._dynamic_config.ignored_binaries.setdefault(
                                 pkg, {}
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -7910,6 +7910,10 @@ class depgraph:
                         if in_usepkg_exclude or not in_usepkg_include:
                             break
 
+                        # do not select binpkgs if user patches exist (see bug #917047)
+                        if pkg in pkgsettings._user_patches:
+                            break
+
                     # We can choose not to install a live package from using binary
                     # cache by disabling it with option --usepkg-exclude-live in the
                     # emerge call.
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -29,7 +29,9 @@ from portage.dbapi.DummyTree import DummyTree
 from portage.dbapi.IndexedPortdb import IndexedPortdb
 from portage.dep import (
     Atom,
+    _build_id_separator,
     _repo_separator,
+    _slot_separator,
     best_match_to_list,
     check_required_use,
     extract_affecting_use,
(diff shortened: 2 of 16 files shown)


jeth-ro:
There remain a few loose ends here, but am removing the draft status as would like feedback as to whether this is even remotely on the right track before making any attempt at final polish...

Such loose ends include things like:
 - output when `-K` is used (currently reports all masked with no reason)
 - whether or not the python `sorted()` on bytearrays actually is equivalent to POSIX order (`LC_COLLATE=C` or such)
 - better test coverage, currently only really testing resolver and config changes

When it comes to `-K` the default behaviour has changed to fail in the absence of suitably user-patched binpkgs, unless overridden with `--binpkg-user-patches=n`. Am not a huge fan of the option name `--binpkg-user-patches` as it doesn't clearly say what it does, but am trying to follow the naming pattern of options that fulfill similar purposes and behave similarly.

Last commit on this branch adds additional output from emerge when installing binpkgs created from ebuilds with user patches applied. That output looks like this, with or without the warning:

20260601_23h08m39s_crop

lib/portage/package/ebuild/_config/UserPatches.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # copy() so as not to mutate on each lookup
+        patches = self._patch_sets.get(key.cp, {}).copy()
+        patches |= self._patch_sets.get(key.cpv, {})
+        if hasattr(key.cpv, "slot"):
+            slot = _slot_separator + key.cpv.slot
+            patches |= self._patch_sets.get(key.cp + slot, {})
+            patches |= self._patch_sets.get(key.cpv + slot, {})
+
+        hashes = [patches[p] for p in sorted(patches.keys())]
```
jeth-ro: This `sorted()` should be working on encoded binary strings here (see `_load_patch_set()` below) which I'm thinking more or less emulates the POSIX order described in phase-helpers.sh where the shell command `sort` is used with `LC_ALL=C`. Have to admit I did not (yet) read up to confirm beyond doubt what the C locale does for sorting, but my understanding is that it is by byte value.

jeth-ro: Have dug around POSIX and C standards and found in the glibc documentation that `strcoll()` under the "C" locale is equivalent to `strcmp()`. So am taking that as confirmation that LC_ALL=C is a byte-value sort and so python `sorted()` over binary strings should be doing what is needed here, unless anyone else knows better :)

jeth-ro:
Now with more useful output when `--usepkgonly` is in effect. Below example with `-Gav`:

20260611_01h09m28s_grim


The "(masked by: )" output for specific versions looks to be an existing limitation with ignored binaries and is also seen in the messaging for binpkg-respect-use ignored binaries. Is caused by interactions between `_eliminate_ignored_binaries()` and `_show_unsatisfied_dep()`, with the latter consulting the ignored binaries list to report this kind of masking. Am tempted to file a bug for that.

There remains a gap in testing for this change with nothing really checking the changes to `phase-helpers.sh`. Have not found much in the way of existing tests over `ebuild.sh` to build on. Presumably would require minimal ebuilds, mock distributables that could actually be patched, etc, all to be added or generated. Please do advise if that is desired or required (or already exists and I'm just blind), else I've nothing more to add to this PR at this time - ready for review.

thesamesam:
I keep thinking about https://bugs.gentoo.org/917047#c8 (around there) and I can't figure out if it's actually worth having this per-repository or not. Maybe we won't worry about that for now until someone asks for it.

> Presumably would require minimal ebuilds, mock distributables that could actually be patched, etc, all to be added or generated

I've been wanting something like this for some time. You don't need to do that, but if you have the energy to work on it (independent of this PR, or with it), that'd be most welcome.

zmedico (approved):

jeth-ro:
> I've been wanting something like this for some time. You don't need to do that, but if you have the energy to work on it (independent of this PR, or with it), that'd be most welcome.

I had a few ideas back when this one was fresher in my mind, will check my notes. Timeline likely to be weeks rather than days, so independent of this PR would be preferred.

thesamesam:
> One open question might be how this should interact with --usepkgonly. Current implementation will revert to ignoring user patches, but some may prefer it to fail rather than effectively reverting user patches.

User configuration of a daemon or tool might depend on something patched in. I'd say let's default it to _respect_ user patches. We already want to fix the respect-use behaviour (https://bugs.gentoo.org/934962).

lib/_emerge/main.py
```diff
@@ -150,6 +150,8 @@ def __contains__(self, s):
         "--autounmask-unrestricted-atoms": y_or_n,
         "--autounmask-write": y_or_n,
         "--binpkg-changed-deps": y_or_n,
+        "--binpkg-respect-use": y_or_n,
+        "--binpkg-user-patches": y_or_n,
```
thesamesam: `--binpkg-respect-user-patches` was suggested on IRC to make clear what its effect is.

lib/_emerge/main.py
```diff
@@ -389,7 +390,15 @@ def parse_opts(tmpcmdline, silent=False):
...
-            "help": ("reject binary packages with outdated " "dependencies"),
+            "help": "ignore binary packages with outdated dependencies",
+            "choices": true_y_or_n,
+        },
+        "--binpkg-respect-use": {
+            "help": "ignore binary packages if their use flags don't match the current configuration",
+            "choices": true_y_or_n,
+        },
+        "--binpkg-user-patches": {
+            "help": "ignore from binary packages for which user patches exist",
```
thesamesam: "Ignore binary packages for which user patches exist."

(And then update that text later because we make it more intelligent later in this series, right?)

lib/_emerge/depgraph.py
```diff
@@ -1347,6 +1354,59 @@ def _show_ignored_binaries_changed_deps(self, changed_deps):
...
+
+        messages = []
+        patchset = set()
+
+        for pkg, patches in user_patches.items():
+            msg = f"     {pkg.cpv}"
+            if hasattr(pkg.cpv, "slot") and pkg.slot != "0":
+                msg += _slot_separator + pkg.cpv.slot
+            if hasattr(pkg, "build_id") and pkg.build_id:
+                msg += f"-{pkg.build_id}"  # FIXME: build-id-sep?
```
thesamesam: FIXME

thesamesam (commented):
Some comments based on what Kerin (monkfish) said on IRC.

bin/phase-helpers.sh
```diff
@@ -1185,13 +1185,24 @@ if ___eapi_has_eapply_user; then
 		done
 
 		if (( ${#patch_by[@]} > 0 )); then
+			if [[ ! -d "${PORTAGE_BUILDDIR}"/build-info ]]; then
```
thesamesam: from Kerin:

----
 
- Uncalled-for proxy check and TOCTOU
- No localisation of userpatch_digests
- No localisation of userpatch_hash
- No error checking

Consider: `mkdir -p -- "${PORTAGE_BUILDDIR}"/build-info || die`

bin/phase-helpers.sh
```diff
@@ -1185,13 +1185,24 @@ if ___eapi_has_eapply_user; then
...
+			userpatch_digests=${PORTAGE_BUILDDIR}/build-info/user_patch.digests
+			userpatches_hash=${PORTAGE_BUILDDIR}/build-info/USER_PATCHES
+
 			printf -v hr "%$(( columns - 3 ))s"
 			hr=${hr//?/=}
 			einfo "${PORTAGE_COLOR_INFO}${hr}${PORTAGE_COLOR_NORMAL}"
 			einfo "Applying user patches from ${basedir} ..."
 			while IFS= read -rd '' basename; do
 				eapply -- "${patch_by[$basename]}"
+				hash=($(sha256sum "${patch_by[$basename]}"))
```
thesamesam: from Kerin:

----

- No localisation of hash
- No localisation of total (which is never otherwise initialised)
- Pathname expansion is not impeded during array population
- Inadequate quoting
- Ambiguous redirect becomes a latent possibility
- The echo utility is rubbish
- No error checking

Consider:
```
read -r hash _ < <(sha256sum -- "${patch_by[$basename]}") \
&& wait "$!" \
&& total+=${hash} \
&& printf '%s\n' "${hash} ${basename}" >> "${userpatch_digests}" \
|| die
```

bin/phase-helpers.sh
```diff
@@ -1185,13 +1185,24 @@ if ___eapi_has_eapply_user; then
...
+				total+=$hash
+				echo $hash "$basename" >> $userpatch_digests
 			done < <(printf '%s\0' "${!patch_by[@]}" | LC_ALL=C sort -z)
+			hash=($(sha256sum <(printf '%s' $total)))
```
thesamesam: from Kerin:

----

Much as before. Consider:
```
read -r hash _ < <(printf %s "${total}" | sha256sum) \
&& wait "$!" \
&& printf '%s\n' "${hash}" > "${userpatches_hash}" \
|| die
```

Finally, IFS should be localised if it isn't already (so that the fields of
sha256sum are guaranteed to be separated correctly).

jeth-ro:
> > One open question might be how this should interact with --usepkgonly. Current implementation will revert to ignoring user patches, but some may prefer it to fail rather than effectively reverting user patches.
> 
> User configuration of a daemon or tool might depend on something patched in. I'd say let's default it to _respect_ user patches. We already want to fix the respect-use behaviour (https://bugs.gentoo.org/934962).

For the avoidance of doubt... This has changed since the PR was opened and is now the default behaviour. It must be explicitly overridden with `--binpkg-respect-user-patches=n` and this is tested for in `test_binpackage_patches.py::testBinPkgUserPatchesOption()`. This also includes the case where a locally built binary has user patches that are no longer (or never were) in the current portage config.

jeth-ro:
Thanks for all the feedback on `phase-helpers.sh`, quite helpful. Bash scripting is obviously not one of my stronger points... I went through each carefully to be sure I understood what is being suggested and have implemented largely as suggested, having hopefully learnt a few things!

That said, am _still_ not sure what is meant by "Ambiguous redirect becomes a latent possibility" but am happy to assume it's solved in the suggested changes. I've also interpreted the suggestion to to localize IFS to mean explicitly define this for `read`, rather than to confine it's scope to `eapply_user()`. Apologies in advance if I've misunderstood.

thesamesam:
Could you check my question at https://bugs.gentoo.org/917047#c10 as well?

jeth-ro:
> Could you check my question at https://bugs.gentoo.org/917047#c10 as well?

Sorry, had forgotten about that. Short answer is there's no differentiation between local and remote binpkgs at all. Answered in detail on bug.

jeth-ro:
Updated with fixes for https://bugs.gentoo.org/917047#c12 and rebased onto latest master.

jeth-ro:
Now with helper for rendering paths for eyeballs, as [suggested](https://bugs.gentoo.org/917047#c16). Perhaps not as envisaged but tested as follows, including on filenames that previously raised `UnicodeEncodeError`.

ASCII control codes, in this case \r, replaced with ?.
20260908_22h48m30s_grim

Bad UTF-8, in this case \xc0\x80 (invalid null byte encoding), replaced with one ? for each byte.
20260908_22h47m18s_grim

Valid UTF-8, in this case \xc3\x83 (code point U+00C3), is rendered as expected.
20260908_22h47m46s_grim

lib/_emerge/main.py
```diff
@@ -840,6 +846,18 @@ def parse_opts(tmpcmdline, silent=False):
         else:
             myoptions.binpkg_changed_deps = "n"
 
+    if myoptions.binpkg_respect_use is not None:
```
thesamesam, line 849: One day we'll clean all of this up

thesamesam (approved):
Thanks, this looks great to me. Much appreciated Jethro.

PR #1577, opened 2026-04-26
spawn: move process supervisor code from pid-ns-init

floppym:
This works around an error when Portage is run using qemu-user.

qemu calls pthread_create() in its startup code, which ends up invoking clone() with CLONE_THREAD in the flags parameter. This will fail in a process which has previously called unshare() with CLONE_NEWPID.

We can avoid triggering the qemu startup code by avoiding any execve() calls in the process that called unshare(). Practically, this means copying the relevant "supervisor" code from pid-ns-init.

Bug: https://bugs.gentoo.org/703278

### DIFF
--- a/bin/pid-ns-init
+++ b/bin/pid-ns-init
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright 2018-2021 Gentoo Authors
+# Copyright 2018-2026 Gentoo Authors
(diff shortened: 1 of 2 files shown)


floppym:
This PR is easier to review if you enable the "ignore whitespace" option.

thesamesam (approved):

PR #1578, opened 2026-04-27
emaint: hide failures to determine ebuild masking reasons (bug #971602)

jeth-ro:
Try/catch band-aid for https://bugs.gentoo.org/971602. Am not particularly satisfied with this and intended to dig a bit more, but as this currently looks more like a data than a code problem this is perhaps required for robustness anyway.

Also fixes for logical error introduced by myself in [8d0633c](https://github.com/gentoo/portage/commit/8d0633c240c7d9dfd0eed459d69f521bb5d635bf) which had emaint report non-installed atoms as masked but with no masking reason given.

Note that only one issue will be reported by `emaint --check world` for each problematic entry in the world file, precedence of issues is:
 - atom is invalid (but see below)
 - missing from tree (installed or not)
 - not installed (masked or not)
 - not visible (masking reason reported)

The code has been structured such that the call to `getmaskingstatus()` is only made if higher priority problems are not found.

In the hope of not introducing another such error, I've also added tests. From this it is apparent that invalid atom strings in the world file are simply omitted from the `@selected` set in the loaded config and so are invisible to emaint. Unsure if this is as intended? Point being I haven't managed to actually trigger the original "invalid atom" path (have marked the test xfail).

### DIFF
--- a/lib/portage/emaint/modules/world/world.py
+++ b/lib/portage/emaint/modules/world/world.py
@@ -40,7 +40,7 @@ class WorldHandler:
         maxval = len(world_atoms)
         if onProgress:
             onProgress(maxval, 0)
-        for i, atom in enumerate(world_atoms):
+        for i, atom in enumerate(sorted(world_atoms)):
             if not isinstance(atom, portage.dep.Atom):
                 if atom.startswith(SETPREFIX):
                     s = atom[len(SETPREFIX) :]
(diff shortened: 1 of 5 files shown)


thesamesam:
> Unsure if this is as intended? 

I think it's intended to avoid other issues down the line.

thesamesam (approved):
Thanks, let's try it out and see what toralf says. It's an improvement in any case.

PR #1579, opened 2026-04-30
emaint: report masking reasons from porttree rather than vartree

jeth-ro:
See https://bugs.gentoo.org/971602 for details

### DIFF
--- a/lib/portage/emaint/modules/world/world.py
+++ b/lib/portage/emaint/modules/world/world.py
@@ -55,16 +55,20 @@ class WorldHandler:
                 continue
 
             installed = vardb.match(atom)
-            if not portdb.xmatch("match-all", atom):
+            existing = portdb.xmatch("match-all", atom)
+            if not existing:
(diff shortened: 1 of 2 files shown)


lib/portage/emaint/modules/world/world.py
```diff
@@ -55,14 +55,20 @@ def _check_world(self, onProgress):
...
                 self.not_installed.append(atom)
             elif not portdb.xmatch("match-visible", atom):
                 try:
                     reasons = ", ".join(
-                        {r for cpv in installed for r in getmaskingstatus(cpv)}
+                        {
+                            r
+                            for cpv in existing
+                            for r in getmaskingstatus(cpv)
+                            if cpv.version != "9999"
```
thesamesam: Please check if it has the "live" property instead.

jeth-ro: Sorted, I think.

thesamesam (approved):

PR #1580, opened 2026-04-30
Add high-performance io_uring support for file operations

ghost:
## Overview
This PR introduces support for the Linux `io_uring` interface to Portage, significantly optimizing I/O-intensive operations. By leveraging the `liburing-py` library, Portage can now perform asynchronous, non-blocking I/O, allowing kernel-side data transfers to overlap with user-space processing (such as hash computations).

### 1. Robust io_uring Utility Wrapper (`portage.util.io_uring`)
Introduced a high-level `IoUring` class that serves as a clean Pythonic interface to `liburing-py`.
- **Batch Submission**: Implemented logic to "prep" multiple operations and submit them in a single system call, drastically reducing context-switch overhead.
- **Extended Opcode Support**: Support for `prep_read`, `prep_write`, `prep_splice`, `prep_openat`, `prep_close`, and `prep_copy_file_range`.
- **Deterministic Cleanup**: Designed as a Context Manager to ensure kernel rings are properly exited and resources released.
- **Graceful Detection**: Features `is_available()` to check for both the Python module and kernel-level support, ensuring Portage remains compatible with all environments.

### 2. Optimized Checksumming Mechanism (`portage.checksum`)
The file hashing logic has been overhauled to maximize throughput during Manifest verification and creation.
- **Double-Buffering Pipeline**: Implemented a `_data_reader` that uses a two-buffer rotation. While user-space is hashing one block, the kernel is concurrently fetching the next block via `io_uring`.
- **Zero-Copy Reads**: Integrated `memoryview` for yielding data chunks. This avoids memory allocations and data copies when passing buffer slices to hash objects.
- **Parallel Multi-Hash**: Refactored `perform_multiple_checksums` to process a file in a single pass. All requested hash objects (MD5, SHA256, BLAKE2B, etc.) are updated concurrently as data arrives from the async pipeline.

### 3. Accelerated File Copying (`portage.util.file_copy`)
Integrated `io_uring` into the `_fastcopy` path to provide faster file duplication.
- **Kernel-Side Copies**: Added `_io_uring_copy` using `prep_copy_file_range`.
- **Optimization Hierarchy**: If reflinks (`FICLONE`) are not supported by the underlying filesystem, Portage now prioritizes asynchronous `io_uring` copies before falling back to standard synchronous `copy_file_range` or `sendfile`.

### 4. Build System Integration
- Updated `portage/lib/portage/util/meson.build` to register the new `io_uring.py` module, ensuring it is correctly installed and packaged.

## Technical Benefits
- **Throughput**: Pipelining I/O and CPU tasks ensures neither the disk nor the processor is idle during large file operations.
- **Efficiency**: Batching requests reduces the "Syscall Tax," leading to lower overall CPU usage for the same I/O volume.
- **Memory Efficiency**: Pre-allocated buffers and `memoryview` slices reduce pressure on the Python Garbage Collector.

## Compatibility & Fallbacks
- **Optional Dependency**: This feature requires `dev-python/liburing`.
- **Robust Fallback**: Portage maintains full backward compatibility. If `liburing-py` is missing or the kernel does not support specific `io_uring` opcodes (e.g., on older kernels or cross-filesystem copies), Portage will automatically and transparently revert to standard synchronous I/O paths.

## Testing
- Verified checksum consistency between standard and `io_uring` paths.
- Tested file copy robustness across different filesystem boundaries (EXT4, BTRFS).
- Confirmed graceful fallback on systems without `liburing-py`.

### DIFF
--- a/lib/portage/checksum.py
+++ b/lib/portage/checksum.py
@@ -1,5 +1,5 @@
 # checksum.py -- core Portage functionality
-# Copyright 1998-2020 Gentoo Authors
+# Copyright 1998-2026 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 # pylint: disable=ungrouped-imports
 
@@ -15,7 +15,7 @@ from portage import _encodings, _unicode_decode, _unicode_encode
 from portage import os
 from portage.const import HASHING_BLOCKSIZE, PRELINK_BINARY
 from portage.localization import _
-
+from portage.util.io_uring import IoUring, is_available as is_uring_available
 
 # Summary of all available hashes and their implementations,
 # most preferred first. Please keep this in sync with logic below.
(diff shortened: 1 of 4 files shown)


mgorny:
Using LLMs is in violation of Gentoo policy and results in immediate ban.

PR #1581, opened 2026-05-01
dispatch-conf: Color on

immolo:
All the required packages are now included by default to support colour in dispatch-conf.

This commit turns on the mode supported by sys-apps/diffutils and adds a link to wiki to show other options that user to can select so there is still user choice for those that need it, and a good default choice for those unaware.

### DIFF
--- a/cnf/dispatch-conf.conf
+++ b/cnf/dispatch-conf.conf
@@ -26,7 +26,8 @@ use-rcs=no
 # %s new file
 # If using colordiff instead of diff, the less -R option may be required
 # for correct display.
-diff="diff -Nu '%s' '%s'"
+# see https://wiki.gentoo.org/wiki/Dispatch-conf for more color options.
+diff="diff --color=always -Nu '%s' '%s'"
 
 # Set the pager for use with diff commands (this will
 # cause the PAGER environment variable to be ignored).


Flowdalic:
Typo in commit message:

> good default choice fot those unaware.

s/fot/for/

I also wonder why we don't use `--color=auto`?

immolo:
> Typo in commit message:
> 
> > good default choice fot those unaware.
> 
> s/fot/for/

One day I'll learn to spell, thanks! :P

> I also wonder why we don't use `--color=auto`?

Being honest I only have this [forum post](https://superuser.com/questions/417957/is-there-any-way-to-keep-text-passed-to-head-tail-less-etc-to-be-colored) and the not very helpful `diff` manpage, so I'm happy to change to what is decided to be the best way forward.

As I understand it, passing `auto` will turn off when passed to another tool like `less` which we use here https://github.com/gentoo/portage/blob/master/bin/dispatch-conf#L144

So if I'm reading this correctly `dispatch-conf` either has a default or fallback behaviour which requires forcing.

Originally though I just used what the wiki suggests and didn't think much pass that as it's been there for 7 years without complaint.

https://wiki.gentoo.org/index.php?title=Dispatch-conf&type=revision&diff=812264&oldid=812246

(Also thanks for asking this as that was a surprisingly fun rabbit hole to enjoy with my morning coffee.)

PR #1584, opened 2026-05-27
Add FEATURES="buildpkg-proactive" to rebuild missing/unusable binpkgs

chewi:
If you build binary packages in order to do a `--usepkgonly` build later, you want to be sure that no packages are missing and that the packages you do have will actually be visible, e.g. they were not built when a now stable package was previously unstable.

This will proactively rebuild matching packages that have no (usable) binary package, even if those packages are already installed.

This required some refactoring of the logic around whether to build a binary package or not. Note that it is not necessary to check whether `--buildpkg=n` was given because `adjust_config()` already takes care of that.

I thought about automatically enabling `buildpkg-proactive` when `--buildpkgonly` is given, but that option applies to / as well as ROOT, so this would have been unhelpful.

Bug: https://bugs.gentoo.org/976083

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,6 +9,14 @@ Release notes take the form of the following optional categories:
 portage-3.0.80 (UNRELEASED)
 --------------
 
+Breaking changes:
+
+* Drop the @unavailable-binaries set in favour of FEATURES="buildpkg-proactive"
+  (see below) because the former cannot check whether an existing binary package
+  is usable or whether a given ebuild will actually produce a binary package or
+  not. This set is broken anyway and probably has been for 11 years since
+  Portage 2.2.19. (bug #976083)
+
 Features:
 
 * Add User-Agent to default wget FETCHCOMMAND and RESUMECOMMAND settings, as
--- a/NEWS
+++ b/NEWS
@@ -15,6 +15,10 @@ Features:
   well as emirrordist, in order to avoid blocking by some web servers,
   particularly crates.io. (bug #682610)
 
+* Add FEATURES="buildpkg-proactive" to proactively rebuild matching packages
+  that have no (usable) binary package, even if those packages are already
+  installed. (bug #976083)
+
 portage-3.0.79 (2026-05-22)
 --------------
 
--- a/cnf/sets/portage.conf
+++ b/cnf/sets/portage.conf
@@ -100,11 +100,6 @@ class = portage.sets.dbapi.DowngradeSet
 [unavailable]
 class = portage.sets.dbapi.UnavailableSet
 
-# Installed packages for which corresponding binary packages
-# are not available.
-[unavailable-binaries]
-class = portage.sets.dbapi.UnavailableBinaries
-
 # Installed packages for which vdb *DEPEND entries are outdated compared
 # to the matching portdb entry.
 [changed-deps]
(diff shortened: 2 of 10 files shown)


chewi:
@thesamesam pointed out that this functionality might be covered by the `@unavailable-binaries` set. I tried it out and found it to be broken. It probably has been for 11 years since Portage 2.2.19 when `binpkg-multi-instance` was introduced.

While that can easily be fixed, it still cannot check whether an existing binary package is usable or whether a given ebuild will actually produce a binary package or not. I think doing so would require the heaving lifting that depgraph.py does.

Even if I could make it work, I think this new feature is nicer for users anyway. A set might feel clever, but it you want to update `@world` and ensure you're not missing any binpkgs, you have to do it with two invocations, because `--update` prevents `@unavailable-binaries` from rebuilding packages that are already updated. `buildpkg-proactive` ensures everything is consistent while you do your regular updates.

On that basis, I have added a commit to drop `@unavailable-binaries`.

chewi:
I've successfully done a full Flatcar build with this change now.

chewi:
I'm dropping this to draft for the moment while I finalise a change requested by @eli-schwartz to consider new/changed USE flags. This will help the binhost case.

thesamesam:
> @thesamesam pointed out that this functionality might be covered by the @unavailable-binaries set. I tried it out and found it to be broken. It probably has been for 11 years since Portage 2.2.19 when binpkg-multi-instance was introduced.

Thanks for doing that. I'd suggest dropping it in a separate commit (same PR) with an explanation of why, as the commit message doesn't currently feature that.

I haven't reviewed the changes, but saying this explicitly may be useful anyway as it occurred to me when I read your bug: does the FEATURES flag pull in packages w/o a binpkg, or does it just affect/inhibit generation upon merging if no acceptable binpkg exists? Will a package be added to the merge plan if it does not otherwise need an update or rebuild, but does need a binpkg to be created?

(Please answer that in the commit message + docs as appropriate.)

chewi:
> Thanks for doing that. I'd suggest dropping it in a separate commit (same PR) with an explanation of why, as the commit message doesn't currently feature that.

But I did do it in a separate commit with an explanation? :confused: 

> Does the FEATURES flag pull in packages w/o a binpkg, or does it just affect/inhibit generation upon merging if no acceptable binpkg exists? Will a package be added to the merge plan if it does not otherwise need an update or rebuild, but does need a binpkg to be created?

I tried to answer this as concisely as possible in the man page. Maybe **matching** doesn't make it clear that it only does this with packages it is already considering for update (e.g. `@world`). Suggestions welcome.

> Proactively rebuild **matching** packages that have no (usable) binary package, even if those packages are already installed.

chewi:
Okay, I've gone with:

> Proactively rebuild packages that have no (usable) binary package. This only applies to installed packages that Portage is already checking but would otherwise not consider rebuilding, e.g. because they're already up-to-date. This does nothing unless `buildpkg` or `buildsyspkg` is also enabled. When `emerge --binpkg-respect-use` is enabled (it usually is), binary packages will also be rebuilt if the USE flags have changed.

@eli-schwartz You might want to check this actually helps with the binhost use case. Add `FEATURES="buildpkg-proactive"` and drop `--changed-use` from your emerge arguments.

thesamesam:
>  But I did do it in a separate commit with an explanation? 😕

No idea, I must have looked at the wrong tab. Thanks and sorry!

PR #1585, opened 2026-05-28
Add User-Agent to default FETCHCOMMAND/RESUMECOMMAND and emirrordist

chewi:
This is to avoid blocking by some web servers, particularly crates.io.

Bug: https://bugs.gentoo.org/682610

### DIFF
--- a/cnf/make.conf.example
+++ b/cnf/make.conf.example
@@ -154,12 +154,12 @@
 # at \${DISTDIR}/\${FILE}.
 #
 # Default fetch command (3 tries, passive ftp for firewall compatibility)
-#FETCHCOMMAND="wget -t 3 -T 60 --passive-ftp -O \"\${DISTDIR}/\${FILE}\" \"\${URI}\""
-#RESUMECOMMAND="wget -c -t 3 -T 60 --passive-ftp -O \"\${DISTDIR}/\${FILE}\" \"\${URI}\""
+#FETCHCOMMAND="wget -t 3 -T 60 --passive-ftp -U \"Portage (Gentoo, https://www.gentoo.org) distfile-fetch\" -O \"\${DISTDIR}/\${FILE}\" \"\${URI}\""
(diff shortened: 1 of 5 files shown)


zmedico (approved):

eli-schwartz (approved):
This is a bit "damned if you do, damned if you don't". Some sites also allow curl/wget, but reject UAs that they have never heard of because they assume it must be a spambot. Some sites only allow "genuine browsers" and "Mozilla" is the only decent UA.

I do think this change is good, though, because friendly sites will accept anything and announcing what we are will empower site operators to understand what the traffic is doing when trying to understand e.g. abuse patterns.

cnf/make.conf.example
```diff
@@ -154,12 +154,12 @@
 # at \${DISTDIR}/\${FILE}.
 #
 # Default fetch command (3 tries, passive ftp for firewall compatibility)
-#FETCHCOMMAND="wget -t 3 -T 60 --passive-ftp -O \"\${DISTDIR}/\${FILE}\" \"\${URI}\""
-#RESUMECOMMAND="wget -c -t 3 -T 60 --passive-ftp -O \"\${DISTDIR}/\${FILE}\" \"\${URI}\""
+#FETCHCOMMAND="wget -t 3 -T 60 --passive-ftp -U \"Portage (Gentoo, https://www.gentoo.org) distfile-fetch\" -O \"\${DISTDIR}/\${FILE}\" \"\${URI}\""
```
eli-schwartz, line 157: Note based on IRC, a suggestion was made to explicitly omit the portage version.

Including a version is common, because it allows site operators to understand which bugs the software might have, which is sometimes relevant for abuse analysis.

Not sure what to think about in this specific scenario, given that we also ideally only download any given url once. For sure if we were communicating with an API and sending queries, I'd say we would need the version number included.

chewi: I think we can trust Portage to not do anything stupid here. I'm considering a later change to have Portage batch fetches, allowing wget to reuse connections, but if there's any communication issue with that, it will be with wget, not Portage.

floppym (approved):
This seems ok to me.

In the future, maybe we could add a variable like `PORTAGE_USER_AGENT` to allow users to customize this without overriding `FETCHCOMMAND`. We could also use said variable for the urllib calls embedded in Portage's Python code.

PR #1590, opened 2026-06-10
doebuild: Re-export SYSROOT, ESYSROOT, and BROOT in EAPI 9

Flowdalic:
It turns out that SYSROOT, ESYSROOT, and BROOT need to be exported for cross compilation. Re-export those PMS variables again in EAPI 9.

Bug: https://bugs.gentoo.org/977170


floppym:
Looks like Chewi beat you to it with https://github.com/gentoo/portage/pull/1589.

Flowdalic:
> Looks like Chewi beat you to it with #1589.

Let's go with Chewi's PR then!

PR #1594, opened 2026-06-20
Atom: drink up the string soup (rebase of #1497 + more)

mattst88:


Jannik2099:
thanks matt for picking up on this - I've been very short on time lately, sorry.

In the long term we should probably get rid of all `sorted(atoms, key=str)` aswell, as the only legitimate use of atom ordering is version comparisons (right?).

For the public record, I approve of reusing and modifying my commits for this.

lib/_emerge/depgraph.py
```diff
@@ -4495,8 +4499,8 @@ def _wrapped_add_pkg_dep_string(
                 # from dep_check, map it back to the original, in
                 # order to avoid distortion in places like display
                 # or conflict resolution code.
-                is_virt = hasattr(atom, "_orig_atom")
-                atom = getattr(atom, "_orig_atom", atom)
+                is_virt = atom.orig_atom is not None
+                atom = atom.orig_atom if atom.orig_atom is not None else atom
```
thesamesam: Match the above hunk:
```suggestion
                atom = atom.orig_atom if is_virt else atom
```

lib/_emerge/depgraph.py
```diff
@@ -6030,7 +6035,7 @@ def _expand_virt_from_graph(self, root, atom):
 
             for atoms in rdepend.values():
                 for atom in atoms:
-                    if hasattr(atom, "_orig_atom"):
+                    if getattr(atom, "orig_atom", None) is not None:
```
thesamesam: What about this instead (and elsewhere)?

```suggestion
                    if getattr(atom, "orig_atom", False):
```

Or, wait, why not just use hasattr again?

lib/_emerge/depgraph.py
```diff
@@ -6844,7 +6849,7 @@ def _show_unsatisfied_dep(
             mask_docs = True
         else:
             cp_exists = False
-            if atom.package and not atom.cp.startswith("null/"):
+            if atom.package and atom.category != "null":
```
thesamesam, line 6854: I'm not asking you to fix this now, but this null crap really needs to go in the future.

thesamesam (commented):
Some small comments from a first pass. Thanks a lot to you and Jannik for doing this.

Please make sure you're daily driving it if you're not already (I recommend this in general for Portage patches), it's invaluable testing.

lib/_emerge/depgraph.py
mattst88: Done.

lib/_emerge/depgraph.py
mattst88: Done.

lib/_emerge/depgraph.py
```diff
@@ -5946,9 +5951,9 @@ def _select_atoms_highest_available(
                 chain(
                     (id(atom) for atom in mycheck[1]),
                     (
-                        id(atom._orig_atom)
+                        id(atom.orig_atom)
                         for atom in mycheck[1]
-                        if hasattr(atom, "_orig_atom")
+                        if getattr(atom, "orig_atom", None) is not None
```
thesamesam: This could again be False not None, but why are we changing from hasattr here anyway? Can't all of these checks be hasattr but with the new attribute name?

lib/portage/dep/__init__.py
```diff
@@ -1605,77 +1753,64 @@ def __init__(
...
                 use = _use
             else:
                 use = _use_dep(use_str[1:-1].split(","), eapi_attrs)
-            without_use = Atom(
-                blocker_prefix + m.group("without_use"), allow_repo=allow_repo
-            )
         else:
             use = None
-            if unevaluated_atom is not None and unevaluated_atom.use is not None:
-                # unevaluated_atom.use is used for IUSE checks when matching
```
thesamesam, line 1661: Should this comment be preserved? It gets lost in the new `without_use` but AFAICT, the comment here was referring to the attribute, so I _think_ it would make sense to keep the comment and move it over there.

lib/portage/versions.py
```diff
@@ -19,9 +19,19 @@
...
-from typing import Any, Optional, Union
+from typing import TYPE_CHECKING, Any, Optional, Union
 from collections.abc import Sequence
 
+if TYPE_CHECKING:
+    from portage.dep import Atom
+
+import portage
+
+portage.proxy.lazyimport.lazyimport(
```
thesamesam: I think this is from a bad rebase. We got rid of almost all lazyimport in 0526272794560e37e242a34a42522f891229887f..cbfbdc017247e6e4f71beea67c2400fa56d934fe.

lib/_emerge/resolver/slot_collision.py
```diff
@@ -520,7 +520,7 @@ def highlight_violations(atom, version, use, slot_violated):
                         if version:
                             op = atom.operator
                             ver = None
-                            if atom.cp != atom.cpv:
+                            if atom.is_versioned:
                                 ver = cpv_getversion(atom.cpv)
```
thesamesam: Can't `atom.version` be used directly now instead, no need for `cpv_getversion`?

so just:
```python
ver = atom.version if atom.is_versioned else None
```

But given what `is_versioned` does, I think even this may be fine:
```python
ver = atom.version
```

(Repeated elsewhere.)

lib/portage/versions.py
mattst88: Oops, thanks for noticing.

lib/_emerge/resolver/slot_collision.py
mattst88: Good call. Done.

lib/portage/dep/__init__.py
mattst88: Fixed.

lib/_emerge/depgraph.py
mattst88: `orig_atom` is always defined (to `None`) on `Atom`, so `hasattr(atom, "orig_atom")` would always return `True`.

These checks are in loops where you may be processing non-Atom types as well (e.g. `SonameAtom`, I think) that don't have the `orig_atom` attribute.

thesamesam: Ah, duh, thanks.

thesamesam (approved):
This looks good to me now, thanks! I assume all of @zmedico's comments from #1497 are handled (at a glance it looks like they are).

I wouldn't be surprised if there's some fallout, so be prepared for that.

mattst88:
Merged!

PR #1595, opened 2026-06-21
Add FEATURES="observability" for emerge status reporting

mattst88:
I've been playing with writing an external portage monitoring utility. This PR adds `FEATURES="observability"` that lets external tools inspect a running emerge process without instrumenting emerge itself.

When enabled, the Scheduler publishes a machine-readable JSON snapshot of its current state to `/run/portage/emerge-<pid>.json` on every job state change, listing packages being built or merged, their current ebuild phase, elapsed time, and packages waiting to merge. The file is removed when emerge exits.

A Unix-domain socket at `/run/portage/emerge-<pid>.sock` streams JSON snapshots to connected clients: the current snapshot on connect, then one line per update.

Two new query interfaces are added:
- `emerge --status`: human-readable table of active builds across all running emerge processes
- `portageq jobs [--json]`: same data in machine-readable form

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -183,6 +183,11 @@ Breaking changes:
 
 Features:
 
+* Add FEATURES="observability" to publish a machine-readable snapshot of the
+  running emerge process (which packages are building/merging, their current
+  ebuild phase, and elapsed time) to /run/portage/emerge-<pid>.json. Query it
+  with the new "portageq jobs" command or "emerge --status".
+
 * Add User-Agent to default wget FETCHCOMMAND and RESUMECOMMAND settings, as
   well as emirrordist, in order to avoid blocking by some web servers,
   particularly crates.io. (bug #682610)
--- a/NEWS
+++ b/NEWS
@@ -186,7 +186,8 @@ Features:
 * Add FEATURES="observability" to publish a machine-readable snapshot of the
   running emerge process (which packages are building/merging, their current
   ebuild phase, and elapsed time) to /run/portage/emerge-<pid>.json. Query it
-  with the new "portageq jobs" command or "emerge --status".
+  with the new "portageq jobs" command or "emerge --status". A Unix socket
+  streams the same snapshots live to connected clients.
 
 * Add User-Agent to default wget FETCHCOMMAND and RESUMECOMMAND settings, as
   well as emirrordist, in order to avoid blocking by some web servers,
--- a/bin/portageq
+++ b/bin/portageq
@@ -219,6 +219,25 @@ try:
 		"""
     mass_best_version.__doc__ = docstrings["mass_best_version"]
 
+    def jobs(argv):
+        import json as _json
+        from _emerge._observability import read_snapshots, format_snapshots
+        from portage.const import EPREFIX
+
+        snapshots = read_snapshots(EPREFIX)
+        if "--json" in argv:
+            print(_json.dumps(snapshots, sort_keys=True, indent=2))
+        else:
+            sys.stdout.write(format_snapshots(snapshots))
+        return 0
+
+    docstrings["jobs"] = """[--json]
+		Show packages currently being built/merged by running emerge
+		processes (requires FEATURES="observability").  With --json, emit
+		the raw status snapshots.
+		"""
+    jobs.__doc__ = docstrings["jobs"]
+
     @uses_eroot
     def metadata(argv):
         if len(argv) < 4:
--- a/lib/_emerge/EbuildPhase.py
+++ b/lib/_emerge/EbuildPhase.py
@@ -141,6 +141,14 @@ class EbuildPhase(CompositeTask):
     _locked_phases = ("setup", "preinst", "postinst", "prerm", "postrm")
 
     def _start(self):
+        # Notify the observability monitor (if any) that this package has
+        # entered a new ebuild phase, so "what is building?" queries reflect
+        # the live phase.  Guarded since not every SchedulerInterface that
+        # runs phases (e.g. the standalone `ebuild` command) provides it.
+        notify_phase = getattr(self.scheduler, "notifyPhase", None)
+        if notify_phase is not None:
+            notify_phase(self.settings.mycpv, self.phase)
+
         future = asyncio.ensure_future(self._async_start(), loop=self.scheduler)
         self._start_task(AsyncTaskFuture(future=future), self._async_start_exit)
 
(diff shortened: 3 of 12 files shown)


mattst88:
@robbat2 wondered whether using OpenTelemetry would make sense here, so I looked into it.

As far as I can tell, it's not a good match:

- OpenTelemetry is designed around (a number of producers) pushing time series data to a collector over the network. This feature is intended to just be local inspection of the current state.
- This feature is just providing a snapshot of the full status of portage on each update. OpenTelemetry is designed around measurements at a particular instant, not "here's the whole state".
- Dependencies... this PR adds no dependencies. Adding dependencies to portage is not something to take lightly. Using OpenTelemetry would require installing `dev-python/opentelemetry-{sdk,semantic-conventions,api}`.
- OpenTelemetry requires some configuration, and a running collector. This PR just works without any config. `portageq jobs` just works.

I think it might make sense to layer OpenTelemetry on top of this by consuming the JSON from the socket and emitting OpenTelemetry, but I don't think OpenTelemetry is the right fit for this problem.

desultory:
I'd be interested in some way to have portage emit events to a system like Loki to track the status of portage. Somewhat like what you get in /var/log/emerge.log, but more tuned to be hooked up to a modern monitoring system (with minimal middleware)

Maybe expanding current logging could be more of a benefit than adding a local socket to track the current state? 
Given how portage is generally used, I think using a push model makes more sense, that way you can be sure the system is collecting important events and not missing things because of bad luck/insufficient polling

mattst88:
This branch implements a push model, to be clear.

As far as I understand Loki (which I'm generally unfamiliar with) ingests log lines. So it's taking in discrete events, whereas this PR is providing full state snapshots.

Pushing to Loki would mean, like with OpenTelemetry, that you have to configure an address, auth token, etc.

Note that if a job finished in <1 second that it could be missed in the current PR. I think we should be so lucky if that happened, and I am fine with it not appearing.

If you want auditable history, I agree that discrete events is what you want, and that's what emerge.log is for.

If we did want to emit the discrete events the snapshots in the current PR contain to emerge.log, we'd need to add

- Start and end of each phase (unpack, prepare, et al)
- merge-wait enter/exit
- emerge PID

This would give us persistent history, no missed transitions (things happening in <1 second), and a natural fit for log-based tooling. It would mean that tooling would have to reconstruct state by replaying events -- which is more work than "just display the current state" -- but would also mean you could do it at arbitrary points. Not sure if that's valuable though?

It would also mean that the log would be pretty verbose, and given the number of external tools that read emerge.log already, I'd probably want to put this in a separate log file (emerge-events.log?)

I think all this comes down to the question: is it valuable to reconstruct state and audit emerge events, or do we just want to monitor it in the moment? I've been targeting the latter, and it's not clear to me why we would necessarily want the former at a granularity greater than what emerge.log currently provides.

mattst88:
Maybe the socket should give the full state on connect (which is current does), and then discrete events afterwards?

mattst88:
This would mean a monitoring program would have to build its own state model, so a lot of additional complexity on the consumer side just for displaying what's currently happening.

Seems like it gets back to

> I think all this comes down to the question: is it valuable to reconstruct state and audit emerge events, or do we just want to monitor it in the moment? I've been targeting the latter, and it's not clear to me why we would necessarily want the former at a granularity greater than what emerge.log currently provides.

desultory:
Yes, if all events were logged, something like promtail or alloy could be used to scrape those and ship them to loki. I don't think a pull model makes a lot of sense here. The pipeline could have its own filtering mechanism as well, so more info than needed shouldn't be a huge problem. This setup is a bit atypical because generally "push" based monitoring systems will send the data directly to a system, rather than to a socket which can be used. Because it's buffered it doesn't have the typical disadvantages of data loss you'd get with a typical pull model, but it would require middleware to actually transport that data to a monitoring system, unless it was operating purely locally.

I think it would be very cool if portage had some more native integration so it could ship things to a system like loki natively, but I can see why that would be hard to implement. I'm coming at this from the angle that many projects will natively expose a "metrics" interface that is compatible with Prometheus.

At the very least, having the logs JSON formatted as an option should greatly simplify the scraping logic and make it more interoperable. I don't know if it makes more sense to have this new feature be a new mechanism built around sockets, or if it could make more sense to just give the logging an option to use JSON and/or sockets instead of a plain text file

ajakk:
I like it!

I don't really see the point of discussing higher level observability technologies here. Seems like a bad idea to tie ourselves to one or the other and to carry all of the implementation complexity. With the source state exposed, push or pull middleware shouldn't be hugely complex to implement.

With how simple this is I almost think it should be on by default (eg the log isn't optional) but I don't feel strongly.

mattst88:
To close the loop on the events question, since I raised it twice and then talked myself out of my own suggestion.

The snapshot carries gauges, not just transitions. I've got work in progress code that adds per-task CPU time, current/peak memory and I/O to each task entry, and build_snapshot() samples them inline:

    if cgroup is not None:
        res = cgroup.read_stats(str(pkg.cpv))

There is no scheduler state change to hang those off. "Memory is now 4.2 GB" is not an event. Emitting them as one would mean inventing a periodic sample event fired on a timer, which is a snapshot with extra steps and a worse schema.

That also kills my own suggestion upthread of "full state on connect, then discrete events afterwards" -- that is the one variant that breaks gauge consumers, since their numbers would freeze at connect time.

The two want opposite cadences, which is the interesting part. Snapshots are expensive (read_stats() is per-task cgroupfs I/O every time) and want the existing 1s rate limit. Events are cheap (notify_phase already has the cpv and phase in hand, no I/O) and must be unthrottled, since not missing sub-second transitions is the entire point of having them. So it's both, not either -- one stream, each kind at its natural rate.

Concretely, what I plan to do:

1. This PR stays snapshot-only. I have added a "type": "snapshot" field to each object so the socket can interleave other kinds later without breaking existing clients. Consumers should dispatch on it and ignore what they do not recognize.

2. Events land separately. @desultory, for the Loki case I think the socket is the wrong surface anyway -- promtail and alloy scrape files. What you want is an emerge-events.log with phase start/end, merge-wait enter/exit and the emerge PID, JSON-formatted, separate from emerge.log so the existing parsers do not have to care. That is a self-contained change and I would rather not grow this PR into it.

Agreed with @ajakk that the right move is to expose the source state and let middleware decide, rather than tie portage to any particular stack.

PR #1596, opened 2026-06-22
Respect --usepkg=n and --getbinpkg=n on command line if FEATURES=getbinpkg

Flowdalic:
Portage would ignore --usepkg=n or --getbinpkg=n on the command line, if the user has FEATURES=getbinpkg in their make.conf. So, instead of

    emerge --usepkg=n <package>

users with FEATURES=getbinpkg in make.conf would need to use

    FEATURES=-getbinpkg emerge <package>

to perform a source-based emerge.

This was caused by Portage's option parser converting 'n' to None, which prevented it from being added to myopts, because None values are ignored.

Fix this by converting 'n' to False instead of None. This ensures that the users intend is recored in myopts.

As a consequence, all value retrieval checks need to be adjusted, to check of True, instead of the sole existences of the option in myopts.

Finally, the configuration override via features only kicks in if the command line option does not disaggree with it.

Closes: https://bugs.gentoo.org/759067

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -37,6 +37,9 @@ Bug fixes:
   reduced-privilege phases can source them where the process TMPDIR is not
   traversable by the portage user. (bug #977245)
 
+* Respect --usepkg=n and --getbinpkg=n on command line if the user has set
+  FEATURES=getbinpkg (bug #759067)
+
 portage-3.0.79 (2026-05-22)
 --------------
 
--- a/lib/_emerge/Scheduler.py
+++ b/lib/_emerge/Scheduler.py
@@ -849,7 +849,7 @@ class Scheduler(PollScheduler):
 
         elif (
             pkg.type_name == "binary"
-            and "--getbinpkg" in self.myopts
+            and self.myopts.get("--getbinpkg") is True
             and pkg.root_config.trees["bintree"].download_required(pkg.cpv)
         ):
             prefetcher = BinpkgPrefetcher(
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -145,7 +145,7 @@ def action_build(
         + os.path.sep
     )
     quickpkg_direct = (
-        "--usepkg" in emerge_config.opts
+        emerge_config.opts.get("--usepkg") is True
         and emerge_config.opts.get("--quickpkg-direct", "n") == "y"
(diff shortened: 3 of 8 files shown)


chewi (approved):
This looks fine. It's strange that these options also accept `True` but not `False`. Furthermore, `man emerge` says the options are `y`, `n`, `true`, and `false` (note the casing). I'll drop the mention of `true` and `false`. I don't think `True` should be mentioned at all because it's just weird.

Flowdalic:
> I'll drop the mention of `true` and `false`. I don't think `True` should be mentioned at all because it's just weird.

Ah, I forgot about this portage behavior. I don't think we could ever drop the behavior, but we could drop it being mentioned in the man page (and if it's just for simplification purposes).

chewi:
Umm, I don't think we should add `False` to the mix. I was just remarking that it was strange. I've already pulled your earlier version of this, and I'm happy with it.

thesamesam:
The argparsing really needs redoing. Anyway, this looks fine, thanks.

Flowdalic:
> Umm, I don't think we should add `False` to the mix.

If we don't, then `emerge --getbinpkg=n <package>` behaves potentially different than `emerge --getbinpkg=False`, which is rather surprising (at least to me).

> I've already pulled your earlier version of this, and I'm happy with it.

You mean you pulled the earlier version in a local branch of yours?

> Furthermore, `man emerge` says the options are `y`, `n`, `true`, and `false`

I just looked again in the man page, could you point me to where this is stated? I only find a note about this for the `--alert` argument.

chewi:
> If we don't, then `emerge --getbinpkg=n <package>` behaves potentially different than `emerge --getbinpkg=False`, which is rather surprising (at least to me).

It is if you know about `True`, but strictly speaking, that isn't even documented. I've only ever seen people use y/n, and I'd rather keep it that way.

> You mean you pulled the earlier version in a local branch of yours?

Yeah, there was a regression in the last release, so I'm about to get another one out. I've been sent a bunch of other fixes though, so there's a pile to go through.

> I just looked again in the man page, could you point me to where this is stated? I only find a note about this for the `--alert` argument.

I think the idea was to mention these options in detail under `--alert` because it's the first alphabetically. I couldn't see anything special about `--alert` from the code. This could be made clearer.

PR #1597, opened 2026-06-22
Update black to 26.5.1 and re-format

Flowdalic:


thesamesam (commented):
Could you check out the CI failure? I think this is OK otherwise but it may be best to wait until the next release is cut, as it would require rebasing some patches that need to be merge shortly.

Flowdalic:
I've (hopefully) fixed with this `pre-commit run --all-files`.

Flowdalic:
Yep, it's green now

thesamesam:
Thanks. Do think we need to reconsider black overall (see IRC discussion overnight) but this looks fine for now.

But I think maybe we should wait until matt's mega unicode PR is in, to avoid inflicting pain.

mattst88:
> But I think maybe we should wait until matt's mega unicode PR is in, to avoid inflicting pain.

Thanks, planning to land it today.

PR #1599, opened 2026-06-24
Deduplicate the ld.so.conf.d include

anthonyryan1:
Regression was introduced during iteration. An earlier iteration of 4e154adc45136c040308e536367bf92e142f4c5b had attempted to solve the problem differently, and these lines were an artifact of an inferior approach.

During visual inspection of /etc/ld.so.conf after changes I misread the prepended include as part of the initial comment block, and inspected all lines beginning with /usr and afterward.

Part-of: #1592


anthonyryan1:
@thesamesam

thesamesam:
Thanks for fixing it quickly!

PR #1600, opened 2026-06-25
fetch: add FETCH_WRAPPER support for command wrapping

mattst88:
When `FETCHCOMMAND` contains a wrapper that re-interprets its arguments through a shell (most commonly ssh) the fetch breaks with a shell syntax error.

ssh joins all post-hostname arguments with spaces to form the remote command string, then passes that string to the remote shell. Arguments that were separate tokens when passed to execve() end up concatenated without quoting. The default `FETCHCOMMAND` includes:

```
  -U "Portage (Gentoo, https://www.gentoo.org) distfile-fetch"
```

After `shlex.split()`, the user-agent is correctly one argument, but once ssh joins everything for the remote shell the parentheses are unquoted:

```
  wget ... -U Portage (Gentoo, ...) distfile-fetch -O ...
```

The remote shell then fails with:

```
  bash: -c: line 1: syntax error near unexpected token `('
```

A `FETCHCOMMAND` such as the following worked before the user-agent was added in commit 421ab5457 ("Add User-Agent to default FETCHCOMMAND/RESUMECOMMAND and emirrordist"):
```
  FETCHCOMMAND="ssh -i ~/.ssh/id_ed25519 user@host ${FETCHCOMMAND}"
```
When `FETCH_WRAPPER` is set, the `FETCHCOMMAND` is expanded and each argument individually shell-quoted via `shlex.quote()`, then joined into a single string passed as one argument to the wrapper. The wrapper (e.g. ssh) forwards this single string to the remote shell, which parses it correctly.

Example make.conf usage:
```
  FETCH_WRAPPER="ssh -i ~/.ssh/id_ed25519 user@host"
```
`FETCHCOMMAND` and `RESUMECOMMAND` can then be left as their `make.globals` defaults, including any future changes to the user-agent string.

### DIFF
--- a/lib/portage/package/ebuild/_config/special_env_vars.py
+++ b/lib/portage/package/ebuild/_config/special_env_vars.py
@@ -287,6 +287,7 @@ environ_filter = frozenset(
         "EMERGE_DEFAULT_OPTS",
         "EMERGE_LOG_DIR",
         "EMERGE_WARNING_DELAY",
+        "FETCH_WRAPPER",
         "FETCHCOMMAND",
         "FETCHCOMMAND_FTP",
         "FETCHCOMMAND_HTTP",
--- a/lib/portage/package/ebuild/fetch.py
+++ b/lib/portage/package/ebuild/fetch.py
@@ -1813,8 +1813,16 @@ async def async_fetch(
                         if v is not None:
                             variables[k] = v
 
-                    myfetch = varexpand(locfetch, mydict=variables)
-                    myfetch = shlex.split(myfetch)
+                    fetch_wrapper = mysettings.get("FETCH_WRAPPER")
+                    if fetch_wrapper:
+                        wrapper_args = shlex.split(
+                            varexpand(fetch_wrapper, mydict=variables)
+                        )
+                        fetch_args = shlex.split(varexpand(locfetch, mydict=variables))
+                        myfetch = wrapper_args + [shlex.join(fetch_args)]
+                    else:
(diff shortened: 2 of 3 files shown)


thesamesam:
I have no opinion on this but just noting: are you aware of @hartwork's app-portage/fetchcommandwrapper? It looks quite aria2-specific so maybe there's not really any overlap here.

(Also, it looks like fcw needs to set UA too.)

mattst88:
> I have no opinion on this but just noting: are you aware of @hartwork's app-portage/fetchcommandwrapper? It looks quite aria2-specific so maybe there's not really any overlap here.

Yep, I've used it in the past. It's orthogonal to `FETCH_WRAPPER`. You could even use them together. You use fetchcommandwrapper by setting `FETCHCOMMAND`.

chewi:
I'm not against this as such, but did you know this works?
```sh
FETCHCOMMAND="ssh localhost \"$(portageq envvar FETCHCOMMAND | sed 's:":\\\\":g')\"" emerge -f hello
```

Or in full:
```sh
FETCHCOMMAND="ssh localhost \"wget -t 3 -T 60 --passive-ftp -U \\\"Portage (Gentoo, https://www.gentoo.org) distfile-fetch\\\" -O \\\"\${DISTDIR}/\${FILE}\\\" \\\"\${URI}\\\"\"" emerge -f hello
```

mattst88:
That is hilariously gross :)

PR #1601, opened 2026-06-25
Remove Python 2 unicode compatibility layer

mattst88:
This finishes the work attempted in #803.

Python 3 strings are unicode natively. The portage unicode wrapper infrastructure -- `_unicode_func_wrapper`, `_unicode_module_wrapper`, `utf8_mode`, `_encodings`, `_unicode_encode()`, `_unicode_decode()` -- was Python 2 compat that has been dead weight for years.

This series removes it in logical stages: rename the module wrappers to make their shadowing of stdlib names explicit, strip the compat functions and `encode()` call sites, then delete the wrapper infrastructure itself.

Notable changes:
- `movefile()` drops the `encoding=` parameter (unused by any in-tree caller)
- `portage._encodings`, `portage._unicode_encode()`, `portage._unicode_decode()`, and `portage._native_string()` are retained as deprecated stubs with DeprecationWarning for third-party compatibility (e.g. gentoolkit)

The best part is the diff stat:

```
263 files changed, 1354 insertions(+), 2348 deletions(-)
```


lib/portage/xpak.py
```diff
@@ -52,23 +48,29 @@
...
-        _unicode_decode(curdir, encoding=_encodings["fs"], errors="strict")
-    )
+    if isinstance(curdir, bytes):
+        if isinstance(curdir, bytes):
+            curdir = curdir.decode("utf-8", "strict")
+    curdir = normalize_path(curdir)
     for parent, dirs, files in os.walk(curdir):
-        parent = _unicode_decode(parent, encoding=_encodings["fs"], errors="strict")
+        if isinstance(parent, bytes):
+            if isinstance(parent, bytes):
```
thesamesam: Dupe.

lib/portage/xpak.py
```diff
@@ -52,23 +48,29 @@
 def addtolist(mylist, curdir):
     """(list, dir) --- Takes an array(list) and appends all files from dir down
     the directory tree. Returns nothing. list is modified."""
-    curdir = normalize_path(
-        _unicode_decode(curdir, encoding=_encodings["fs"], errors="strict")
-    )
+    if isinstance(curdir, bytes):
+        if isinstance(curdir, bytes):
```
thesamesam: Dupe.

thesamesam (commented):
Just noticed this quickly - mentioning it now so you don't waste any time debugging any other possible substitution failures, will do a proper first pass later

bin/dispatch-conf
```diff
@@ -178,10 +178,11 @@ class dispatch:
 
             try:
                 # Find existing configs
-                path_list = _unicode_decode(
-                    subprocess.check_output(["find", path] + find_opts),
-                    errors="strict",
-                ).splitlines()
+                path_list = (
+                    subprocess.check_output(["find", path] + find_opts)
```
thesamesam: Should we pass `encoding=` (and `errors=`) to `check_output` instead for these?

lib/_emerge/actions.py
```diff
@@ -2037,7 +2036,7 @@ def action_info(settings, trees, myopts, myfiles):
         except OSError:
             pass
         else:
-            output = _unicode_decode(proc.communicate()[0]).splitlines()
+            output = proc.communicate()[0].decode("utf-8", "replace").splitlines()
```
thesamesam: This is another `encoding=` candidate if moved to `Popen`.

lib/_emerge/resolver/output_helpers.py
```diff
@@ -9,7 +9,7 @@
 
 import re
 
-from portage import os
+import os
```
thesamesam, line 12: Nit: I think this should now be grouped with `import re`.

lib/_emerge/Binpkg.py
```diff
@@ -376,15 +373,13 @@ async def _unpack_metadata(self, loop=None):
...
-                    os.path.join(infloc, k), encoding=_encodings["fs"], errors="strict"
-                ),
+                os.path.join(infloc, k),
                 mode="w",
-                encoding=_encodings["content"],
+                encoding="utf-8",
                 errors="backslashreplace",
             )
             try:
-                f.write(_unicode_decode(v + "\n"))
```
thesamesam, line 387: This should use a context manager.

lib/_emerge/Binpkg.py
```diff
@@ -376,15 +373,13 @@ async def _unpack_metadata(self, loop=None):
...
+                f.write(v + "\n")
```
thesamesam, line 382: You could do `f.write(f"{v}\n")` for consistency with other uses here.

lib/_emerge/JobStatusDisplay.py
```diff
@@ -126,7 +125,7 @@ def _init_term(self):
         term_codes = {}
         for k, capname in self._termcap_name_map.items():
             # Use _native_string for PyPy compat (bug #470258).
```
thesamesam, line 127: Stale comment.

lib/portage/_sets/files.py
```diff
@@ -159,49 +156,44 @@ def multiBuilder(self, options, settings, trees):
                 )
 
         try:
-            directory = _unicode_decode(
-                directory, encoding=_encodings["fs"], errors="strict"
-            )
+            if isinstance(directory, bytes):
+                if isinstance(directory, bytes):
```
thesamesam: Dupe.

lib/portage/_sets/files.py
```diff
@@ -159,49 +156,44 @@ def multiBuilder(self, options, settings, trees):
...
+                    directory = directory.decode("utf-8", "strict")
             # Now verify that we can also encode it.
-            _unicode_encode(directory, encoding=_encodings["fs"], errors="strict")
+            directory.encode("utf-8", "strict")
         except UnicodeError:
-            directory = _unicode_decode(
-                directory, encoding=_encodings["fs"], errors="replace"
-            )
+            if isinstance(directory, bytes):
+                if isinstance(directory, bytes):
```
thesamesam: Dupe.

lib/portage/_sets/shell.py
```diff
@@ -37,7 +36,7 @@ def load(self):
         pipe = subprocess.Popen(self._command, stdout=subprocess.PIPE, shell=True)
         stdout, stderr = pipe.communicate()
         if pipe.wait() == os.EX_OK:
-            self._setAtoms(_unicode_decode(stdout).splitlines())
+            self._setAtoms(stdout.decode("utf-8", "replace").splitlines())
```
thesamesam: Hoist into Popen again here.

lib/portage/dbapi/bintree.py
```diff
@@ -780,38 +766,36 @@ def move_ent(self, mylist, repo_match=None):
...
+                    "utf-8", "backslashreplace"
                 )
             if mynewpkg != myoldpkg:
-                ebuild_data = mydata.pop(
-                    _unicode_encode(
-                        myoldpkg + ".ebuild", encoding=_encodings["repo.content"]
-                    ),
-                    None,
+                ebuild_key = (
+                    (myoldpkg + ".ebuild")
```
thesamesam: I'd f-string here but it can be done later. It's just that I have to read the code anyway, so I may as well say it.

lib/portage/dbapi/vartree.py
```diff
@@ -6581,24 +6353,22 @@ def tar_contents(contents, root, tar, protect=None, onProgress=None, xattrs=Fals
...
         for x in contents:
-            _unicode_encode(x, encoding=_encodings["merge"], errors="strict")
+            x
     except UnicodeEncodeError:
         # The package appears to have been merged with a
         # different value of sys.getfilesystemencoding(),
         # so fall back to utf_8 if appropriate.
         try:
             for x in contents:
-                _unicode_encode(x, encoding=_encodings["fs"], errors="strict")
```
thesamesam, line 6596: Isn't this `except` path now the same as the `try`? The encoding will be equal on both attempts.

lib/portage/dep/__init__.py
```diff
@@ -1648,7 +1647,8 @@ def __init__(
 
         if not isinstance(s, str):
             # Avoid TypeError from str.__init__ with PyPy.
```
thesamesam: I suspect the comment is obsolete now but I don't blame you if you don't want to check.

lib/portage/package/ebuild/config.py
```diff
@@ -556,12 +557,7 @@ def __init__(
             if env is None:
                 env = os.environ
 
-            # Avoid potential UnicodeDecodeError exceptions later.
-            env_unicode = {
-                _unicode_decode(k): _unicode_decode(v) for k, v in env.items()
-            }
-
-            self.backupenv = env_unicode
+            self.backupenv = dict(env)
```
thesamesam: Should this be `env.copy()` or `env.deepcopy()` instead?

lib/portage/package/ebuild/digestcheck.py
```diff
@@ -108,14 +108,15 @@ def digestcheck(myfiles, mysettings, strict=False, justmanifest=None, mf=None):
 
     for parent, dirs, files in os.walk(filesdir):
         try:
-            parent = _unicode_decode(parent, encoding=_encodings["fs"], errors="strict")
+            if isinstance(parent, bytes):
+                if isinstance(parent, bytes):
```
thesamesam, line 112: Dupe.

lib/portage/package/ebuild/doebuild.py
```diff
@@ -2892,45 +2862,40 @@ def _post_src_install_uid_fix(mysettings, out):
...
+            dirs = [os.fsencode(value) for value in dirs]
+            files = [os.fsencode(value) for value in files]
             try:
-                parent = _unicode_decode(
-                    parent, encoding=_encodings["merge"], errors="strict"
-                )
+                if isinstance(parent, bytes):
+                    parent = parent.decode("utf-8", "strict")
             except UnicodeDecodeError:
-                new_parent = _unicode_decode(
```
thesamesam, line 2904: This pattern shows up in a few places and I'm not sure if it can be simplified further or not. Just commenting to say someone (maybe me, maybe not) needs to look at it (again).

lib/portage/sync/modules/git/git.py
```diff
@@ -239,12 +240,14 @@ def update(self) -> tuple[int, bool]:
                 # shallow one. And do not perform a shallow update if
                 # the target repository is not shallow.
                 is_shallow_cmd = ["git", "rev-parse", "--is-shallow-repository"]
-                is_shallow_res = portage._unicode_decode(
+                is_shallow_res = (
                     subprocess.check_output(
                         is_shallow_cmd,
-                        cwd=portage._unicode_encode(self.repo.location),
+                        cwd=self.repo.location.encode("utf-8", "backslashreplace"),
```
thesamesam: This is done repeatedly, maybe do it on the attribute itself instead.

lib/portage/sync/modules/git/git.py
```diff
@@ -239,12 +240,14 @@ def update(self) -> tuple[int, bool]:
...
                     )
-                ).rstrip("\n")
+                    .decode("utf-8", "replace")
```
thesamesam: Hoist into the check_output call (lots of these candidates).

lib/portage/tests/dbapi/test_portdb_cache.py
```diff
@@ -344,7 +345,7 @@ def testPortdbCache(self):
                     proc.stdout.close()
                     if proc.returncode != step.returncode:
                         for line in output:
-                            sys.stderr.write(portage._unicode_decode(line))
+                            sys.stderr.write(line.decode("utf-8", "replace"))
```
thesamesam: Have Popen do it?

lib/portage/tests/lint/test_bash_syntax.py
```diff
@@ -31,25 +31,16 @@ def testBashSyntax(self):
...
+                f = open(x, "rb")
+                line = f.readline().decode("utf-8", "replace")
                 f.close()
                 if line[:2] == "#!" and "bash" in line:
                     cmd = [BASH_BINARY, "-n", x]
-                    cmd = [
-                        _unicode_encode(x, encoding=_encodings["fs"], errors="strict")
-                        for x in cmd
-                    ]
+                    cmd = [x for x in cmd]
```
thesamesam, line 39: Obsolete.

lib/portage/tests/lint/test_compile_modules.py
```diff
@@ -50,16 +50,10 @@ def testCompileModules(self):
...
-                            ),
+                            x,
                             "rb",
                         ) as f:
-                            line = _unicode_decode(
-                                f.readline(),
-                                encoding=_encodings["content"],
-                                errors="replace",
-                            )
+                            line = f.readline().decode("utf-8", "replace")
```
thesamesam: Don't you want the open call to do this instead?

lib/portage/util/_desktop_entry.py
```diff
@@ -47,9 +46,9 @@ def parse_desktop_entry(path):
 def validate_desktop_entry(path):
     args = ["desktop-file-validate", path]
 
-    args = [_unicode_encode(x, errors="strict") for x in args]
+    args = [x for x in args]
```
thesamesam, line 49: Obsolete.

lib/portage/util/cpuinfo.py
```diff
@@ -1,6 +1,7 @@
 # Copyright 2015-2019 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+import os
```
thesamesam: Any reason to not just leave this lazy?

lib/portage/util/movefile.py
```diff
@@ -326,10 +321,7 @@ def movefile(
         renamefailed = False
     if not hardlinked and (selinux_enabled or sstat.st_dev == dstat.st_dev):
         try:
-            if selinux_enabled:
-                selinux.rename(src, dest)
```
thesamesam, line 330: Funny how this went unnoticed before!

lib/portage/__init__.py
```diff
@@ -177,29 +177,9 @@
     sys.stderr.write(f"    {e}\n\n")
     raise
 
-utf8_mode = sys.getfilesystemencoding() == "utf-8"
```
thesamesam, line 180: https://peps.python.org/pep-0686/ targets Python 3.15. You may have covered this elsewhere but I'm noting it now in case not: do we need to do something to enable UTF 8 mode explicitly for earlier Pythons?

lib/portage/checksum.py
```diff
@@ -189,11 +182,10 @@ def checksum_file(self, filename):
 
 # end actual hash functions
 
-
 prelink_capable = False
 if os.path.exists(PRELINK_BINARY):
     cmd = [PRELINK_BINARY, "--version"]
-    cmd = [_unicode_encode(x, encoding=_encodings["fs"], errors="strict") for x in cmd]
+    cmd = [x for x in cmd]
```
thesamesam, line 188: Obsolete.

lib/portage/checksum.py
```diff
@@ -217,9 +209,7 @@ def perform_md5(x, calc_prelink=0):
 
 
 def _perform_md5_merge(x, **kwargs):
```
thesamesam, line 211: Just kill this and inline?

lib/portage/data.py
```diff
@@ -336,13 +328,13 @@ def _init(settings):
 
         v = settings.get("PORTAGE_GRPNAME", "portage")
         if native_string:
```
thesamesam, line 330: Useless loop.

lib/portage/dispatch_conf.py
```diff
@@ -37,9 +39,9 @@ def diffstatusoutput(cmd, file1, file2):
     # raise a UnicodeDecodeError which makes the output inaccessible.
     args = shlex.split(cmd % (file1, file2))
 
-    args = (portage._unicode_encode(x, errors="strict") for x in args)
+    args = (x for x in args)
```
thesamesam, line 42: Obsolete unless you really want a tuple there?

lib/portage/localization.py
thesamesam, line 12: Your next cleanup can be killing this, if you dare. I doubt we're ever going to do translations, but if we are, there's so many missing `_()` calls anyway.

lib/portage/manifest.py
```diff
@@ -422,9 +417,9 @@ def _stat(path):
             # self.pkgdir is already included via preserved_stats.
             for parent_dir, dirs, files in os.walk(self.pkgdir.rstrip(os.sep)):
                 try:
-                    parent_dir = _unicode_decode(
-                        parent_dir, encoding=_encodings["fs"], errors="strict"
-                    )
+                    if isinstance(parent_dir, bytes):
+                        if isinstance(parent_dir, bytes):
```
thesamesam: Dupe.

lib/portage/manifest.py
```diff
@@ -601,9 +596,9 @@ def _update_thin_pkgdir(self, cat, pn, pkgdir):
 
         def _process_for_cpv(filename):
             try:
-                filename = _unicode_decode(
-                    filename, encoding=_encodings["fs"], errors="strict"
-                )
+                if isinstance(filename, bytes):
+                    if isinstance(filename, bytes):
```
thesamesam: Dupe.

lib/portage/manifest.py
```diff
@@ -646,7 +643,9 @@ def _update_thick_pkgdir(self, cat, pn, pkgdir):
         for parentdir, dirs, files in os.walk(os.path.join(pkgdir, "files")):
             for f in files:
                 try:
-                    f = _unicode_decode(f, encoding=_encodings["fs"], errors="strict")
+                    if isinstance(f, bytes):
+                        if isinstance(f, bytes):
```
thesamesam: Dupe.

lib/portage/process.py
```diff
@@ -48,10 +43,9 @@
 except ImportError:
     max_fd_limit = 256
 
-
 # Support PEP 446 for Python >=3.4
 try:
-    _set_inheritable = _os.set_inheritable
```
thesamesam, line 54: Kill this (perhaps in a followup)?

lib/portage/versions.py
```diff
@@ -404,7 +403,7 @@ def __init__(
 
         if not isinstance(cpv, str):
             # Avoid TypeError from str.__init__ with PyPy.
-            cpv = _unicode_decode(cpv)
+            cpv = cpv
```
thesamesam, line 406: Obsolete.

lib/portage/__init__.py
thesamesam: > utf8_mode was always True on modern Gentoo;

Is it? We encourage people to use a UTF-8 locale but some don't still, plus there's Prefix. I think we need to explicitly enable it.

thesamesam: See @zmedico's comment in the other other PR: https://github.com/gentoo/portage/pull/700#issuecomment-1758461071

thesamesam:
As for testing, obviously daily drive it for a bit, including with a hostile non-UTF-8 locale and merge some packages, but also see how @mjsir911 tested in https://github.com/gentoo/portage/pull/700#issue-866840142 with an evil TMPDIR.

lib/_emerge/resolver/output_helpers.py
mattst88: Done.

lib/_emerge/Binpkg.py
mattst88: Done.

lib/_emerge/JobStatusDisplay.py
mattst88: Removed.

lib/portage/_sets/files.py
mattst88: Fixed.

lib/portage/_sets/files.py
mattst88: Fixed.

lib/portage/package/ebuild/digestcheck.py
mattst88: Fixed.

lib/portage/tests/lint/test_bash_syntax.py
mattst88: Removed.

lib/portage/util/_desktop_entry.py
mattst88: Removed.

lib/portage/checksum.py
mattst88: Removed.

lib/portage/data.py
mattst88: Fixed.

lib/portage/manifest.py
mattst88: Fixed.

lib/portage/manifest.py
mattst88: Fixed.

lib/portage/xpak.py
mattst88: Fixed.

lib/portage/xpak.py
mattst88: Fixed.

lib/portage/checksum.py
mattst88: Removed.

lib/portage/versions.py
mattst88: Removed.

lib/portage/manifest.py
mattst88: Fixed.

lib/portage/dispatch_conf.py
mattst88: Removed.

lib/portage/util/movefile.py
mattst88: Looks like we found a counter example to https://en.wikipedia.org/wiki/Fundamental_theorem_of_software_engineering

lib/portage/manifest.py
mattst88: Fixed.

bin/dispatch-conf
mattst88: Good call. Fixed.

lib/_emerge/actions.py
mattst88: Fixed.

lib/_emerge/Binpkg.py
mattst88: Fixed.

lib/portage/_sets/shell.py
mattst88: Fixed.

lib/portage/dbapi/bintree.py
mattst88: Done.

lib/portage/dbapi/vartree.py
mattst88: Indeed, good catch. Fixed.

lib/portage/package/ebuild/config.py
mattst88: yeah, `env.copy()` is better. deepcopy is unnecessary since the dict just contains strings. Fixed.

lib/portage/sync/modules/git/git.py
mattst88: Fixed.

lib/portage/tests/dbapi/test_portdb_cache.py
mattst88: Done.

lib/portage/tests/lint/test_compile_modules.py
mattst88: Yes, done.

lib/portage/util/cpuinfo.py
mattst88: Fixed.

lib/portage/sync/modules/git/git.py
mattst88: Fixed.

lib/portage/dep/__init__.py
mattst88: 99.99% sure it is. Removed.

lib/portage/package/ebuild/doebuild.py
mattst88: I've added a commit "doebuild: remove dead isinstance checks in unicode path rename" (that I'll squash into an earlier change if you agree). How about that? Same pattern as we're using in vartree.py.

lib/portage/__init__.py
mattst88: Oh, I think you might be right to be skeptical here.

I think we can rely on https://peps.python.org/pep-0540/ which adds the `PYTHONUTF8=1` environment variable and the `-X utf8` flag.

I'll look into this.

mattst88: I've added `bin: re-exec with PYTHONUTF8=1 if not already in UTF-8 mode` which I think should take care of this.

floppym: External consumers of the Portage API (like gentoolkit) would also need to be updated I think.

Maybe it make sense to emit a warning (or error out) from the `portage` module if it is imported with utf8_mode disabled.

lib/portage/process.py
```diff
@@ -48,10 +43,9 @@
...
+    _set_inheritable = os.set_inheritable
```
thesamesam, line 48: Did you accidentally roll this back, or is github showing me an old copy somehow? 

Anyway, I think we can drop `_set_inheritable` and use `os.set_inheritable` directly.

mattst88: I dropped that from the series. I'll make a separate PR for that and some other old-Python cleanups.

thesamesam: wfm, thx

lib/portage/package/ebuild/doebuild.py
thesamesam: Looks good, thanks. Though there's a lot of instances of this that we'd need to switch too, even in vartree.py.

lib/portage/process.py
mattst88: Will do in a follow up.

lib/portage/__init__.py
mattst88: That's a good idea. I've added a commit that does that: `portage: warn if imported without a UTF-8 locale`.

bin/emerge
```diff
@@ -2,9 +2,16 @@
 # Copyright 2006-2025 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-import signal
+import os
 import sys
 
+if not sys.flags.utf8_mode:
```
floppym: Note that this flag is NOT set when Python is executed with a UTF-8 locale set via LANG, LC_CTYPE, or LC_ALL.

That effectively means we will trigger this re-exec every time someone with a properly configured locale runs emerge.

We could avoid that by:
1. Using an appropriate shebang.
`#!/usr/bin/env -S python3 -X utf-8`
Note that this relies on the GNU-specific `env -S` option.
2. Not relying on `sys.flags.utf8_mode` and instead checking the various preferred encodings.
https://docs.python.org/3/library/os.html#utf8-mode

bin/emerge
```diff
@@ -2,9 +2,16 @@
...
+    os.environ["PYTHONUTF8"] = "1"
+    os.execvpe(sys.executable, [sys.executable] + sys.argv, os.environ)
```
floppym: This could be simplified slightly. We don't need to search PATH or override the environment.
```suggestion
    os.execv(sys.executable, [sys.executable] + sys.argv)
```

bin/emerge
mattst88: Good point. Applying this change to all the scripts in `bin/`:

```diff
+import locale
 import os
 import sys
 
-if not sys.flags.utf8_mode:
+if sys.getfilesystemencoding() != "utf-8" or locale.getpreferredencoding(False) != "utf-8":
     os.environ["PYTHONUTF8"] = "1"
-    os.execvpe(sys.executable, [sys.executable] + sys.argv, os.environ)
+    os.execv(sys.executable, [sys.executable] + sys.argv)
```

>  `#!/usr/bin/env -S python3 -X utf-8`

I don't think we can do this actually because it doesn't propagate to subprocesses, whereas setting `PYTHONUTF8=1` does.

bin/emerge
mattst88: Fixed, thanks!

lib/portage/package/ebuild/doebuild.py
mattst88: Handled those as well!

bin/emerge
floppym: > I don't think we can do this actually because it doesn't propagate to subprocesses

Ah that raises another question: do we leak the PYTHONUTF8 envvar into the ebuild environment? That could potentially lead to some weird behavior from Python packages that don't expect it.

I suppose we can cross that bridge when we come to it though.

mattst88: Thinking about this more...

PMS says

> The package manager must ensure that the `LC_CTYPE` and `LC_COLLATE` locale categories are equivalent to the POSIX locale, as far as characters in the ASCII range (U+0000 to U+007F) are concerned. Only for EAPIs listed in such a manner in table [11.6](https://projects.gentoo.org/pms/latest/pms.html#x1-111040x6).

where table 11.6 is EAPI >= 6 → true.

Portage implements this in `lib/_emerge/EbuildPhase.py` by trying (in order) `C.UTF-8`, `en_US.UTF-8`, `en_GB.UTF-8`. If  none of those are available, it tries the non-UTF-8 `C` locale. In that case, leaking `PYTHONUTF8=1` into the ebuild environment would potentially change behavior.

Seems bad to me to try to get a UTF-8 locale before falling back to non-UTF-8 when UTF-8 isn't even required to begin with. I wouldn't want to have to debug something caused by this.

This seems really edge-casey, and I kind of don't want to think about it too deeply, so I'm inclined to add it to `environ_filter` and forget about it :)

(Maybe a future EAPI should require a UTF-8  locale?)

mattst88: I've added it to `environ_filter`.

floppym: Sounds good to me.

mjsir911:
yo this is crazy, good job

PR #1604, opened 2026-06-27
Fix missed Atom str-subtyping call sites

mattst88:
Three call sites missed by the 7e15fc7fa ("dep: Atom: remove str subtyping") audit:

- actions: Atom.startswith() in world_candidates filtering (bug 978088)
- create_world_atom: Atom += str when recording slotted atom to world
- depgraph: Atom.startswith() in saveNomergeFavorites

Regression tests for the first two bugs. (Third would require actual emerging, I think)

Bug: https://bugs.gentoo.org/978088

### DIFF
--- a/lib/_emerge/actions.py
+++ b/lib/_emerge/actions.py
@@ -504,7 +504,9 @@ def action_build(
                         x
                         for x in favorites
                         if not (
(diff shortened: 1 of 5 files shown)


thesamesam (approved):
ok once black is happy (flow's PR landed earlier)

PR #1611, opened 2026-06-30
util: listdir: drop invalid follow_symlinks kwarg from is_symlink()

Flowdalic:
Python's is_symlink() method takes no arguments, unlike is_dir() and is_file(). Passing the 'follow_symlinks' keyword argument results in a TypeError.

This bug surfaces during package tree traversals, notably crashing "eclean distfiles" with:

  TypeError: is_symlink() takes no arguments
      elif entry.is_symlink(follow_symlinks=followSymlinks):

Fix this by removing the invalid argument.

Fixes: ef82d7e97385c7d7f377c0adf6f9cec21f0856bb


thesamesam:
Thanks.

PR #1614, opened 2026-07-01
Remove dead Python <3.9 compatibility guards

mattst88:
Since portage requires Python >=3.9, several compatibility shims for
older Python versions are unreachable dead code. Remove them.

- process: `os.set_inheritable` and `fcntl.FD_CLOEXEC` are guaranteed present since 3.4; drop the `try`/`except` guards and `None`-checks
- `array.tostring()` was removed in 3.9; drop fallbacks from `tobytes()` calls in xpak, AbstractPollTask, xattr-helper, and tests
- socks5-server: `asyncio.ensure_future` (3.4.4) and `asyncio.current_task` (3.7) are always available; remove the compat aliases
- mtimedb: `Unpickler.find_global` is Python 2 only; the assignment always raised `AttributeError` in Python 3. Remove the dead `try`/`except`
- configparser: `ConfigParser.readfp()` was removed in 3.12; `read_file()` has been the right call since 3.2

### DIFF
--- a/lib/portage/process.py
+++ b/lib/portage/process.py
@@ -43,17 +43,6 @@ try:
 except ImportError:
     max_fd_limit = 256
 
-# Support PEP 446 for Python >=3.4
-try:
-    _set_inheritable = os.set_inheritable
-except AttributeError:
-    _set_inheritable = None
-
-try:
-    _FD_CLOEXEC = fcntl.FD_CLOEXEC
-except AttributeError:
-    _FD_CLOEXEC = None
-
 # Prefer /proc/self/fd if available (/dev/fd
 # doesn't work on solaris, see bug #474536).
 for _fd_dir in ("/proc/self/fd", "/dev/fd"):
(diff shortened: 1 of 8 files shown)


lib/portage/process.py
```diff
@@ -43,16 +43,8 @@
...
-    _set_inheritable = os.set_inheritable
-except AttributeError:
-    _set_inheritable = None
-
-try:
-    _FD_CLOEXEC = fcntl.FD_CLOEXEC
-except AttributeError:
-    _FD_CLOEXEC = None
+_set_inheritable = os.set_inheritable
+_FD_CLOEXEC = fcntl.FD_CLOEXEC
```
thesamesam: Just inline these instead?

mattst88: Oh yeah, no reason to keep the indirection. Thanks!

thesamesam (approved):

PR #1615, opened 2026-07-02
porttree: hoist pkgsplit import out of findname2

mattst88:
findname2 is on the resolver hot path (called once per candidate cpv, tens of thousands of times per `@world` resolve) and did 'from portage.versions import pkgsplit' on every call. The import is cheap not free at that call volume. Import pkgsplit at module scope instead; there is no circular-import risk (portage.versions does not import porttree).

Also drop the now-redundant in-function import in the deprecated getname().


thesamesam (approved):
Look forward to real lazy imports w/ 3.15..

zmedico (approved):

PR #1616, opened 2026-07-02
vartree: skip preserve-libs linkmap rebuild for fileless packages

mattst88:
With FEATURES=preserve-libs, every merge and unmerge runs _linkmap_rebuild(), a full-system scan of every installed package's NEEDED.ELF.2. A package that owns no files (empty/virtual packages) provides no libraries to preserve and removes none, so it cannot orphan any preserved library's consumers; the rebuild and the preserve/prune scans it feeds are guaranteed no-ops.

Gate them on self.getcontents() being non-empty at the two per-package sites that run on every merge: the treewalk merge-side rebuild (which only feeds _find_libs_to_preserve() on the replaced instance) and _prune_plib_registry(). unregister()/store() still run so empty registry entries are cleaned up. This also skips the rebuild for fresh installs and when preserve-libs is disabled, where its result was already unused.

Merging an empty virtual on a warm 1851-pkg system: NEEDED.ELF.2 opens
half as many: 3707 -> 1854.

### DIFF
--- a/lib/portage/dbapi/vartree.py
+++ b/lib/portage/dbapi/vartree.py
@@ -2205,25 +2205,36 @@ class dblink:
             try:
                 plib_registry.load()
 
-                unmerge_with_replacement = unmerge and preserve_paths is not None
-                if unmerge_with_replacement:
-                    # If self.mycpv is about to be unmerged and we
-                    # have a replacement package, we want to exclude
-                    # the irrelevant NEEDED data that belongs to
-                    # files which are being unmerged now.
-                    exclude_pkgs = (self.mycpv,)
-                else:
-                    exclude_pkgs = None
+                # An instance that owns no files (e.g. empty/virtual
+                # packages) has no libraries of its own to preserve and
+                # removes no files, so it cannot drop the last consumer of an
+                # already-preserved library and leave that library orphaned.
+                # The expensive system-wide linkmap rebuild and the
+                # preserve/prune scans below are therefore guaranteed no-ops.
+                # Registry unregistration on unmerge is still performed so any
+                # (empty) entry is cleaned up.
+                instance_owns_files = bool(self.getcontents())
 
-                self._linkmap_rebuild(
-                    exclude_pkgs=exclude_pkgs,
-                    include_file=needed,
-                    preserve_paths=preserve_paths,
-                )
+                unmerge_with_replacement = unmerge and preserve_paths is not None
+                if instance_owns_files:
+                    if unmerge_with_replacement:
+                        # If self.mycpv is about to be unmerged and we
+                        # have a replacement package, we want to exclude
+                        # the irrelevant NEEDED data that belongs to
+                        # files which are being unmerged now.
+                        exclude_pkgs = (self.mycpv,)
+                    else:
+                        exclude_pkgs = None
+
+                    self._linkmap_rebuild(
+                        exclude_pkgs=exclude_pkgs,
+                        include_file=needed,
+                        preserve_paths=preserve_paths,
+                    )
 
                 if unmerge:
                     unmerge_preserve = None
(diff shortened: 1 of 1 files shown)


thesamesam:
While I review this: are you sure it's still needed with 98d0362c13903c9d8d04d01d4559568c96379250?

thesamesam:
> While I review this: are you sure it's still needed with https://github.com/gentoo/portage/commit/98d0362c13903c9d8d04d01d4559568c96379250?

My change only inhibits calling `_prune_plib_registry`, not `treetalk`.

A nice thing about `98d0362c13903c9d8d04d01d4559568c96379250` is we avoid taking a lock. Now, for empty contents in the to-be-installed image, it's harmless in the correctness sense to punt on `_prune_plib_registry`, but I was worried it may leave junk in there in the case that we previously installed files and now we don't, if one of those previously-installed files was holding onto a preserved lib. However, I think it's fine because it still gets called on unmerge, just not merge.

Should we revert part of my change given it's unnecessary and yours is better reasoned (and I think easier to follow)?

lib/portage/dbapi/vartree.py
```diff
@@ -2205,25 +2205,35 @@ def _prune_plib_registry(self, unmerge=False, needed=None, preserve_paths=None):
             try:
                 plib_registry.load()
 
+                # An instance that owns no files (e.g. empty/virtual
+                # packages) can neither preserve its own libraries nor
+                # orphan the consumers of any already-preserved library, so
```
thesamesam: By "orphan the consumers of ...", do you mean "we cannot be the last consumer of it, leaving it now-orphaned", or am I misunderstanding? I haven't ever looked at preserve-libs in detail so I might be (obviously I get the principle but I don't know what terms we use internally).

I think what you said is fine but for some reason I couldn't parse it easily (like it would make the consumers orphans somehow).

mattst88:
> Should we revert part of my change given it's unnecessary and yours is better reasoned (and I think easier to follow)?

I don't think it's unnecessary. Your commit does something mine intentionally doesn't.

98d0362 gates the post-merge call to _prune_plib_registry() (and _post_merge_sync()) on if contents:. My change gates the treewalk merge-side rebuild and the scans inside _prune_plib_registry() (on the replaced instance's getcontents(), which serves the unmerge path). So at the post-merge site, your if contents: is what fires. My internal gate isn't even reached there.

The important non-overlap: your change skips the whole call, so it avoids _fs_lock() + the plib-registry lock + load()/store() for a fileless merge. Mine intentionally still takes those and runs unregister()/store() (to clean up any empty registry entry), so it does not give you the lock avoidance. Your change also gates _post_merge_sync(), which is unrelated to preserve-libs and should definitely stay. Reverting would reintroduce both the lock and the post-merge sync on empty installs.

On your correctness worry -- file-owning instance replaced by a now-fileless one, where an old file held a preserved I convinced myself it's fine, same as you: the pruning still happens, just via the replaced instance's unmerge, not the merge. Concretely, during the replacement the old (file-owning) instance is unmerged, and that unmerge calls _prune_plib_registry(unmerge=True) on it; in my gating instance_owns_files is keyed on that instance's getcontents(), which is non-empty, so the scan runs and any now-unused preserved lib is cleaned. The merge-side rebuild is likewise keyed on self._installed_instance (the old, file-owning one), so preservation is still evaluated. Nothing is left behind.

So I'd keep both: yours = "don't take the lock / sync / prune post-merge when we installed nothing"; mine = "don't do the linkmap rebuild at the sites that still run." They dovetail nicely :)

zmedico (approved):

thesamesam (commented):
Thanks, that makes sense.

OK with comment tweak (see above) and maybe commit message tweaked to mention the interaction we just discussed.

PR #1617, opened 2026-07-02
EbuildExecuter: skip no-op src phases for sourceless packages

mattst88:
A package with no source (empty SRC_URI) that defines none of its own src_* phases runs only no-op default bodies for unpack/prepare/configure/ compile/test. EbuildExecuter forked an ebuild.sh subprocess for each, re-sourcing ${T}/environment every time -- a lot of unnecessary work for empty/virtual packages.

Add _can_skip_source_phases(): when SRC_URI is empty and none of those phases are defined, skip straight from pkg_setup to src_install (which creates ${D} and runs install QA). Disabled for FEATURES=noauto and PROPERTIES=live. eapply_user is skipped along with src_prepare, so a virtual's /etc/portage/patches would now be silently ignored.

Merging empty virtuals (warm 1851-pkg vdb, --jobs 1): phase spawns 14 -> 9, wall ~7.2s -> ~6.0s.

### DIFF
--- a/lib/_emerge/EbuildExecuter.py
+++ b/lib/_emerge/EbuildExecuter.py
@@ -18,6 +18,50 @@ class EbuildExecuter(CompositeTask):
 
     _phases = ("prepare", "configure", "compile", "test", "install")
 
+    # src_* phases whose default (EAPI-provided) implementations are no-ops
+    # for a package with no source to act on. src_install is intentionally
+    # excluded: it creates ${D} and drives the install-time QA checks, so it
+    # always runs. Excluding src_install also sidesteps declarative variables
+    # such as DOCS, which can reference FILESDIR (present even for sourceless
+    # packages) and thus may do real work in the default src_install body.
+    _skippable_src_phases = frozenset(
+        ("unpack", "prepare", "configure", "compile", "test")
+    )
+
+    def _can_skip_source_phases(self):
+        """
+        Return True if unpack and the src_prepare/configure/compile/test
+        phases would only run no-op default bodies and can be skipped
+        entirely (avoiding one ebuild.sh subprocess spawn per phase).
+
+        This holds when the package has no source to unpack (empty SRC_URI,
+        hence empty ${A}, hence nothing for the default phase bodies to
+        configure/compile/install) and defines none of those phases itself.
+        Empty/virtual packages (DEFINED_PHASES=-) are the motivating case.
+        """
+        settings = self.settings
+        # The ebuild command's manual mode relies on running each phase.
+        if "noauto" in settings.features:
+            return False
+        # Live ebuilds fetch their source in src_unpack regardless of
+        # SRC_URI, so never skip unpack for them.
+        if "live" in settings.get("PROPERTIES", "").split():
+            return False
+
+        metadata = self.pkg._metadata
+        if metadata.get("SRC_URI", "").strip():
+            return False
+        defined_phases = frozenset(metadata.get("DEFINED_PHASES", "").split())
+        if defined_phases.intersection(self._skippable_src_phases):
+            return False
+
+        # pkg_setup (which has run by now) records whether PATCHES is set;
+        # the default src_prepare would apply it, so don't skip in that case.
+        if os.path.exists(os.path.join(settings["PORTAGE_BUILDDIR"], ".src_patches")):
+            return False
+
+        return True
+
     def _start(self):
         pkg = self.pkg
         scheduler = self.scheduler
@@ -50,6 +94,16 @@ class EbuildExecuter(CompositeTask):
             self.wait()
             return
 
+        if self._can_skip_source_phases():
+            # No source and no ebuild-defined src_* phases: unpack and
+            # src_prepare/configure/compile/test would only run no-op
+            # default bodies. Skip straight to src_install (which creates
+            # ${D} and runs the install QA checks). WORKDIR was already
+            # created by prepare_build_dirs(), so src_install can cd into
+            # it as usual.
+            self._start_phases(("install",))
+            return
+
         unpack_phase = EbuildPhase(
             background=self.background,
             phase="unpack",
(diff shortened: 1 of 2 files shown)


lib/_emerge/EbuildExecuter.py
```diff
@@ -18,6 +18,40 @@ class EbuildExecuter(CompositeTask):
 
     _phases = ("prepare", "configure", "compile", "test", "install")
 
+    # src_* phases whose default (EAPI-provided) implementations are no-ops
+    # for a package with no source to act on. src_install is intentionally
+    # excluded: it creates ${D} and drives the install-time QA checks, so it
```
thesamesam, line 23: An important point is also that it has some interaction with declarative variables, so we have `DOCS` which could I think point to `FILESDIR` (even if bizarre). By punting on `src_install`, you avoid that problem.

(Consider putting a summary of this in a comment or the commit message, as I think it's important to be very clear on why this is safe.)

lib/_emerge/EbuildExecuter.py
```diff
@@ -18,6 +18,40 @@ class EbuildExecuter(CompositeTask):
...
+        configure/compile/install) and defines none of those phases itself.
+        Empty/virtual packages (DEFINED_PHASES=-) are the motivating case.
+        """
+        settings = self.settings
+        # The ebuild command's manual mode relies on running each phase.
+        if "noauto" in settings.features:
+            return False
+        # Live ebuilds fetch their source in src_unpack regardless of
+        # SRC_URI, so never skip unpack for them.
+        if "live" in settings.get("PROPERTIES", "").split():
```
thesamesam, line 48: I think we should punt on any eclasses being inherited as well to be safe. I can't see why you would inherit any eclasses in an ebuild where we're doing nothing.

That said, I can't (yet) think of an example where it would pose a problem, but still.

thesamesam (commented):
Need to think about this more, also cc @eli-schwartz who has proposed this many times.

thesamesam:
`Bug: https://bugs.gentoo.org/937150`

lib/_emerge/EbuildExecuter.py
mattst88: That's actually something I wondered about as well. When I was surveying the virtual/ ebuilds the only ones I noticed with defined `src_*` functions were ruby virtuals that inherit ruby-ng. E.g.: `ruby-ssl/ruby-ssl-15.ebuild` contains

```
inherit ruby-ng

[...]

pkg_setup() { :; }
src_unpack() { :; }
src_prepare() { eapply_user; }
src_compile() { :; }
src_install() { :; }
pkg_preinst() { :; }
pkg_postinst() { :; }
```

So it seems like that is safe in practice, but... ugh. Unsure.

The other case is virtuals inheriting `multilib-build` just to get `ABI_*`.

thesamesam: Ah, yeah.. using it for dependencies makes sense. Let me mull it over more but it may be okay to leave it as-is then.

eli-schwartz: We can't really tell whether eclass phase definitions are "doing anything" without a full bash parser. So if an eclass defines it even if the body is `:`, we should handle this by breaking the optimization, and simply running all phases. How common is this case anyway?

The whole idea of optimizing this away, as I see things, is by applying a kind of "as-if" rule to PMS requirements. If we cannot actually guarantee beyond all shadow of a doubt that the optimization transform is correct, then we should be pursuing changes to PMS rather than making assumptions in portage.

Inheriting eclasses, as long as those eclasses don't export a phase or set SRC_URI, should be fine I guess.

lib/_emerge/EbuildExecuter.py
```diff
@@ -18,6 +18,40 @@ class EbuildExecuter(CompositeTask):
...
+            return False
+
+        metadata = self.pkg._metadata
+        if metadata.get("SRC_URI", "").strip():
```
eli-schwartz, line 52: Should also check ${PATCHES}. (Sorry! :P)

> eapply_user is skipped along with src_prepare, so a virtual's /etc/portage/patches would now be silently ignored.

PMS permits the PM to implement eapply_user as a no-op, if it doesn't wish to support user patches. So ignoring /etc/portage/patches is fine.

eli-schwartz (changes requested):

lib/_emerge/EbuildExecuter.py
mattst88: Actually, I think this is already covered. `DEFINED_PHASES` is built in ebuild.sh via `declare -F` over each phase function, so an eclass that exports `src_prepare`, even `src_prepare() { :; }`, lands prepare in `DEFINED_PHASES`, and the gate then runs all phases as normal. It's not inferring whether the body does anything; it's portage's own record of which phase functions are defined (the same thing `__has_phase_defined` etc. rely on). So "eclass defines a phase -> break the optimization" is exactly what happens, no bash parser required. The only thing an inherited eclass has to avoid to stay eligible is exporting a phase or setting SRC_URI, which matches what you said.

eli-schwartz: Yes, I'd assume so as well since inheriting an eclass is essentially just a framework for sourcing stuff and defining functions. :P

The tricky thing here is, as below, PATCHES (which is bash, not a metadata variable).

lib/_emerge/EbuildExecuter.py
mattst88: Done in the latest push -- I think there's no requirement that `PATCHES` is set at metadata generation time (but in practice always is). I think it would probably be good to tighten up that requirement in the spec in the future.

eli-schwartz: It's either defined at global scope, or defined at local scope immediately before running default_src_prepare.

In the former case it is available whenever the ebuild is sourced at all.

In the latter case someone has defined a custom `src_prepare()` just so they can use `local -x PATCHES=(); default` and confuse people. Obviously this implies that we're already running phases because there is a custom `src_*`, so....

mattst88: Right, and those two cases line up with the two checks: global-scope PATCHES is caught by the pkg_setup marker, and the local-in-custom-src_prepare case is caught earlier by DEFINED_PHASES (a custom src_prepare means we're not skipping anyway). So both are covered. Thanks for thinking through it with me.

thesamesam (commented):
I think this looks OK to me now once https://github.com/gentoo/portage/pull/1617#discussion_r3516421768 is addressed.

(I both want that in there for the record to show we considered it, but also to allow grepping for e.g. declarative in future when we add more things to the spec.)

lib/_emerge/EbuildExecuter.py
mattst88: Very good point. Thank you!

thesamesam (approved):

PR #1618, opened 2026-07-03
install-qa-check.d/60pkgconfig: relax prefix match for libdir check

parona-source:
Otherwise ${exec_prefix}/lib can pass undetected.

Bug: https://bugs.gentoo.org/978404

### DIFF
--- a/bin/install-qa-check.d/60pkgconfig
+++ b/bin/install-qa-check.d/60pkgconfig
@@ -85,7 +85,7 @@ pkgconfig_check() {
 		elif [[ ${f} == *lib64/pkgconfig* ]] ; then
 			# We want to match /lib/, /lib/foo/, but not e.g. /lib64 or /lib64/, or libfoo
 			# Only check lines with a known prefix in order to avoid false positives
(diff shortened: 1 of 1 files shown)


bin/install-qa-check.d/60pkgconfig
```diff
@@ -85,7 +85,7 @@ pkgconfig_check() {
 		elif [[ ${f} == *lib64/pkgconfig* ]] ; then
 			# We want to match /lib/, /lib/foo/, but not e.g. /lib64 or /lib64/, or libfoo
 			# Only check lines with a known prefix in order to avoid false positives
```
floppym, line 87: Sorry for the late review, but this change directly contradicts this comment.

parona-source: Checking the prefix as well would be doable with.

`[[ $(pkg-config --maximum-traverse-depth=1 --with-path=${ED}/usr/lib64/pkgconfig --variable=libdir ${f}) =~ ^${EPREFIX}/usr/lib(/|$) ]]`

floppym: Ah, I think I misinterpreted "known prefix". That's referring to `^libdir=` and `^Libs:`, not to the path itself.

PR #1619, opened 2026-07-03
checksum: read files once and hash large files in parallel

mattst88:
Portage's multi-hash checksum paths reread the file once per hash. With the default BLAKE2B+SHA512 digest that means every fetch verify, Manifest regen, and binpkg index hash streams each file twice, and `perform_all` reads it once per supported hash.

This series makes those paths read the file a single time:

- `perform_multiple_checksums` / `perform_all` now stream the file once, updating every requested hash per block (`_checksum_file_serial`). The size pseudo-hash comes from one stat. Output is unchanged for all callers.
- `verify_all` computes all hashes in one pass instead of calling `perform_checksum` per hash. Behavior change: it no longer short-circuits at the first hash type, but the reported failure reason (first mismatch in sorted order) is unchanged.
- Large files (≥1 MiB, more than one hash) additionally hash in parallel: `hashlib` releases the GIL during `update()`, so a single reader thread feeds one worker thread per hash. Smaller files and single-hash requests keep the serial path.
- OSError toportage-exception mapping is factored out into a shared `_raise_checksum_oserror` helper.

Performance (270 MiB distfile, default BLAKE2B+SHA512):
- Local/tmpfs (CPU-bound): 1.9x (716 ms → 373 ms).
- Slow NFS (I/O-bound, ~3 MiB/s): parallel is ~1x, but the single-read change alone halves wire traffic (~200 s → ~100 s).

Tests: adds coverage for the previously untested multi-hash, parallel, `verify_all`, and `perform_all` paths. Known answer digests plus a serial↔parallel parity check on an over-threshold file.

### DIFF
--- a/lib/portage/tests/util/test_checksum.py
+++ b/lib/portage/tests/util/test_checksum.py
@@ -1,10 +1,23 @@
 # Copyright 2011-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+import os
+import tempfile
+
 from portage.tests import TestCase
 
-from portage.checksum import checksum_str, _apply_hash_filter
-from portage.exception import DigestException
+from portage.checksum import (
+    checksum_str,
+    perform_all,
+    perform_multiple_checksums,
+    verify_all,
+    hashfunc_keys,
+    _apply_hash_filter,
+    _checksum_file_parallel,
+    _checksum_file_serial,
+    _CHECKSUM_PARALLEL_MIN_SIZE,
+)
+from portage.exception import DigestException, FileNotFound
 
 
 class ChecksumTestCase(TestCase):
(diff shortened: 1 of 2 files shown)


lib/portage/tests/util/test_checksum.py
```diff
@@ -122,6 +135,129 @@ def test_sha3_512(self):
...
+
+    def _mydict(self, **overrides):
+        mydict = dict(self.digests)
+        mydict["size"] = len(self.text)
+        mydict.update(overrides)
+        return mydict
+
+    def test_pass(self):
+        path = _write_tempfile(self, self.text)
+        ok, reason = verify_all(path, self._mydict())
```
thesamesam: Nit: `reason` is unused:

```suggestion
        ok, _ = verify_all(path, self._mydict())
```

But then just directly `self.assertTrue` it?

lib/portage/tests/util/test_checksum.py
```diff
@@ -122,6 +135,129 @@ def test_sha3_512(self):
...
+        self.assertTrue(ok)
+
+    def test_checksum_mismatch(self):
+        path = _write_tempfile(self, self.text)
+        ok, reason = verify_all(path, self._mydict(MD5="0" * 32))
+        self.assertFalse(ok)
+        self.assertIn("MD5", reason[0])
+
+    def test_size_mismatch(self):
+        path = _write_tempfile(self, self.text)
```
thesamesam, line 238: I think you could use a fixture for this but it's probably not worth it.

lib/portage/tests/util/test_checksum.py
```diff
@@ -122,6 +135,129 @@ def test_sha3_512(self):
...
+        self.assertRaises(
+            DigestException, perform_multiple_checksums, path, hashes=["BOGUS"]
+        )
+
+    def test_parallel_matches_serial(self):
+        # A file over the parallel threshold with more than one hash takes the
+        # thread-per-hash path; it must agree with the serial single-pass path.
+        content = bytes(range(256)) * ((2 * _CHECKSUM_PARALLEL_MIN_SIZE) // 256)
+        path = _write_tempfile(self, content)
+        hashes = ["MD5", "SHA256", "SHA512"]
```
thesamesam: DRY: `text_digests.keys()`

lib/portage/tests/util/test_checksum.py
```diff
@@ -122,6 +135,129 @@ def test_sha3_512(self):
...
+    def test_single_hash(self):
+        path = _write_tempfile(self, self.text)
+        result = perform_multiple_checksums(path, hashes=["MD5"])
+        self.assertEqual(result, {"MD5": self.text_digests["MD5"]})
+        # Digests are hex strings, not (hash, size) tuples.
+        self.assertIsInstance(result["MD5"], str)
+
+    def test_multiple_hashes(self):
+        path = _write_tempfile(self, self.text)
+        result = perform_multiple_checksums(path, hashes=["MD5", "SHA1", "SHA256"])
```
thesamesam: text_digests.keys()

mattst88: Done.

lib/portage/tests/util/test_checksum.py
mattst88: Done.

lib/portage/tests/util/test_checksum.py
mattst88: Done.

thesamesam (approved):

PR #1621, opened 2026-07-04
emirrordist: fix dropping untried primary URIs

peeweep:
When --tries is less than the number of primary URIs (e.g. emirrordist --tries=1 with kernel-2.eclass's GENPATCHES_URI), this can cause unexpected fetch failures. In this case, when remaining_tries drops to 0, _next_uri() will return None even if untried URIs remain in _primaryuri_stack.


thesamesam:
Thanks!

PR #1622, opened 2026-07-04
special_env_vars.py: add QEMU_* to environ_whitelist_re

peppergrayxyz:
When running emerge in a cross architecture chroot via qemu-static
with extensions that are not enabled by default it fails with error:
```
Sandboxed process killed by signal: Illegal instruction
 * The ebuild phase 'unpack' has been killed by signal 4.
```
This is caused by `qemu-static` being invoked without the `QEMU_*` vars.
They have already been stripped by environ, but not yet being added
back in `_grab_pkg_env`. Thus `qemu-static` falls back to its default
settings for the target architecture and doesn't load the
needed cpu extensions.

Therefore `sandbox` (`environ`) and `bash` (`__filter_readonly_variables`)
fail, when execting in a cross arch chroot.

For qemu-static to work properly:
- allow `QEMU_*` vars through the environ filter
- add `QEMU_*` vars to the readonly vars filter environment

Closes: https://bugs.gentoo.org/978685

### DIFF
--- a/bin/phase-functions.sh
+++ b/bin/phase-functions.sh
@@ -86,14 +86,19 @@ portage_mutable_filtered_vars=( AA HOSTNAME )
 # is to preserve various variables as they were at the time that the binary
 # package was built while protecting against the application of package renames.
 __filter_readonly_variables() {
-	local -a filtered_vars bash_vars
-	local IFS
+	local -a filtered_vars bash_vars qemu_env
+	local IFS x
+
+	# preserve QEMU vars for qemu-static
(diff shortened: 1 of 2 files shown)


peppergrayxyz:
Do you want to add tests for qemu-static or is this too specific? I'd zip the files from the x86 CI, put them into an arm runner and rerun them with qemu-static as binfmt_misc handler.

zmedico (approved):

thesamesam:
Nah, no need for tests for this. Thanks.

PR #1623, opened 2026-07-04
dispatch-conf: fix AttributeError in massage() on str from newconfigs

mattst88:
path_list is decoded to str before splitlines(), so nconf elements are already str. The .decode() call added in commit 30bbecfbc is wrong; drop it.

Fixes: 30bbecfbc ("Remove Python 2 unicode compatibility layer")


thesamesam (approved):
Sure.

PR #1628, opened 2026-07-08
Fix two regressions from skip-no-op-src-phases

mattst88:
Two bugs introduced by the sourceless-package phase-skip work:

- EbuildPhase: missing FILESDIR symlink: When `_can_skip_source_phases()` bypasses unpack, `_prepare_fake_filesdir` never runs. The `${PORTAGE_BUILDDIR}/files` → `${O}/files` symlink is absent at install time, breaking any ebuild that references `${FILESDIR}` in `src_install` (e.g. `newins "${FILESDIR}/...`"). Fixed by calling `_prepare_fake_filesdir` at the top of the install phase when the symlink is absent.
- phase-functions.sh: spurious cp failure during binary merge: move the build-info creation block to the outer src-phase case so it covers all src phases including clean. For binary merges `Binpkg.__init__` sets `EBUILD` to the not-yet-extracted `build-info/${PF}.ebuild`; when the clean phase fires before `unpack_metadata()` the cp fails with "cannot stat". Fixed by guarding the cp with `[[ -e ${EBUILD} ]]`.

### DIFF
--- a/bin/phase-functions.sh
+++ b/bin/phase-functions.sh
@@ -1088,12 +1088,14 @@ __ebuild_main() {
 		# reachable when earlier no-op src phases are skipped. Guard on
 		# PORTAGE_BUILDDIR: the clean phase can run before prepare_build_dirs()
 		# (and tolerates a missing builddir), so an unguarded cd would create
-		# a stray build-info/ in the cwd.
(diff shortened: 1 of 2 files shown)


thesamesam:
CI should pass on rebase.

thesamesam (approved):

PR #1629, opened 2026-07-11
isolated-functions.sh: Don't call sed in die()

ulm:
Use pure Bash to extract the failing code snippet.

Restoration of PATH and unsetting of QA_INTERCEPTORS should no longer be needed, since the function doesn't invoke any external commands.


thesamesam (approved):
Thanks!

PR #1632, opened 2026-07-22
resolver: compute cycle-breaking closures in one SCC pass

mattst88:
## What

`depgraph._serialize_tasks` breaks dependency cycles by finding the smallest group of mutually-RDEPENDing nodes that can be merged as a unit. This branch replaces the per-node recursive search with a single Tarjan SCC pass over the induced subgraph, plus two supporting `digraph` primitives.

Measured on a `@world` resolve, `_serialize_tasks` drops from ~62s to ~30s.

## The old algorithm

`gather_deps(ignore_priority, mergeable_nodes, selected_nodes, node)` is a recursive DFS. Starting at `node`, it walks child edges (filtered by `ignore_priority`) and returns `False` as soon as it reaches a node outside `mergeable_nodes`, or reaches the `replacement_portage` node while that node still has non-uninstall RDEPENDs in the graph. On success, `selected_nodes` holds the full reachable closure.

The caller ran that from scratch for every candidate:

```python
for priority in (soft, medium_soft, medium_soft+1):   # 3 levels
    for node in sorted(nodes):                        # N candidates
        selected_nodes = set()
        if gather_deps(priority, mergeable_nodes, selected_nodes, node):
            if smallest_cycle is None or len(selected_nodes) < len(smallest_cycle):
                smallest_cycle = selected_nodes
```

Each `gather_deps` call is O(V+E) worst case, so the block is O(N*(V+E)) per priority level, and the whole block re-runs on every iteration of the serialization loop. Adjacent candidates re-walk almost identical subgraphs, and every node in a cycle re-derives the same closure.

In the diamond below, `A` reaches `{A,B,C,D}`, `B` reaches `{B,D}`, `C` reaches `{C,D}` and `D` reaches `{D}`. The old code walks `D` four times, once per start node:

```mermaid
graph TD
    A --> B
    A --> C
    B --> D
    C --> D
```

In a cycle the redundancy is total. Every start node has the same closure `{A,B,C}`, computed three times:

```mermaid
graph LR
    A --> B
    B --> C
    C --> A
```

## The new algorithm

`_gather_deps_closures(graph, valid_nodes, ignore_priority, blocked_nodes)` computes, for all candidates at once, whether `gather_deps` would succeed and how big the resulting closure would be, in O(V+E) per priority level.

It rests on one property: `gather_deps(n)` succeeds iff every node reachable from `n` stays inside `valid_nodes` and is not blocked. That depends only on the reachable set, so it can be propagated bottom-up instead of re-derived top-down from each start node.

The four steps:

1. Build the filtered induced subgraph. For each node in `valid_nodes`, collect its `ignore_priority`-filtered children that are also in `valid_nodes`. A node with a filtered child outside `valid_nodes` is flagged `escapes`, since following that edge always fails.

2. Run iterative Tarjan SCC over that subgraph. Components come out in reverse-topological order (a component is emitted only after everything reachable from it), which is the order the next step needs. The implementation is iterative rather than recursive, so deep graphs cannot exhaust the Python stack.

3. Sweep the condensation in that order. A component is `ok` iff no member is blocked or escaping and every successor component is `ok`. Its closure is its own members plus the closures of its successor components, all of which were computed earlier in the sweep.

4. Fan the result out to nodes. Every node in an `ok` component gets that component's closure size, and all members of a cycle share one frozenset object, since their closures are identical.

The caller then does a lookup per candidate:

```python
ok_nodes, closure_size, closure_members = _gather_deps_closures(
    mygraph, mergeable_nodes, priority, blocked_nodes)
for node in nodes:
    if node not in ok_nodes:
        continue
    if smallest_cycle is None or closure_size[node] < smallest_size:
        smallest_cycle = closure_members[node]
```

### Worked example

```mermaid
graph LR
    subgraph valid["mergeable_nodes"]
        A --> B
        B --> A
        B --> C
        C --> D
        D --> C
        E --> F
        F --> E
        G --> H
    end
    H --> X["X (outside mergeable_nodes)"]
```

Condensation, swept in reverse-topological order:

| SCC | members | escapes? | successors | ok | closure | size |
|-----|---------|----------|------------|----|---------|------|
| 0 | `{C,D}` | no | none | yes | `{C,D}` | 2 |
| 1 | `{A,B}` | no | SCC 0 | yes | `{A,B,C,D}` | 4 |
| 2 | `{E,F}` | no | none | yes | `{E,F}` | 2 |
| 3 | `{H}` | yes (edge to X) | none | no | | |
| 4 | `{G}` | no | SCC 3 (not ok) | no | | |

`C`, `D`, `E` and `F` tie at size 2; `sorted(nodes)` order picks the first deterministically, as before. `G` and `H` are rejected: the failure at `H` propagates to `G` through the condensation instead of being rediscovered by a second DFS.

### Hoisted `replacement_portage` check

The old special case lived inside the recursion but closed over the outer `priority_range`, not the per-level `priority`, so its result was invariant across the whole search. It is now evaluated once before the priority loop and passed in as `blocked_nodes`:

```python
blocked_nodes = None
if (replacement_portage is not None
        and replacement_portage in mergeable_nodes
        and any(getattr(rdep, "operation", None) != "uninstall"
                for rdep in mygraph.child_nodes(
                    replacement_portage,
                    ignore_priority=priority_range.ignore_medium_soft))):
    blocked_nodes = frozenset((replacement_portage,))
```

Same semantics: new portage must get all its RDEPENDs installed first, so it may never be gathered into a cycle, and uninstalls (which occur when new portage blocks an older version) do not count.

`gather_deps` itself remains, since the later asap-node path still calls it.

## digraph support

`child_nodes_iter(node, ignore_priority)` is a generator form of `child_nodes()` with identical child order and no intermediate list. It is used by the closure builder (called once per node per priority level) and by `bfs()`.

`induced_subgraph(nodes)` builds the subgraph directly in O(len(nodes) + internal edges). It replaces the old copy-then-delete idiom, which cost O(whole graph) regardless of how small the cycle was:

```python
# before
cycle_digraph = mygraph.copy()
cycle_digraph.difference_update([x for x in cycle_digraph if x not in selected_nodes])

# after
cycle_digraph = mygraph.induced_subgraph(selected_nodes)
```

It preserves the parent graph's insertion order restricted to `nodes`, and shares one fresh priorities list between the child and parent views of each edge, as `add()` and `clone()` do, so the result matches the old construction.

## Tests

- `lib/portage/tests/resolver/test_gather_deps_closures.py`: differential tests asserting `_gather_deps_closures` agrees with the reference recursive `gather_deps` across escapes, blocked nodes, nested cycles and priority filters.
- `lib/portage/tests/util/test_digraph.py`: coverage for `child_nodes_iter` ordering and filtering, and for `induced_subgraph` equivalence to copy plus `difference_update`.

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -10,6 +10,7 @@ import time
 import warnings
 import collections
 from collections import deque, OrderedDict
+from collections.abc import Callable, Iterable
 from itertools import chain
 
 import os
@@ -66,6 +67,7 @@ from portage.util.portage_lru_cache import show_lru_cache_info
 from portage.versions import _pkg_str, catpkgsplit
 from portage.binpkg import get_binpkg_format
 
+from _emerge.AbstractDepPriority import AbstractDepPriority
 from _emerge.AtomArg import AtomArg
 from _emerge.Blocker import Blocker
 from _emerge.BlockerCache import BlockerCache
@@ -92,6 +94,7 @@ from _emerge.RootConfig import RootConfig
 from _emerge.search import search
 from _emerge.SetArg import SetArg
 from _emerge.show_invalid_depstring_notice import show_invalid_depstring_notice
+from _emerge.Task import Task
 from _emerge.UnmergeDepPriority import UnmergeDepPriority
 from _emerge.UseFlagDisplay import pkg_use_display
 from _emerge.UserQuery import UserQuery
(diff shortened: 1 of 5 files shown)


lib/_emerge/depgraph.py
```diff
@@ -123,6 +123,152 @@
...
+    escapes = {}
+    for node in valid_nodes:
+        kids = []
+        escaped = False
+        for child in graph.child_nodes_iter(node, ignore_priority=ignore_priority):
+            if child in valid_nodes:
+                kids.append(child)
+            else:
+                escaped = True
+        children[node] = kids
```
thesamesam, line 183: lol

lib/_emerge/depgraph.py
```diff
@@ -123,6 +123,152 @@
 )
 
 
+def _gather_deps_closures(graph, valid_nodes, ignore_priority, blocked_nodes=None):
```
thesamesam: I know we're not very typed in Portage yet, but could you try add some here? I think they're useful as documentation even if we're nowhere near using `mypy` yet.

zmedico (approved):

lib/_emerge/depgraph.py
```diff
@@ -9601,8 +9747,28 @@ def find_smallest_cycle(mergeable_nodes, local_priority_range):
                     # smallest cycle in order to try and identify and prefer
                     # these smaller independent cycles.
                     smallest_cycle = None
+                    smallest_size = None
                     ignore_priority = None
 
+                    # The replacement_portage special-case from gather_deps.
```
thesamesam, line 9770: Could you add a comment at the other site saying that we're usually not going to reach it?

thesamesam (approved):
This is lovely, thanks.

Some small comments. I'd also suggest to drop the number of tests that pass from the commit message, it doesn't matter; the size of the merge plan when testing _does_ matter, but not the testsuite count.

lib/_emerge/depgraph.py
eli-schwartz: (For generic annotations like Iterable, there's an implied "of typing.Any". It is, likewise, useful to document with more granularity here.)

lib/_emerge/depgraph.py
mattst88: Done, good idea.

lib/_emerge/depgraph.py
mattst88: :D

lib/_emerge/depgraph.py
mattst88: Done, (I think)

eli-schwartz: Sorry, when I said "with more granularity" I meant "Any is bad, because it's just explicitly saying that there is NO annotations. And if you don't actually define the type parameters for a generic type, it is implicitly generic across Any, which is not as good as actually defining what it's intended to be generic across".

We could add explicit `var: Any` annotations to every single variable in the codebase. It could be automated even. But it would not be very useful. The type checker would use it to suppress "variable is untyped" warnings, but the actual resolution would be the same: nothing whatsoever is known.

Any is useful in certain contexts, such as for functions that genuinely expect to take any input whatsoever and do "at least something" with it. 

For example locals() and globals() return dicts (a generic object) declared as `dict[str, Any]` since it is *truthful* to say that they can contain anything.

print() accepts object, not Any, because it can print anything that is a subclass of object but doesn't assume anything about its input other than that the interfaces provided by all objects no matter what, exist. (Like getting the string or repr value, and printing it. ;))

mattst88: Ah, understood. How's this then?

PR #1633, opened 2026-07-23
process: reset pid_ns_for_children before exec'ing the pid-ns supervisor

mattst88:
unshare(CLONE_NEWPID) does not move the caller into the new pid namespace; it only sets pid_ns_for_children so that subsequently forked children land there. That setting survives execve(2), and for as long as it differs from the caller's active pid namespace the kernel rejects CLONE_THREAD with EINVAL.

With FEATURES="pid-sandbox", _exec() unshares, forks the namespace's init process, and then execs the supervisor in the process that did the unshare. The supervisor therefore runs a program that cannot create threads. CPython never does so during startup, so this has gone unnoticed, but qemu-user spawns an RCU thread as it starts and dies:

    qemu: qemu_thread_create: Invalid argument

which makes pid-sandbox fail for every package built in a binfmt_misc emulated chroot.

Once init has been forked, this process has no further use for pid_ns_for_children, so point it back at the pid namespace the process already belongs to. That is a no-op for the process itself and lifts the CLONE_THREAD restriction for whatever it execs.

### DIFF
--- a/lib/portage/process.py
+++ b/lib/portage/process.py
@@ -937,6 +937,47 @@ def _exec_wrapper(
         raise
 
 
+def _reset_pid_ns_for_children(libc):
+    """
+    Restore this process's pid_ns_for_children to its own pid namespace.
+
+    unshare(CLONE_NEWPID) does not move the caller into the new pid
+    namespace, it only sets pid_ns_for_children so that future children
+    are created there. That setting survives execve(2), and while it
+    differs from the caller's active pid namespace the kernel rejects
+    CLONE_THREAD with EINVAL. Any program exec'd in that state which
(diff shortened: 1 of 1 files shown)


mattst88:
I ran into this yesterday when setting up an alpha nspawn container locally. Looking at the releng.git repo, I assume this is why we have `FEATURES=-pid-sandbox`. cc: @akhuettel

thesamesam:
https://github.com/gentoo/portage/pull/1577

floppym:
> #1577

Yeah, this is a different way of solving the same problem. This does seem like a simpler solution.

mattst88:
Closing in favor of #1577

PR #1634, opened 2026-07-23
resolver: speed up _serialize_tasks with an incremental leaf frontier

mattst88:
The merge-order calculation in _serialize_tasks repeatedly asks the digraph which nodes are leaves under a given ignore_priority filter, via digraph.leaf_nodes(). Each such query is an O(V) scan of the whole graph, and the asap/leaf-scan selection loop issues many of them, so the leaf-finding cost grows quadratically on large graphs.

Add lib/_emerge/_serialize_frontier.py, a Kahn-style leaf frontier. For every node it keeps a per-level count of how many of its children have an edge surviving that level's filter; a node is a leaf under level L iff its count for L is zero. Removing a node decrements its parents' counts, and adding an edge (the uninstall-reversal and blocker paths) increments the parent's counts, computed from a per-edge survival bitmask. Each level also owns a min-heap of the order-indices of its current leaves, so leaves can be enumerated in mygraph.order without an O(V) walk. A _FrontierDigraph wrapper keeps the frontier in sync on every mutation.

_serialize_tasks now drives the asap child-test (frontier.is_leaf) and the leaf-scan level loop (frontier.ready_nodes) from the frontier instead of leaf_nodes(). The selection logic downstream of the leaf list is unchanged, so the chosen node, and thus the merge order, is identical. Set PORTAGE_SERIALIZE_FRONTIER_DISABLE to fall back to the leaf_nodes() path.

Merge order is unchanged, verified byte-identical against emerge -uDNp @world (307-package graph, PYTHONHASHSEED=0). On that graph _serialize_tasks drops from ~9.09s to ~8.22s total across its 4 calls (~9.5%). The win scales with graph size; _serialize_tasks is only ~16% of the ~52s @world resolution, so total wall-clock moves little here.

### DIFF
--- a/lib/_emerge/_serialize_frontier.py
+++ b/lib/_emerge/_serialize_frontier.py
@@ -1,0 +1,257 @@
+# Copyright 2026 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+"""
+Incrementally-maintained leaf frontier for depgraph._serialize_tasks.
+
+_serialize_tasks repeatedly asks digraph.leaf_nodes() which nodes are leaves
+under an ignore_priority filter. Each query is an O(V) scan, and the selection
+loop issues many of them.
+
+The frontier keeps, per node and per level, a count of how many children have an
+edge surviving that level's filter; the node is a leaf under level L iff its
+count for L is zero. Removing a node decrements its parents' counts; adding an
+edge increments the parent's counts.
+
+Each level owns a min-heap of the order-indices of its current leaves, so leaves
+can be enumerated in mygraph.order without an O(V) walk. Heaps use lazy deletion:
+an entry is valid iff the node is still in the graph, still leaf under the level,
+and the entry's index is the node's current order-index.
+
+order-index is a node's position in mygraph.order; a node introduced later
+(uninstall reversal / blocker edges) gets an appended index, as in digraph.add().
+
+PORTAGE_SERIALIZE_FRONTIER_DISABLE makes _serialize_tasks fall back to the
+digraph.leaf_nodes() path (frontier not built).
+"""
+
+from heapq import heappop, heappush
+
+from _emerge.DepPriorityNormalRange import DepPriorityNormalRange
+from _emerge.DepPrioritySatisfiedRange import DepPrioritySatisfiedRange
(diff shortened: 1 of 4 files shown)


zmedico (approved):

lib/_emerge/depgraph.py
```diff
@@ -9461,6 +9462,17 @@ def _serialize_tasks(self):
 
         mygraph = self._dynamic_config.digraph.copy()
 
+        # Wrap mygraph so an incremental leaf frontier stays in sync with every
+        # mutation, then drive the asap/leaf-scan selection from it instead of
+        # O(V) leaf_nodes() scans. PORTAGE_SERIALIZE_FRONTIER_DISABLE falls back
+        # to leaf_nodes(). See _serialize_frontier.
+        use_frontier = "PORTAGE_SERIALIZE_FRONTIER_DISABLE" not in os.environ
```
thesamesam, line 9469: Please document this as an internal/experimental environment variable that may be removed or changed later in `emerge(1)` (see e.g. `PORTAGE_SHOW_HTTP_TRACE`).

Is the intention to just have this for a little while to allow debugging any problems? I assume so.

thesamesam (commented):
A tiny comment, not looked at this properly yet.

lib/_emerge/depgraph.py
mattst88: Oh yes, will do.

> Is the intention to just have this for a little while to allow debugging any problems? I assume so.

Yes, exactly. It was useful for benchmarking the commit as well.

thesamesam (approved):

lib/_emerge/_serialize_frontier.py
```diff
@@ -0,0 +1,257 @@
+# Copyright 2026 Gentoo Authors
```
thesamesam, line 1: Need to update `meson.build` for this new file.

mattst88: Oops, done, thanks.

lib/_emerge/depgraph.py
mattst88: Documented it.

PR #1635, opened 2026-07-26
Add FEATURES="cgroup" for per-build resource accounting

mattst88:
Add monitor-only cgroup v2 resource accounting for package builds.

When FEATURES="cgroup" is set and emerge runs as root on a cgroup v2 system, each build gets a leaf cgroup under PORTAGE_CGROUP_ROOT (default /sys/fs/cgroup/portage). The build's process tree is placed in that cgroup from within the child process (before any unshare/fork), so accounting is correct even with pid-sandbox enabled. No resource limits are imposed.

On build completion, emerge prints:
```
  >>> Resource usage for x11-libs/vte-0.82.3: cpu 22.3s, peak-mem 233MiB, io 4/0MiB r/w
```
Live snapshots are also included in observability output. Degrades gracefully to a no-op when not root or cgroup v2 is unavailable.

Adds PORTAGE_CGROUP_ROOT make.conf variable to override the base path.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -23,6 +23,12 @@ Breaking changes:
 
 Features:
 
+* Add FEATURES="cgroup" to account each source build's CPU time, peak memory,
+  and I/O in a per-build cgroup v2. No limits are imposed. Totals are written
+  to the build log and, with FEATURES="observability", included in the live
+  status snapshot. Configurable via PORTAGE_CGROUP_ROOT (default
+  /sys/fs/cgroup/portage); requires root and cgroup v2.
+
 * vartree: Avoid pruning preserved-libs database or syncing to disk if we
   installed no files.
 
--- a/NEWS
+++ b/NEWS
@@ -29,6 +29,12 @@ Features:
   status snapshot. Configurable via PORTAGE_CGROUP_ROOT (default
   /sys/fs/cgroup/portage); requires root and cgroup v2.
 
+* Add FEATURES="observability" to publish a machine-readable snapshot of the
+  running emerge process (which packages are building/merging, their current
+  ebuild phase, and elapsed time) to /run/portage/emerge-<pid>.json. Query it
+  with the new "portageq jobs" command or "emerge --status". A Unix socket
+  streams the same snapshots live to connected clients.
+
 * vartree: Avoid pruning preserved-libs database or syncing to disk if we
   installed no files.
 
@@ -192,12 +198,6 @@ Breaking changes:
   Portage 2.2.19. (bug #976083)
 
 Features:
-
-* Add FEATURES="observability" to publish a machine-readable snapshot of the
-  running emerge process (which packages are building/merging, their current
-  ebuild phase, and elapsed time) to /run/portage/emerge-<pid>.json. Query it
-  with the new "portageq jobs" command or "emerge --status". A Unix socket
-  streams the same snapshots live to connected clients.
 
 * Add User-Agent to default wget FETCHCOMMAND and RESUMECOMMAND settings, as
   well as emirrordist, in order to avoid blocking by some web servers,
--- a/lib/_emerge/Scheduler.py
+++ b/lib/_emerge/Scheduler.py
@@ -24,15 +24,17 @@ from portage.output import colorize, create_color_func, red
 bad = create_color_func("BAD")
 from portage._sets import SETPREFIX
 from portage._sets.base import InternalPackageSet
+from portage.util import ensure_dirs, writemsg, writemsg_level
+from portage.util.cgroup import CgroupManager, DEFAULT_CGROUP_ROOT
+from portage.util.futures import asyncio
+from portage.util.human_readable import bytes_to_human
+from portage.util.path import first_existing
+from portage.util.SlotObject import SlotObject
+from portage.util._async.SchedulerInterface import SchedulerInterface
 from portage.package.ebuild.digestcheck import digestcheck
 from portage.package.ebuild.digestgen import digestgen
 from portage.package.ebuild.doebuild import _check_temp_dir, _prepare_self_update
 from portage.package.ebuild.prepare_build_dirs import prepare_build_dirs
-from portage.util import ensure_dirs, writemsg, writemsg_level
-from portage.util._async.SchedulerInterface import SchedulerInterface
-from portage.util.futures import asyncio
-from portage.util.path import first_existing
-from portage.util.SlotObject import SlotObject
 
 import _emerge
 from _emerge._find_deep_system_runtime_deps import _find_deep_system_runtime_deps
@@ -229,6 +231,18 @@ class Scheduler(PollScheduler):
             xterm_titles=("notitles" not in settings.features)
         )
         self._observability = ObservabilityMonitor(self)
+        self._cgroup = None
+        if "cgroup" in settings.features:
+            cg = CgroupManager(
+                settings.get("PORTAGE_CGROUP_ROOT", DEFAULT_CGROUP_ROOT),
+                verbose="--verbose" in myopts,
+            )
+            if cg.setup():
+                self._cgroup = cg
+                settings.unlock()
+                settings["PORTAGE_CGROUP_ROOT"] = cg.emerge_root
+                settings.backup_changes("PORTAGE_CGROUP_ROOT")
+                settings.lock()
         self._max_load = myopts.get("--load-average")
         max_jobs = myopts.get("--jobs", 1)
         self._set_max_jobs(max_jobs)
@@ -404,6 +418,36 @@ class Scheduler(PollScheduler):
         callback so the observability snapshot reflects the live phase.
         """
         self._observability.note_phase(cpv, phase)
+
+    def _cgroup_finish(self, build):
+        """Log the final cgroup resource summary for a build and remove it."""
+        if self._cgroup is None:
+            return
+        cpv = build.pkg.cpv
+        stats = self._cgroup.read_stats(cpv)
+        if stats:
+            parts = []
+            if "cpu_usec" in stats:
+                parts.append(f"cpu {stats['cpu_usec'] / 1e6:.1f}s")
+            if "mem_peak" in stats:
+                parts.append(f"peak-mem {bytes_to_human(stats['mem_peak'])}")
+            if "mem_swap_peak" in stats and stats["mem_swap_peak"]:
+                parts.append(f"peak-swap {bytes_to_human(stats['mem_swap_peak'])}")
+            if "mem_zswap_current" in stats and stats["mem_zswap_current"]:
+                parts.append(f"zswap {bytes_to_human(stats['mem_zswap_current'])}")
+            if "io_read_bytes" in stats:
+                parts.append(
+                    "io {} / {} r/w".format(
+                        bytes_to_human(stats["io_read_bytes"]),
+                        bytes_to_human(stats.get("io_write_bytes", 0)),
+                    )
+                )
+            if parts:
+                msg = f"=== Resource usage for {cpv}: {', '.join(parts)}"
+                log_path = build.settings.get("PORTAGE_LOG_FILE")
+                self._sched_iface.output(f"{msg}\n", log_path=log_path)
+                self._logger.log(f" {msg}")
+        self._cgroup.destroy(cpv)
 
     def _init_graph(self, graph_config):
         """
@@ -1576,6 +1620,7 @@ class Scheduler(PollScheduler):
     def _build_exit(self, build):
         self._running_tasks.pop(id(build), None)
         self._observability.note_task_finished(build)
+        self._cgroup_finish(build)
         self._release_job_token(id(build))
         if build.returncode == os.EX_OK and self._terminated_tasks:
             # We've been interrupted, so we won't
@@ -1737,6 +1782,8 @@ class Scheduler(PollScheduler):
         finally:
             self._main_loop_cleanup()
             self._observability.close()
+            if self._cgroup is not None:
+                self._cgroup.close()
             portage.locks._quiet = False
             portage.elog.remove_listener(self._elog_listener)
             if display_callback.handle is not None:
@@ -1955,8 +2002,6 @@ class Scheduler(PollScheduler):
 
                         if actual_free_bytes < required_free_bytes:
                             if not self._warned_tmpdir_free_space:
-                                from portage.util.human_readable import bytes_to_human
-
                                 actual_free_bytes_hr = bytes_to_human(actual_free_bytes)
                                 required_free_bytes_hr = bytes_to_human(
                                     required_free_bytes
--- a/lib/portage/util/cgroup.py
+++ b/lib/portage/util/cgroup.py
@@ -1,0 +1,507 @@
+# Copyright 2026 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+"""cgroup v2 resource accounting for package builds (monitor-only).
+
+When FEATURES="cgroup" is enabled and emerge runs as root on a cgroup v2
+system, each source build gets its own leaf cgroup under
+PORTAGE_CGROUP_ROOT/emerge-<pid>/. The build's process tree is placed in
+that cgroup, so the kernel accounts its CPU time, peak memory, and I/O.
+These are read back for live observability snapshots and a per-build
+summary; no limits are imposed.
+
+On systems with CONFIG_FAIR_GROUP_SCHED the per-build cgroup hierarchy also
+improves scheduling fairness: the kernel distributes CPU time evenly across
+build slots rather than proportionally to thread count, which improves system
+responsiveness during parallel emerges.
+
+When running under a cgroup v2 manager that has already placed this process
+in a cgroup from which every controller we need can reach our leaves, the
+hierarchy is anchored under that cgroup rather than directly under the
+cgroup v2 root, keeping the manager's view of the hierarchy consistent.
+In practice that means a unit started with systemd's Delegate=yes, where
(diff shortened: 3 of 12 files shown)


thesamesam:
See f8e3b11496bd6d602a690535c4a3bb32bb8e9744 and some followups.

lib/portage/util/cgroup.py
```diff
@@ -0,0 +1,213 @@
...
+    return out
+
+
+def read_stats(path: str) -> dict | None:
+    """Read accounting from a leaf cgroup path. None if the cgroup is gone.
+
+    Keys (any may be absent depending on kernel/controllers):
+      cpu_usec       - total CPU time (microseconds)
+      mem_current    - current memory (bytes)
+      mem_peak       - peak memory (bytes; kernel >= 5.19)
```
Flowdalic: Maybe also consider adding `memory_swap_current` and `memory_swap_peak`. Those would be of interest to me. Maybe even also `memory_zswap_current` for a complete(r) overview of the memory subsystem.

lib/_emerge/Scheduler.py
```diff
@@ -405,6 +411,31 @@ def _observability_phase(self, cpv, phase):
...
         self._observability.note_phase(cpv, phase)
 
+    def _cgroup_finish(self, build):
+        """Log the final cgroup resource summary for a build and remove it."""
+        if self._cgroup is None:
+            return
+        cpv = build.pkg.cpv
+        stats = self._cgroup.read_stats(cpv)
+        if stats:
+            parts = []
```
Flowdalic, line 429: Reporting of bytes could use `bytes_to_human` from `lib/portage/util/human_readable.py`

Flowdalic (commented):
I really like this. Can we even extend the reporting of cgroup accounting to `FEATURES=observability`, e.g., so that the cgroup information is also shown via `emerge   --status`?

I did a brief review, but haven't tested it yet. Though, I intend to test this soon. Some minor remarks inline, nothing blocking.

> Failures to join are warned and ignored.
I am not sure if I understand this part of the commit message.

This re-uses the previously existing `FEATURES=cgroup`, which should be fine, but maybe provide a brief rationale why it is ok to re-use and we don't use e.g., `FEATURES=cgroup2` instead. `FEATURES=cgroup` got removed a while ago and this provides similar functionality and the underlying kernel mechanism is the successor that the previous FEATURE was based on.

lib/portage/util/cgroup.py
mattst88: Oh, nice. Added!

lib/_emerge/Scheduler.py
mattst88: Done.

mattst88:
> Can we even extend the reporting of cgroup accounting to FEATURES=observability, e.g., so that the cgroup information is also shown via emerge   --status?

Yep, good call. Added.

> This re-uses the previously existing FEATURES=cgroup, which should be fine, but maybe provide a brief rationale why it is ok to re-use and we don't use e.g., FEATURES=cgroup2 instead. FEATURES=cgroup got removed a while ago and this provides similar functionality and the underlying kernel mechanism is the successor that the previous FEATURE was based on.

Yep. The old FEATURES=cgroup was dropped in 2023, so it's unlikely anyone still has it in their FEATURES. It was cgroup v1, which is deprecated (and removed?). It was also about containing / limiting builds AFAIU rather than simply reporting resource usage.

We could probably also consider extending this cgroupv2 support to containing / limiting in the future, but that wasn't my aim here. Perhaps that should be a different FEATURES setting.

Flowdalic:
Note that this will also have an impact on the scheduling,  in a desirable way (and if CONFIG_FAIR_GROUP_SCHED is enabled). The scheduler will distribute computer resource much more fairly over the emerge jobs ( I don't think this was previously the case). For users not already scoping portage invocations, e.g. via systemd-run, this would also help with general system responsiveness. The commit message and probably the feature description should reflect that.

lib/portage/util/cgroup.py
```diff
@@ -0,0 +1,225 @@
...
+            ) as fh:
+                fh.write(" ".join(f"+{c}" for c in wanted))
+        except OSError as e:
+            # Non-fatal: e.g. controllers already enabled, or EBUSY.
+            _warn(f"cannot enable controllers on {root}: {e}")
+    return True
+
+
+def leaf_path(root: str, cpv: str) -> str:
+    return os.path.join(root, sanitize_name(cpv))
```
Flowdalic, line 259: If a build fails and leaves behind stray processes, then the `os.rmdir()` later will fail with `EBUSY` (as noted). However, this also means that the cgroup will stay alive and a subsequent emerge of the same package will be dumped into this cgroup. Because cgroup accounting metrics are kept, the new build's stat will be cluttered with the stats of the previous build.

I am not saying that this is a fatal design flaw, but certainly something to keep in mind. It could probably also be simply mitigated by adding a timestamp to the path (either unix or, maybe even better, one formatted as ISO).

Like, for example

```python
def leaf_path(root: str, cpv: str) -> str:
    """Constructs a new leaf path with an ISO timestamp prefix."""
    # Example: 2026-07-28T08:53:16_dev-libs_foo-1.2
    timestamp = datetime.datetime.now().replace(microsecond=0).isoformat()
    return os.path.join(root, f"{timestamp}_{sanitize_name(cpv)}")
```

lib/portage/util/cgroup.py
```diff
@@ -0,0 +1,225 @@
...
+"""
+
+from __future__ import annotations
+
+import os
+import re
+
+from portage.util import writemsg_level
+
+DEFAULT_CGROUP_ROOT = "/sys/fs/cgroup/portage"
```
Flowdalic: I wonder if it would be sensible if portage wouldn't place itself under its current cgroup (if it's not one of the root cgroups). On systemd systems, this should systemd aware of the cgroups since they are then not created outside of systemd manged cgroups. So maybe something like this untested code?

```python
def get_default_cgroup_root() -> str:
    try:
        with open("/proc/self/cgroup", encoding="ascii") as f:
            for line in f:
                if not line.startswith("0::"):
                    continue

                path = line.strip().split("::", 1)[1]
                if path and path not in ("/", "/init.scope"):
                    return f"/sys/fs/cgroup{path}/portage"
    except OSError:
        pass
    return "/sys/fs/cgroup/portage"

DEFAULT_CGROUP_ROOT = get_default_cgroup_root()
```

lib/portage/util/cgroup.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def read_stats(self, cpv: str) -> dict | None:
+        if not self.enabled:
+            return None
+        self._seen.add(cpv)
+        return read_stats(leaf_path(self.root, cpv))
+
+    def destroy(self, cpv: str) -> None:
+        self._seen.discard(cpv)
+        try:
```
Flowdalic, line 504: Cgroup v2 supports `cgroup.kill`. Thinking about it, we may could avoid the entire `FEATURES=pid-sandbox` machinery if we simply did

```python
with open(os.path.join(path, "cgroup.kill"), "w", encoding="ascii") as f:
    f.write("1")
```

here. CC'ing @mgorny as pid-sandbox author. I'd love to hear your thoughts on this.

Assuming the change in `leaf_path()` suggested above, this requires that we save the timed-based cgroup paths somewhere or use some other technique to obtain it.

lib/portage/util/cgroup.py
Flowdalic: Or, maybe even better, use `os.getpid` in the cgroup name and create a hierarchy under a single cgroup

```python
def get_default_cgroup_root() -> str:
    base = "/sys/fs/cgroup"
    try:
        with open("/proc/self/cgroup", encoding="ascii") as f:
            for line in f:
                if not line.startswith("0::"):
                    continue
                path = line.strip().split("::", 1)[1]
                if path and path not in ("/", "/init.scope"):
                    base = f"/sys/fs/cgroup{path}"
                    break
    except OSError:
        pass
        
    return os.path.join(base, "portage", f"emerge-{os.getpid()}")
```

This requires that we do a top-down walk enabling the cgroup controllers though.

lib/portage/util/cgroup.py
mattst88: Implemented in CgroupManager.destroy(): writes 1 to cgroup.kill before rmdir if the file exists (kernel ≥ 5.14). On older kernels the file is absent and we fall through to rmdir as before. Regarding replacing FEATURES=pid-sandbox: that's tempting but out of scope here. pid-sandbox also provides PID namespace isolation, not just cleanup, so they address different things.

lib/portage/util/cgroup.py
mattst88: Done.

_detect_portage_root() reads /proc/self/cgroup, skips / and /init.scope, and anchors under the current slice (e.g. …/user@1000.service/portage). Falls back to /sys/fs/cgroup/portage on bare systems. Computed once at module import.

Hierarchy is now <detected-base>/portage/emerge-<pid>/<cpv>. CgroupManager.setup() creates both levels and calls _enable_subtree_control() at each, so controller delegation flows top-down without a manual walk. After setup, the Scheduler writes emerge_root into settings["PORTAGE_CGROUP_ROOT"] so doebuild's module-level ensure_leaf call lands in the same subtree.

lib/portage/util/cgroup.py
mattst88: Good point. I went with the emerge-<pid> approach suggested in r3663652246 rather than a timestamp: each emerge invocation gets its own portage/emerge-<pid>/ intermediate cgroup, so stale leaves from a previous failed build are in a different subtree. Timestamps would work too but PIDs are cheaper and already unique per-invocation.

mattst88:
> Note that this will also have an impact on the scheduling, in a desirable way (and if CONFIG_FAIR_GROUP_SCHED is enabled). The scheduler will distribute computer resource much more fairly over the emerge jobs ( I don't think this was previously the case). For users not already scoping portage invocations, e.g. via systemd-run, this would also help with general system responsiveness. The commit message and probably the feature description should reflect that.

Good point, added to the module docstring: "On systems with CONFIG_FAIR_GROUP_SCHED the per-build cgroup hierarchy also improves scheduling fairness: the kernel distributes CPU time evenly across build slots rather than proportionally to thread count, which improves system responsiveness during parallel emerges." Will add to NEWS/make.conf.5 as well.

Flowdalic:
```
flo@neo-pc eix 1 $ sudo emerge -av socat
[gentoobinhost] Local copy of remote index is up-to-date and will be used (local: 2026-08-02T15:37:20+02:00, remote: 2026-08-02T15:37:20+02:00).

Calculating dependencies ... done!
Dependency resolution took 6.63 s (backtrack: 0/20).

[binary    gU  ] net-misc/socat-1.8.1.3-1::gentoo [1.8.1.1::gentoo] USE="ipv6 readline ssl -tcpd" 350 KiB

Total: 1 package (1 upgrade, 1 binary), Size of downloads: 350 KiB

Would you like to merge these packages? [Yes/No] yes
  File "/usr/lib/python-exec/python3.13/emerge", line 62, in main
    retval = emerge_main()
  File "/usr/lib/python3.13/site-packages/_emerge/main.py", line 1358, in emerge_main
    return run_action(emerge_config)
  File "/usr/lib/python3.13/site-packages/_emerge/actions.py", line 4285, in run_action
    retval = action_build(emerge_config, spinner=spinner)
  File "/usr/lib/python3.13/site-packages/_emerge/actions.py", line 679, in action_build
    mergetask = Scheduler(
        settings,
    ...<5 lines>...
        graph_config=mydepgraph.schedulerGraph(),
    )
  File "/usr/lib/python3.13/site-packages/_emerge/Scheduler.py", line 239, in __init__
    settings["PORTAGE_CGROUP_ROOT"] = cg.emerge_root
    ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/portage/package/ebuild/config.py", line 3258, in __setitem__
    self.modifying()
    ~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/portage/package/ebuild/config.py", line 1604, in modifying
    raise Exception(_("Configuration is locked."))
Exception: Configuration is locked.
```

So you need

```
diff --git a/lib/_emerge/Scheduler.py b/lib/_emerge/Scheduler.py
index 8ef9eccd3876..ab9f9cd53c6f 100644
--- a/lib/_emerge/Scheduler.py
+++ b/lib/_emerge/Scheduler.py
@@ -236,7 +236,10 @@ class Scheduler(PollScheduler):
             cg = CgroupManager(settings.get("PORTAGE_CGROUP_ROOT", DEFAULT_CGROUP_ROOT))
             if cg.setup():
                 self._cgroup = cg
+                settings.unlock()
                 settings["PORTAGE_CGROUP_ROOT"] = cg.emerge_root
+                settings.backup_changes("PORTAGE_CGROUP_ROOT")
+                settings.lock()
         self._max_load = myopts.get("--load-average")
         max_jobs = myopts.get("--jobs", 1)
         self._set_max_jobs(max_jobs)
```

Flowdalic:
We currently leave cgroups behind. There is no reliable way to prevent this, due to how cgroups v2 are designed, but we could at least try to clean up a bit. See attached patch
[0001-cleanup-cgroups.patch](https://github.com/user-attachments/files/30659122/0001-cleanup-cgroups.patch)

mattst88:
Thank you! Both should be fixed now (insofar as the cleanup can be fixed).

Flowdalic:
Looking good. I just found another issue when this is used under a systemd service with `Delegate=yes` (which want to set if `FEATURES=cgroup`). Portage needs to prepare the init.scope in this case. Please see attached patch for the logic to do so. It also includes some logging

[cgroup.patch](https://github.com/user-attachments/files/30699829/cgroup.patch)

mattst88:
Thank you so much for your review and advice. I think I've addressed this in the latest push. Please have a look.

Flowdalic:
Looking good. Unfortunately, I notice another thing:

Shouldn't the "Resource usage for {cpv}" line also appear in `emerge.log`? It seems to contain interesting information that other tools, like `emlop` could use. Maybe consider the following change


```
diff --git a/lib/_emerge/Scheduler.py b/lib/_emerge/Scheduler.py
index 87222d20ffc5..ae025574677e 100644
--- a/lib/_emerge/Scheduler.py
+++ b/lib/_emerge/Scheduler.py
@@ -443,9 +443,10 @@ class Scheduler(PollScheduler):
                     )
                 )
             if parts:
-                msg = f">>> Resource usage for {cpv}: {', '.join(parts)}\n"
+                msg = f"=== Resource usage for {cpv}: {', '.join(parts)}"
                 log_path = build.settings.get("PORTAGE_LOG_FILE")
-                self._sched_iface.output(msg, log_path=log_path)
+                self._sched_iface.output(f"{msg}\n", log_path=log_path)
+                self._logger.log(f" {msg}")
         self._cgroup.destroy(cpv)
 
     def _init_graph(self, graph_config):
 ```
 
 notice that I also changed the zing to `===`.

**Comment by mattst88** [comment]
I didn't consider that, but that does seem like a valuable thing to log! Thank you!

File: lib/_emerge/Scheduler.py [hunk 6]
```diff
@@ -405,6 +419,36 @@ def _observability_phase(self, cpv, phase):
...
+            if "cpu_usec" in stats:
+                parts.append(f"cpu {stats['cpu_usec'] / 1e6:.1f}s")
+            if "mem_peak" in stats:
+                parts.append(f"peak-mem {bytes_to_human(stats['mem_peak'], 0)}")
```
Flowdalic: Sorry, I only noticed that after using this for a while: This currently produces output like

1786082879:  === Resource usage for app-editors/vscode-1.132.0: cpu 728.9s, peak-mem 2 GiB, peak-swap 18 MiB, io 1 GiB / 344 MiB r/w
1786082892:  === Resource usage for app-containers/containerd-2.3.2-r1: cpu 393.5s, peak-mem 1 GiB, peak-swap 7 MiB, io 1 GiB / 153 MiB r/w
1786082919:  === Resource usage for gnome-base/nautilus-49.6-r1: cpu 199.0s, peak-mem 551 MiB, peak-swap 5 MiB, io 345 MiB / 42 MiB r/w

can we please drop the second argument to `bytes_to_human()` and use the default of two decimal places? Or, at least use one decimal place (i.e., change the `0` to `1`)? But I'd prefer simply the default of two decimal places. Thanks.

Here, and at the call sites below.

mattst88: Yeah, nice find! Thank you. I've dropped the 0 argument to let it default to two decimal places now.

NEWS
```diff
@@ -187,6 +187,12 @@ Breaking changes:
 
 Features:
 
+* Add FEATURES="cgroup" to account each source build's CPU time, peak memory,
```
thesamesam: Wrong release ;)

NEWS
```diff
@@ -187,6 +187,12 @@ Breaking changes:
...
+  and I/O in a per-build cgroup v2. No limits are imposed. Totals are written
+  to the build log and, with FEATURES="observability", included in the live
+  status snapshot. Configurable via PORTAGE_CGROUP_ROOT (default
+  /sys/fs/cgroup/portage); requires root and cgroup v2.
+
 * Add FEATURES="observability" to publish a machine-readable snapshot of the
```
thesamesam: Wrong release too, actually. Sorry for only noticing this now. Please check it all over.

mattst88: Oops. Fixed.

NEWS
mattst88: Oops. Fixed.

Flowdalic (approved):

PR #1636, opened 2026-07-28
FEATURES: add merge-use-vdb to bypass identical file rewrites

Flowdalic:
When FEATURES="merge-use-vdb" is enabled, Portage will use the hashes and mtimes recorded in VDB for installed packages. If the mtime on the live filesystem matches the VDB, it trusts the VDB hash rather than reading the live file and computing a new hash.

This allows portage to skip rewriting files that are identical, saving disk I/O and reduces pressure on the filesystem cache, while still correctly falling back to checking live hashes if the user has touched or modified the file.

Also update dblink._needs_move() to return a MoveReason Enum indicating why a file should be moved or not. This allows the caller to inspect the specific reason while maintaining compatibility with boolean checks via a custom __bool__ implementation.

When a file is skipped specifically because of a VDB hash match, mergeme() now outputs an '=V=' prefix instead of '==='.

This is similar to what @chewi proposed in https://github.com/chewi/portage/commit/f7e3e3753ab3a808865c2e351898b3054fd638ec but has an additional optimization to not compute `destmd5` if enabled, not only avoiding disk writes, but also unnecessary disk reads.

@floppym raised the concern that this feature, if enabled, prevents the restoration of corrupted package installs, since portage would not merge and override corrupted files when it relies only on the VDB entry. This is a valid concern. However, the feature is *disabled by default*. The tradeoff is also explicitly documented in the man page so users can make an informed choice before opting *in.*

### DIFF
--- a/lib/portage/dbapi/vartree.py
+++ b/lib/portage/dbapi/vartree.py
@@ -25,6 +25,7 @@ import tempfile
 import textwrap
 import time
 import warnings
+from enum import Enum, auto
 from itertools import chain
 
 from _emerge.EbuildBuildDir import EbuildBuildDir
@@ -345,6 +346,23 @@ def _explode_metadata_file(dbdir):
 
     os.unlink(path)
     return missing
+
+
+class MoveReason(Enum):
+    # Falsy reasons (Move not needed)
+    VDB_HASH_MATCHES = auto()
+    CONTENT_MATCHES = auto()
+
+    # Truthy reasons (Move needed)
+    FILE_MISSING_OR_NOT_REGULAR = auto()
+    MODE_DIFFERS = auto()
+    VDB_HASH_DIFFERS = auto()
+    XATTR_DIFFERS = auto()
+    CONTENT_DIFFERS = auto()
+    COMPARISON_EXCEPTION = auto()
+
+    def __bool__(self):
+        return self.value >= self.__class__.FILE_MISSING_OR_NOT_REGULAR.value
 
 
 class vardbapi(dbapi):
@@ -5623,7 +5641,26 @@ class dblink:
                         ).hexdigest()
 
                     elif stat.S_ISREG(mydmode):
-                        destmd5 = perform_md5(mydest)
+                        destmd5 = None
+                        # If the file hasn't been modified since it was installed, we can safely
+                        # reuse the MD5 hash recorded in the var database (VDB) instead of
+                        # reading the file from disk to compute it.
+                        if (
+                            "merge-use-vdb" in self.settings.features
+                            and self._installed_instance is not None
+                        ):
+                            k = self._installed_instance._match_contents(myrealdest)
+                            if k is not False:
+                                data = self._installed_instance.getcontents()[k]
+                                if data[0] == "obj":
+                                    vdb_mtime = data[1]
+                                    if (
+                                        str(mydstat.st_mtime_ns // 1000000000)
+                                        == vdb_mtime
+                                    ):
+                                        destmd5 = data[2]
+                        if destmd5 is None:
+                            destmd5 = perform_md5(mydest)
             except (FileNotFound, OSError) as e:
                 if isinstance(e, OSError) and e.errno != errno.ENOENT:
                     raise
@@ -5916,7 +5953,10 @@ class dblink:
                 # same way.  Unless moveme=0 (blocking directory)
                 if moveme:
                     # only replace the existing file if it differs, see #722270
-                    if self._needs_move(mysrc, mydest, mymode, mydmode):
+                    needs_move_reason = self._needs_move(
+                        mysrc, mydest, mymode, mydmode, mymd5, myrealdest
+                    )
+                    if needs_move_reason:
                         # Create hardlinks only for source files that already exist
                         # as hardlinks (having identical st_dev and st_ino).
                         hardlink_key = (mystat.st_dev, mystat.st_ino)
@@ -5945,7 +5985,11 @@ class dblink:
                         except OSError:
                             # utime can fail here with EPERM
                             pass
-                        zing = "==="
+                        zing = (
+                            "=V="
+                            if needs_move_reason == MoveReason.VDB_HASH_MATCHES
+                            else "==="
+                        )
 
                     try:
                         self._merged_path(mydest, os.lstat(mydest))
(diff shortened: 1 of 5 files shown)


Flowdalic:
Some benchmark results (also in commit message):

```
$ sudo ./bench-merge-use-vdb.sh --min-runs 15
Benchmarking emerge for: dev-lang/rust-bin
Cold cache (drop VM caches): 0
Extra hyperfine args: none
--------------------------------------------------------
Benchmark 1: with    merge-use-vdb
  Time (mean ± σ):     80.588 s ±  5.651 s    [User: 73.699 s, System: 10.421 s]
  Range (min … max):   73.536 s … 90.338 s    15 runs

Benchmark 2: without merge-use-vdb
  Time (mean ± σ):     90.187 s ±  3.641 s    [User: 74.933 s, System: 12.688 s]
  Range (min … max):   86.546 s … 96.082 s    15 runs

Summary
  with    merge-use-vdb ran
    1.12 ± 0.09 times faster than without merge-use-vdb
    ```

**Comment by thesamesam** [comment]
Please tag https://bugs.gentoo.org/722270.

Should we remove the old implementation? I think if people really want the optimisation, it makes sense to enable this. If not, they can live with the writes and may even prefer that over the comparison for speed.

I'm still not totally sure about this just in terms of more complexity + FEATURE knobs. I'd prefer it if we were removing something related in the process given how similar they are.

Can you benchmark this PR with the removal as an alternative for comparison too?

**Comment by chewi** [comment]
> Should we remove the old implementation? I think if people really want the optimisation, it makes sense to enable this. If not, they can live with the writes and may even prefer that over the comparison for speed.

Personally, I'm happy with what we have, despite my initial concerns about performance. It's fast enough that I'd prefer to take the performance hit over risking the corruption issue, at least on an SSD.

**Comment by Flowdalic** [comment]
> Can you benchmark this PR with the removal as an alternative for comparison too?

I have benchmarked https://github.com/Flowdalic/portage/commit/eca2fdcc9b852515c99b5ba6577680469be39b45:

```
$ sudo ./bench-merge-use-vdb.sh --min-runs 15 --emerge-bin ~/repos/gentoo/portage/bin/emerge
Password: 
Benchmarking emerge for: dev-lang/rust-bin
EMERGE_BIN: /home/flo/repos/gentoo/portage/bin/emerge
Cold cache (drop VM caches): 0
Extra hyperfine args: none
--------------------------------------------------------
Benchmark 1: with	 merge-use-vdb
  Time (mean ± σ):     84.805 s ±  2.345 s    [User: 79.364 s, System: 10.670 s]
  Range (min … max):   82.465 s … 92.068 s    15 runs
 
Benchmark 2: without merge-use-vdb
  Time (mean ± σ):     285.552 s ± 22.809 s    [User: 90.409 s, System: 30.482 s]
  Range (min … max):   244.298 s … 323.734 s    15 runs
 
Summary
  with	 merge-use-vdb ran
    3.37 ± 0.28 times faster than without merge-use-vdb
```

I absolutely support the goal of reducing codebase's complexity. However, those numbers show that removing filecmp incurs a significant performance penalty for users on the default configuration (assuming FEATURES=merge-use-vdb remains optional).

Because the existing filecmp code path has low cyclomatic complexity, removing it yields minimal maintenance gains while sacrificing a rather huge performance advantage and it comes with increasing disk writes. Given this disproportionate cost-to-benefit ratio, we should be pragmatic and retain the filecmp path.


> > Should we remove the old implementation? I think if people really want the optimisation, it makes sense to enable this. If not, they can live with the writes and may even prefer that over the comparison for speed.
> 
> Personally, I'm happy with what we have, despite my initial concerns about performance. It's fast enough that I'd prefer to take the performance hit over risking the corruption issue, at least on an SSD.

I think we have a good compromise here: by keeping the default behavior as-is, we prioritize safety, users take the performance hit, but we avoid the risk of corruption.

At the same time, we shouldn't remove the optimization entirely. Keeping the filecmp implementation as default seems sensible, while FEATURE=merge-use-vdb provides a significant speedup.

File: lib/portage/dbapi/vartree.py [hunk 1]
```diff
@@ -5416,7 +5434,23 @@ def mergeme(
                         ).hexdigest()
 
                     elif stat.S_ISREG(mydmode):
-                        destmd5 = perform_md5(mydest)
+                        destmd5 = None
+                        if (
```
thesamesam, line 5648: Please add some commentary here. One can figure out what it does but it'd be nice to have it said.

lib/portage/dbapi/vartree.py
```diff
@@ -6172,13 +6213,38 @@ def _needs_move(self, mysrc, mydest, mymode, mydmode):
...
+        if mymode != mydmode:
+            return MoveReason.MODE_DIFFERS
+
+        if (
+            "merge-use-vdb" in self.settings.features
+            and myrealdest is not None
+            and mymd5 is not None
+            and self._installed_instance is not None
+        ):
+            k = self._installed_instance._match_contents(myrealdest)
```
thesamesam, line 6435: commentary again for these additions please?

thesamesam:
Alright, we can do it. Can you include that information in the commit message too? (Maybe separate it with a `--` or something).

Is the mtime still updated with this (https://bugs.gentoo.org/906978)?

Flowdalic:
Addressed the review comments

lib/portage/dbapi/vartree.py
```diff
@@ -6391,9 +6464,13 @@ def _needs_move(self, mysrc, mydest, mymode, mydmode):
...
             )
-            return True
+            return MoveReason.COMPARISON_EXCEPTION
 
-        return not files_equal
+        return (
+            MoveReason.CONTENT_DIFFERS
+            if not files_equal
+            else MoveReason.CONTENT_MATCHES
+        )
```
zmedico, line 6473: It used to return False here when the content matched, but now it returns MoveReason.CONTENT_MATCHES. Is the `if needs_move_reason:` condition always True now, and is that intentional?

Flowdalic: `MoveReason` has

```python3
    def __bool__(self):
        return self.value >= self.__class__.FILE_MISSING_OR_NOT_REGULAR.value
```
so if `needs_move_reason` is `MoveReason.CONTENT_MATCHES` then `if needs_move_reason` would evaluate to `false`.

zmedico: I see. Thank you!

zmedico (approved):

thesamesam:
Thanks. I'll merge after the release (as master is already huge, and this touches a critical component where bugs may not be noticed immediately).

lib/portage/tests/emerge/test_merge_use_vdb.py
```diff
@@ -0,0 +1,198 @@
+# Copyright 2026 Gentoo Authors
```
thesamesam, line 1: I've added this to meson.build too so it gets installed.

PR #1639, opened 2026-08-02
dep: add a native C dependency-string parser

mattst88:
Dependency-string parsing is one of the largest single costs in a `@world` resolve. This adds a recursive-descent parser in C and routes `use_reduce()` and `Atom` construction through it, falling back to the existing pure-Python implementation whenever the fast path cannot serve a call.

The two paths are meant to be indistinguishable. `PORTAGE_NATIVE_DEP_PARSER=0` forces the pure-Python one, both so that can be verified and so a user who hits a defect in the C parser has a way to keep working while it is fixed. It is read once, when `portage.dep` is imported, long before make.conf is parsed, so it is an environment variable only, documented in `emerge(1)` alongside the other Portage development variables.

## Results

Over the 80065 DEPEND/RDEPEND/BDEPEND/PDEPEND/IDEPEND strings in the ::gentoo md5-cache (`token_class=Atom`, `matchall=True`, `eapi=8`), best of three, `-O2`. The pure-Python path is memoized, so its `lru_cache` is cleared between runs:

| | time | str/s | |
|---|---|---|---|
| `_parser.parse()` alone | 0.167 s | 479124 | |
| `use_reduce`, Python path | 2.908 s | 27536 | |
| `use_reduce`, C path | 1.178 s | 67990 | **2.5x** |

End to end, unprofiled, `emerge -uDNp @world` goes from **54.3s to 50.1s** (best of three, `PORTAGE_NATIVE_DEP_PARSER=0` as the baseline). Under cProfile, `Atom.__init__` drops from 14.65s to 6.98s cumulative and the Atom fast path serves 663538 of 779655 constructions (85%) — nearly every call not excluded by an injected `_use` or a virtual-expansion original. The `@world` merge list is unchanged.

### Why 2.5x and not 2.9x

The C path is fastest partway through the series — 78938 str/s after commit 2 — and settles at 67990 by the end. Two commits account for the difference, and both buy correctness:

- **Commit 5** (~14%, the bulk of it). The parser originally inlined naked all-of groups into their parent, which is right in an all-of context but silently drops the conjunction inside an any-of: `|| ( ( a b ) c )` came back as `["||", ["a", "b", "c"]]` instead of `["||", [["a", "b"], "c"]]`. Fixing it means emitting explicit group boundaries and then normalizing that tree back into `use_reduce`'s shape on the Python side. That normalization is the cost.
- **Commit 7** (~2%). Making `scan_atom` byte-exact with the regex atom grammar needs validations the low-level scanner did not have — a version requires an operator, `=*` is only valid with `=`, and a package name must not end in a hyphen followed by a version, which means checking every `-` position rather than just the last.

Both are the difference between a fast parser and a fast parser that returns a different answer. Since the entire design rests on the two paths being indistinguishable — that is what makes the kill switch meaningful and what the fuzzing checks — paying ~16% for it is the right trade. The end-to-end `@world` number above is measured on the as-merged series, so it already includes this cost.

## Commits

1. `dep: add native C dep-string parser` — the scanner, the Python extension, and C unit tests
2. `dep: wire C parser into use_reduce fast path` — gated to EAPI 5+, whose atom grammar the scanner implements
3. `dep: add a PORTAGE_NATIVE_DEP_PARSER kill switch`
4. `tests: add C parser parity and classify_use_deps tests`
5. `dep: preserve conjunctions inside any-of groups`
6. `dep: extend C parser fast path to flat=True`
7. `dep: add C fast path for Atom construction`
8. `dep: widen the Atom C fast path to wildcard-flag and is_valid_flag callers`
9. `NEWS: mention the native C dependency-string parser`

Every commit builds warning-free and passes the full test suite on its own.

## Testing

`_UseReduceTests` is a mixin run twice — `TestUseReducePythonPath` and `TestUseReduceCPath` — so every correctness assertion executes on both code paths. On top of that, `TestCFastVsPython` and `TestScanAtom` assert the two paths are byte-identical field by field and reject the same malformed input, `TestCParserEapiGuard` covers EAPI 0–8, and `TestDeepNesting` checks nesting parity across the point where the C parser's group stack spills to the heap.

Beyond the suite: a differential fuzz of 192000 combinations (nesting depths 0–5, varied USE lists, matchall on/off, EAPI 8/6/5/None, flat and non-flat) found no differences against the pure-Python path, and the same fuzz under ASan+UBSan is clean. The standalone `test_parser` binary runs 127 C-level cases and is wired into `meson test`.

## Notes for reviewers

- **License headers.** The new files use `SPDX-License-Identifier: GPL-2.0-or-later OR MIT` rather than the tree's usual GPL-2 boilerplate, since this is all new code. GPL-2.0-or-later is because there have been some discussions on trying to make this change more generally for portage. MIT is so the code could be included in pkgcraft or other tools if wanted.
- **Nonstandard C.** `__attribute__((cleanup))` for scope-bound `Py_DECREF`, and `__VA_OPT__` (C2x) in one helper macro. Both work with GCC and Clang, which is what Portage builds with in practice, but no `c_std` is set anywhere in the build so it is worth a conscious decision.
- **No hardcoded optimization flags.** The extension inherits the buildtype, so Gentoo's `--buildtype plain` plus user `CFLAGS` applies. Note this means a bare `meson setup build` (default `optimization=0`) produces an extension about 1.6x slower than `-O2`; `-O2` and `-O3` are within noise of each other. Same as `_whirlpool` today.
- **Nesting depth is unbounded** (heap-grown group stack) rather than capped, because the pure-Python path has no limit either and rejecting a dep string it accepts would be a divergence.

### DIFF
--- a/lib/portage/dep/__init__.py
+++ b/lib/portage/dep/__init__.py
@@ -200,9 +200,24 @@ def _c_normalize_seq(seq, empty_always_true, under_anyof=False):
     return out
 
 
-def _c_fast_use_reduce(depstr, uselist, matchall, eapi):
+def _c_flatten_result(items, out, eapi, eapi_attrs, uselist, matchall):
+    """Recursively flatten the C parse tree into use_reduce(flat=True) form.
+    Nested || groups produce repeated '||' tokens, matching the Python path."""
+    for item in items:
+        if isinstance(item, str):
+            out.append(item)
+        elif isinstance(item, list):
+            _c_flatten_result(item, out, eapi, eapi_attrs, uselist, matchall)
+        else:
+            out.append(_c_atom_from_c(item, eapi, eapi_attrs, uselist, matchall))
+    return out
+
+
+def _c_fast_use_reduce(depstr, uselist, matchall, eapi, flat):
     raw = _c_dep_parser.parse(depstr, uselist=uselist, matchall=matchall)
     eapi_attrs = _get_eapi_attrs(eapi)
+    if flat:
+        return _c_flatten_result(raw, [], eapi, eapi_attrs, uselist, matchall)
     result = _c_convert_result(raw, eapi, eapi_attrs, uselist, matchall)
     return _c_normalize_seq(result, eapi_attrs.empty_groups_always_true)
 
@@ -1109,7 +1124,6 @@ def use_reduce(
         and token_class is Atom
         and not is_src_uri
         and not opconvert
-        and not flat
         and is_valid_flag is None
         and subset is None
         and not matchnone
@@ -1119,7 +1133,7 @@ def use_reduce(
         and (eapi is None or _get_eapi_attrs(eapi).slot_operator)
     ):
         try:
-            return _c_fast_use_reduce(depstr, uselist, matchall, eapi)
+            return _c_fast_use_reduce(depstr, uselist, matchall, eapi, flat)
         except ValueError as e:
             raise InvalidDependString(str(e)) from e
 
--- a/lib/portage/dep/__init__.py
+++ b/lib/portage/dep/__init__.py
@@ -1826,7 +1826,23 @@ class Atom:
         def __init__(self, forbid_overlap=False):
             self.overlap = self._overlap(forbid=forbid_overlap)
 
-    def _c_fast_init(self, catom, eapi, eapi_attrs, unevaluated_atom):
+    def _validate_conditional_flags(self, is_valid_flag):
+        """Raise if a USE-conditional flag in this atom is not in IUSE.
+
+        Mirrors the check the regex path performs in __init__.
+        """
+        for conditional_type, flags in self._use.conditional.items():
+            for flag in flags:
+                if is_valid_flag(flag):
+                    continue
+                conditional_str = _use_dep._conditional_strings[conditional_type]
+                raise InvalidAtom(
+                    f"USE flag '{flag}' referenced in conditional "
+                    f"'{conditional_str % flag}' in atom '{self}' is not in IUSE",
+                    category="IUSE.missing",
+                )
+
+    def _c_fast_init(self, catom, eapi, eapi_attrs, is_valid_flag, unevaluated_atom):
         """Populate this Atom from a C scan_atom result.
 
         Raises InvalidAtom/TypeError for the same cases as the regex path.
@@ -1868,6 +1884,8 @@ class Atom:
                     f"Use dep defaults are not allowed in EAPI {eapi}: '{self}'",
                     category="EAPI.incompatible",
                 )
+            if is_valid_flag is not None and self._use.conditional:
+                self._validate_conditional_flags(is_valid_flag)
 
         if (
             self._blocker_obj
@@ -1918,17 +1936,13 @@ class Atom:
             if allow_build_id is None:
                 allow_build_id = True
 
-        # Fast path: parse the atom in C for the common case. scan_atom is
-        # byte-exact with the regex path or raises ValueError (repo specs,
-        # build-ids, wildcards, anything invalid), in which case we fall
-        # through to the pure-Python regex path below. Wildcards, an injected
-        # _use, flag validation, and virtual-expansion originals are handled
-        # only by the regex path.
+        # Try the C fast path. scan_atom raises ValueError for anything it
+        # can't handle (wildcards, repo specs, build-ids); we fall through to
+        # the regex path. allow_wildcard is safe: __init__ tries strict regex
+        # first, so actual wildcard atoms fail scan_atom and fall back naturally.
         if (
             _c_dep_parser is not None
-            and not allow_wildcard
             and _use is None
-            and is_valid_flag is None
             and orig_atom is None
             and (eapi is None or eapi_attrs.slot_operator)
         ):
@@ -1937,7 +1951,9 @@ class Atom:
             except ValueError:
                 catom = None
             if catom is not None:
-                self._c_fast_init(catom, eapi, eapi_attrs, unevaluated_atom)
+                self._c_fast_init(
+                    catom, eapi, eapi_attrs, is_valid_flag, unevaluated_atom
+                )
                 return
 
         if s[:1] == "!":
--- a/lib/portage/dep/__init__.py
+++ b/lib/portage/dep/__init__.py
@@ -30,6 +30,7 @@ __all__ = [
     "use_reduce",
 ]
 
+import os
 import re
 import warnings
 from functools import lru_cache
@@ -57,6 +58,10 @@ try:
     # extension has not been built, as in the lint-only CI job.
     import portage.dep._parser as _c_dep_parser
 except ImportError:
+    _c_dep_parser = None
+
+# PORTAGE_NATIVE_DEP_PARSER=0 forces the pure-Python path.  See emerge(1).
+if os.environ.get("PORTAGE_NATIVE_DEP_PARSER") == "0":
     _c_dep_parser = None
 
 if TYPE_CHECKING:
--- a/lib/portage/dep/__init__.py
+++ b/lib/portage/dep/__init__.py
@@ -130,10 +130,81 @@ def _c_convert_result(items, eapi, eapi_attrs, uselist, matchall):
     return result
 
 
+def _c_normalize_alts(group, empty_always_true):
+    """Reduce the alternatives of one || group to use_reduce's non-flat form."""
+    alts = []
+    i = 0
+    n = len(group)
+    while i < n:
+        item = group[i]
+        if isinstance(item, str):  # a nested '||'
+            i += 1
+            nested = _c_normalize_alts(group[i], empty_always_true)
+            if nested:
+                alts.extend(nested)
+            elif not empty_always_true:
+                # empty nested || -> placeholder atom (EAPI 7+)
+                alts.append([Atom("__const__/empty-any-of")])
+        elif isinstance(item, list):
+            # conjunction: normalize under_anyof to suppress bracket removal
+            sub = _c_normalize_seq(item, empty_always_true, under_anyof=True)
+            if len(sub) == 1:
+                # ( X ) alternative -> X
+                alts.append(sub[0])
+            elif len(sub) == 2 and sub[0] == "||":
+                # ( || ( ... ) ) alternative -> its alternatives flatten into
+                # the enclosing any-of.
+                alts.extend(sub[1])
+            elif sub:
+                alts.append(sub)
+        else:
+            alts.append(item)
+        i += 1
+    return alts
+
+
+def _c_normalize_seq(seq, empty_always_true, under_anyof=False):
+    """Reduce a C parse (sub)tree to use_reduce's non-flat form.
+
+    under_anyof: this sequence is an alternative of an enclosing || group;
+    use_reduce keeps redundant brackets there, suppressing conjunction inlining."""
+    out = []
+    i = 0
+    n = len(seq)
+    while i < n:
+        item = seq[i]
+        if isinstance(item, str):  # '||'
+            i += 1
+            alts = _c_normalize_alts(seq[i], empty_always_true)
+            if not alts:
+                # || ( ) -> dropped (EAPI < 7) or a const placeholder (EAPI 7+).
+                if not empty_always_true:
+                    out.append(Atom("__const__/empty-any-of"))
+            elif len(alts) == 1:
+                # || ( X ) -> X
+                alt = alts[0]
+                if isinstance(alt, list) and not under_anyof:
+                    out.extend(alt)
+                else:
+                    out.append(alt)
+            else:
+                out.append("||")
+                out.append(alts)
+        elif isinstance(item, list):
+            out.extend(
+                _c_normalize_seq(item, empty_always_true, under_anyof=under_anyof)
+            )
+        else:
+            out.append(item)
+        i += 1
+    return out
+
+
 def _c_fast_use_reduce(depstr, uselist, matchall, eapi):
     raw = _c_dep_parser.parse(depstr, uselist=uselist, matchall=matchall)
     eapi_attrs = _get_eapi_attrs(eapi)
-    return _c_convert_result(raw, eapi, eapi_attrs, uselist, matchall)
+    result = _c_convert_result(raw, eapi, eapi_attrs, uselist, matchall)
+    return _c_normalize_seq(result, eapi_attrs.empty_groups_always_true)
 
 
 # \w is [a-zA-Z0-9_]
--- a/lib/portage/dep/__init__.py
+++ b/lib/portage/dep/__init__.py
@@ -68,34 +68,46 @@ if TYPE_CHECKING:
     from _emerge.Package import Package
 
 
-def _c_atom_from_c(catom, eapi, eapi_attrs, uselist, matchall):
-    """Construct a portage.dep.Atom from a _parser.Atom, bypassing __init__ regex."""
-    a = Atom.__new__(Atom)
-    a._string = str(catom)
-    a._cp = catom.cp
-    a._repo = None
-    a._slot = catom.slot
-    a._sub_slot = catom.sub_slot
-    a._slot_operator = catom.slot_operator
-    ver = catom.version
+def _c_fill_atom_fields(atom, catom, eapi):
+    """Copy the scalar fields of a _parser.Atom onto a portage.dep.Atom.
+
+    Everything except _string, _use and _unevaluated_atom, which the two
+    callers set differently.  The C scanner reports the "=*" glob as operator
+    "=" with a trailing "*" on the version, so undo that here.
+    """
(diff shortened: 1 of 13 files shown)


eli-schwartz:
> ## Notes for reviewers
> 
> * **License headers.** The new files use `SPDX-License-Identifier: GPL-2.0-or-later OR MIT` rather than the tree's usual GPL-2 boilerplate, since this is all new code. GPL-2.0-or-later is because there have been some discussions on trying to make this change more generally for portage. MIT is so the code could be included in pkgcraft or other tools if wanted.

I'm not convinced this is meaningful, e.g. pkgcraft already statically links libscallop.a -- a fork of bash intended to be embedded and run in-process.


In general I'm somewhat skeptical of allowing GPL code to be alternatively available under permissive licenses "in case people want to use it in permissive codebases". It basically runs counter to the point of having a GPL at all. And I think other tools can simply use the GPL and have no problems, it's open source written for the open source world anyway.

I mean, it's your code anyway so you can always allow it regardless of what license the portage copy identifies as, and omitting MIT here "for project policy" or similar is a meaningless restriction if you want to offer both licenses.

Just not really sure I see what your motivation for doing so is. :P

lib/portage/dep/__init__.py
```diff
@@ -1627,6 +1824,75 @@ def __init__(self, forbid=False):
...
+        if use_tokens is not None:
+            # _use_dep validates conflicting flags, same as the regex path.
+            self._use = _use_dep(list(use_tokens), eapi_attrs)
+        else:
+            self._use = None
+
+        _c_fill_atom_fields(self, catom, eapi)
+        self._unevaluated_atom = unevaluated_atom if unevaluated_atom else self
+
+        if eapi is not None:
```
thesamesam: Can you early-return here?

lib/portage/dep/__init__.py
```diff
@@ -1627,6 +1824,75 @@ def __init__(self, forbid=False):
...
+                        category="EAPI.incompatible",
+                    )
+                if is_valid_flag is not None and self._use.conditional:
+                    invalid_flag = None
+                    try:
+                        for conditional_type, flags in self._use.conditional.items():
+                            for flag in flags:
+                                if not is_valid_flag(flag):
+                                    invalid_flag = (conditional_type, flag)
+                                    raise StopIteration()
```
thesamesam: This is a trick because of a needed double-break? Maybe it'd be better as a function we call instead. I'm really trying to reduce the huge depth we have in Portage, but especially for new code.

lib/portage/dep/__init__.py
```diff
@@ -1627,6 +1824,75 @@ def __init__(self, forbid=False):
...
+                    except StopIteration:
+                        pass
+                    if invalid_flag is not None:
+                        conditional_type, flag = invalid_flag
+                        conditional_str = _use_dep._conditional_strings[
+                            conditional_type
+                        ]
+                        msg = _(
```
thesamesam: Don't bother adding new `_()` just f-string it from the off. We've never had translations in Portage and I don't see us starting any time soon.

src/dep_parser.c
```diff
@@ -0,0 +1,701 @@
...
+#define AUTO_PY __attribute__((cleanup(py_decref_p))) PyObject *
+
+static struct {
+    /* operator strings */
+    PyObject *op_lt, *op_gt, *op_le, *op_ge, *op_eq, *op_tilde;
+    /* blocker strings */
+    PyObject *blocker_weak, *blocker_strong;   /* "!" weak, "!!" strong */
+    /* slot operator strings */
+    PyObject *slot_op_eq, *slot_op_star;
+} G;
```
thesamesam: Why `G`?

src/dep_parser.c
```diff
@@ -0,0 +1,701 @@
...
+        int main_len = (int)(slash - raw);
+        if (main_len > 0 && raw[main_len - 1] == '=') {
+            slot_op = '=';
+            main_len--;
+        }
+        *out_slot = PyUnicode_FromStringAndSize(raw, main_len);
+
+        const char *sub = slash + 1;
+        int sub_len = rlen - (int)(sub - raw);
+        if (sub_len == 1 && (sub[0] == '*' || sub[0] == '=')) {
```
thesamesam: One can work this out quite quickly but I'd like some comments for parsing steps where you show the input/output transformations. It makes it easier to follow and debug in future.

(This applies in a few places.)

src/dep_parser.c
```diff
@@ -0,0 +1,701 @@
...
+
+static PyObject *build_atom_obj(const AtomInfo *a, const char *tok, int tok_len)
+{
+    PyObject *py_str = PyUnicode_FromStringAndSize(tok, tok_len);
+    if (!py_str)
+        return NULL;
+
+    /* cp = "cat/pkg" */
+    char cp_buf[512];
+    int  cp_len = a->cat_len + 1 + a->pkg_len;
```
thesamesam: Are `cat_len` and `pkg_len` bounded a while before this point?

src/dep_parser.c
```diff
@@ -0,0 +1,701 @@
...
+    if (cp_len >= (int)sizeof(cp_buf)) {
+        PyErr_SetString(PyExc_ValueError, "atom string too long");
+        return NULL;
+    }
+
+    memcpy(cp_buf, a->cat, a->cat_len);
+    cp_buf[a->cat_len] = '/';
+    memcpy(cp_buf + a->cat_len + 1, a->pkg, a->pkg_len);
+
+    PyObject *py_cp = PyUnicode_FromStringAndSize(cp_buf, cp_len);
```
thesamesam: ~`PyUnicode_FromStringAndSize` takes a `const char *` but we're operating on `cp_buf` with `memcpy`. Where does the nul-terminator come from? Is it always preserved?~

(Nevermind: https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_FromStringAndSize documents its input buffer differently from https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_FromString.)

src/dep_parser.c
```diff
@@ -0,0 +1,701 @@
...
+        if (in_set < 0)
+            return -1;
+    }
+    return is_neg ? !in_set : in_set;
+}
+
+static int dep_parse(const char *s, Py_ssize_t n, const char **err,
+                     PyObject *result, UseCtx *use)
+{
+    P p = { s, s + n, NULL };
```
thesamesam: Can we have more descriptive struct names please?

src/dep_parser.c
```diff
@@ -0,0 +1,701 @@
...
+        if (err)
+            *err = p.err ? p.err : "parse error";
+        return 0;
+    }
+
+    py_ctx_free(&ctx);
+
+    skip_whitespace(&p);
+
+    if (p.p < p.pe) {
```
thesamesam: Need more descriptive member names here I think?

src/dep_parser.c
```diff
@@ -0,0 +1,701 @@
...
+    if (!scan_atom(&p, &info) || p.p != p.pe) {
+        PyErr_SetString(PyExc_ValueError, p.err ? p.err : "invalid atom");
+        return NULL;
+    }
+    return build_atom_obj(&info, s, (int)n);
+}
+
+static PyMethodDef methods[] = {
+    {
+        "parse",
```
thesamesam: Designated initialisers here please. You can figure out the meaning easily but still.

src/dep_parser_core.c
```diff
@@ -0,0 +1,608 @@
...
+ * conditionals are reported inactive so their bodies are skipped too. */
+static int skip_on_atom(void *ctx, const char *start, int len, const AtomInfo *info)
+{
+    (void)ctx; (void)start; (void)len; (void)info;
+    return 1;
+}
+
+static int skip_on_group_start(void *ctx, const char *op, int op_len)
+{
+    (void)ctx; (void)op; (void)op_len;
```
thesamesam: Just use `attribute((unused))` (or the GNU/C23-style form of that attribute). But what's the point of this?

thesamesam (commented):
Some initial quick thoughts. Main thing so far  is I'd like to see more comments when parsing.

thesamesam:
> Beyond the suite: a differential fuzz of 192000 combinations (nesting depths 0–5, varied USE lists, matchall on/off, EAPI 8/6/5/None, flat and non-flat) found no differences against the pure-Python path, and the same fuzz under ASan+UBSan is clean. The standalone test_parser binary runs 127 C-level cases and is wired into meson test.

If it's possible for you to do so, including some libfuzzer harness might be nice here. It doesn't need to be run by default in `meson test`.

mgorny:
Since this is new code, could you consider making it freethreading friendly as well?

mattst88:
> Since this is new code, could you consider making it freethreading friendly as well?

Yes, I'll look into how to do that.

mattst88:
> I mean, it's your code anyway so you can always allow it regardless of what license the portage copy identifies as, and omitting MIT here "for project policy" or similar is a meaningless restriction if you want to offer both licenses.
>
> Just not really sure I see what your motivation for doing so is. :P

So future contributions to these files are also under MIT.

mattst88:
> If it's possible for you to do so, including some libfuzzer harness might be nice here. It doesn't need to be run by default in meson test.

Done, behind `-Dfuzzing=true`.

mattst88:
> Since this is new code, could you consider making it freethreading friendly as well?

Done, I think. Seems like all that's needed here is to use `Py_MOD_GIL_NOT_USED`.

lib/portage/dep/__init__.py
mattst88: Yep, good call.

lib/portage/dep/__init__.py
mattst88: Oh, thanks. Much cleaner as a separate function.

lib/portage/dep/__init__.py
mattst88: Done.

src/dep_parser.c
mattst88: Changed to `interned`. Sorry for being lazy :)

src/dep_parser.c
mattst88: Done. Let me know if it still needs more.

src/dep_parser.c
mattst88: Nope! (Oops)

Fixed by removing the limit.

src/dep_parser.c
mattst88: Done: `DepScanner`, `UseContext`, `PyParseContext`, `PyGroupFrame`.

src/dep_parser.c
mattst88: Done: `p`/`pe` -> `cur`/`end`.

src/dep_parser.c
mattst88: Good call. Done.

src/dep_parser_core.c
mattst88: These no-op functions defined in `src/dep_parser_core.c` implement the interface required by `DepVisitor`. Since they're no-ops, they don't use the function arguments. With `-Wextra`, we get `-Wunused-parameter` warnings unless we use `attribute((unused))` or similar.

The no-op functions are used in cases where a USE conditional (e.g. `x? ( bogus )`) is inactive but the body still has to be validated. The alternative is we add a conditional to the hot path to check whether the visitor function is null. For good measure, I'll check how that affects the performance.

(Added and used `#define UNUSED __attribute__((unused))` to resolve this)

mgorny:
Yeah, if the code is thread-safe. The main guide is https://docs.python.org/3/howto/free-threading-extensions.html.

There are also modern module initialization methods, here's a dummy example: https://github.com/Quansight-Labs/stable-abi-testing/blob/2282ea96611c35a81f6c57476ef2875988b58b4e/common/limited.c#L17-L54

That said, multi-phase init seems to be supported since Python 3.5, so I don't think you need to keep the old init method.

mattst88:
Nice. Added a commit that addresses that.

mgorny:
Nice, thanks. Sorry I'm only able to give general comments about random PEPs :-).

lib/portage/tests/dep/test_c_parser.py
```diff
@@ -0,0 +1,1117 @@
...
+                "_slot_operator",
+                "_repo",
+                "_build_id",
+                "_extended_syntax",
+            ):
+                self.assertEqual(getattr(py, attr), getattr(c, attr), f"{s!r}: {attr}")
+            self.assertEqual(str(py._use or ""), str(c._use or ""), f"{s!r}: use")
+            self.assertEqual(str(py), str(c))
+
+    def test_valid_atoms(self):
```
thesamesam, line 967: https://github.com/pkgcore/pkgcore/blob/master/tests/ebuild/test_cpv.py has some good other cases to check too.

lib/portage/tests/dep/test_c_parser.py
```diff
@@ -0,0 +1,1117 @@
...
+            self.skipTest("_parser extension not available")
+
+    def _reduce(self, depstr, eapi, use_c):
+        _dep_mod._c_dep_parser = _orig_c_dep_parser if use_c else None
+        try:
+            return use_reduce(
+                depstr, token_class=Atom, matchall=True, eapi=eapi, uselist=[]
+            )
+        finally:
+            _dep_mod._c_dep_parser = _orig_c_dep_parser
```
thesamesam: I guess this is fine for xdist because all toggling for `use_c` is done on the level of a subtest.

thesamesam (commented):
Thanks for addressing those! Just 2 small comments from going over it again. I'll need some time to go over it properly, but another question occurred to me: could you try look at the coverage attained for the C parts by the tests added?

lib/portage/tests/dep/test_c_parser.py
mattst88: Nice, thanks. Added.

mattst88:
> Thanks for addressing those! Just 2 small comments from going over it again. I'll need some time to go over it properly, but another question occurred to me: could you try look at the coverage attained for the C parts by the tests added?

Measured with a coverage build: `meson setup -Db_coverage=true`, ran `pytest lib/portage/tests/dep` (230 tests + 432 subtests), then the full suite.

The new C files have a coverage of ~93.8%. The remaining bits are error conditions. E.g. 5 `return 0` when a visitor fails. 38 are `PyErr_NoMemory()` / `return null`. I'm not sure it's feasible to test the out of memory conditions, and likely not possible without a separate build flag.

I'll see about any low hanging fruit for improving coverage.

mattst88:
Oh, I think we can do the trick where we run a sweep, causing each memory allocation to fail in isolation. Let me give that a try.

mattst88:
Added a patch that implements the memory failure sweep I mentioned. Here's the coverage now.

|         File          |     Lines     | Branches taken |
| --------------------- | ------------- | -------------- |
| src/dep_parser_core.c | 99.70% of 338 | 91.18% of 408  |
| src/dep_parser.c      | 97.97% of 345 | 85.45% of 378  |
| src/dep_atom.h        | 100% of 53    | 100% of 10     |
| src/dep_parser_core.h | 100% of 2     | 87.50% of 8    |
| total                 | 98.92% of 738 | 88.56% of 804  |

That seems pretty great to me.

thesamesam (approved):
I think I'm happy but I'd like @zmedico and any others to comment if they can.

Thanks a lot!

src/dep_parser_core.c
```diff
@@ -0,0 +1,661 @@
...
+        return 1;
+    }
+    return 0;
+}
+
+/* PMS 8.3.3: the text after ':' -- a slot, an optional /sub-slot, and an
+ * optional '=' operator, or a bare ':=' / ':*'.
+ *
+ *   "0"     "myslot"   "0/53"   "0="   "0/53="   "="   "*"   -> consumed
+ *   "/slot"  "-slot"   "0/="    "0/*"  "0=/53"              -> rejected
```
thesamesam, line 138: Could you silence GCC's `-Wcomment` here? It doesn't like the `/*` (a pragma will hopefully work).

mattst88:
@zmedico: Any comments?

src/dep_parser_core.c
mattst88: Oh, hah! Yes, will fix.

zmedico:
There are test failures with pypy:
https://github.com/gentoo/portage/actions/runs/31047172949/job/92445419806?pr=1639

SUBFAILED(s="'sys-apps/portage'") ../../home/runner/work/portage/portage/lib/portage/tests/dep/test_c_parser.py::TestCParserAllocFailure::test_atom_fast_path - AttributeError: module '_testcapi' has no attribute 'set_nomemory'

mattst88:
> There are test failures with pypy:
https://github.com/gentoo/portage/actions/runs/31047172949/job/92445419806?pr=1639

Oops. Thanks. Should be fixed now.

zmedico (approved):

PR #1641, opened 2026-08-03
man/ebuild: document 'fetchall' command

Flowdalic:
The 'fetchall' command was previously not mentioned in the ebuild(1) man page. Fix this by documenting it.

### DIFF
--- a/man/ebuild.1
+++ b/man/ebuild.1
@@ -66,6 +66,9 @@ just\-downloaded sources' checksums don't match those recorded
 in files/digest\-[package]\-[version\-rev], a warning is printed
 and ebuild exits with an error code of 1.
 .TP
+.BR fetchall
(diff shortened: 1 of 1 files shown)


zmedico (approved):

man/ebuild.1
```diff
@@ -66,6 +66,9 @@ just\-downloaded sources' checksums don't match those recorded
 in files/digest\-[package]\-[version\-rev], a warning is printed
 and ebuild exits with an error code of 1.
 .TP
+.BR fetchall
+Like \fBfetch\fR, but fetches all URIs (even ones invalid due to USE conditionals).
```
eli-schwartz, line 70: I don't think invalid was the word you were looking for.

Flowdalic: Fair point, happy about suggestions on how to improve. Or even better, a patch.

thesamesam: Irrelevant maybe?

PR #1642, opened 2026-08-03
Make "emerge --status" error out if not FEATURES=observability

Flowdalic:
Make "emerge --status" error out with a helpful message if FEATURES=observability is not enabled


mattst88:
This is a good idea, but the wrong way of handling it. We shouldn't check the reader's `FEATURES`, but instead check whether there are snapshots we can read. E.g. consider `make.conf`'s `FEATURES` doesn't contain `observability`, but we're running `FEATURES=observability emerge ...`.

How about https://github.com/gentoo/portage/pull/1643 instead? (which also fixes `portageq`)

Flowdalic:
> How about #1643 instead? (which also fixes `portageq`)

wfm, closing

PR #1644, opened 2026-08-05
emerge: Don't report unsatisfied unrestricted downloads as restricted

chewi:
Otherwise you get output like this when the restricted downloads are already present but an unrestricted download is not.

```
$ USE=gog emerge -pv unreal-tournament-data

[ebuild   RF   ] games-fps/unreal-tournament-data-436_p4::gentoo  USE="gog* -bonuspacks -cdinstall doc" L10N="es fr it" 11,005 KiB

Total: 1 package (1 reinstall), Size of downloads: 11,005 KiB
Fetch Restriction: 1 package (1 unsatisfied)

Fetch instructions for games-fps/unreal-tournament-data-436_p4:
 * Please download setup_ut_goty_2.0.0.5.exe from GOG and move it
 * to your distfiles directory. Note that you can no longer buy this
 * game from GOG, so disable USE="gog" to use OldUnreal's free ISO if
 * you don't already own the game.
```

Now it looks like this.

```

[ebuild   Rf   ] games-fps/unreal-tournament-data-436_p4::gentoo  USE="gog* -bonuspacks -cdinstall doc" L10N="es fr it" 11,005 KiB

Total: 1 package (1 reinstall), Size of downloads: 11,005 KiB
Fetch Restriction: 1 package
```

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -65,6 +65,9 @@ Bug fixes:
 * Set a non-zero window size for newly allocated ptys, so build processes
   can more reliably query their terminal dimensions (bug #978222).
 
+* emerge: Don't report that there are unsatisfied fetch restricted downloads
+  when all the remaining unsatisfied downloads are not restricted (bug #830763).
+
 Cleanups:
 
 * Atom: De-stringify, add type annotations, use proper attributes at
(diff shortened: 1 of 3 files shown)


thesamesam (commented):
Does this fix https://bugs.gentoo.org/830763?

chewi:
> Does this fix https://bugs.gentoo.org/830763?

Why, yes it does. :slightly_smiling_face: Amended accordingly.

thesamesam (approved):
Sorry, thought I'd approved this. Thanks!

zmedico (approved):

PR #1647, opened 2026-08-09
tests/resolver: give binhost packages a fixed mtime

mattst88:
ResolverPlayground creates the same binary packages twice when a test uses binrepos: once in PKGDIR and once in the remote binhost directory. A binary package instance is keyed on (cpv, build_id, file_size, build_time, mtime), and the playground sets BUILD_TIME to 0 with no BUILD_ID, so two copies of one package differ only by file mtime.

bintree._populate_remote_repo() drops a remote package whose instance key already exists locally. Whether that happened depended on whether the PKGDIR copy and the binhost copy of a given package were written in the same wall-clock second: usually they were not and both instances survived, but under load the two writes could land in the same second, the remote instance was discarded as a duplicate, and the affected package resolved to [binary] where the test expected [binary,remote].

This showed up as intermittent failures in test_binpackage_selection, one package at a time and a different package each occurrence -- about 2 in 8 runs of that file on a loaded machine, rarer in a full suite run. It reproduces deterministically by forcing the two mtimes to match with os.utime.

Give the binhost copies a fixed mtime well in the past instead. PKGDIR copies always carry the current time, so the two instance keys can never collide, whatever the timing.

10 consecutive runs of test_binpackage_selection pass; full resolver suite: 230 passed, 7 xfailed, 222 subtests.

### DIFF
--- a/lib/portage/tests/resolver/ResolverPlayground.py
+++ b/lib/portage/tests/resolver/ResolverPlayground.py
@@ -43,6 +43,11 @@ from portage.tests import cnf_path
 from portage.util import ensure_dirs, normalize_path
 from portage.versions import catsplit
 
+# Fixed mtime for binary packages created in a remote binhost, so that
+# they never collide with the local PKGDIR copies of the same packages
+# (see _create_binpkgs).
+_BINREPO_MTIME = 1000000000
+
 
 def _combine_repo_config(conf, lines):
     merged = lines.copy()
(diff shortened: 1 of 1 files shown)


thesamesam:
cc @jeth-ro

jeth-ro:
Oops.

I'm pretty sure I *did* actually end up digging through all that bintree _pkg_instance stuff while trying to figure out why `testUsePkgExcludeMultiRepo()` and similar tests were failing github CI despite working locally (unrelated). So certainly won't claim ignorance of the mechanics of this but I obviously didn't see this angle on it back then, sorry.

This reproduces for me if matching filesystem times are forced with os.utime() as described, but in practice I've never once hit it by chance. Am usually doing this on slower hardware which can insert anywhere from 2-10 seconds between binpkgs and binrepos in setting up ResolverPlayground, so I guess I wouldn't expect to.

I can confirm that all of `test_binpackage_selection.py` was written with the assumption that pkgdir would not collide with binrepos in this way, which is to say I see no issue with the forced mtime fix as a transparent default either. I'm also pretty sure that particular test module is currently the only use of this ResolverPlayground feature.

PR #1648, opened 2026-08-09
tests: shut down gpg-agent before removing the temporary GNUPGHOME

mattst88:
Signing spawns a gpg-agent and scdaemon for the temporary home directory that the session fixture creates. They daemonize, so removing the directory leaves them running with a home directory that no longer exists, and a test run aborted before the fixture teardown leaks the directory too. Kill them with "gpgconf --kill all" before removing it.

Also move the mkdtemp() call out of the try block, since the finally block already assumes gpg_path is bound.

### DIFF
--- a/lib/portage/tests/conftest.py
+++ b/lib/portage/tests/conftest.py
@@ -9,6 +9,7 @@ import os.path as osp
 import pwd
 import shutil
 import signal
+import subprocess
 import sys
 import tempfile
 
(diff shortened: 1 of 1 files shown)


thesamesam (commented):
On mobile so forgive the question: does this duplicate some existing cleanup we do in some tests or complement it (wasn't running for those tests before)?

Also, thanks, I've ended up with many leftover processes and it's been annoying me..

mattst88:
Complements. Nothing in the test suite shut the agent down before. The only `gpgconf --kill` in the tree is in `bin/emerge-webrsync`, for its own ephemeral `GNUPGHOME`, and `GPG.stop()` in `lib/portage/gpg.py` only stops the keepalive thread. The session fixture's finally just rmtree'd the temp homedir, leaving the gpg-agent and scdaemon it spawned running against a directory that no longer exists, which apparently kills them (most of the time?)

thesamesam (approved):
OK, thanks. Maybe we want to make `stop` do something better (but this requires some careful checking as we don't want to interfere with anything the user has running, possibly specifically for Portage's benefit, like an already-unlocked signing keyring) or have a `GPG` context manager but you don't need to do that here.

PR #1649, opened 2026-08-09
vartree: replace vdb_metadata.pickle with a per-package metadata file

mattst88:
The VDB stores one file per metadata field per installed package, so reading
every field costs ~24 `openat` per package, multiplied across the whole
installed set on each resolve.

This adds a supplemental per-package `metadata` file (`KEY=value` per line)
that `_aux_get()` reads instead, and removes the `vdb_metadata.pickle` /
`vdb_metadata_delta.json` cross-session cache and `VdbMetadataDelta` along
with it. Per-package state needs no delta bookkeeping, no whole-file rewrite,
and no lock.

Measured on this machine: 1742 installed packages, 23 keys, best of 5
full-VDB passes through `_aux_get()`, syscalls via `strace -f -c`.

|                         | time     | openat | syscalls |
|-------------------------|----------|--------|----------|
| master                  | 306.7 ms |  41008 |   203022 |
| this series             |  46.2 ms |   2683 |    20123 |

6.6x faster, 15.3x fewer `openat`, 10.1x fewer syscalls overall: one open per
package also means one read and one fstat instead of ~24 of each.

Whole process via hyperfine (10 runs, no shell): 527.7 ms ± 9.3 to
268.1 ms ± 16.9.

The cross-session case is the one `vdb_metadata.pickle` existed to serve, and
the metadata file beats it with no cross-session state to invalidate: a fresh
process reading every package takes 137.6 ms against a warm pickle on master,
62.2 ms on this series.

**Disk cost.** 4.30 MiB of metadata files across 1742 packages, ~2.6 KiB per
package. On this machine (ZFS, compressed) `/var/db/pkg` grows 1.7 MiB; on a
4K-block filesystem the worst case is 7.21 MiB. `CONTENTS` staying out of the
file is what keeps this small.

**Trust model.** The file is used only when both hold: its `#format=` matches
`_METADATA_FILE_FORMAT_VERSION`, and its recorded `#dir_mtime=` matches the
package directory. The version check makes it a complete snapshot, so a field
absent from it is genuinely empty and not merely unknown. The mtime check is
the same freshness signal `vdb_metadata.pickle` validated against, now
per-package instead of one global file. A file failing either check is
ignored and the caller falls back to the individual per-field files, which
are kept intact for quickpkg, xpak, bintree and external consumers such as
portage-utils and pkgcore. `emaint vdb` populates existing VDBs and upgrades
files left by an older format.

Only single-line fields go in the file. `CONTENTS` and bare `NEEDED` are
excluded by name, and dotted names (`NEEDED.ELF.2`, `environment.bz2`,
`*.ebuild`) fail the name pattern by construction; all are read from their own
files as before.

`_mtime_` is unchanged for callers: `aux_get()` sourced it from the pickle on
a warm cache, so `_aux_get()` now returns the float `st_mtime` instead of the
truncated int a cache miss used to yield. Verified field-for-field across
every installed package and 24 keys against a warm pickle: no differences.

**Known limitation, unchanged from the pickle:** rewriting a field file in
place does not change the directory mtime and so goes unnoticed. Portage has
always used dir mtime as the VDB freshness signal, and this series does not
narrow that.

13 commits, including unit tests for the metadata helpers and the
version/mtime guards.

Bug: https://bugs.gentoo.org/321317


mattst88:
This was something I thought about a few years ago: https://public-inbox.gentoo.org/gentoo-portage-dev/faa2dab9-f467-f81a-bd27-ab712ed7ed7f@gentoo.org/T/

I decided against using a more complex format (e.g. JSON, YAML, TOML, tar, etc) for a few reasons:

- Most of the space savings were from nesting the `CONTENTS` which allowed not repeating ourselves, but `CONTENTS` really does make sense as a separate file.
- Flat files make grepping trivial
- some others that I forgot already?

thesamesam:
Have you measured how much space the new file takes (across your whole VDB on say your daily driver)? I imagine without `CONTENTS`, it is insignificant, fortunately.

bin/vdb-benchmark
```diff
@@ -0,0 +1,180 @@
...
+from portage.const import VDB_PATH
+from portage.dbapi.vartree import _METADATA_FILE
+
+
+def _count_metadata_files(dbroot):
+    """Return (with_metadata, without_metadata) package counts."""
+    with_meta = 0
+    without_meta = 0
+    try:
+        for cat in _os.listdir(dbroot):
```
thesamesam: Not that it really matters here but I think scandir will be faster and let you avoid a stat call for `isdir`.

bin/vdb-benchmark
```diff
@@ -0,0 +1,180 @@
...
+        for cpv in cpvs:
+            vardb._aux_get(cpv, keys)
+        durations.append(time.perf_counter() - t0)
+    return cpvs, durations
+
+
+def _strace_open_count(script_body):
+    """Run script_body via strace and return the openat() call count."""
+    strace_cmd = [
+        "strace",
```
thesamesam, line 80: Oh black..

bin/vdb-benchmark
```diff
@@ -0,0 +1,180 @@
...
+        result = subprocess.run(
+            strace_cmd,
+            capture_output=True,
+            text=True,
+        )
+    except FileNotFoundError:
+        return None
+
+    # strace -c summary goes to stderr; look for the openat line.
+    for line in result.stderr.splitlines():
```
thesamesam, line 99: I'm a little surprised strace doesn't support JSON output or something!

bin/vdb-benchmark
```diff
@@ -0,0 +1,180 @@
...
+    best = min(durations)
+    avg = sum(durations) / len(durations)
+    print(f"Packages:    {npkgs}")
+    print(f"Best run:    {best:.3f}s  ({best / npkgs * 1000:.2f} ms/pkg)")
+    print(f"Average:     {avg:.3f}s  ({avg / npkgs * 1000:.2f} ms/pkg)")
+
+    if opts.strace:
+        print("\nCounting openat() syscalls via strace (single pass)…")
+        # Build a self-contained script for strace to execute.
+        script = f"""\
```
thesamesam: I mention this just because you may find it interesting, not because you need to use it here: you can use `textwrap.dedent` for cases like this to make it prettier.

lib/portage/emaint/modules/vdb/__init__.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2025 Gentoo Authors
```
thesamesam: 2026, but I guess 2025-2026 if you started work on it last year.

lib/portage/emaint/modules/vdb/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+            "sourcefile": "vdb",
+            "class": "VdbMetadata",
+            "description": doc,
+            "functions": ["check", "fix", "remove"],
+            "func_desc": {
+                "delete_individual_files": {
+                    "long": "--delete-individual-files",
+                    "help": "(fix only): also remove per-field files after writing "
+                    "the metadata file. WARNING: breaks tools that read "
+                    "individual VDB files directly (pkgcore, paludis, shell "
```
thesamesam: portage-utils is probably more relevant than paludis here

lib/portage/emaint/modules/vdb/vdb.py
```diff
@@ -0,0 +1,120 @@
...
+    except OSError:
+        return False
+
+
+def _iter_pkg_dirs(dbroot):
+    try:
+        categories = os.listdir(dbroot)
+    except OSError:
+        return
+    for cat in sorted(categories):
```
thesamesam: scandir here please, but also, I thought we had a function for this already

lib/portage/emaint/modules/vdb/vdb.py
```diff
@@ -0,0 +1,120 @@
...
+        except OSError:
+            continue
+        for pkg in sorted(pkgs):
+            pkgdir = os.path.join(catdir, pkg)
+            if os.path.isdir(pkgdir):
+                yield cat, pkg, pkgdir
+
+
+class VdbMetadata:
+    short_desc = "Manage VDB consolidated metadata files"
```
thesamesam: I think swap "VDB" and "consolidated" here?

lib/portage/tests/resolver/ResolverPlayground.py
```diff
@@ -497,9 +498,18 @@ def _create_installed(self, installed):
                 )
 
             metadata["repository"] = repo
+            metadata_kv = {}
             for k, v in metadata.items():
+                # Write the individual file for every field, the way a real
+                # merge does. The metadata file is a read optimization layered
+                # on top of them, not a replacement: quickpkg, xpak and other
```
thesamesam: "Both Portage (e.g. quickpkg, xpak binpkgs) consumers and non-Portage consumers (e.g. pkgcore) still read the per-field fields directly."

thesamesam (commented):
Please tag https://bugs.gentoo.org/321317 as it's relevant to that bug and may even make the (possible) need for it obsolete.

Please add `Signed-off-by`.

I would be interested to see the total syscall count before+after too.

I've left some quick initial comments, not done a proper pass yet.

lib/portage/dbapi/vartree.py
```diff
@@ -1067,23 +1067,30 @@ def aux_update(self, cpv, values):
...
-                try:
-                    existing = _read_metadata_file(metadata_path)
-                    if k in existing:
-                        del existing[k]
-                        _write_metadata_file(self.getpath(cpv), existing)
-                except OSError:
-                    pass
+                if k != "CONTENTS" and _METADATA_FIELD_RE.match(k) is not None:
+                    metadata_deletes.append(k)
+        if metadata_updates or metadata_deletes:
```
thesamesam: Ah, nice! I did something like this in 3b975f9b28d1aa6e40c93431086669ad23f6f460 and had a feeling there may be other places to update too.

lib/portage/dbapi/vartree.py
```diff
@@ -847,12 +914,35 @@ def _aux_get(self, mycpv, wants, st=None):
...
         env_keys = []
         for x in wants:
             if x == "_mtime_":
                 results[x] = st[stat.ST_MTIME]
                 continue
+
+            # Only fields actually present in the metadata file may be served
+            # from it. A field missing there is not known to be empty: the
+            # metadata file can predate the field (written by an older portage,
+            # or by a merge that set it afterwards), while the individual file
```
thesamesam: I'm not following what the "predate the field (...)" part is referring to.

The field may be missing because an older Portage was used even after a metadata file was generated for a package, that's fine and we allow that (disallowing it would be nasty for testing for regressions, maybe we can make this an assumption when the feature has been around a while, but it would imply we can never remove a key from it, so..)

What does the "merge that set it afterwards" part mean?

lib/portage/dbapi/vartree.py
```diff
@@ -62,14 +61,147 @@
...
+
+_METADATA_FILE = "metadata"
+# Matches all-caps field names (and lowercase "repository") that belong in the
+# consolidated metadata file. Files with dots (NEEDED.ELF.2, environment.bz2,
+# *.ebuild) and CONTENTS (multi-line) are intentionally excluded.
+_METADATA_FIELD_RE = re.compile(r"^([A-Z][A-Z0-9_]*|repository)$")
+
+
+_METADATA_FILE_FORMAT_VERSION = 1
+
```
thesamesam: I'd say drop the newlines here but up to you.

lib/portage/dbapi/vartree.py
```diff
@@ -62,14 +61,147 @@
...
 
 # Made global to fix importing on Python version upgrade:
 # https://bugs.gentoo.org/970375
 from ._SyncfsProcess import SyncfsProcess
-from ._VdbMetadataDelta import VdbMetadataDelta
+
+_METADATA_FILE = "metadata"
+# Matches all-caps field names (and lowercase "repository") that belong in the
+# consolidated metadata file. Files with dots (NEEDED.ELF.2, environment.bz2,
+# *.ebuild) and CONTENTS (multi-line) are intentionally excluded.
```
thesamesam: Why are those with dots excluded? It seems odd that we can have `NEEDED` in there but not `NEEDED.ELF.2`. I assume it's because it makes field name parsing uglier and it's not worth it because of where we use `NEEDED.ELF.2`?

lib/portage/dbapi/vartree.py
```diff
@@ -62,14 +61,147 @@
...
+                    try:
+                        dir_mtime = int(line[len(_METADATA_DIR_MTIME_PREFIX) :])
+                    except ValueError:
+                        return None
+                continue
+            if "=" not in line:
+                continue
+            k, v = line.split("=", 1)
+            result[k] = v
+    if version != _METADATA_FILE_FORMAT_VERSION or dir_mtime is None:
```
thesamesam: You could return early in the loop rather than iterating when it's known to be useless.

lib/portage/dbapi/vartree.py
```diff
@@ -847,12 +840,40 @@ def _aux_get(self, mycpv, wants, st=None):
...
                 continue
+
+            # _read_metadata_file only returns a dict for a file whose format
+            # version matches, and such a file is a complete snapshot of the
+            # matching fields.  A field missing from it therefore had no
+            # individual file either, so serve it as empty instead of paying
+            # an open() that would just fail.
+            if (
+                metadata_data is not None
+                and x != "CONTENTS"
```
thesamesam: Why don't you have to exempt `NEEDED.ELF.2` and `repository` here (and elsewhere)?

lib/portage/tests/dbapi/test_vdb_metadata.py
```diff
@@ -0,0 +1,244 @@
...
+    def test_keys_sorted_in_file(self):
+        data = {"SLOT": "0", "EAPI": "8", "USE": "foo"}
+        _write_metadata_file(self._tmpdir, data)
+        path = os.path.join(self._tmpdir, _METADATA_FILE)
+        with open(path) as f:
+            lines = [l.rstrip("\n") for l in f if not l.startswith("#")]
+        keys = [l.split("=", 1)[0] for l in lines if "=" in l]
+        self.assertEqual(keys, sorted(keys))
+
+    def test_value_with_equals_sign(self):
```
thesamesam, line 169: Perhaps test these also:
* `https://example.com/#anchor` (because `#`)
* `http://127.0.0.1/?a=1#anchor` (because dots)

bin/meson.build
```diff
@@ -19,7 +19,9 @@ py_sbins = [
     'emaint',
     'env-update',
     'fixpackages',
-    'regenworld'
+
+    'regenworld',
+    'regenworld',
```
thesamesam: This looks off :)

lib/portage/emaint/modules/vdb/__init__.py
mattst88: Nope, just copy and paste. Fixed.

lib/portage/emaint/modules/vdb/vdb.py
mattst88: I guess it's a question of whether the "VDB" or the "metadata files" are consolidated :)

mattst88:
Thanks for the review. Pushed a rebase addressing all of it; every commit now has `Signed-off-by`, and the first one carries `Bug: https://bugs.gentoo.org/321317`.

> I would be interested to see the total syscall count before+after too.

One full `_aux_get()` pass over 1742 installed packages, 23 keys, `strace -f -c`, master vs. this series:

|          | master  | series |
|----------|---------|--------|
| total    | 203068  | 20092  |
| openat   |  41018  |  2689  |
| read     |  62434  |  4607  |
| fstat    |  31458  |  2538  |

10.1x fewer syscalls overall, not just opens: one open per package also means one read and one fstat instead of ~24 of each.

> Have you measured how much space the new file takes (across your whole VDB on say your daily driver)? I imagine without `CONTENTS`, it is insignificant, fortunately.

4.30 MiB across 1742 packages, ~2.6 KiB each. `/var/db/pkg` here grows by 1.7 MiB (ZFS, compressed); the worst case on a 4K-block filesystem is 7.21 MiB.

You're right that `CONTENTS` staying out is what keeps it small, and so does keeping bare `NEEDED` out, which was 28% of the file's bytes until your `NEEDED.ELF.2` question led me to drop it.

bin/meson.build
mattst88: Oops. Fixed.

lib/portage/tests/dbapi/test_vdb_metadata.py
mattst88: Thanks. Good idea.

lib/portage/dbapi/vartree.py
mattst88: You identified something important here, but maybe unintentionally :)

`NEEDED.ELF.2` is excluded because the regex doesn't match dots:

```python
_METADATA_FIELD_RE = re.compile(r"^([A-Z][A-Z0-9_]*|repository)$")
```

`repository` is the opposite case. It's matched deliberately, by that alternation. It's in `_aux_cache_keys`, so it's read for every package, and it's a single short line, so it belongs in the file. Nothing to exempt in either case, just for opposite reasons.

But the thing you spotted: bare `NEEDED`, the legacy pre-`NEEDED.ELF.2` file, has no dots, so it did match and was going in. It's line-oriented like `CONTENTS`, and nothing reads it, absent from `_aux_cache_keys`, and `_aux_cache_keys_re` requires a dot. It was 28% of the file's bytes serving no read at all. Oops. Fix coming up.

lib/portage/dbapi/vartree.py
mattst88: Great catch. Nope, this was a real problem. Fix incoming.

lib/portage/dbapi/vartree.py
mattst88: Done. Honestly surprised that of all the things black complains about this wasn't one.

lib/portage/dbapi/vartree.py
mattst88: Done fo rthe version check. We can't for the mtime because it's written last on purpose, so a file that is truncated by some interruption will be rejected. So we can't know the mtime until we've read the whole file.

lib/portage/dbapi/vartree.py
mattst88: Bad wording on my part. You're on commit 1, where the metadata file isn't yet trusted as a complete snapshot, so a missing field falls through to the per-field read. The comment justifies that fallback, naming two ways the file can be older than a field:

1. An older portage wrote it, with a field set lacking that key.
2. The field was written after the file was -- at that commit `aux_update()`/`setfile()` update the individual file without touching the metadata file.

"a merge that set it afterwards" was meant to be (2), and says it terribly. I meant "a write that happened after the metadata file was written". I'll reword it.

Both close later in the series: "batch metadata file writes in aux_update" syncs those two paths, and "gate the metadata file on its format version" makes a matching file a complete snapshot, dropping the x in metadata_data condition and this comment with it.

Your aside: an older portage after generation stays supported via `#dir_mtime=` It writes the individual file, that changes the directory mtime, snapshot rejected, fall back. The one hole is an in-place rewrite of an existing field file, which changes no directory entry -- same limitation the pickle had.

On never removing a key: that's what `_METADATA_FILE_FORMAT_VERSION` is for. Removing one means bumping it; existing files fail the check and get rewritten by the next merge or `emaint vdb --fix`.

lib/portage/emaint/modules/vdb/__init__.py
mattst88: True. Fixed.

bin/vdb-benchmark
mattst88: Oh yeah. Thank you!

bin/vdb-benchmark
mattst88: No kidding. It has `-X <format>` but the only options are raw, abbrev, and verbose. :(

bin/vdb-benchmark
mattst88: Done. Thanks!

lib/portage/emaint/modules/vdb/vdb.py
mattst88: Done. Thanks!

lib/portage/tests/resolver/ResolverPlayground.py
mattst88: Done.

thesamesam:
```
# emaint vdb

2639 packages in VDB
  0 have consolidated metadata file
  2639 are missing, stale, or use an older format
Run 'emaint vdb --fix' to populate missing metadata files.


```

Can we drop the newlines before/after the output?

lib/portage/emaint/modules/vdb/vdb.py
```diff
@@ -0,0 +1,102 @@
...
+            except Exception as e:
+                errors.append(f"{cpv}: {e}")
+
+        if errors:
+            return (False, errors)
+        return (True, None)
+
+    def remove(self, **kwargs):
+        """Remove consolidated metadata files from all VDB package directories."""
+        settings = kwargs.get("settings", getattr(portage, "settings", {}))
```
zmedico, line 94: Maybe remove should perform the reverse of _consolidate_to_metadata_file?

mattst88: Yeah, and it turned out to be a bug, not just an asymmetry. Thanks for catching it.

mattst88:
Pretty big rework, with one idea stolen from grobian's 'gtree' from portage-utils: drop derived and unread keys (they strip _md5_, reduce _eclasses_ hashes to bare names). Doing that uncovered a couple of real bugs.

lib/portage/dbapi/vartree.py
```diff
@@ -633,119 +855,22 @@ def findname(self, mycpv, myrepo=None):
...
-        if owners is None:
-            owners = {"base_names": {}, "version": self._owners_cache_version}
-            aux_cache["owners"] = owners
-
-        aux_cache["modified"] = set()
-        self._aux_cache_obj = aux_cache
+        self._aux_cache_obj = {
+            "packages": {},
+            "owners": {"base_names": {}},
+        }
```
zmedico, line 881: It's worth noting that without vdb_metadata.pickle, owners data does not persist. This unfortunately forces all CONTENTS to be read if we need to look up owners for some reason.

zmedico (approved):

lib/portage/emaint/modules/vdb/vdb.py
```diff
@@ -0,0 +1,111 @@
...
+        for cpv, pkgdir in _iter_pkg_dirs(settings):
+            # A current metadata file means there is nothing to populate, but
+            # --delete-individual-files may still have per-field files to
+            # remove, so only the plain populate path can skip the package.
+            if not delete_individual and _has_usable_metadata(pkgdir):
+                continue
+            try:
+                _consolidate_to_metadata_file(
+                    pkgdir, delete_individual=delete_individual
+                )
```
zmedico, line 79: It would be neat if _consolidate_to_metadata_file could detect a valid metadata file and skip in this case.

lib/portage/dbapi/vartree.py
mattst88: Right, owners data no longer persists. The callers are all one-shot queries, though: `portageq owners`, `emerge --unmerge`/`--deselect` given a file path rather than an atom, and `OwnerSet` for a set declared with `class = portage.sets.dbapi.OwnerSet`. Preserved-libs handling uses `LinkageMapELF.getOwners()`, and `equery belongs` walks CONTENTS itself in `gentoolkit.helpers.FileOwner`, so neither is affected. Nothing in the resolve or merge path calls `populate()`.

Cost is ~4s here for 1743 packages, and it's CPU in `getcontents()` parsing and the per-path md5, not I/O (reading all 88 MB of CONTENTS takes 51ms). If that's worth fixing it's fixable in `populate()` itself, which would help master too. I'd rather do that than reintroduce a global cache file.

lib/portage/emaint/modules/vdb/vdb.py
mattst88: Done. `_consolidate_to_metadata_file()` now returns early when the existing file still validates, so the check benefits every caller rather than just emaint, and `fix()` no longer needs its own pre-check.

`delete_individual=True` deliberately does not take the shortcut: a valid file says nothing about whether the per-field files are still there to remove. Tests cover the skip, the rebuild when the file is stale, and that case.

lib/portage/dbapi/vartree.py
thesamesam: Thanks for explaining. Happy now!

PR #1650, opened 2026-08-10
Add cgroup stats to "emerge --status" output

Flowdalic:
Best combined with https://github.com/gentoo/portage/pull/1651 (and vice versa). Split into two commits for better reviewability.

### DIFF
--- a/lib/_emerge/_observability.py
+++ b/lib/_emerge/_observability.py
@@ -33,6 +33,7 @@ from portage import os
 from portage.const import PORTAGE_RUN_PATH
 from portage.util import atomic_ofstream, ensure_dirs, writemsg_level
 from portage.util.futures import asyncio
+from portage.util.human_readable import bytes_to_human
 
 from _emerge.PackageMerge import PackageMerge as _PackageMerge
 
@@ -181,6 +182,27 @@ def _pid_alive(pid):
     return True
 
 
+def _format_cpu(val, task):
+    cpu_s = val / 1e6
+    elapsed = task.get("elapsed")
+    if elapsed is not None and elapsed > 0:
+        parallelism = cpu_s / elapsed
+        return f"{cpu_s:.2f}s ({parallelism:.2f}x)"
+    return f"{cpu_s:.2f}s"
+
+
+_RESOURCE_FIELDS = (
+    ("cpu_usec", "CPU", _format_cpu),
+    ("mem_current", "Mem", lambda v, t: bytes_to_human(v)),
+    ("mem_peak", "MaxMem", lambda v, t: bytes_to_human(v)),
+    ("mem_swap_current", "Swap", lambda v, t: bytes_to_human(v)),
+    ("mem_swap_peak", "MaxSwap", lambda v, t: bytes_to_human(v)),
+    ("mem_zswap_current", "ZSwap", lambda v, t: bytes_to_human(v)),
+    ("io_read_bytes", "I/O R", lambda v, t: bytes_to_human(v)),
+    ("io_write_bytes", "I/O W", lambda v, t: bytes_to_human(v)),
+)
+
+
 def format_snapshots(snapshots):
     """Render snapshots as a human-readable table."""
     if not snapshots:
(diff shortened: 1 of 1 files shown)


lib/_emerge/_observability.py
```diff
@@ -203,7 +205,47 @@ def format_snapshots(snapshots):
...
+                        f"ZSwap: {bytes_to_human(resources['mem_zswap_current'])}"
+                    )
+                if resources.get("io_read_bytes"):
+                    res_strs.append(
+                        f"I/O R: {bytes_to_human(resources['io_read_bytes'])}"
+                    )
+                if resources.get("io_write_bytes"):
+                    res_strs.append(
+                        f"I/O W: {bytes_to_human(resources['io_write_bytes'])}"
+                    )
```
mattst88: I think you can simplify all this with a table-drive approach:

```python
_RESOURCE_FIELDS = (
    ("mem_current", "Mem"),
    ("mem_peak", "MaxMem"),
    ("mem_swap_current", "Swap"),
    ("mem_swap_peak", "MaxSwap"),
    ("mem_zswap_current", "ZSwap"),
    ("io_read_bytes", "I/O R"),
    ("io_write_bytes", "I/O W"),
)
```

lib/_emerge/_observability.py
```diff
@@ -203,7 +205,47 @@ def format_snapshots(snapshots):
...
-            lines.append(f"  {task.get('cpv', '?'):<45} {phase:<10} {elapsed_str:>7}")
+            line = f"  {task.get('cpv', '?'):<45} {phase:<10} {elapsed_str:>7}"
+
+            resources = task.get("resources")
+            if resources:
+                res_strs = []
+                if resources.get("cpu_usec"):
+                    cpu_s = resources["cpu_usec"] / 1000000.0
+                    if elapsed is not None and elapsed > 0:
+                        parallelism = cpu_s / elapsed
```
mattst88: `parallelism = cpu_s / elapsed` breaks for merge-wait tasks. `elapsed` is intentionally frozen at build completion but `resources` comes from a live `cgroup.read_stats()`. So I think reported parallelism will just keep growing while the package is in the merge-wait/merge states.

I think you either want to freeze `cpu_usec` the same way, or only print parallelism when not `task["merge_wait"]`.

mattst88 (changes requested):

lib/_emerge/_observability.py
```diff
@@ -186,6 +186,8 @@ def format_snapshots(snapshots):
     if not snapshots:
         return "No emerge processes are currently running.\n"
 
+    from portage.util.human_readable import bytes_to_human
```
mattst88: I'd move this import to the top. The reasons (AFAIU) to import within a function are

1. it's very rarely needed (optimization)
2. avoiding circular imports

lib/_emerge/_observability.py
```diff
@@ -203,7 +205,47 @@ def format_snapshots(snapshots):
...
             elapsed_str = f"{max(0, int(elapsed))}s" if elapsed is not None else "-"
             phase = task.get("phase") or task.get("kind") or "-"
-            lines.append(f"  {task.get('cpv', '?'):<45} {phase:<10} {elapsed_str:>7}")
+            line = f"  {task.get('cpv', '?'):<45} {phase:<10} {elapsed_str:>7}"
+
+            resources = task.get("resources")
+            if resources:
+                res_strs = []
+                if resources.get("cpu_usec"):
+                    cpu_s = resources["cpu_usec"] / 1000000.0
```
mattst88: might want to `s/1000000.0/1e6/` to go with the `usec` suffix.

Or, maybe even better: add this as a per-field formatter to the table so you don't have to special case it at all.

lib/_emerge/_observability.py
```diff
@@ -203,7 +205,47 @@ def format_snapshots(snapshots):
...
+                    )
+                if resources.get("mem_zswap_current"):
+                    res_strs.append(
+                        f"ZSwap: {bytes_to_human(resources['mem_zswap_current'])}"
+                    )
+                if resources.get("io_read_bytes"):
+                    res_strs.append(
+                        f"I/O R: {bytes_to_human(resources['io_read_bytes'])}"
+                    )
+                if resources.get("io_write_bytes"):
```
mattst88: 0 IO bytes is actually signal of its own. Unsure whether dropping it is what we want?

lib/_emerge/_observability.py
Flowdalic: I've went with freezing the cgroup resource information at build completion.

mattst88 (approved):

Flowdalic:
Closing in favor of #1661

PR #1651, opened 2026-08-10
Make "emerge --status" output fresh

Flowdalic:
Best combined with https://github.com/gentoo/portage/pull/1650 (and vice versa). Split into two commits for better reviewability.

### DIFF
--- a/lib/_emerge/_observability.py
+++ b/lib/_emerge/_observability.py
@@ -414,6 +414,7 @@ class ObservabilityMonitor:
             )
 
     async def _client_connected(self, reader, writer):
+        self.update()
         if self._last_snapshot is not None:
             data = (json.dumps(self._last_snapshot, sort_keys=True) + "\n").encode(
                 "utf_8"
--- a/lib/_emerge/_observability.py
+++ b/lib/_emerge/_observability.py
@@ -152,8 +152,27 @@ def status_dir(eprefix=""):
 def read_snapshots(eprefix=""):
     """Read all live emerge status files; return a list of snapshot dicts."""
     import glob
+    import socket
 
     snapshots = []
+
+    # Try reading from sockets first to trigger a fresh snapshot
+    for path in sorted(glob.glob(os.path.join(status_dir(eprefix), "emerge-*.sock"))):
+        try:
+            with socket.socket(socket.AF_UNIX, socket.SOCK_STREAM) as s:
+                s.settimeout(0.5)
+                s.connect(path)
+                with s.makefile("r", encoding="utf_8") as f:
+                    line = f.readline()
+                    if line:
+                        snapshot = json.loads(line)
+                        pid = snapshot.get("emerge_pid")
+                        if isinstance(pid, int) and pid > 0 and _pid_alive(pid):
+                            snapshots.append(snapshot)
+        except (OSError, ValueError, socket.timeout, json.JSONDecodeError):
+            pass
+
+    # Fall back to json files, e.g., in case we run into the socket timeout above.
     for path in sorted(glob.glob(os.path.join(status_dir(eprefix), "emerge-*.json"))):
         try:
             with open(path, encoding="utf_8") as f:
@@ -163,7 +182,14 @@ def read_snapshots(eprefix=""):
         pid = snapshot.get("emerge_pid")
         if not isinstance(pid, int) or pid <= 0 or not _pid_alive(pid):
             continue
-        snapshots.append(snapshot)
+        # If we successfully read a live snapshot for this PID via the socket,
+        # skip the JSON fallback to avoid duplicating the same process in the output.
+        # Note: the socket connect above triggers an update() server-side which rewrites
+        # the JSON status file. So by the time this loop runs, the JSON is fresh too
+        # and either would do, but the socket snapshot is the one we know is current.
+        if not any(s.get("emerge_pid") == pid for s in snapshots):
+            snapshots.append(snapshot)
+
     return snapshots
 
 


lib/_emerge/_observability.py
```diff
@@ -152,8 +152,27 @@ def status_dir(eprefix=""):
...
 
     snapshots = []
+
+    # Try reading from sockets first to trigger a fresh snapshot
+    for path in sorted(glob.glob(os.path.join(status_dir(eprefix), "emerge-*.sock"))):
+        try:
+            with socket.socket(socket.AF_UNIX, socket.SOCK_STREAM) as s:
+                s.settimeout(0.5)
+                s.connect(path)
+                f = s.makefile("r", encoding="utf_8")
```
mattst88: `f` is never closed. Wrap it with `with`?

lib/_emerge/_observability.py
```diff
@@ -152,8 +152,27 @@ def status_dir(eprefix=""):
...
+                line = f.readline()
+                if line:
+                    snapshot = json.loads(line)
+                    pid = snapshot.get("emerge_pid")
+                    if isinstance(pid, int) and pid > 0 and _pid_alive(pid):
+                        snapshots.append(snapshot)
+        except Exception:
```
mattst88: Bare except Exception: pass will silently swallow real regressions here. The fallback loop below is specific except `OSError`, `ValueError`. Please match that, plus `socket.timeout` and `json.JSONDecodeError`.

lib/_emerge/_observability.py
```diff
@@ -163,7 +182,11 @@ def read_snapshots(eprefix=""):
         pid = snapshot.get("emerge_pid")
         if not isinstance(pid, int) or pid <= 0 or not _pid_alive(pid):
             continue
-        snapshots.append(snapshot)
+        # If we successfully read a live snapshot for this PID via the socket,
+        # skip the JSON fallback to avoid duplicating the same process in the output.
+        if not any(s.get("emerge_pid") == pid for s in snapshots):
+            snapshots.append(snapshot)
```
mattst88, line 191: Worth a word in the comment on why the socket entry wins rather than the JSON one: the connect above triggers update() server-side, which rewrites the status file, so by the time this loop runs the JSON is fresh too and either would do, but the socket snapshot is the one we know is current. Not a change request, just a note for whoever reads this in a year.

lib/_emerge/_observability.py
```diff
@@ -383,6 +406,7 @@ def _server_ready(self, future):
             )
 
     def _client_connected(self, reader, writer):
+        self.update()
```
mattst88, line 443: Okay, this is the important bit.

`.update()` is previously only called on phase changes, even though cgroup stats like CPU and memory continue to change during e.g. a long compile.

Refresh-on-connect is also the better of the two obvious fixes: a periodic timer would refresh for all consumers but would write /run forever whether or not anyone is watching, which is the churn `_min_write_latency` exists to bound. This costs nothing when nobody asks.

Please put that reasoning in the commit message. And to be explicit that I checked it: not passing `force=True` is correct. The rate limit only suppresses a refresh when the snapshot is already under a second old, i.e. when refreshing would buy nothing.

mattst88 (changes requested):

lib/_emerge/_observability.py
Flowdalic: I think I addressed all review remarks, but I am unsure if what I did in the commit message sufficiently addresses this comment. I did not mention a periodic timer, as we now seem to have one.

mattst88 (commented):
Socket-first is the right shape, but I don't think it delivers freshness yet.

`_client_connected` calls `self.update()`, which returns without publishing when `now - self._last_write < self._min_write_latency` (1.0s). The client then gets the previous `_last_snapshot` unchanged, still up to a second stale. It also makes the new comment in `read_snapshots` inaccurate: the connect doesn't reliably rewrite the JSON file.

`self.update(force=True)` fixes both. The rate limit is there to bound IO from bursty task events; a client connect is human-paced, and the socket is 0600 so only root can abuse it.

lib/_emerge/_observability.py
```diff
@@ -152,8 +152,27 @@ def status_dir(eprefix=""):
...
+                s.settimeout(0.5)
+                s.connect(path)
+                with s.makefile("r", encoding="utf_8") as f:
+                    line = f.readline()
+                    if line:
+                        snapshot = json.loads(line)
+                        pid = snapshot.get("emerge_pid")
+                        if isinstance(pid, int) and pid > 0 and _pid_alive(pid):
+                            snapshots.append(snapshot)
+        except (OSError, ValueError, socket.timeout, json.JSONDecodeError):
```
mattst88, line 172: Both extras are already covered: `socket.timeout` subclasses `OSError` (since 3.3), `JSONDecodeError` subclasses `ValueError`. `except (OSError, ValueError):` matches the JSON loop below.

lib/_emerge/_observability.py
```diff
@@ -152,8 +152,27 @@ def status_dir(eprefix=""):
 def read_snapshots(eprefix=""):
     """Read all live emerge status files; return a list of snapshot dicts."""
     import glob
+    import socket
 
     snapshots = []
+
+    # Try reading from sockets first to trigger a fresh snapshot
+    for path in sorted(glob.glob(os.path.join(status_dir(eprefix), "emerge-*.sock"))):
```
mattst88, line 160: - Timeouts are serial: worst case 0.5s per socket that exists but doesn't answer. Fine for one or two emerges, worth a comment.
- Sockets before JSON means the result is no longer path-ordered. Sorting by `emerge_pid` before returning keeps output stable regardless of which transport answered.
- Socket filename pid isn't checked against `snapshot["emerge_pid"]`. `_pid_alive` covers the important case; a mismatch means a stale socket on a recycled pid.

Flowdalic:
> Socket-first is the right shape, but I don't think it delivers freshness yet.

Well, it is bound to min_write_latency, which I think is fine (and I assume you did so as well since you wrote not using `Force=true` is ok). Personally, so staleness is acceptable. But we could certainly go with `Force=true` if that's what you now suggest?

mattst88:
Sorry, you're right, I blessed no-force earlier and then contradicted it without saying so. My fault.

My original reasoning was that the rate limit only suppresses a refresh when the snapshot is already under a second old, so refreshing buys nothing. On freshness alone that holds, and with _refresh_interval at 2s the worst case without any connect-time update is 2s, so this commit buys something either way.

What I got wrong is narrower than my last comment made it sound. The new comment in read_snapshots says the connect triggers a server-side update that rewrites the status file, so the JSON is fresh too. Under the rate limit that isn't guaranteed: if a task event published within the last second, update() returns early and neither the snapshot nor the file is rewritten.

Two ways to close it, and I'm fine with either:

1. force=True. The comment becomes true, the client gets a snapshot built at the moment it asked, and the cost is bounded: the socket is 0600, connects are human-paced, and one extra build_snapshot plus status file write per connect is noise next to a compile.
2. Keep the rate limit, reword the comment to claim only what holds: the socket snapshot is at most _min_write_latency old, and the connect may not have rewritten the JSON file.

I lean toward 1 because "client sees state at the moment it asked" is easier to reason about than a staleness bound, but 2 is defensible and your call.

Flowdalic:
Closing in favor of #1661

PR #1652, opened 2026-08-10
Scheduler: add parallelism to cgroup log message

Flowdalic:
Calculate the average parallelism of the build by dividing the cgroup CPU time by the wall-clock time and append it to the resource usage log message in emerge.log.

### DIFF
--- a/lib/_emerge/Scheduler.py
+++ b/lib/_emerge/Scheduler.py
@@ -413,9 +413,19 @@ class Scheduler(PollScheduler):
         cpv = build.pkg.cpv
         stats = self._cgroup.read_stats(cpv)
         if stats:
+            elapsed = None
+            times = self._observability._build_times.get(str(cpv))
+            if times is not None and times[0] is not None and times[1] is not None:
+                elapsed = times[1] - times[0]
+
             parts = []
             if "cpu_usec" in stats:
-                parts.append(f"cpu {stats['cpu_usec'] / 1e6:.1f}s")
+                cpu_s = stats["cpu_usec"] / 1e6
+                if elapsed is not None and elapsed > 0:
+                    parallelism = cpu_s / elapsed
+                    parts.append(f"cpu {cpu_s:.1f}s ({parallelism:.2f}x)")
+                else:
+                    parts.append(f"cpu {cpu_s:.1f}s")
             if "mem_peak" in stats:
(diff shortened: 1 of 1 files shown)


thesamesam (commented):
I don't feel strongly either way. If @mattst88 is happy then OK.

mattst88 (commented):
Ordering is fine: `_build_exit` calls `note_task_finished(build)` before `_cgroup_finish`, so `times[1]` is set when you read it. Two issues above that.

This is dead unless `FEATURES="observability"` is also set. `cgroup` and `observability` are independent (`Scheduler.py:233-238`), but `_build_times` is only populated by `note_task_started`/`note_task_finished`, which return early when the monitor is disabled. Cgroup-only users get the old `cpu 412.3s` with no parallelism and no hint why. The wall time is cheap to record: track it in the Scheduler, or populate `_build_times` unconditionally and gate only publishing on `enabled`.

Reaching into `_observability._build_times` also pins this to a private dict's positional layout, which #1650 changes to three elements. A `build_elapsed(cpv)` accessor decouples both.

Whichever of this and #1650 lands second: the cpu/elapsed -> parallelism math is duplicated here and in `_observability._format_cpu`.

lib/_emerge/Scheduler.py
```diff
@@ -413,9 +413,19 @@ def _cgroup_finish(self, build):
         cpv = build.pkg.cpv
         stats = self._cgroup.read_stats(cpv)
         if stats:
+            elapsed = None
+            times = self._observability._build_times.get(str(cpv))
```
mattst88, line 417: `_build_times` is empty whenever `FEATURES="observability"` is off, so `elapsed` is always `None` for cgroup-only users and `(N.NNx)` never appears. The private-attribute access also pins this to the current list layout, which #1650 changes. A `self._observability.build_elapsed(str(cpv))` accessor, backed by timing that doesn't require the monitor, solves both.

Flowdalic:
> Whichever of this and #1650 lands second: the cpu/elapsed -> parallelism math is duplicated here and in `_observability._format_cpu`.

It is not only the parallelism math, but the whole cgroup stats to human readable string logic. My plan would be to get 1650 in and then re-use code from it here.

mattst88:
Sounds good.

Flowdalic:
Closing in favor of #1661

PR #1653, opened 2026-08-11
cgroup: drop the "using <root>" message

mattst88:
It reads like leftover debugging output and tells the user nothing that PORTAGE_CGROUP_ROOT does not already say.

It was emitted only under --verbose, and was the only use of both CgroupManager's verbose parameter and the _info() helper, so drop those as well.

### DIFF
--- a/lib/portage/util/cgroup.py
+++ b/lib/portage/util/cgroup.py
@@ -59,10 +59,6 @@ def _warn(msg: str) -> None:
     writemsg_level(f"!!! cgroup: {msg}\n", level=logging.WARNING, noiselevel=-1)
 
 
-def _info(msg: str) -> None:
-    writemsg_level(f"cgroup: {msg}\n", level=logging.INFO)
-
-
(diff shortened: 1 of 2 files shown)


thesamesam (approved):
thanks matt

lib/portage/util/cgroup.py
```diff
@@ -453,19 +449,16 @@ class CgroupManager:
...
         self.enabled = False
-        self.verbose = verbose
         self._seen: set[str] = set()
 
     def setup(self) -> bool:
         """Initialize the base and per-emerge cgroups. Sets self.enabled."""
         if not ensure_base(self.root):
             return False
-        if self.verbose:
-            _info(f"using {self.root}")
```
Flowdalic, line 468: I found the output quite helpful when working on this code, so maybe we could at least

```
_debug(f"using {self.root}")
```

mattst88: Done.

PR #1655, opened 2026-08-13
emaint: drop the blank lines around results output

mattst88:
print_results() padded every module's output with one blank line before and two after, so a one-line result printed as five lines. Print just the result.

This is shared by every emaint module and by the 'emerge --sync' message path in _emerge/actions.py, so their output loses the padding too.


mattst88:
Sam suggested this in https://github.com/gentoo/portage/pull/1649#issuecomment-5268975574

thesamesam (approved):

PR #1656, opened 2026-08-13
observability: forget clients that disconnect

mattst88:
_client_connected appended each client's writer to self._writers with no removal path other than _broadcast dropping a writer whose write() raised. It never does: asyncio's stream protocol keeps the transport open after EOF, to permit half-close, and once connection_lost() has fired a write is counted and discarded rather than raising. Every client that connected therefore stayed in the list for the lifetime of the emerge, and every update walked all of them.

Await EOF on the client's reader and drop the writer when it arrives. Also do not record a writer whose initial snapshot could not be sent.

This is minor today, but becomes routine once "emerge --status" reads the socket, where each invocation connects, reads one line, and exits.

### DIFF
--- a/lib/_emerge/_observability.py
+++ b/lib/_emerge/_observability.py
@@ -382,13 +382,29 @@ class ObservabilityMonitor:
                 noiselevel=-1,
             )
 
-    def _client_connected(self, reader, writer):
-        self._writers.append(writer)
+    async def _client_connected(self, reader, writer):
         if self._last_snapshot is not None:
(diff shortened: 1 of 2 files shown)


thesamesam:
> This is minor today, but becomes routine once "emerge --status" reads the socket, where each invocation connects, reads one line, and exits.

"happens more often once ..." or "becomes problematic when..", I think "becomes routine" is technically correct but a strange way of putting it.

thesamesam (approved):
It doesn't look particularly pretty but I'll trust that you couldn't figure out a nicer way of doing it.

PR #1657, opened 2026-08-13
depgraph: skip the dynamic-deps preload for already-applied packages

mattst88:
_load_vdb() runs once per depgraph, but frozen_config, and therefore the
FakeVartree it holds, is shared by every backtracking depgraph. Each new
depgraph re-walked every installed package, calling findname2() and
_pull_valid_cache() and re-applying the live ebuild dependencies that an
earlier pass had already applied.

The repeat passes are not merely redundant. The first pass rewrites
Package._metadata via aux_update(), so later passes derive dynamic deps
from an installed instance whose metadata is no longer the vdb's.

_aux_get_history already records which instances have had dynamic deps
applied, so use it to skip them. _sync() discards a changed instance via
cpv_discard(), which clears its entry, so an instance that really did
change is still re-derived.

An "emerge -p -uDN @world" resolve against a root with 1736 installed
packages ran the preload seven times per root before this change. Mean of
five interleaved runs of that resolve: 110.2s before, 101.5s after, with
run-to-run spread under 1.2s on either side.

### DIFF
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -937,6 +937,12 @@ class depgraph:
         for pkg in fake_vartree.dbapi:
             self._dynamic_config._package_tracker.add_installed_pkg(pkg)
             self._add_installed_sonames(pkg)
+            if fake_vartree.dynamic_deps_applied(pkg):
+                # An earlier depgraph sharing this FakeVartree already applied
+                # dynamic deps to this instance, and doing it again would only
+                # re-derive the same result from metadata that first pass
+                # rewrote.
+                continue
             ebuild_path, repo_path = portdb.findname2(pkg.cpv, myrepo=pkg.repo)
             if ebuild_path is None:
                 fake_vartree.dynamic_deps_preload(pkg, None)
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -913,36 +913,42 @@ class depgraph:
             if preload_installed_pkgs:
                 vardb = fake_vartree.dbapi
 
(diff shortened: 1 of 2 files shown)


lib/_emerge/depgraph.py
```diff
@@ -937,6 +937,12 @@ def _dynamic_deps_preload(self, fake_vartree):
         for pkg in fake_vartree.dbapi:
             self._dynamic_config._package_tracker.add_installed_pkg(pkg)
             self._add_installed_sonames(pkg)
```
thesamesam: I was going to ask if the check can be done first, above these two function calls. I couldn't yet convince myself it's safe to do so.

Out of interest, did you explore that? It's fine if you didn't.

mattst88: It can't move up. `_package_tracker` and `_installed_sonames` are on `self._dynamic_config`, which is rebuilt for every backtracking depgraph, so they have to be repopulated on every pass. The `FakeVartree` is in `frozen_config` and shared across all of them, and the dynamic-deps apply is the only thing in the loop that mutates it. So the check goes between them.

I've pushed a follow-up that makes that split structural: the tracker/soname loop is hoisted out and unconditional, `_dynamic_deps_preload()` takes the pending list, and the `TaskScheduler` isn't constructed when it's empty.

I also instrumented it to see if anything was left. On the 1736-package root the two real passes are 1.34s and 0.76s; the twelve already-applied passes are ~2ms each. Nothing more to get.

Output unchanged, `lib/portage/tests/resolver` passes.

zmedico (approved):

PR #1658, opened 2026-08-15
vartree: fix the match cache and the vdb syscalls it hides

mattst88:
`vardbapi.match()` stores and returns its cached result under
`cache_key = (mydep, mydep.unevaluated_atom)`, but tests membership with the
bare atom, which is never a key in that dict. The test always succeeds, so
every call recomputes and the cache has done nothing since commit
d603f1440c81 introduced the key in 2012.

Recomputing means calling `cp_list()`, which stat()s the category directory
`match()` has just stat()ed itself: the doubled stat in bug 980860. The third
commit stops both of them reading a category directory that does not exist,
which a news item's Display-If-Installed lines hit repeatedly (bug 980777).
That skip is gated on ENOENT alone, since any other stat() failure has to
reach listdir(), which is the call that turns EACCES into PermissionDenied.

Syscalls under /var/db/pkg, per commit:

| | `emerge --info` | match loop | news-style |
|---|---|---|---|
| before | 251 stat | 12245 stat | 240 stat, 60 opendir |
| cache key | 251 stat | 8829 stat | 120 stat, 60 opendir |
| single stat | 215 stat | 7121 stat | 120 stat, 60 opendir |
| listdir skip | 215 stat | 7121 stat | 120 stat, 0 opendir |

The match loop calls `vdb.match(cp)` three times for each of the 1708
installed cps, which is where a working match cache shows up and an
`emerge --info` that matches each atom once does not. News-style is 60 atoms
across three categories that have no directory, matched twice.

Wall clock barely moves: 0.247s to 0.230s for that match loop on a warm page
cache, and `emerge -p -uDN @world` is unchanged at about 29s. The cost was
always syscalls and strace noise, which is what the bugs are about.

match() and cp_list() results for 8550 atoms are identical before and after
(every installed cp as bare, `>=`, `:0` and `=cpv`, plus missing-category and
`*/*` cases), and PermissionDenied is still raised both for an unreadable
category directory and for an unsearchable /var/db/pkg. Test suite passes.

Bug: https://bugs.gentoo.org/980777
Bug: https://bugs.gentoo.org/980860

### DIFF
--- a/lib/portage/dbapi/vartree.py
+++ b/lib/portage/dbapi/vartree.py
@@ -689,16 +689,24 @@ class vardbapi(dbapi):
 
         return moves
 
-    def cp_list(self, mycp, use_cache=1):
+    def cp_list(self, mycp, use_cache=1, _cat_mtime=None):
+        """
+        @param _cat_mtime: st_mtime_ns of the category directory, for internal
+                callers that have already stat()ed it. Saves a redundant stat()
+                of the same directory.
+        """
         from portage.versions import _pkg_str, catsplit, pkgsplit
 
         mysplit = catsplit(mycp)
         if mysplit[0] == "*":
             mysplit[0] = mysplit[0][1:]
-        try:
-            mystat = os.stat(self.getpath(mysplit[0])).st_mtime_ns
-        except OSError:
-            mystat = 0
+        if _cat_mtime is not None:
+            mystat = _cat_mtime
+        else:
+            try:
+                mystat = os.stat(self.getpath(mysplit[0])).st_mtime_ns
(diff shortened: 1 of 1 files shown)


thesamesam (approved):
Thank you! I'd started to suspect the caching just wasn't doing anything but didn't get any further with it.

Definitely worth fixing even if it doesn't seem to help perf here, seeing crap in strace is both actively distracting and also makes it harder to follow what we're doing on the Python side (and why).

(Also far easier to debug issues later on when you don't have redundant work done on either side as it's clear what is canonical and what isn't.)

PR #1660, opened 2026-08-17
doebuild: rebind loaded module paths for self update

mattst88:
_prepare_self_update() copies the running version of portage to a
temporary directory and prepends it to sys.path, so that a portage
update cannot pull modules out from under the running process.

sys.path only governs top-level imports. By the time the update runs,
the portage package is already imported, so an import of a submodule
such as portage.util.env_update is resolved via portage.util.__path__,
which still points at the installed copy. That copy disappears when the
new version installs somewhere else, as happens when PYTHON_TARGETS
changes, and the import then fails with:

    ModuleNotFoundError: No module named 'portage.util.env_update'

Only the multiprocessing forkserver and spawn start methods escaped
this, since their children import portage from scratch and pick up the
sys.path entry. The parent process, and anything forked from it, kept
the stale paths.

Rewrite `__path__` of the already imported portage and _emerge packages
to point at the backup copy, so that lazily imported submodules come
from the same snapshot as the rest of the running process.

Bug: https://bugs.gentoo.org/976616

### DIFF
--- a/lib/portage/package/ebuild/doebuild.py
+++ b/lib/portage/package/ebuild/doebuild.py
@@ -3352,6 +3352,34 @@ def _merge_unicode_error(errors):
     return lines
 
 
+def _rebind_loaded_modules(orig_pym_path, new_pym_path):
+    """
+    Rewrite the __path__ of loaded portage packages to point at a
+    temporary backup copy of the running version of portage. Submodule
+    imports are resolved via the __path__ of the parent package rather
+    than via sys.path, so this is what allows a process which has
+    already imported portage to import anything else after the
+    installed copy is replaced or removed (bug 976616).
+    """
+
+    orig_prefix = orig_pym_path.rstrip(os.sep) + os.sep
+
+    def _remap(path):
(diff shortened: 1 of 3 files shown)


thesamesam:
It indeed fixes (as expected) the PYTHON_TARGETS case too:
```
#!/bin/bash
export FEATURES="-buildpkg -test"
export USE="-native-extensions" # speed up the re-emerge

if [[ -f /usr/lib/python3.15/site-packages/_emerge/depgraph.py ]] ; then
        export PYTHON_TARGETS=python3_14
else
        export PYTHON_TARGETS=python3_15
fi

exec emerge -v1 portage -O --usepkg=n "$@"
```

thesamesam (approved):
Thanks, this looks good. I'd say add a `Suggested-by: Kerin Millar <kfm@plushkava.net>` (or a `Thanks-to`). I don't know if you debugged it in parallel but given we were stuck on this for ages and he nailed the analysis earlier, I think preserving some credit for him is nice.

Much much appreciated, as this has been bugging me for a while, and I felt rather embarrassed by it :(

PR #1661, opened 2026-08-20
observability: cgroup resources in emerge --status, plus follow-ups

mattst88:
Includes @Flowdalic's #1650, #1651 and #1652, carried unmodified and with authorship intact:

* #1650: human-readable cgroup resources in `emerge --status`, frozen at build completion
* #1651: read status over the socket first, refreshing on client connect
* #1652: average parallelism in the cgroup summary logged to emerge.log

The rest is on top of those, kept as separate commits so @Flowdalic's patches stay as authored.

### Fixes

* Parallelism was wrong once a package started merging: `elapsed` keeps running through the merge while the counters are frozen, so 20x decayed to 8x. The build's own duration is published as `build_elapsed` and used as the denominator (additive key, schema unchanged).
* `FEATURES="cgroup"` never showed the `(N.NNx)` suffix without `FEATURES="observability"`, since only the latter recorded the build timing it divides by. Timing is now recorded either way.
* Freezing kept `memory.current` and friends, which describe a cgroup that is destroyed moments later, so `--status` rendered them as live usage for the whole merge window. Dropped when freezing; peaks and totals kept.
* A snapshot read from `emerge-<pid>.sock` was never checked against the pid in its name, so a socket left by a killed emerge is answered by whatever inherits the pid. The pid now has to match, for status files too, and results are ordered by pid rather than by transport.
* Refresh on connect was rate-limited, leaving the client with the older snapshot and the status file unrewritten. Forced now.
* `_build_times`/`_phases` leaked for a build that produces no merge task (it failed, or emerge was interrupted).
* The cgroup was read twice per build, to freeze and to log. Its owner reads it once and hands the result to the monitor.

### Cleanups

* `_build_times` is a `_BuildTimes` object instead of a positional list that grew from two elements to three.
* emerge.log and `--status` disagreed on labels, CPU precision and whether to print zeros. Both call `format_resources()`:

  === Resource usage for dev-libs/foo-1.2 [CPU: 800.00s (20.00x), MaxMem: 4.00 KiB, I/O R: 0.00 B, I/O W: 8.00 KiB]

Zeros are logged, and `memory.zswap.current` is no longer summarized (no peak to report instead, and it says as little as `memory.current` by then). `FEATURES="cgroup"` has not been released, so nothing parses the old format.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -25,15 +25,17 @@ Features:
 
 * Add FEATURES="cgroup" to account each source build's CPU time, peak memory,
   and I/O in a per-build cgroup v2. No limits are imposed. Totals are written
-  to the build log and, with FEATURES="observability", included in the live
-  status snapshot. Configurable via PORTAGE_CGROUP_ROOT (default
+  to the build log, along with the average number of CPUs the build kept
+  busy, and, with FEATURES="observability", included in the live status
+  snapshot. Configurable via PORTAGE_CGROUP_ROOT (default
   /sys/fs/cgroup/portage); requires root and cgroup v2.
 
 * Add FEATURES="observability" to publish a machine-readable snapshot of the
   running emerge process (which packages are building/merging, their current
-  ebuild phase, and elapsed time) to /run/portage/emerge-<pid>.json. Query it
-  with the new "portageq jobs" command or "emerge --status". A Unix socket
-  streams the same snapshots live to connected clients.
+  ebuild phase, elapsed time, and cgroup resource usage where available) to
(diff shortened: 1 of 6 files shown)


Flowdalic (approved):

thesamesam (approved):
Thanks. I've only given it a cursory look but it looks reasonable to me and I couldn't spot anything wrong.

mattst88:
Thanks @Flowdalic! Thanks @thesamesam!

PR #1662, opened 2026-08-21
NEWS: document changes made since portage-3.0.81

mattst88:
Add the user-facing changes which had not yet been recorded in the 3.0.82 section: the prelink removal, the VDB metadata file rework and its emaint module, the checksum and resolver speedups, and several bug fixes.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -21,6 +21,15 @@ Breaking changes:
 * Importing `portage` as a library will require a UTF-8 locale be in effect:
   Portage will warn when this isn't the case.
 
+* Remove prelink support. sys-devel/prelink was treecleaned in 2022, so none
+  of this code could run any more. FEATURES="prelink-checksums" and the
+  PRELINK_BINARY, PRELINK_PATH and PRELINK_PATH_MASK variables are gone, as
+  is the generation of /etc/prelink.conf by env-update (bug #833909).
+
+* vartree: Remove the /var/cache/edb/vdb_metadata.pickle cross-session cache
+  (and vdb_metadata_delta.json). The per-package metadata file described
+  below replaces it. Tools reading the pickle directly will need updating.
+
 Features:
 
 * Add FEATURES="cgroup" to account each source build's CPU time, peak memory,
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -71,6 +114,25 @@ Bug fixes:
 * Set a non-zero window size for newly allocated ptys, so build processes
   can more reliably query their terminal dimensions (bug #978222).
 
+* vartree: Remove preserved libraries which are only consumed by each other.
+  A group of preserved libraries forming a consumer cycle was never removed
+  from the preserved libs registry, and neither were the libraries consumed
+  only by that cycle (bug #652382).
+
+* Fix "cp: cannot stat '.../build-info/<PF>.ebuild'" printed when merging a
```
thesamesam: I think this one was both introduced-and-fixed in this cycle, so you can drop this entry.

NEWS
```diff
@@ -71,6 +114,25 @@ Bug fixes:
...
+  binary package, where build-info is populated by the binpkg extraction
+  rather than by copying the ebuild.
+
+* vartree: Fix the vardbapi.match() cache being looked up with a key it never
+  stores, so that every match() call recomputed its result and re-stat()ed
+  the category directory.
+
+* observability: Republish the status snapshot on a timer, so a long-running
```
thesamesam: Ditto.

NEWS
```diff
@@ -71,6 +114,25 @@ Bug fixes:
...
+  phase such as a gcc compile no longer leaves the snapshot untouched (and
+  its derived CPU usage reading at zero) for hours.
+
+* observability: Drop clients from the broadcast list when they disconnect.
```
thesamesam: Ditto.

thesamesam (commented):
Thanks for doing these!

NEWS
mattst88: Removed.

NEWS
mattst88: Removed.

NEWS
mattst88: Removed.

NEWS
```diff
@@ -21,21 +21,28 @@ Breaking changes:
 * Importing `portage` as a library will require a UTF-8 locale be in effect:
   Portage will warn when this isn't the case.
 
+* Remove prelink support. sys-devel/prelink was treecleaned in 2022, so none
```
thesamesam, line 24: End of an era :)

thesamesam (approved):
Lovely, thank you again!

PR #1664, opened 2026-08-21
Scheduler: tear down observability once, not once per keep-going pass

mattst88:
"emerge --status" reported nothing at all for the rest of an emerge that had restarted after a failure with --keep-going.

merge() runs _merge() once per pass of its --keep-going loop, but the ObservabilityMonitor and the CgroupManager are built once, in the Scheduler's constructor, and nothing recreates them. Tearing them down in _merge()'s finally block therefore disabled them permanently at the end of the first pass, taking emerge-<pid>.json and emerge-<pid>.sock with it. Everything merged after the first failure was published nowhere, so a status client found no file to read and printed "Nothing to report...".

The cgroup manager was lost the same way, along with the per-emerge cgroup tree: builds in later passes were accounted for nowhere and their resource lines disappeared from both "emerge --status" and the build log.

Move both calls out to merge(), around the loop rather than inside it, so that they run once the last pass is done. _main_loop_cleanup() stays in _merge(), since that one is per-pass. A comment in CgroupManager.setup() already claimed close() was called from merge(); now it is.

### DIFF
--- a/lib/portage/tests/emerge/test_observability.py
+++ b/lib/portage/tests/emerge/test_observability.py
@@ -2,6 +2,8 @@
 # Distributed under the terms of the GNU General Public License v2
 
 import asyncio
+import contextlib
+import io
 import json
 import os
 import socket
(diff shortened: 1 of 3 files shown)


lib/portage/tests/emerge/test_observability.py
```diff
@@ -908,3 +910,93 @@ def test_no_parallelism_without_a_build_duration(self):
...
+                    cgroup.enabled,
+                    os.path.exists(monitor._status_path),
+                )
+            )
+            if len(seen) == 1:
+                sched._failed_pkgs.append(failed_pkg)
+                return 1
+            return os.EX_OK
+
+        sched.myopts = {"--keep-going": True}
```
thesamesam, line 953: Holy mock :)

zmedico (approved):

thesamesam (approved):

PR #1665, opened 2026-08-22
Run pkg_pretend phases in parallel (bug 579526)

mattst88:
pkg_pretend runs one package at a time. Ebuilds doing real work there mean minutes of waiting before the first package is built.

The phases have run in a coroutine since c7fa3f1eb, so this drives them with async_iter_completed(), which already implements `--jobs` and `--load-average` throttling. Output needs more care: _run_pkg_pretend() sends everything to stdout regardless of `--quiet` or `--jobs`, and concurrent phases sharing a terminal interleave. So with more than one job, each phase runs in background mode with PORTAGE_LOG_FILE pointed at a capture file in its own build dir, and results are replayed to stdout in mergelist order as they become available. Failures are recorded in that same ordered pass, so _failed_pkgs and `--keep-going` are unchanged. With `--jobs=1` the original direct-to-stdout path is kept.

The first commit is a pure extraction of the loop body into _run_pkg_pretend_one(). It carries the reindentation, so `git show -w` on the second commit is identical to `git show`.

Over 91 installed acct-user/acct-group packages on a 20-core machine, with `--jobs=8 --load-average=16`, the pkg_pretend phase drops from 51.8s to 5.9s and the whole run from 391.2s to 353.8s.

I checked every pkg_pretend in ::gentoo for anything that breaks under concurrency. Nothing reads stdin, writes outside ${T}, takes a lock, or mounts anything. Two things worth knowing:

dev-lang/perl warns and sleeps two seconds to let you abort. Buffered, the warning is replayed after the sleep has elapsed, so it no longer does its job.

linux-info.eclass's getfilevar() runs emake in ${KERNEL_DIR}, reached from 24 pkg_pretend functions. This concurrency is new, since _schedule_setup() puts setup phases on the merge queue to serialize them. I ran it 16-wide against a 6.18 tree: identical output, empty stderr, nothing written to the kernel tree or M=${T}.

Bug: https://bugs.gentoo.org/579526

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -26,6 +26,11 @@ Features:
   atoms, USE dependencies, USE flag names and flag sets that packages have in
   common cuts the peak of "emerge -puDN @world" by about 46% (bug #240320).
 
+* emerge: Run pkg_pretend phases concurrently, subject to --jobs and
+  --load-average. Output from each package is still shown one package at a
+  time, in merge order (bug #579526).
+
+
 portage-3.0.82 (2026-08-22)
 --------------
 
--- a/lib/_emerge/Scheduler.py
+++ b/lib/_emerge/Scheduler.py
@@ -876,263 +876,275 @@ class Scheduler(PollScheduler):
 
         return prefetcher
 
+    def _pkg_needs_pretend(self, pkg):
+        return (
+            isinstance(pkg, Package)
+            and pkg.operation != "uninstall"
+            and pkg.eapi not in ("0", "1", "2", "3")
+            and "pretend" in pkg.defined_phases
+        )
+
     async def _run_pkg_pretend(self, loop=None):
         """
         Since pkg_pretend output may be important, this method sends all
         output directly to stdout (regardless of options like --quiet or
         --jobs).
         """
+        loop = asyncio._wrap_loop(loop or self._sched_iface)
 
         failures = 0
-        sched_iface = loop = asyncio._wrap_loop(loop or self._sched_iface)
 
         for x in self._mergelist:
-            if not isinstance(x, Package):
+            if not self._pkg_needs_pretend(x):
                 continue
 
-            if x.operation == "uninstall":
-                continue
+            failures += await self._run_pkg_pretend_one(x, loop)
 
-            if x.eapi in ("0", "1", "2", "3"):
-                continue
+        if failures:
+            return FAILURE
+        return os.EX_OK
 
-            if "pretend" not in x.defined_phases:
-                continue
+    async def _run_pkg_pretend_one(self, x, loop):
+        """
+        Run the pkg_pretend phase for a single package.
 
-            self._termination_check()
-            if self._terminated_tasks:
-                raise asyncio.CancelledError
+        @rtype: int
+        @return: the number of failures (0 or 1)
+        """
+        sched_iface = loop
 
-            root_config = x.root_config
-            settings = self._allocate_config(root_config.root)
-            settings.setcpv(x)
+        self._termination_check()
+        if self._terminated_tasks:
+            raise asyncio.CancelledError
 
-            color = "PKG_BINARY_MERGE" if x.built else "INFORM"
-            self._status_msg(f"Running pre-merge checks for {colorize(color, x.cpv)}")
+        failures = 0
 
-            if not x.built:
-                # Get required SRC_URI metadata (it's not cached in x.metadata
-                # because some packages have an extremely large SRC_URI value).
-                portdb = root_config.trees["porttree"].dbapi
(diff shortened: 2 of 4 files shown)


thesamesam:
> dev-lang/perl warns and sleeps two seconds to let you abort. Buffered, the warning is replayed after the sleep has elapsed, so it no longer does its job.

I'm a little worried about this one. I nearly added such a warning the other day (sorry, I can't remember where yet). But I don't know if it's a deal-breaker (hopefully not). Having a `PROPERTIES` flag to disable it could work? Like `interactive` but milder?

I'm not sure I really believe in `sleep` for this but I think allowing it to work may be useful?

mattst88:
CTRL+C'ing during the pkg_pretend sleep vs a little while after doesn't seem like a problem to me.

I think sleeping and hoping that people read the message is kind of nasty. If it's that big a deal, can we just `die` in pkg_pretend instead?

eli-schwartz:
> I think sleeping and hoping that people read the message is kind of nasty. If it's that big a deal, can we just `die` in pkg_pretend instead?

I think I agree with this. The pkg_pretend is supposed to handle the case where users have NOT read the news, as a last ditch attempt to prevent you wedging your system. If we're that concerned about it, we should just die unless the user sets an environment variable indicating "yes I'm aware that my USE flags aren't consistently migrated here, I actually chose to change it by coincidence at this time".

eli-schwartz:
> I checked every pkg_pretend in ::gentoo for anything that breaks under concurrency. Nothing reads stdin, writes outside ${T}, takes a lock, or mounts anything.

If there were any that did most of this, if would be a PMS violation. This phase function is banned from "writing to the filesystem".

Which is consistent with the fact that its purpose is to sanity *check* things as part of dependency calculation and graph resolution, it certainly shouldn't be prepping the build environment for any reason and even environment saving is off the table, so you can't actually do anything other than choose to abort the emerge.

eli-schwartz:
> CTRL+C'ing during the pkg_pretend sleep vs a little while after doesn't seem like a problem to me.

Thinking a bit more about this specific point.

Since the purpose of pkg_pretend is to do sanity checks and abort if unsatisfiable, but do no actual work... by the time we finish buffering and then printing all pretend phases, the user sees the message as phases start to run, and in practice there is sufficient time to interrupt even a binpackage install, let alone a from source install, of packages such as perl etc.

We could perhaps also create a wrapper specifically for sleep, and have the wrapper trigger immediately dumping the buffer or else just rerun the sleep after all pretend phases have finished (to give the user time to react). It's a hack but a practical one?

mattst88:
Flushing the buffer immediately would mean an ebuild-ipc round trip to the scheduler, emitting a package's partial log out of mergelist order, and then emitting the rest of it later, split across other packages' output. Two ebuilds sleeping at once puts us back to interleaved output, which is the thing the buffering exists to prevent.

Re-running the sleep afterwards is mechanically easy, but it means shadowing a coreutils binary in the ebuild environment, which Portage has never done. It also changes the meaning of every other sleep in a pkg_pretend, and it's bypassed by /bin/sleep, command sleep, a sleep inside a called script, or read -t, so it's unreliable in exactly the cases it targets.

If we want the pause, it's simpler to do it entirely on the scheduler side: after the ordered replay finishes, pause for a few seconds if any pkg_pretend produced output. No wrapper, no IPC, nothing for an ebuild to bypass, and the ebuild-side sleeps still overlap across jobs so the speedup is unaffected.

That said, I think sleeping to give users the opportunity to read text and CTRL+C is really distasteful, so I'd rather take this opportunity to stop doing that.

thesamesam:
> That said, I think sleeping to give users the opportunity to read text and CTRL+C is really distasteful, so I'd rather take this opportunity to stop doing that.

[02:32] <sam_> anyway, I think I agree it's a stupid thing for ebuilds to do
[02:32] <sam_> it's not very accessible, it assumes you are actively watching the terminal, and any use should surely be replaced with a news item / ewarn / eerror + die

Now, that said, can we account for PROPERTIES=interactive and disable it in that case, perhaps? I don't expect ebuilds in ::gentoo to be doing this but it seems reasonable to have an opt-out like that.

NEWS
```diff
@@ -107,16 +92,14 @@ Features:
   every backtracking pass. Besides being wasted work, repeat passes derived
   dependencies from metadata an earlier pass had already rewritten.
 
+* emerge: Run pkg_pretend phases concurrently, subject to --jobs and
```
thesamesam: Wrong section.

thesamesam (approved):
Looks good to me but @zmedico knows more about the coroutine stuff in Portage than I do.

PR #1666, opened 2026-08-23
emirrordist: tolerate unreadable distfiles db entries

mattst88:
The distfiles db is a shelve mapping a distfile name to the cpv that pulled it in, and the cpv is only ever used to annotate a log line. Until 8b5f96782 (2021), FetchTask stored self.cpv directly rather than str(self.cpv), so shelve pickled the whole _pkg_str object. That was cheap until 27eeeb6e4 (2018) taught portdbapi.cp_list() to pass db=self, at which point every cpv carried a live portdbapi and the stored pickle grew to include its object graph.

Since 7e15fc7fa, Atom is no longer a str subclass and has no __dict__, so such a pickle no longer loads:
```
  File "/usr/lib/python3.12/shelve.py", line 114, in __getitem__
    value = Unpickler(f).load()
  AttributeError: 'Atom' object has no attribute '__dict__'
```
Only entries written between those two commits are affected; values written before 27eeeb6e4 hold derived scalars only, and values written after 8b5f96782 are plain strings, so both still load.

An affected entry aborts the run. MirrorDistTask._scheduled_deletion_log() re-reads every pending file on every run and never drops a row, so a single one of them breaks emirrordist until the file is finally deleted. DeletionTask._success() does drop the row, but it raises on the read before reaching the delete, so it never recovers either.

Add Config.lookup_cpv() and use it for both reads. A row that cannot be unpickled now yields "unknown" instead of aborting, which lets DeletionTask._success() reach its delete and drop the row for good. Warn once per run so a genuinely damaged db is still visible.

Bug: https://bugs.gentoo.org/981223
Bug: https://bugs.gentoo.org/766459


lib/portage/_emirrordist/Config.py
```diff
@@ -88,6 +89,28 @@ def __init__(self, options, portdb, event_loop):
...
+        Return the cpv that filename belongs to, or "unknown" if the
+        distfiles db holds no usable entry for it.
+        """
+        if self.distfiles_db is None:
+            return "unknown"
+        try:
+            return self.distfiles_db[filename]
+        except KeyError:
+            return "unknown"
+        except Exception:
```
thesamesam, line 103: I guess `AttributeError` is too risky here but do you have an example of what else could be raised here?

lib/portage/_emirrordist/Config.py
```diff
@@ -88,6 +89,28 @@ def __init__(self, options, portdb, event_loop):
...
+            # Entries written before bug 766459 was fixed are pickled
+            # _pkg_str object graphs, and classes in those graphs have
+            # since changed shape, so they no longer unpickle. The cpv
+            # only annotates a log line, so carry on without it.
+            logger.debug(f"unreadable distfiles db entry for '{filename}'")
+            if not self.distfiles_db_unreadable:
+                self.distfiles_db_unreadable = True
+                logger.warning("distfiles db has unreadable entries")
```
thesamesam, line 119: This fix is probably more suitable for backporting, but as a followup, could we version the DB and delete old ones / rewrite them (it's okay if the next run does the rewrite as if no DB existed)?

thesamesam (commented):
Looks like it works on dipper!

thesamesam:
The pypy test failure looks like an actual problem.

lib/portage/_emirrordist/Config.py
thesamesam: Nevermind, I read more closely now:
> A row that cannot be unpickled now yields "unknown" instead of aborting, which lets DeletionTask._success() reach its delete and drop the row for good.

lib/portage/_emirrordist/Config.py
mattst88: I've only confirmed `AttributeError`, from restoring the old attributes onto an Atom that no longer has a `__dict__`. But I can't rule out a class in the graph having moved or gone since, which would fail differently, but I have no damaged db to confirm that with.

What convinced me to keep the catch broad is the pypy failure: the same pickle raises on CPython and unpickles cleanly on pypy. Whether an old entry fails at all, never mind how, depends on the interpreter reading it.

thesamesam (approved):

PR #1667, opened 2026-08-23
util: accept Atom instances in stack_dictlist()

mattst88:
VirtualsManager._read_dirVirtuals() builds a dict whose keys and values are Atom instances and hands it to stack_dictlist(), which inspects each value with thing[:1] to implement the incremental "-" prefix. That worked only while Atom was a str subclass. Since commit 7e15fc7fa ("dep: Atom: remove str subtyping") every emerge invocation on a host whose profile ships a non-empty 'virtuals' file fails during config setup with:

    TypeError: 'Atom' object is not subscriptable

Wrap the value with str() for the two subscripts. The values themselves are stored unchanged, so providers remain Atom instances, and removing an entry by its "-cat/pkg" form still works because Atom compares equal to the matching str.

get_virts_p() has the same problem one call further on: it splits the keys of the compiled virtuals dict, which come from _dirVirtuals and are therefore Atom instances too. Wrap that in str() as well.

Bug: https://bugs.gentoo.org/981272
Fixes: 7e15fc7fa4c1 ("dep: Atom: remove str subtyping")

### DIFF
--- a/lib/portage/package/ebuild/_config/VirtualsManager.py
+++ b/lib/portage/package/ebuild/_config/VirtualsManager.py
@@ -232,7 +232,7 @@ class VirtualsManager:
         virts = self.getvirtuals()
         virts_p = {}
         for x in virts:
-            vkeysplit = x.split("/")
+            vkeysplit = str(x).split("/")
             if vkeysplit[1] not in virts_p:
                 virts_p[vkeysplit[1]] = virts[x]
         self._virts_p = virts_p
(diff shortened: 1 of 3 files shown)


thesamesam (approved):

thesamesam (commented):
Uh, should we be warning on use of this `virtuals` file? See 7ddb34f2d7729a632e1fdf2cbbcbf3b860b4d5ff.

`portage(5)` says:
> virtuals
> The virtuals file controls default preferences for virtuals that are defined via the PROVIDE ebuild variable (see ebuild(5)). Since Gentoo now uses GLEP 37 virtuals instead of PROVIDE virtuals, the virtuals file is irrelevant for all Gentoo ebuilds. However, it is still >possible for third-parties to distribute ebuilds that make use of PROVIDE.

But that appears to be outdated now.

See e.g. how we warn for `package.keywords` in `lib/portage/package/ebuild/_config/KeywordsManager.py` from 7e50a057958624dd728e858a3c07517240f85d02.

mattst88:
Yeah, we should be warning. I'll make another PR.

PR #1671, opened 2026-08-24
Warn about deprecated 'virtuals' files

mattst88:
Follow-up to the review of #1667.

Support for PROVIDE virtuals was removed in portage-2.3.25 (7ddb34f2d772,
"Remove support for PROVIDE virtuals"), so nothing can define such virtuals
any more, but portage still reads a profile 'virtuals' file silently.
Warn when a non-empty one is found, in the same way KeywordsManager warns
about package.keywords.

portage(5) claimed the file was only irrelevant to Gentoo ebuilds and that
third parties could still ship PROVIDE ebuilds, which has not been true
since 2.3.25. The /etc/portage/virtuals notice in config.py told users to
move the file to /etc/portage/profile/virtuals, which is now deprecated
too. Both are updated.

VirtualsManager had no test coverage, which is how the "'Atom' object is
not subscriptable" regression fixed in #1667 went unnoticed. The first
commit adds tests for the parsing path and for getvirtuals() and
get_virts_p(), both of which that bug touched.

Bug: https://bugs.gentoo.org/981272

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,6 +9,12 @@ Release notes take the form of the following optional categories:
 portage-3.0.83 (UNRELEASED)
 --------------
 
+Breaking changes:
+
+* Output a deprecation warning when a profile 'virtuals' file is used.
+  Support for PROVIDE virtuals was removed in portage-2.3.25, so these
+  files are obsolete; use GLEP 37 virtual packages instead (bug #981272).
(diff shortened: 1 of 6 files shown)


thesamesam (approved):
Thank you!

PR #1672, opened 2026-08-24
dep: fix three C dep-string parser divergences from the regex path

mattst88:
The C dep-string parser accepted or rejected a few atoms differently from the regex path. Because `Atom()` falls back to the regex path whenever `scan_atom` rejects a string, the existing parity tests could not see a scanner that rejects too much, and three divergences reached a release with a name-grammar corpus that already listed the affected names.

`use_reduce` has no such fallback, so the rejections were user-visible: installing a package with a dependency on `games-arcade/2048` aborted in `_post_src_install_write_metadata` with `expected '?'` (bug 981298), the misleading message coming from `scan_item` falling through to the use-conditional production after `scan_atom` bailed.

Fixed here:

- An all-digit first word of a package name (`games-arcade/2048`, `games-emulation/81-libretro`) is accepted. PMS constrains only the first character, and a digit is allowed there.
- A hyphen that does not begin a version is treated as an ordinary name character, so names ending in a hyphen or containing a hyphen run (`dev-util/timidity--`, `dev-util/diff-mode-`, `dev-util/diffball-9-`) scan. PMS only forbids a name from *starting* with `-`.
- A use dep that spells one flag's missing-USE default two different ways (`cat/pkg[a(+),-a]`) is rejected, matching `_use_dep`. `classify_use_deps` built the sets in C and handed them to `_use_dep`'s shortcut constructor, which trusts them. Repeating a flag with a consistent default (`cat/pkg[a,-a]`) stays valid.

The last commit closes the hole that hid all three: `_assert_same` now also compares acceptance through `use_reduce` for every case that passes nothing but an EAPI. Cases using `allow_wildcard` or `allow_repo` opt out, since those are `Atom()`-only options.

Bug: https://bugs.gentoo.org/981298

### DIFF
--- a/lib/portage/tests/dep/test_c_parser.py
+++ b/lib/portage/tests/dep/test_c_parser.py
@@ -1181,6 +1181,22 @@ class TestScanAtomNameGrammar(_AtomParityMixin, TestCase):
             with self.subTest(s=s):
                 self._assert_same(s, eapi="8")
 
+    def test_hyphens_that_do_not_start_a_version(self):
+        # A '-' that is not followed by a complete version is an ordinary
+        # name character, wherever it appears, so these names run to the end
+        # of the token rather than ending at the '-'.
+        for s in (
+            "dev-util/diff-mode-",
+            "dev-util/timidity--",
+            "dev-util/frob---",
+            "dev-util/diffball-9-",
+            "dev-util/foo-2048+",
+            "dev-util/foo-9-bar",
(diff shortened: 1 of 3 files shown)


thesamesam (approved):

mackal:
This appears to fix the issues I was experiencing.

mattst88:
Thank you for testing!

PR #1673, opened 2026-08-25
dep, Package: intern the values that packages have in common

mattst88:
A dependency calculation keeps far more memory alive than it needs to, because it builds a fresh object for every occurrence of a value that thousands of packages share: the same atom, the same USE dependency, the same USE flag name, the same set of flags. CPython shares none of them on its own, not even the empty frozenset.

These four commits intern each of those by value. For `emerge -puDN @world` here:

    dep: intern _use_dep instances             292.3MB -> 206.9MB
    dep: intern Atom instances                 206.9MB -> 172.5MB
    Package: intern IUSE and USE flag names    172.5MB -> 165.4MB
    Package: intern the frozensets ...         165.4MB -> 158.1MB

That is 46% off the peak. Interning Atom also takes the run time from 29.2s to 27.0s, because a cache hit skips the regex parse as well; the other three leave it alone.

The two caches in portage.dep hold weak references, so an entry goes away with the last user of its value. The one in Package holds strong references, because frozenset does not support weak references; it is bounded by the number of distinct IUSE and USE values in the repositories, which came to about 2200 entries here.

Every peak RSS figure above is the median of five runs of the same `emerge -puDN @world`, all measured in one session, so they chain. The five runs of a revision spanned at most 2.8MB. Every revision printed the same merge list, and the full test suite passes at each one.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -22,6 +22,10 @@ Features:
   VDB (var database) to skip writing files that are already identical on
   disk, which saves I/O and reduces filesystem cache pressure.
 
+* emerge: Reduce the memory used by a dependency calculation. Interning the
+  atoms, USE dependencies, USE flag names and flag sets that packages have in
+  common cuts the peak of "emerge -puDN @world" by about 46% (bug #240320).
+
 portage-3.0.82 (2026-08-22)
 --------------
 
--- a/lib/_emerge/Package.py
+++ b/lib/_emerge/Package.py
@@ -23,6 +23,24 @@ from portage.versions import _pkg_str, _unknown_repo
 
 from _emerge.Task import Task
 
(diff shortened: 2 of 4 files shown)


NEWS
```diff
@@ -113,6 +113,10 @@ Features:
   every backtracking pass. Besides being wasted work, repeat passes derived
   dependencies from metadata an earlier pass had already rewritten.
 
+* emerge: Reduce the memory used by a dependency calculation. Interning the
```
thesamesam: Wrong section!

thesamesam (approved):
Lovely, thank you!

zmedico (approved):

PR #1674, opened 2026-08-25
tests: cover merge order around circular dependencies

mattst88:
Two long standing reports describe packages being merged before the dependency they link against, when cycles are present in the graph: a slot operator dependency during an --emptytree run, and a sub-slot triggered rebuild next to an unrelated cycle.

Neither reproduces with the current resolver. Add the scenarios as regression tests so that the ordering stays correct.

Bug: https://bugs.gentoo.org/463976
Bug: https://bugs.gentoo.org/468052

### DIFF
--- a/lib/portage/tests/resolver/meson.build
+++ b/lib/portage/tests/resolver/meson.build
@@ -40,6 +40,7 @@ py.install_sources(
         'test_installkernel.py',
         'test_keywords.py',
         'test_merge_order.py',
+        'test_merge_order_cycles.py',
         'test_missing_iuse_and_evaluated_atoms.py',
         'test_multirepo.py',
         'test_multislot.py',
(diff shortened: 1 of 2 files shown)


thesamesam:
I assume these failed before (to know the test is useful). Can you bisect?

zmedico (approved):

mattst88:
Took some work, but that was a good idea and really worth it. I made the test actually reproduce the issue reported in bug 463976. I couldn't reproduce the problem from 468052 though.

thesamesam:
Yay! Thank you. It's OK if you want to keep the test for 468052 as well given it covers a scenario that isn't unlikely, and I would generally feel better about touching this area if I knew we had better coverage, just perhaps as a second commit given the analysis in the commit message won't apply there now.

mattst88:
Done!

thesamesam (approved):
Thank you!

PR #1675, opened 2026-08-25
tests: add has_ipv6() tests

mattst88:
Cover the unsupported, socket failure, bind failure, and supported cases, and that the probe result is cached after the first call.

Bug: https://bugs.gentoo.org/981369


zmedico (approved):

thesamesam (approved):
Thanks for doing that.

PR #1676, opened 2026-08-26
Scheduler: show cgroup resource summary with --jobs when --verbose 

Flowdalic:


zmedico (approved):

lib/_emerge/Scheduler.py
```diff
@@ -1670,6 +1672,7 @@ def _merge_wait_exit_handler(self, task):
     def _merge_exit(self, merge):
         self._running_tasks.pop(id(merge), None)
         self._observability.note_task_finished(merge)
+        self._cgroup_finish(merge.merge, action="install")
```
mattst88: This only works because of the line above it. `note_task_finished(merge)` takes the `_PackageMerge` branch and pops `_build_times[cpv]`, so the `note_build_resources()` inside `_cgroup_finish()` stores nothing and `build_elapsed()` returns `None`.

Swap the two lines and it breaks silently: the build's `times.resources` gets overwritten with the install counters, and `format_resources()` divides install CPU by the build's wall clock, producing a meaningless "(Nx)".

Better to not depend on the ordering. The install summary wants to render and discard:


```python
    def _cgroup_finish(self, build, action="emerge", record=True):
        ...
        stats = self._cgroup.read_stats(cpv)
        resources = (
            self._observability.note_build_resources(cpv, stats)
            if record
            else freeze_resources(stats)
        )
```

and call it here with `record=False`.

lib/_emerge/JobStatusDisplay.py
```diff
@@ -153,13 +153,17 @@ def _update(self, msg):
...
 
-    def displayMessage(self, msg):
+    def displayMessage(self, msg, prefix=None):
         was_displayed = self._displayed
 
         if self._isatty and self._displayed:
             self._erase()
 
-        self._write(self._format_msg(msg) + self._term_codes["newline"])
+        if prefix is None:
```
mattst88: This branch has no test. All three new tests replace `_status_display` with a `SimpleNamespace` whose `displayMessage` is a lambda, so the code changed in this file never runs. Please add a JobStatusDisplay test against a StringIO asserting the presence/absence of the ">>> " prefix.

lib/_emerge/JobStatusDisplay.py
```diff
@@ -153,13 +153,17 @@ def _update(self, msg):
 
         self._display(self._format_msg(msg))
 
-    def displayMessage(self, msg):
+    def displayMessage(self, msg, prefix=None):
```
mattst88: `prefix` conflates two things. Every caller passes `""`, and what they want is "don't apply `_format_msg`". As written a non-empty prefix would be a second formatting path that nothing wants and nothing tests.

```python
    def displayMessage(self, msg, raw=False):
        ...
        formatted_msg = msg if raw else self._format_msg(msg)
```

lib/_emerge/Scheduler.py
```diff
@@ -422,10 +422,12 @@ def _cgroup_finish(self, build):
         )
         rendered = format_resources(resources, self._observability.build_elapsed(cpv))
         if rendered:
-            msg = f"=== Resource usage for {cpv} [{rendered}]"
+            msg = f"=== Resource usage for {action} of {cpv}: {rendered}"
```
mattst88, line 433: This changes the format of a line `self._logger.log()` writes to emerge.log:

```diff
-    === Resource usage for foo-1.2 [CPU: 800.00s (20.00x), ...]
+    === Resource usage for emerge of foo-1.2: CPU: 800.00s (20.00x), ...
```

I think that's fine. Just calling it out. I don't really know what our stability guarantees are regarding emerge.log formatting.

lib/_emerge/Scheduler.py
```diff
@@ -422,10 +422,12 @@ def _cgroup_finish(self, build):
...
             log_path = build.settings.get("PORTAGE_LOG_FILE")
             self._sched_iface.output(f"{msg}\n", log_path=log_path)
             self._logger.log(f" {msg}")
+            if self._background and "--verbose" in self.myopts:
```
mattst88, line 437: 1. Title says "with --jobs", but the gate is `self._background`, also true for `--quiet-build` and whenever stdout isn't a tty. Fix the title or the condition.
2. New user-visible effect of `--verbose`, undocumented. man/emerge.1 needs a line saying background runs echo the resource summary to the terminal.

lib/_emerge/Scheduler.py
```diff
@@ -409,8 +409,8 @@ def _observability_phase(self, cpv, phase):
         """
         self._observability.note_phase(cpv, phase)
 
-    def _cgroup_finish(self, build):
-        """Log the final cgroup resource summary for a build and remove it."""
+    def _cgroup_finish(self, build, action="emerge"):
```
mattst88: `action="emerge"` reads oddly next to `"install"` -- this branch is the build, as the docstring below says. `action="build"` makes the call sites parallel.

lib/_emerge/Scheduler.py
mattst88: Nit: `_cgroup_finish()` runs before `_do_merge_exit()`, so a failed merge prints its resource summary right before the failure message. Moving it after `_do_merge_exit()` is safe. `_deallocate_config()` is later still.

lib/portage/tests/emerge/test_observability.py
```diff
@@ -818,6 +818,9 @@ def _cgroup_finish(self, features, stats):
 
         messages = []
         sched = Scheduler.__new__(Scheduler)
+        sched.myopts = {}
```
mattst88: `Scheduler.__new__(Scheduler)` plus the same three assignments now appears five times, and every future attribute `_cgroup_finish` touches has to be added to all five. Worth a `_make_cgroup_sched(...)` helper.

mattst88 (changes requested):
All in all, looks good. Just a few comments.

Also, the merge-phase leaf is created by `doebuild_spawn()` -> `ensure_leaf()` for pkg_preinst/pkg_postinst, after `_build_exit()` destroyed the build leaf. Before this change nothing reaped it until `CgroupManager.close()` swept `emerge_root` at end of run. So this fixes a per-package cgroup leak for the whole emerge, which is a better justification than the reporting change, and it belongs in the commit message.

mattst88:
Looks great. Thank you! Merged.

PR #1679, opened 2026-08-26
emerge: add --merge-wait-scope to limit merge-wait barriers

mattst88:
Packages in _deep_system_deps are held in _merge_wait_queue and merged only once every build slot has gone idle, and never in parallel with another merge, because dependencies on system packages are frequently unspecified and merging them alongside unrelated builds risks failures such as bug 256616.

That set is the `@system` set plus its transitive runtime dependencies, which on a typical installation is around a quarter of all packages: an `emerge -e @world` graph here has 429 of 1674 merges in it. Each one spends a long time built but not yet installed, which leaves all of its dependents ineligible and collapses the schedulable frontier. Modelling that graph with build and merge times taken from emerge.log gives 23% build slot utilization at --jobs=16, and a makespan that stops improving past --jobs=16 and gets worse at --jobs=32.

Make the scope selectable. "deep" is the default and preserves the current behavior exactly. "system" restricts it to the `@system` set itself (50 packages here), "toolchain" to the core toolchain (11 of the 12 listed, since only one libc is installed), and "none" treats no package as a system package. In the same model at --jobs=16, against a 4.24h critical path lower bound, the makespan goes from 8.57h to 6.74h for "system" and 6.31h for "toolchain", and to 4.81h for "toolchain" combined with FEATURES="parallel-install".

Note that FEATURES="merge-wait" is enabled by default and holds back every package regardless of this option, so narrowing the scope only increases parallelism for packages which are not already held back: either with FEATURES="-merge-wait", or with FEATURES="parallel-install", which merges the packages outside the scope in parallel batches.

Narrowing the scope trades away the protection described above, so the default is unchanged and the tradeoff is left to the user.

Note that the merge order preference in depgraph, which sorts system packages first, deliberately keeps using the full deep set.


zmedico (approved):

thesamesam:
I am not sure if the granularity here is right, but I don't know how to improve it.

The largest non-system/toolchain issue was Python-related, see https://bugs.gentoo.org/902713 and https://bugs.gentoo.org/663324. It happened often enough that tinderboxes actually hit it a few times.

I don't have a better idea for handling or mitigating that though. If we could mitigate or reduce the chance of the Python problem happening (perhaps by reducing the possible window that may or may not exist from bug 902713), then we could even consider the default being more relaxed.

So, if you choose "system" or "toolchain" even as a user, you still are at risk of the big problem that motivated us changing the behaviour to begin with, not just some fringe case :(

eli-schwartz:
> Note that FEATURES="merge-wait" is enabled by default and holds back every package regardless of this option,

Right, because by default we assume it is "likely horribly unsafe". So by default this PR is inert and all it actually does is shift around the existing latent breakage possibilities even more.

I... get why you want to add this, but at the same time it makes me kind of uncomfortable to keep making the "I agree to risk breakage" case even more complicated to understand. Honestly, if you're convinced that nothing is going to break (maybe because you looked at the merge list with -p first?) or you think it's fine to backtrack and re-run emerge to resolve any packages whose src_compile chokes, why wouldn't you just be happy disabling all kinds of merge-wait?

A better long term fix which is semantically correct would be to unlock merging packages if none of the currently running jobs have the to-be-merged package in the (recursive) deptree. It's entirely plausible to merge disjoint branches of `@world` (and possibly even `@system`) without breaking the compilation phase of the other branch.

Modulo the setuptools plugins case, of course. Which probably needs EAPI support for some form of "BREAKS=". But if everything other than setuptools worked, people might be more tempted to feel safe disabling merge-wait...

mattst88:
I look at this as providing a well-defined flag multiple people can use and test (and be sure they're testing the same thing), with the end goal of flipping the default. It's an experiment. If it doesn't pan out, we rip it out.

thesamesam:
Let's document it as experimental please in the man page entries.

mattst88:
@thesamesam: I looked at whether this actually widens the exposure for those two bugs, and I don't think it does.

Both of them involve packages that are already outside the deep set. On this installation, `@system` plus its transitive runtime dependencies is 380 packages out of 1756 installed, and:

| package | in deep set |
| ------- | ----------- |
| dev-lang/python | yes |
| dev-python/setuptools | yes |
| dev-python/cython | no |
| dev-python/pythran | no |
| dev-python/setuptools-scm | no |

(Computed from the installed RDEPEND closure rather than a resolver graph, but the shape is the same.)

For bug 902713, what protects you today is `FEATURES="merge-wait"` itself, not the deep scope: pythran and cython are not in the deep set, so with `-merge-wait` their merges run concurrently with builds regardless of what this option is set to. For bug 663324 the victim is setuptools, which is in the set, but the package whose merge interferes is the plugin (setuptools-scm, pbr, setuptools-git), and those are all outside it, so the same holds.

So for these two, "system" and "toolchain" are not giving up a protection that "deep" was providing. The deep set is protecting against the bug 256616 class, where the interfering package is a system package. Someone running `-merge-wait` today already has the full Python exposure, and this option does not change it in either direction.

On mitigating it, I agree that closing the window from 902713 is the interesting case, since it is what would make relaxing the default something we would consider at all. The summary names the mechanism: portage creates a directory and then pauses before installing files into it, and an empty directory in site-packages is an importable namespace package. Deferring the mkdir until the contents are ready, or staging and renaming, would shrink or remove it. Renaming is awkward for a shared directory like site-packages, so it needs care, but the deferral part looks tractable.

I'll mark the option experimental in the man pages.

I'll also do some more investigation around the bugs you linked and how we might be able to handle those cases better.

@eli-schwartz: deptree-based unlocking wouldn't cover bug 902713. pythran gets onto the system as a BDEPEND of scipy/scikit-learn/scikit-image, but the build that breaks is shapely, which depends on cython and not pythran, and cython picks pythran up with a bare `try: import pythran / except: pythran = None`. So it is never in shapely's recursive deptree and reachability would unlock the merge. Same undeclared-interference class as the setuptools plugins. So it wants `BREAKS=`, or the 902713 window closed so a half-merged directory is not importable.

mattst88:
@thesamesam, @eli-schwartz: any other thoughts?

thesamesam (commented):
I think I'm OK with it but I still find the situation quite complex and hope we can simplify it in future.

PR #1680, opened 2026-08-26
vartree, movefile: survive a portage self update

mattst88:
A merge aborted mid-install with:

    TypeError: movefile() got an unexpected keyword argument 'encoding'

vartree imported portage.util.movefile from inside the merge loop, late enough that a portage self update had already installed the new movefile.py. The vartree in memory then called it with a signature 8d416df5 removed.

1b188f43 mostly handles this by rebinding `__path__` of the imported portage packages to a backup of the running version. The report predates it. This closes the rest:

- Import movefile and _cmpxattr in vartree at module scope, so the import cannot happen late. No cycle: movefile imports portage and leaf modules only.
- Accept an unused encoding parameter in movefile() again, so a vartree loaded from an older installed version can finish its merge.
- Test the same-location update. The existing case deletes the installed tree, catching a module that fails to import, not one that imports and is the wrong version. The new case overwrites the installed modules after _prepare_self_update() and checks that a later import comes from the backup, in the process and in a child, for fork and forkserver. It fails with the __path__ rebinding disabled.

Bug: https://bugs.gentoo.org/976616

### DIFF
--- a/lib/portage/dbapi/vartree.py
+++ b/lib/portage/dbapi/vartree.py
@@ -62,6 +62,7 @@ from portage.exception import (
 from portage.localization import _
 from portage.util.futures import asyncio
 from portage.util.futures.executor.fork import ForkExecutor
+from portage.util.movefile import _cmpxattr, movefile
 
 from ._ContentsCaseSensitivityManager import ContentsCaseSensitivityManager
 
@@ -5457,7 +5458,6 @@ class dblink:
         from portage.checksum import _perform_md5_merge as perform_md5
         from portage.eapi import eapi_rewrites_symlinks
         from portage.util import normalize_path
-        from portage.util.movefile import movefile
         from portage.versions import pkgsplit
 
         showMessage = self._display_merge
@@ -6325,7 +6325,6 @@ class dblink:
         Should only be used for regular files.
         """
         from portage.util import writemsg
-        from portage.util.movefile import _cmpxattr
 
         if mydmode is None or not stat.S_ISREG(mydmode):
             return MoveReason.FILE_MISSING_OR_NOT_REGULAR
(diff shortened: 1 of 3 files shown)


zmedico (approved):

thesamesam (commented):
Please add the `Bug` tags to each commit.

I would note in the commit message that this is a crutch for those upgrading from <3.0.81.3 (i.e. lacking 1b188f43eb40a3c59c306b0fbd0407435b6cd4a7).

(Edited because I talked myself into getting it.)

mattst88:
Added the Bug tags.

You're right about the cause. I reproduced it both ways: install a copy of portage, run _prepare_self_update(), overwrite the installed movefile.py in place, then do the late import.

```
before 1b188f4: .../lib/portage/util/movefile.py                    (installed, poisoned)
after  1b188f4: .../tmp/portage/._portage_reinstall_.../movefile.py (backup, clean)
```

So with 1b188f4 the late import does come from the backup, and the skew can't happen.

The catch is that _prepare_self_update() runs out of the old copy, so 1b188f4 only helps when the running version already has it. Both 8d416df50 (which dropped the parameter) and 1b188f4 are in 3.0.82, so every upgrade from 3.0.81.x is exposed, which fits the 3.0.81.2 -> 3.0.82.1 stage build.

That's why I'd still like the compat shim: not as ongoing defense, but to unbreak that upgrade for everyone not yet on 3.0.82. I've rewritten its commit message to say so explicitly rather than leaving the reader to re-derive it.

One thing worth flagging about the testcase, since it doesn't do what the shim commit might suggest: it doesn't exercise encoding at all. It asserts that a post-update import resolves to the backup copy and that vartree.movefile is util.movefile.movefile. So it covers 1b188f4 and the module-scope import, not the shim.

thesamesam (approved):

PR #1681, opened 2026-08-26
Improve the circular dependency report

mattst88:
The circular dependency error tells the user that a cycle exists, but little about why it exists or what to do about it. This series adds that information.

The before/after output below is real: a stage3-amd64-systemd container on the stock default/linux/amd64/23.0/desktop/systemd profile, resolving against a current ::gentoo checkout, running emerge from this branch and from master. Where a case needed extra configuration to reproduce, that is stated.

### Name the USE flags responsible for each edge

Bug [310613](https://bugs.gentoo.org/310613). The report showed only the priority of each edge, so the user had to read the ebuilds to work out which flag pulled the dependency in, and had no way to match a suggested USE change to the edge it applies to.

Bug [222601](https://bugs.gentoo.org/222601), open since 2008, is still reproducible: on the desktop profile, avahi[gtk] needs gtk+, gtk+[cups] needs cups, and cups[zeroconf] needs avahi. Only zeroconf is off by default, so the whole config is the stock profile plus one line:

```
net-print/cups zeroconf
```

Before, `emerge -p net-dns/avahi`:

```
 * Error: circular dependencies:

(x11-libs/gtk+-3.24.52:3/3::gentoo, ebuild scheduled for merge) depends on
 (net-print/cups-2.4.19:0/0::gentoo, ebuild scheduled for merge) (buildtime)
  (net-dns/avahi-0.9_rc3:0/0::gentoo, ebuild scheduled for merge) (buildtime)
   (x11-libs/gtk+-3.24.52:3/3::gentoo, ebuild scheduled for merge) (buildtime)

It might be possible to break this cycle
by applying any of the following changes:
- net-dns/avahi-0.9_rc3 (Change USE: -gtk)
- x11-libs/gtk+-3.24.52 (Change USE: -cups)
- net-print/cups-2.4.19 (Change USE: -zeroconf)
```

After:

```
 * Error: circular dependencies:

(x11-libs/gtk+-3.24.52:3/3::gentoo, ebuild scheduled for merge) depends on
 (net-print/cups-2.4.19:0/0::gentoo, ebuild scheduled for merge) (buildtime, USE=cups)
  (net-dns/avahi-0.9_rc3:0/0::gentoo, ebuild scheduled for merge) (buildtime, USE=zeroconf)
   (x11-libs/gtk+-3.24.52:3/3::gentoo, ebuild scheduled for merge) (buildtime, USE=gtk)

It might be possible to break this cycle
by applying any of the following changes:
- net-dns/avahi-0.9_rc3 (Change USE: -gtk)
- x11-libs/gtk+-3.24.52 (Change USE: -cups)
- net-print/cups-2.4.19 (Change USE: -zeroconf)
```

Three edges, three flags, and the suggestions now line up with the edges they came from.

A flag is named when the dependency goes away without it, so a flag under "!flag? ( )" is named while it is disabled, and a dependency that is also pulled in unconditionally is blamed on nothing. When several flags pull in the same dependency and no single one of them is enough, all of them are named.

### Report masked packages that could be used instead

Bug [971256](https://bugs.gentoo.org/971256). || ( ) preferences are evaluated with autounmask disabled, so a bootstrap package that is not keyworded is never chosen and the resolver picks the branch that closes the cycle. Nothing told the user that keywording the bootstrap package would help.

Reproduced by masking the binary bootstrap package, which is the situation on an arch where it is not keyworded. `emerge -p dev-java/openjdk` with dev-java/openjdk-bin masked, before:

```
 * Error: circular dependencies:

(dev-java/openjdk-25.0.3_p9-r1:25/25::gentoo, ebuild scheduled for merge) depends on
 (dev-java/openjdk-25.0.3_p9-r1:25/25::gentoo, ebuild scheduled for merge) (buildtime)

 * Note that circular dependencies can often be avoided by temporarily
 * disabling USE flags that trigger optional dependencies.
```

There is no USE flag to disable here, so the advice is a dead end.

After:

```
 * Error: circular dependencies:

(dev-java/openjdk-25.0.3_p9-r1:25/25::gentoo, ebuild scheduled for merge) depends on
 (dev-java/openjdk-25.0.3_p9-r1:25/25::gentoo, ebuild scheduled for merge) (buildtime, USE=system-bootstrap)

 * The following masked packages could be used instead of a
 * package in the cycle. Unmasking them, for example via
 * /etc/portage/package.accept_keywords, might solve it:
- dev-java/openjdk-bin-25.0.4_p7 (instead of dev-java/openjdk-25.0.3_p9-r1)

 * Note that circular dependencies can often be avoided by temporarily
 * disabling USE flags that trigger optional dependencies.
```

dev-lang/go with go-bootstrap masked, which is the case in the bug, names dev-lang/go-bootstrap-1.24.6. dev-lang/rust with rust-bin masked lists both rust-bin-1.77.1-r102 and rust-bin-1.78.0-r102.

### Point at FEATURES=test for test dependency cycles

Bugs [416871](https://bugs.gentoo.org/416871), [604566](https://bugs.gentoo.org/604566), [703348](https://bugs.gentoo.org/703348). A cycle that only exists because USE=test pulls in a test dependency cannot be solved by the printed "Change USE: -test" suggestion alone, since FEATURES=test is what enables the flag in the first place. The report now names the packages whose test dependencies are in the cycle and prints the config lines that turn their test suites off:

```
 * This cycle involves test dependencies. If you do not
 * need to run the test suites of the packages listed below,
 * disable FEATURES=test for them. Put FEATURES="-test" in
 * /etc/portage/env/no-test.conf, then add the following to
 * /etc/portage/package.env (see "package.env" in the
 * portage(5) man page for more details):

=dev-python/py-1.11 no-test.conf
```

I did not find a package in the current tree that reproduces this from a stage3, so this one is covered by resolver tests.

### Say when the search for a USE change was given up on

The number of USE combinations explored per dependency is capped, because the search is exponential in the number of flags. Past the cap the dependency was skipped silently and the user was told there was no solution, when in fact none was looked for. Now:

```
 * The search for USE changes was given up on for the following
 * packages, because they have too many USE flags that affect the
 * dependency. Raise PORTAGE_CIRCULAR_MAX_USE_FLAGS (currently 10)
 * to search harder:
- x11-libs/gtk+-3.24
```

`PORTAGE_CIRCULAR_MAX_USE_FLAGS` is new and documented in `make.conf(5)`. Also covered by tests.

### Always show a cycle

Bugs [877549](https://bugs.gentoo.org/877549), [929010](https://bugs.gentoo.org/929010). circular_dependency_handler looked for cycles only at ignore_medium_soft, so a cycle made up entirely of dependencies that are ignored at that level (PDEPEND edges, satisfied deps, optional deps) left shortest_cycle unset. That produced an error with the "circular dependencies:" header and nothing after it, and debug_print() could prune the graph down to nothing as well. The handler now walks DepPrioritySatisfiedRange.ignore_priority down from ignore_medium_soft until it finds a cycle, and debug_print() stops pruning before it empties the graph.

### --circular-deps-report=json

Circular dependency failures are common enough in tinderboxes and CI that a machine readable form is worth having. The cycles, the USE flags responsible for each edge, the suggested USE changes, the test dependency parents, the masked alternatives and the truncated searches are written to stdout as a JSON object. For the avahi cycle above:

```json
{
  "shortest_cycle": [
    {
      "parent": "net-dns/avahi-0.9_rc3",
      "child": "x11-libs/gtk+-3.24.52",
      "priority": "buildtime",
      "affecting_use": ["gtk"]
    },
    {
      "parent": "x11-libs/gtk+-3.24.52",
      "child": "net-print/cups-2.4.19",
      "priority": "buildtime",
      "affecting_use": ["cups"]
    },
    {
      "parent": "net-print/cups-2.4.19",
      "child": "net-dns/avahi-0.9_rc3",
      "priority": "buildtime",
      "affecting_use": ["zeroconf"]
    }
  ],
  "solutions": [
    {"package": "net-dns/avahi-0.9_rc3", "use_changes": [{"gtk": false}]},
    {"package": "net-print/cups-2.4.19", "use_changes": [{"zeroconf": false}]},
    {"package": "x11-libs/gtk+-3.24.52", "use_changes": [{"cups": false}]}
  ],
  "test_dep_parents": [],
  "masked_alternatives": [],
  "search_truncated": [],
  "max_affecting_use": 10
}
```

"cycles" holds every cycle found, not just the shortest one. In this run there were 15.

`emerge(1)`, `make.conf(5)` and `NEWS` are updated to match.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -24,6 +24,14 @@ Features:
   --load-average. Output from each package is still shown one package at a
   time, in merge order (bug #579526).
 
+* emerge: Improve the circular dependency report. It now names the USE flags
+  responsible for each edge (bug #310613), points at FEATURES=test for test
+  dependency cycles (bug #416871, bug #604566, bug #703348), lists masked
+  packages that could be used instead (bug #971256), says when the search for
+  a USE change was given up on (configurable with
+  PORTAGE_CIRCULAR_MAX_USE_FLAGS), and is never empty (bug #929010, bug
+  #877549). Add --circular-deps-report=json for automated consumers.
+
 portage-3.0.82.2 (2026-08-28)
 ----------------
 
--- a/lib/_emerge/depgraph.py
+++ b/lib/_emerge/depgraph.py
@@ -4,6 +4,7 @@
 import collections
 import errno
 import functools
+import json
 import logging
 import os
 import stat
@@ -10415,6 +10416,10 @@ class depgraph:
         )
         handler = self._dynamic_config._circular_dependency_handler
 
+        if self._frozen_config.myopts.get("--circular-deps-report") == "json":
+            self._show_circular_deps_json(handler)
+            return
+
         self._frozen_config.myopts.pop("--quiet", None)
         self._frozen_config.myopts["--verbose"] = True
         self._frozen_config.myopts["--tree"] = True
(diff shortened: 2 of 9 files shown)


zmedico (approved):

thesamesam:
For `emerge -pv1 mono`:
> Name the USE flags responsible for each edge
>   (dev-dotnet/libgdiplus-6.2:0/0::gentoo, ebuild scheduled for merge) (buildtime, USE=minimal)

Do you think we could colour the `USE=minimal` there?

thesamesam (approved):
This is great, thanks! Sorry it took a little while to get to it.

kevinlipe:
I just wanted to say... I'm the guy that opened 222601 in 2008 and it's very cool to see this. I haven't owned that PowerBook for 17 years so I won't be able to test on PowerPC, unfortunately.

mattst88:
Hah, thanks! :D

Sorry it took 18 years!

PR #1682, opened 2026-08-29
dep: partition the Atom constructor cache by class

mattst88:
_atom_ctor_key (added in 9d109000e) does not include the constructor's class, so a subclass's constructor can be handed back a cached instance built by a different class's constructor (Atom itself, or another subclass) for the same string and flags. Callers then get an object that is not an instance of the class they asked for, and methods the subclass overrides silently resolve to the base class's implementation instead.

Reproduced with gentoolkit.atom.Atom, a portage.dep.Atom subclass: constructing it with a string already interned elsewhere as a plain Atom returned that plain Atom, so gentoolkit's own intersects() override was never called.

Fix by including cls in the cache key. This keeps the cache correctly partitioned per class while leaving the interning behavior for repeat constructions of the same class unchanged. Benchmarked 40000 constructions across 2000 distinct atoms (5 trials): 0.104s before, 0.109s after, within noise.

Bug: https://bugs.gentoo.org/981521

### DIFF
--- a/lib/portage/dep/__init__.py
+++ b/lib/portage/dep/__init__.py
@@ -1606,7 +1606,7 @@ def _intern_atom(atom):
 
 
 def _atom_ctor_key(
-    s, unevaluated_atom, allow_wildcard, allow_repo, _use, eapi, allow_build_id
+    cls, s, unevaluated_atom, allow_wildcard, allow_repo, _use, eapi, allow_build_id
 ):
     """
     Cache key for Atom instances created via the Atom constructor. The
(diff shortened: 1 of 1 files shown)


mattst88:
This is an alternative fix for one of the problems reported in bug [981521](https://bugs.gentoo.org/981521), but I think we should land it in addition in case there are other pieces of code that subclass `Atom`.

zmedico (approved):

thesamesam (approved):

PR #1683, opened 2026-08-31
emerge: calculate dependencies in a child process

mattst88:
A dependency calculation allocates far more memory than the merge that follows it, and CPython cannot return that memory to the operating system: the objects that survive are spread across many partially used arenas. So the merge runs, for hours on a large build, with the peak of the calculation still resident. For `emerge -uDN @world`, dropping the depgraph, dropping the scheduler graph and calling `clear_caches()` leaves only 79000 objects and 21MB of live data, and RSS stays at 168MB.

Calculating in a child process avoids this, since the child exits and the kernel reclaims what it allocated. The child sends the scheduler graph over a socketpair as plain data; the parent rebuilds the Package instances, the digraph and the fake vartrees from it. The child inherits stdio, so resolution output and the --ask prompt behave as before. The parent ends up at 78MB rather than 168MB, and pays 4.1MB of pickle and 0.5s to rebuild the graph. The peak itself is unchanged, but it now belongs to a process that exits before the first package is built.

The child signals as soon as it owns stdio, which separates the two failure modes: before that point it has shown nothing and asked nothing, so the parent falls back to calculating in the current process; after it, the child may already have displayed the merge list and asked for confirmation, so the parent reports the failure rather than repeating any of it. The child also keeps the `SIGINT` and `SIGTERM` handlers it inherited, unlike `ForkProcess._bootstrap()`, because it holds the terminal and an interrupt has to reach `UserQuery` as a `KeyboardInterrupt` for the `--ask` prompt to quit the way it did before.

`--pretend` and `--resume` calculate in the current process, as does `PORTAGE_FORK_DEP_CALC=0`.

The first four commits are preparatory refactors with no behavior change.

Bug: https://bugs.gentoo.org/549906


lib/_emerge/_depgraph_fork.py
```diff
@@ -0,0 +1,576 @@
...
+    try:
+        parent_sock, child_sock = socket.socketpair()
+    except OSError as e:
+        raise ForkFailed(f"socketpair failed: {e}") from e
+
+    sys.stdout.flush()
+    sys.stderr.flush()
+
+    try:
+        pid = os.fork()
```
zmedico, line 492: Why not use multprocessing.Process instead of os.fork?

mattst88: The child process needs the `emerge_config` state that already exists in the parent, which is not pickle'able. Plain `os.fork` gives the child copy-on-write access to this for free with no serialization. Forcing `multprocessing`'s `fork` method directly could avoid that, but at that point we're not gaining anything from using `multiprocessing`.

It also allows us to take over stdio and signal handling so that `--ask` behaves exactly as if it was run in the parent.

zmedico: We could alternatively use the portage.process._start_fork function which has support to suppress https://bugs.gentoo.org/345289.

mattst88: Oh, I didn't know about that. Thank you! Done. (I also split `_run_in_child` out into a separate module).

thesamesam:
It looks fine to me but I haven't looked closely, I would prefer Zac review this one if possible.

zmedico (approved):

PR #1684, opened 2026-09-01
phase-functions.sh: don't filter ebuild-set QEMU_* variables

mattst88:
`__filter_readonly_variables()` determines the set of bash's own special variables by asking a clean instance of bash to report them. Since commit 7d73f9873 the QEMU_* variables of the calling environment are passed to that instance, so bash reports them back and they end up filtered from the saved ebuild environment.

That breaks any ebuild setting a QEMU_* variable of its own. `sys-firmware/edk2-bin` sets QEMU_TARGETS at global scope; it is stripped when the environment is saved after src_unpack(), leaving it empty for every subsequent phase and failing the build.

Exclude the injected variables from the reported ones, as PATH and SHELL already are. They remain visible to qemu-static.

Bug: https://bugs.gentoo.org/978685
Fixes: 7d73f9873 ("special_env_vars.py: preserve QEMU_* for qemu-static")

### DIFF
--- a/bin/phase-functions.sh
+++ b/bin/phase-functions.sh
@@ -86,12 +86,15 @@ portage_mutable_filtered_vars=( AA HOSTNAME )
 # is to preserve various variables as they were at the time that the binary
 # package was built while protecting against the application of package renames.
 __filter_readonly_variables() {
-	local -a filtered_vars bash_vars qemu_env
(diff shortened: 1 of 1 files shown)


chewi (approved):
Nice, thanks! That saves me from having to fix up the other two affected packages.

PR #1685, opened 2026-09-04
Make the scheduler and depclean cycle-aware

mattst88:
`strongly_connected_components()` lets the scheduler and `--depclean` cheaply test whether a package is part of a dependency cycle. This series adds that check where it was missing.

### Build cycle members in parallel

Bug [452172](https://bugs.gentoo.org/452172). The merge order calculation already breaks cycles by ignoring PDEPEND edges, but the scheduler treated every edge as blocking, so packages in a cycle were built one after another even with `--jobs`. The scheduler now uses `strongly_connected_components()` to find cycle members and ignores post-merge dependencies for them when deciding whether a build can start, matching what the merge order already assumes.

### Name the cycle when depclean can't remove a package

Bug [346351](https://bugs.gentoo.org/346351). `--depclean` with arguments only considers the packages given, and protects everything else installed. A package kept alive only by a dependency cycle looked unremovable, and the output listed the other cycle members as ordinary reverse dependencies without saying they're in the same cycle. Depclean now detects that case and prints the command that removes the whole cycle at once.

### Colorize the USE flags named in the cycle display

Color each flag named in `(buildtime, USE=...)` the same way `emerge -pv` already does elsewhere: red for enabled, blue with a `-` prefix for disabled ([#1681 comment](https://github.com/gentoo/portage/pull/1681#issuecomment-5536974214)). This lines up the flags with the colorized `Change USE:` suggestions printed below them.

`NEWS` is updated to match.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -44,6 +44,11 @@ Features:
   PORTAGE_CIRCULAR_MAX_USE_FLAGS), and is never empty (bug #929010, bug
   #877549). Add --circular-deps-report=json for automated consumers.
 
+* emerge: Packages in a dependency cycle no longer wait for each other's
+  post-merge dependencies during --jobs builds (bug #452172), and --depclean
+  now names the whole cycle when a package cannot be cleaned because of one
+  (bug #346351).
+
 portage-3.0.82.2 (2026-08-28)
 ----------------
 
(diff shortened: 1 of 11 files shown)


thesamesam (commented):
This looks OK but please wait for Zac before merging.

zmedico (approved):

mattst88:
Thanks guys!

PR #1686, opened 2026-09-05
BinpkgFetcher: fetch binpkgs as regular fetch user

thesamesam:
…userpriv)

Bug: https://bugs.gentoo.org/930179

### DIFF
--- a/lib/_emerge/BinpkgFetcher.py
+++ b/lib/_emerge/BinpkgFetcher.py
@@ -9,7 +9,9 @@ from urllib.parse import urlparse as urllib_parse_urlparse
 
 import portage
 from portage.binpkg import get_binpkg_format
+from portage.data import portage_gid, portage_uid, userpriv_groups
 from portage.exception import FileNotFound
+from portage.package.ebuild.fetch import _want_userfetch
 from portage.util._async.AsyncTaskFuture import AsyncTaskFuture
 from portage.util._async.FileCopier import FileCopier
 from portage.util._pty import _create_pty_or_pipe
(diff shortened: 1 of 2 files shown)


thesamesam:
A question is what to do about compatibility. Users might be using SSH for their binhost and rely on the keys being accessible as root.

EDIT: though FEATURES=userfetch can still be used, so users who are disrupted can just turn it off until they have time to sort out their config.

zmedico (approved):

lib/_emerge/BinpkgFetcher.py
```diff
@@ -118,6 +120,11 @@ async def _main(self) -> int:
                 if copier.returncode == os.EX_OK:
                     fetcher.sync_timestamp()
             else:
+                if _want_userfetch(self.pkg.root_config.settings):
+                    portage.util.ensure_dirs(
+                        os.path.dirname(self.pkg_path), uid=portage_gid, gid=portage_gid
+                    )
+
```
zmedico, line 127: We should probably set the group write bit on the created directory.

thesamesam: Maybe I was too clever and should've used bintree's `_ensure_dir` after all, as it handles that for us. Thanks, will fix.

zmedico: I've pushed a fix in b83ea973d09450bb09d01d85125e9713ca16e76f.

thesamesam: Oh, thank you! Sorry I hadn't got to it yet.

PR #1688, opened 2026-09-06
sync: drop privileges for gemato

thesamesam:
Wrap _import_key, _refresh_keys, and _get_openpgp_env with a ForkExecutor so we can drop privileges for FEATURES=usersync.

This achieves two things:
1) avoids running gpg as root;
2) the permissions of the keyring are then accessible to git running
  a as non-root too (bug #838223).

Bug: https://bugs.gentoo.org/877345


thesamesam:
@floppym Does this fix the issue you had in https://github.com/gentoo/portage/pull/818?

PR #1689, opened 2026-09-09
gpkg: reject archive members that escape through a symlink

mattst88:
tar_safe_extract.extractall() rejected members with an absolute path, a
".." component, a device node, or a hardlink pointing outside of the
archive, but it never looked at symlinks. An archive containing

    image/x -> /
    image/x/etc/cron.d/evil

therefore extracted the second member through the symlink created by the
first, writing outside of the temporary directory as root, before any of
the checks that guard the final move to dest_dir ever ran. A binary
package from a compromised or malicious binhost could overwrite any file
on the system.

Add the missing check to _check_member(): resolve the directory the
member will be created in and require it to stay inside the extraction
root. That catches the escape regardless of how many symlinks are
chained or whether the target is absolute or relative.

tarfile's own filters cannot be used for this: tarfile.data_filter
rejects absolute symlink targets and strips the setuid/setgid/sticky
bits that ordinary binary packages legitimately contain, so the
extraction keeps using fully_trusted_filter and validates members itself.

Also tighten the path traversal check, which only looked for a "../"
prefix and "/../" anywhere in the name, and so missed "./../evil".

Add tests covering each rejected shape, and one asserting that an
ordinary archive with a setuid file and an absolute symlink is still
extracted unchanged.

### DIFF
--- a/lib/portage/gpkg.py
+++ b/lib/portage/gpkg.py
@@ -676,6 +676,27 @@ class tar_safe_extract:
         self.closed = False
         self.file_list = []
 
+    def _check_member(self, member: tarfile.TarInfo):
+        """
+        Raise ValueError if member is not safe to extract.
+        """
+        name = member.name
+        if (name in self.file_list) or (os.path.join(".", name) in self.file_list):
+            writemsg(colorize("BAD", f"Danger: duplicate files detected: {name}\n"))
+            raise ValueError("Duplicate files detected.")
+        if name.startswith("/"):
+            writemsg(colorize("BAD", f"Danger: absolute path detected: {name}\n"))
+            raise ValueError("Absolute path detected.")
+        if name.startswith("../") or ("/../" in name):
+            writemsg(colorize("BAD", f"Danger: path traversal detected: {name}\n"))
+            raise ValueError("Path traversal detected.")
+        if member.isdev():
+            writemsg(colorize("BAD", f"Danger: device file detected: {name}\n"))
+            raise ValueError("Device file detected.")
+        if member.islnk() and (member.linkname not in self.file_list):
+            writemsg(colorize("BAD", f"Danger: hardlink escape detected: {name}\n"))
+            raise ValueError("Hardlink escape detected.")
+
     def extractall(self, dest_dir: str):
         """
         Extract all files to a temporary directory in the dest_dir, and move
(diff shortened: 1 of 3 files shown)


thesamesam:
cc @RinCat

lib/portage/gpkg.py
```diff
@@ -684,8 +721,8 @@ def extractall(self, dest_dir: str):
         if self.closed:
             raise OSError("Tar file is closed.")
         temp_dir = tempfile.TemporaryDirectory(dir=dest_dir)
-        # The below tar member security checks can be refactored as a filter function
-        # that raises an exception. Use tarfile.fully_trusted_filter for now, which
+        # Members are validated by _check_member() before extraction, so
+        # the extraction itself must not mangle them. fully_trusted_filter
         # is simply an identity function:
```
thesamesam, line 729: Can we switch to one of tarfile's filters to also avoid people assuming there's a problem here? And defence in depth, though it does seem to often have bugs in CPython too :)

EDIT: Ah, nevermind:
> tarfile's own filters cannot be used for this: tarfile.data_filter
> rejects absolute symlink targets and strips the setuid/setgid/sticky
> bits that ordinary binary packages legitimately contain, so the
> extraction keeps using fully_trusted_filter and validates members itself.

Perhaps that should be a comment in place of the existing one then.

eli-schwartz:
> therefore extracted the second member through the symlink created by the first, writing outside of the temporary directory as root, before any of the checks that guard the final move to dest_dir ever ran. A binary package from a compromised or malicious binhost could overwrite any file on the system.

Does that include running before PGP signature verification?

RinCat:
> > therefore extracted the second member through the symlink created by the first, writing outside of the temporary directory as root, before any of the checks that guard the final move to dest_dir ever ran. A binary package from a compromised or malicious binhost could overwrite any file on the system.
> 
> Does that include running before PGP signature verification?



> > therefore extracted the second member through the symlink created by the first, writing outside of the temporary directory as root, before any of the checks that guard the final move to dest_dir ever ran. A binary package from a compromised or malicious binhost could overwrite any file on the system.
> 
> Does that include running before PGP signature verification?

No, the step is after PGP signature verification.

RinCat:
I found more edge cases, do you want to me fix them all in one PR or merge this first?

thesamesam:
I'd say:
1) @mattst88, could you rewrite the description to not refer to malicious archives, just corrupted or unusual (e.g. unexpected tar options used, or a buggy implementation)? I'd note explicitly that PGP verification is done before any of this in the commit message so people don't worry, then
2) once that's done, we merge this as it looks OK to me, and @RinCat opens their own with followups?

Thank you both!

RinCat:
My PR is up https://github.com/gentoo/portage/pull/1690

mattst88:
> @mattst88, could you rewrite the description to not refer to malicious archives, just corrupted or unusual (e.g. unexpected tar options used, or a buggy implementation)? I'd note explicitly that PGP verification is done before any of this in the commit message so people don't worry, then

Sure thing. Done.

lib/portage/gpkg.py
mattst88: Good call. Added.

eli-schwartz:
Thanks for fixing the commit message. My concern was that we have a security boundary for when to trust remote content (it has to be code-signed) and anything else can already "do anything" at merge time.

Adding additional checks for wonky / corrupted archives is still useful, but it is more of a QA check. We shouldn't advertise that we've fixed a security issue because (in addition to not being a problem that requires an urgent upgrade) it would confuse people about what guarantees they can actually rely on when installing unsigned gpkg archives.

thesamesam:
Merged via https://github.com/gentoo/portage/pull/1690. I plan to write some commentary about it not being a security issue in the bug I've filed at https://bugs.gentoo.org/982208.

PR #1690, opened 2026-09-09
gpkg: prevent symlink and hardlink escape

RinCat:
Prevent specially crafted or placed symlinks and hardlinks from causing writes outside the staging directory

### DIFF
--- a/lib/portage/gpkg.py
+++ b/lib/portage/gpkg.py
@@ -676,6 +676,27 @@ class tar_safe_extract:
         self.closed = False
         self.file_list = []
 
+    def _check_member(self, member: tarfile.TarInfo):
+        """
+        Raise ValueError if member is not safe to extract.
+        """
+        name = member.name
+        if (name in self.file_list) or (os.path.join(".", name) in self.file_list):
+            writemsg(colorize("BAD", f"Danger: duplicate files detected: {name}\n"))
+            raise ValueError("Duplicate files detected.")
+        if name.startswith("/"):
+            writemsg(colorize("BAD", f"Danger: absolute path detected: {name}\n"))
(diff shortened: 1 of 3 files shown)


lib/portage/tests/gpkg/test_gpkg_path.py
```diff
@@ -369,3 +372,321 @@ def test_gpkg_long_filename(self):
...
+                    [
+                        ("image/link", tarfile.SYMTYPE, str(marker)),
+                        ("image/alias", tarfile.SYMTYPE, "."),
+                        ("image/alias/link", tarfile.REGTYPE, b"overwritten"),
+                    ],
+                )
+            self.assertEqual(marker.read_bytes(), b"original")
+
+    def test_contained_symlink_parent(self):
+        for mode in ("r|", "r:"):
```
thesamesam: You can use itertools.product to drop a level of indentation here.

thesamesam (commented):
Could you rebase this on top of Matt's work so we keep credit for what he did too? Or just wait until his is merged then rebase?

Or do you prefer that we close Matt's and merge this? I'd like to avoid that option if possible.

lib/portage/tests/gpkg/test_gpkg_path.py
RinCat: done

RinCat:
That is fine, rebased.

mattst88:
Please line wrap the commit message ("Prevent specially crafted or placed symlinks and hardlinks from causing writes outside the staging directory") to 72 characters.

Otherwise this looks great to me. Nice finds!

RinCat:
Fixed and rebased

PR #1693, opened 2026-09-12
BinpkgVerifier: refuse a remote package that has no digests

mattst88:
_get_digests() returns whatever the Packages index recorded for the package. With no digests at all, or with no recorded SIZE, the verifier returned EX_OK without hashing anything, so the package was merged with nothing checked.

Using a binhost means trusting it either way, but an entry missing either field turns the check off quietly rather than saying so, and the index is not itself verified against anything. A remote gpkg still has its signature checked, which is the default for remote binpkgs; for a remote xpak these digests are all there is.

Fail instead, for a remote package whose index entry does not carry both a size and at least one checksum. Local packages keep the old behavior: PKGDIR is under the administrator's control, and a missing index entry there is a configuration problem.

Package.remote is used rather than bintree.isremote() because Package already has it, and because it is the value from before the package was fetched, which is the one that matters here.

Add tests covering each case against a stubbed bintree.


thesamesam:
> Using a binhost means trusting it either way, but an entry missing either field turns the check off quietly rather than saying so, and the index is not itself verified against anything. A remote gpkg still has its signature checked, which is the default for remote binpkgs; for a remote xpak these digests are all there is.

Indeed, see https://bugs.gentoo.org/877241 too, but it's not a problem if we require signing. Nonetheless, I agree it's worth fixing.

lib/_emerge/BinpkgVerifier.py
```diff
@@ -128,6 +136,19 @@ def _display_success(self):
             out.getvalue(), log_path=self.logfile, background=self.background
         )
 
+    def _missing_digests_exception(self):
+        self.scheduler.output(
+            "\n!!! Digest verification failed:\n"
+            f"!!! {self._pkg_path}\n"
+            "!!! Reason: The binhost's Packages index records no size and\n"
```
thesamesam: Just say "Reason: binhost's Packages index lacks size and checksum for this package." or similar, no need to say the rest

lib/portage/tests/emerge/test_binpkg_verifier.py
```diff
@@ -0,0 +1,65 @@
...
+class _FakePkg:
+    def __init__(self, digests, remote, cpv="dev-libs/A-1"):
+        self.root_config = _FakeRootConfig(_FakeBintree(digests))
+        self.cpv = cpv
+        self.remote = remote
+
+
+class BinpkgVerifierMissingDigestTestCase(TestCase):
+    """
+    An index entry without both a SIZE and a digest leaves nothing for
```
thesamesam: Just say "Verify we need both SIZE and digests for remote packages."

thesamesam (commented):
OK with these fixed, but see my comment above https://github.com/gentoo/portage/pull/1693#issuecomment-5647092582 for some useful colour too.

thesamesam (approved):
Thanks!

PR #1694, opened 2026-09-12
gpkg: verify the image against the Manifest while extracting it

mattst88:
_verify_binpkg() hashes the gpkg file's members and closes it. decompress()
then reopens the file by path and extracts it, so the Manifest check does not
cover the bytes that get extracted.

This is not exploitable today: PKGDIR is documented as trusted and binpkgs in
it are root-owned (EbuildBinpkg creates the tmpfile in the privileged process,
and the package phase is not run under userpriv). My original description
claiming PKGDIR is portage-writable was wrong. It is DISTDIR that has the
weaker ownership story. https://bugs.gentoo.org/982441 has the discussion of
how much we want to rely on that.

The image tar now goes through a hashing reader, and extractall() takes a
verify callback that runs after the last member is extracted and before
anything is moved into the destination, so a mismatch raises DigestException
with nothing installed.

The image ends up hashed twice per merge. Collapsing that means moving the
image's signature verification into the extraction-time pass, which I left
for later.

The first two commits are a dead `except KeyError` in the same code and the
factoring the third one needs.


thesamesam:
> PKGDIR is writable by the portage user and builds run concurrently, so if the file changes in between, the
new contents are extracted with their checksums never checked.

I'm not sure this is currently something we worry about. We may have similar problems with DISTDIR. Nonetheless, I think removing TOCTOU issues is worthwhile. I would also like us to get to a point where we can trust them more, so that we can trust `portage` a bit less. See https://bugs.gentoo.org/915330 too.

Would it perhaps be better to open an fd for the binpkg, verify it, and then keep it open the whole time?

EDIT: guess that may not be possible because of https://bugs.gentoo.org/915330#c11

thesamesam:
cc @rincat @mgorny

RinCat:
I think this was discussed when implanting it. The conclusion is that if something can modify the downloaded files owned by root or portage, then it is already unsafe by definition.

thesamesam:
I would like to get to a point where we can trust `portage` less because I want a malicious ebuild running as non-root, but not yet installing any files, to not be able to interfere with DISTDIR and PKGDIR, even if unlikely. Another example could be a compromised fetch process that overwrites files it shouldn't.

But that is more likely to be achieved via running build steps as a separate user (`portagebuild`, see https://bugs.gentoo.org/600804 too).

thesamesam:
This is probably worth filing a bug for so we can discuss there and have the discussion preserved.

@mattst88 can you do that for this (and anything else which seems like it may invite discussion)? Thank you!

RinCat:
>  I want a malicious ebuild running as non-root, but not yet installing any files, to not be able to interfere with DISTDIR and PKGDIR

Did this got changed somehow? I think when I design it the binpkg owned by root, so build package under portage user cannot write it

thesamesam:
No, looking at it, I think you're right. DISTDIR has this problem but not PKGDIR.

However, see https://bugs.gentoo.org/915330#c2 and the rest of that bug. We documented PKGDIR to be trusted. I don't recall if it was because of unsigned binpkgs or something else.

RinCat:
I think `PKGDIR` is documented as a trusted location. because either:

* you built or copied it yourself
* you downloaded it from a trusted remote source and verified it with a hash or signature
* it is owned by root, so if an attacker is able to modify it, the system is already compromised

mattst88:
Filed bug 982441. Added `Bug: ...` to the commits and updated the PR summary text.

thesamesam (commented):
It looks OK to me and it seems reasonable to do to be more robust and make it clearer what we do (and don't do) but please let @RinCat check.

It's a shame it'll slow down merging binpkgs a little bit though.

RinCat:
I am ok, just hash twice. This used to be a cost we were unwilling to pay. But with current hardware, it should be fine. You'll just run into more user complaints that binpkg installations are slower. 😹

RinCat:
Also I do not think you should do "moving the image's signature verification into the extraction-time pass". if there are something bad happened during extracting, the damage already done.

mattst88:
> Also I do not think you should do "moving the image's signature verification into the extraction-time pass". if there are something bad happened during extracting, the damage already done.

Right, the signature has to be checked before any of the tar is parsed, so that stays a separate pass and the image just gets hashed twice. I've dropped the "left for later" plan from the commit message and the PR description.

Rebased on master.

eli-schwartz:
I do not see the point of checking this *at all*, since checksums exist solely for sanity checking against corruption.

Unless the gpkg is PGP-signed, which this PR doesn't use as a precondition.

PR #1695, opened 2026-09-14
tests: use zstd for the 10 GiB gpkg size test

mattst88:
test_gpkg_large_size only checks the pre-image scan and that the container needs the GNU tar format, so the compressor does not matter. gzip spent about 14s compressing 10 GiB of zeros, or about 54s of CPU across every core with pigz, which starved the rest of a parallel run. zstd, which ResolverPlayground already requires, takes about 4s.


thesamesam (approved):
Should we do `BINPKG_COMPRESS_FLAGS_ZSTD="-1"` or `BINPKG_COMPRESS_FLAGS_GZIP="-0"` perhaps?

I don't really care which, not like we need to avoid zstd, but it's cheap to add either way?

Also, great find, thank you!

mattst88:
Oh good idea. I'll check how those affect the wall time.

mattst88:
I gave the `BINPKG_COMPRESS_FLAGS_ZSTD="-1"` and couldn't measure any difference. Zeros compress really quickly regardless of the level, I guess.

`BINPKG_COMPRESS_FLAGS_GZIP="-1"` speeds up gzip, but it's still slower the zstd.

thesamesam:
Thanks for checking!

PR #1696, opened 2026-09-14
locale: look up libc before forking the locale check

mattst88:
async_check_locale() forks a child for each locale not already in its cache, and the child calls load_libc(), whose find_library() lookup spawns "ldconfig -p" again in every child. egencache runs metadata phases in parallel, and they all miss the per-locale cache until the first check has finished, so each EAPI 6 or later ebuild spawned its own ldconfig.

Look up libc in the parent before forking. An egencache run over 20 EAPI 8 ebuilds now spawns one ldconfig instead of 20 and uses about 10% less CPU time. ResolverPlayground runs egencache when it creates ebuilds, so those tests benefit too.


thesamesam (approved):
If you look at how `find_library` in Python works, it's completely nuts. It even invokes both `ld` and `objdump`. We should probably move away from it, as I believe they've deprecated it now too, because they didn't want to fix it for musl.

Anyway, lovely, thanks!

PR #1698, opened 2026-09-14
Fix EbuildFetcher cancellation hang in test_fetch interrupt test

mattst88:
Cancelling a fetch just after wget is spawned can hang the fetch child until wget gives up on its own (-t 3 -T 60). The interrupt test in `test_fetch.py` hit this in 5 of 25 runs.

Two causes: `ForkProcess` children could lose a `SIGTERM` that arrived before their handlers were reset, and `EbuildFetcher`'s `join(timeout)` blocks in `waitpid()` once the child has exec'd, so the `SIGKILL` fallback never ran. Either fix alone stops the hang; with both, 0 of 25 runs hung.


thesamesam:
Could you tag https://bugs.gentoo.org/982512?

zmedico (approved):

PR #1699, opened 2026-09-15
Scheduler: fetch `pkg_pretend` binary packages one at a time

mattst88:
Since pkg_pretend phases run in parallel, each job that needs a binary package starts its own download, so with --jobs > 1 several binpkgs are fetched at once and their progress output overlaps on the terminal.

Queue those fetches so they run one at a time, with output in emerge-fetch.log, and print the "Fetching in the background" notice right away instead of after pkg_pretend finishes.

A job waiting on the fetch queue still holds its --jobs slot, so other pkg_pretend phases can wait behind a slow download.

Bug: https://bugs.gentoo.org/579526


zmedico (approved):

thesamesam (commented):
I hope the feature can return as a real feature :)

PR #1701, opened 2026-09-15
observability: sort "emerge --status" tasks by phase lifecycle

Flowdalic:
When running parallel builds, tasks in emerge --status are rendered in whatever order they appear in the snapshot (currently sorted chronologically by task start_time). This causes packages in different lifecycle phases to be mixed together.

Sort tasks in format_snapshots() using a forward lifecycle phase order:

build/setup -> unpack -> compile -> install
            -> merge-wait -> merge -> postinst

Within each phase, order tasks by the timestamp they entered that phase (phase_start_time), keeping the visual order of tasks within a phase constant and stable. Fall back to package CPV as a tie-breaker.


lib/_emerge/_observability.py
```diff
@@ -483,6 +551,7 @@ def note_task_finished(self, task):
         cpv = str(pkg.cpv)
         if isinstance(task, _PackageMerge):
             self._phases.pop(cpv, None)
```
mattst88: `_phases` is only cleared when the merge task finishes, so between `_build_exit()` and the `preinst` notification the entry still holds the build's last phase (`install`, `instprep`, or `package` under `buildpkg`). Previously that was a stale label in the phase column; now it also picks the group, so a package sitting in `_task_queues.merge` renders among the still-compiling ones. The `waiting` override at line 156 covers `merge-wait` but not plain merges.

Popping `_phases` in the `else` branch here makes the row fall back to `kind`, which ranks correctly. `_phase_start` needs nothing: `note_task_started(merge)` sets it a moment later.

lib/_emerge/_observability.py
```diff
@@ -355,6 +361,63 @@ def format_resources(resources, build_elapsed=None):
...
+    # Maintenance & misc
+    "config",
+    "info",
+    "other",
+)
+_PHASE_RANKS = {phase: idx for idx, phase in enumerate(_PHASE_LIFECYCLE_ORDER)}
+_DEFAULT_PHASE_RANK = len(_PHASE_LIFECYCLE_ORDER)
+
+
+def _task_phase_sort_key(task):
```
mattst88: `phase = task.get("phase") or task.get("kind") or "-"` is now duplicated at line 443. A `_task_phase(task)` helper would keep the sort and the rendered column from drifting.

lib/_emerge/_observability.py
```diff
@@ -373,7 +436,8 @@ def format_snapshots(snapshots):
                 failed=jobs.get("failed", 0),
             )
         )
-        for task in snapshot.get("tasks", []):
+        tasks = sorted(snapshot.get("tasks", []), key=_task_phase_sort_key)
```
mattst88: `build_snapshot()` still sorts by `start_time` at line 166, so the JSON on the socket and the rendered table disagree. Deliberate? Sorting there instead would give every consumer the same order and let this line go back to plain iteration.

lib/_emerge/_observability.py
```diff
@@ -355,6 +361,63 @@ def format_resources(resources, build_elapsed=None):
     return ", ".join(parts)
 
 
+# Note: portage.const.EBUILD_PHASES also defines ebuild phases in rough
```
mattst88: Can `rpm` and `other` reach a snapshot? `other` looks like an elog phase rather than anything `notifyPhase` reports, and `rpm` only runs under the standalone `ebuild` command, which has no `notifyPhase`.

lib/_emerge/_observability.py
```diff
@@ -140,6 +140,11 @@ def build_snapshot(monitor):
         else:
             elapsed = None
 
+        if waiting and build_finished is not None:
```
mattst88, line 205: `note_task_started(merge)` runs in `_build_exit()` right after `times.finished` is set, so `_phase_start[cpv]` already holds the same value give or take microseconds. If the exact build-completion timestamp is the point, say so in the comment; `test_snapshot_merge_wait_phase_start_time_is_build_finished` is the only thing pinning it.

mattst88: `_build_exit()` calls `note_task_started(merge)` (Scheduler.py:1815) before appending to `_merge_wait_queue` (line 1822), so `_phase_start[cpv]` is already set by the time `waiting` can be true. The branch is redundant unless the exact build-completion timestamp matters; if it does, say that in the comment instead of restating what merge-wait means.

mattst88:
1. Commit message and PR description say the sort is in `format_snapshots()`; it is in `build_snapshot()` now. The commit also adds `phase_start_time` and clears `_phases` for builds, neither of which is mentioned.
2. `_SCHEMA_VERSION` is still 1. The new field is additive, but the task order in the JSON changes too, which consumers will see.
3. Pre-existing: `_phases` and `_phase_start` are keyed by cpv alone, so two tasks for the same cpv in different ROOTs collide. That used to mislabel the phase column; now it also picks the sort position.

lib/_emerge/EbuildPhase.py
```diff
@@ -148,7 +148,9 @@ def _start(self):
         # runs phases (e.g. the standalone `ebuild` command) provides it.
         notify_phase = getattr(self.scheduler, "notifyPhase", None)
         if notify_phase is not None:
-            notify_phase(self.settings.mycpv, self.phase)
+            notify_phase(
+                self.settings.mycpv, self.phase, self.settings.get("ROOT", "/")
+            )
```
mattst88: `pkg.root` is `root_config.root`, which is `settings["EROOT"]` (RootConfig.py:20), and every other `_pkg_key()` is built from `pkg.root`. Passing `ROOT` here means that with a non-empty EPREFIX, `note_phase()` writes under a key `build_snapshot()` never looks up. Prefix users would lose the phase column and every task would sort as `build`/`merge`.

```suggestion
            notify_phase(self.settings.mycpv, self.phase, self.settings["EROOT"])
```

mattst88 (commented):
The earlier inline comments are addressed.

1. The commit message says `build_snapshots()`; the function is `build_snapshot()`. It also doesn't mention the new `phase_start_time` field, the schema bump, the `(cpv, root)` keying, or the extra `notifyPhase` argument. The diagram leaves out `merge` between `merge-wait` and `preinst`, and "as a tie-breakers" should be "as tie-breakers".
2. The PR description still says `format_snapshots()` and shows the old diagram.
3. The `(cpv, root)` keying fixes an existing bug independently of the sort. Could it be its own commit, before this one?
4. The merge-wait `phase_start` branch in `build_snapshot()` still looks redundant, for the reason in the earlier thread. Not a blocker.

lib/portage/tests/emerge/test_observability.py
```diff
@@ -645,6 +648,218 @@ def test_zero_valued_resources_are_still_reported(self):
...
+                        "cat/setup-1",
+                        "cat/compile-early",
+                        "cat/compile-late",
+                        "cat/merge-1",
+                    ],
+                )
+            finally:
+                monitor.close()
+
+    def test_multi_root_same_cpv_no_collision(self):
```
mattst88, line 821: This calls `note_phase(..., root="/")` directly, so it can't catch a mismatch between the root `EbuildPhase` reports and `pkg.root`. Could a test go through the `notifyPhase` path with an EPREFIX set, so that EROOT differs from ROOT?

lib/_emerge/_observability.py
```diff
@@ -97,6 +102,67 @@ def elapsed(self, now):
...
+    "setup",
+    "nofetch",
+    # Build & compilation
+    "unpack",
+    "prepare",
+    "configure",
+    "compile",
+    "test",
+    "install",
+    "instprep",
```
mattst88, line 125: `instprep` runs at merge time, not build time. `dblink.treewalk()` spawns it just before `preinst` (vartree.py:4371), so the merge task reports it. Ranked here, a merging package moves up above rows still in `package` and `merge-wait` while `instprep` runs.

I think it belongs after `"merge"`. The diagram in the commit message would need the same fix.

lib/_emerge/_observability.py
```diff
@@ -97,6 +102,67 @@ def elapsed(self, now):
...
+# lifecycle order. However, _PHASE_LIFECYCLE_ORDER is maintained separately
+# because observability tracks scheduler states and fallback task kinds
+# ("merge-wait", "build", "merge") as well as clean phases ("clean",
+# "cleanrm") that are not in EBUILD_PHASES, and places "nofetch" before
+# "unpack" rather than after "postrm".
+_PHASE_LIFECYCLE_ORDER = (
+    # Pre-build / early setup
+    "build",
+    "pretend",
+    "clean",
```
mattst88, line 115: `clean` doesn't only run first. It also runs after `postinst` (vartree.py:6148) and after `package` under `--buildpkgonly` (EbuildBuild.py:530). Both go through `notifyPhase`, so a package that has just finished merging moves to the top of the list until its task exits.

The rank can't distinguish the two cases because it only sees the phase name. `note_phase()` can: the pre-clean is always the first notification for a key. Something like this would do it:

```python
if phase == "clean" and key in self._phases:
    return
```

A same-version reinstall has a smaller version of this problem. The replaced instance runs `prerm`/`postrm` under the same `(cpv, root)`, so the row goes past `postinst` and then back.

lib/_emerge/_observability.py
```diff
@@ -40,7 +40,12 @@
 
 from _emerge.PackageMerge import PackageMerge as _PackageMerge
 
-_SCHEMA_VERSION = 1
+_SCHEMA_VERSION = 2
+
+
+def _pkg_key(cpv, root=None):
```
mattst88, line 46: Every non-test caller now passes `root`. The `"/"` fallback is what let the ROOT/EROOT mismatch fail silently: the lookup fell through to a key nothing had written. If `root` were required here and in `note_phase()`, `note_build_resources()` and `build_elapsed()`, the next mistake like that would raise an error instead of blanking the phase column.

mattst88 (commented):
Still open from the last round:

1. The commit message diagram still leaves out `merge` between `merge-wait` and `preinst`. It also still doesn't mention `phase_start_time`, the schema bump, the `(cpv, root)` keying, or the new `notifyPhase` argument.
2. The PR description still says `format_snapshots()` and shows the old diagram.
3. The `(cpv, root)` keying would still be better as its own commit, placed before this one.

PR #1702, opened 2026-09-15
ctypes: simplify load_libc()

floppym:
Avoid the need to figure out libc's soname by passing None as the library name. This provides a handle representing the "global symbol table", which will include Python binary and all shared libs loaded at startup.

We risk a conflict if a called symbol is defined in more than one loaded object, but that seems unlikely for most libc-provided functions we would need to call.


zmedico (approved):

thesamesam (approved):
Perhaps note that we're now dropping the use of a soft-deprecated function too (https://docs.python.org/3.15/library/ctypes.html#ctypes.util.find_library). It's broken on musl and does weird things anyway (spawns objdump).

ormaaj:
I always thought it was a special feature but it's really just normal `dlopen` with NULL `path`.

Nice

PR #1703, opened 2026-09-16
tests: lower the passphrase KDF cost of the test signing keys

mattst88:
The two test secret keys were protected with 37281792 SHA-1 S2K iterations, which gpg-agent re-derived for every signature the tests make. A gpg --detach-sign with either key took about 77 ms; it now takes about 31 ms.

The suite signs 1412 times, and running it with -n auto on a 20 core machine saves about a minute and a half of CPU time: over five alternating runs of each, the median system-wide CPU time went from 1040s to 934s and the median wall time from 104s to 97s.

Re-protect both keys with the same passphrase at the minimum count:

```
  echo s2k-count 65536 > "$GNUPGHOME"/gpg-agent.conf
  gpg --batch --pinentry-mode loopback --passphrase GentooTest \
      --passwd <keyid>
```
The key material and the public keyring are unchanged. The key files are now in the extended format with openpgp-s2k3-ocb-aes protection, which gpg-agent has been able to read since GnuPG 2.1.12.


thesamesam (approved):
🚀

PR #1705, opened 2026-09-17
tests: share generated playground metadata within a test run

mattst88:
Every ResolverPlayground runs egencache, which runs a depend phase per ebuild.
A run of the suite creates 1720 playground ebuilds, of which 1024 are distinct.

This series caches the files that egencache generates, keyed by what they are
generated from, and restores them into later playgrounds of the same run
(below pytest's basetemp, so xdist workers share it). A playground whose files
are all cached needs no egencache run at all. 660 depend phases per run are
skipped.

CPU time for the suite with `pytest -n 20`, best of two runs on 20 cores.
Wall time is dominated by the slowest emerge tests and does not change
measurably.

| | CPU time |
|-| -------- |
| master | ~1050s |
| share metadata between playgrounds | ~950s |
| skip egencache when fully cached | ~905s |

`PORTAGE_TEST_VERIFY_METADATA_CACHE=1` restores nothing, lets egencache
generate every file and compares the two, so the assumptions behind the cache
are checkable rather than only documented. A full run under it passes.
TEST-NOTES describes both.


mattst88:
Force-pushed a rewrite. Against the previous version:

- Dropped the persistent `.pytest_cache` commit. It pays off only on repeated
  local runs against unchanged sources, never on CI or the first run after an
  edit, and costs about 50M per source tree. Sending it separately so this part
  can be judged on its own.
- Dropped the forked-egencache commit, for the reason I gave when posting it:
  it shares the parent's event loop and multiprocessing state, and does not
  help wall time.
- Added a commit that skips egencache entirely for a playground whose files are
  all cached, rather than only saving the depend phases.
- Added `PORTAGE_TEST_VERIFY_METADATA_CACHE`.

That last one found a bug in the previous version. A depcachedir entry names
the eclass directory it was generated from, so a playground with eclasses was
restoring another playground's temporary path. Tests still passed, because
portage revalidates and regenerates, but once egencache is skipped nothing
revalidates. Playgrounds that define eclasses, or that configure the auxdb
module, now share nothing.

All numbers re-measured for this version.

zmedico (approved):

thesamesam (commented):
I'm wondering if we want to exempt tests like `lib/portage/tests/dbapi/test_portdb_cache.py` where we're deliberately testing metadata generation as opposed to generating it to use later.

I also wonder if this might mask problems like 9c63c4541ffb891dcb6cbf94c8f5ad8d643c684f or make them harder to debug.

mattst88:
Both fair.

`test_portdb_cache.py` and `test_auxdb.py` were already excluded, but only because one defines eclasses and the other sets the `modules` user config. That is fragile, so `ResolverPlayground` now takes `share_metadata=False` and both tests pass it, which puts the reason at the call site.

On masking: an egencache failure is still fatal, since the playground checks the return code, but you are right that the cache reduces how often we run egencache and so how often CI can hit something like bug 965132. CI now runs one job with `PORTAGE_TEST_VERIFY_METADATA_CACHE=1`, which restores nothing and runs egencache for every playground, so that coverage stays and the cache's assumptions get checked at the same time.

Reproducing a failure should not get harder: the cache lives below the current run's pytest basetemp and is discarded with it, so rerunning a single test or file starts cold and behaves as it did before.

Your comment also turned up a hole. Verification only compared files for a fully cached repository, so a depcachedir entry seeded into a partially cached one was never checked. Every cached entry is compared now.

thesamesam (approved):
Thanks!

PR #1706, opened 2026-09-18
observability: serve the status socket without asyncio streams

mattst88:
emerge restores SIGPIPE to SIG_DFL, but asyncio's transports send without MSG_NOSIGNAL and rely on Python ignoring SIGPIPE. A write to a status socket client that has already disconnected therefore killed emerge outright.

"watch portageq jobs" hits this readily: read_snapshots() gives up on a socket after half a second, so whenever emerge's main loop is busy for longer than that, the snapshot it then sends goes to a closed peer. A client that disconnects just before a broadcast does the same. asyncio offers no way to reach the transport's socket, since get_extra_info() yields a TransportSocket, which does not expose send().

So serve the socket without asyncio's streams: accept from a non-blocking listening socket watched with add_reader, and send each snapshot with MSG_NOSIGNAL, so that a departed client is an EPIPE to handle rather than a signal. Nothing is buffered; a snapshot is far smaller than the socket buffer, so a client that cannot take one whole is dropped. Owning the listening socket also allows restricting it to 0600 before listen() rather than after, closing the window in which it was reachable with whatever permissions the umask gave it.

Accepting by hand brings back an error asyncio handled: a connection that cannot be accepted for want of a descriptor leaves the listening socket readable, so accepting again at once spins until one comes free. The socket is left unwatched for a second instead, as asyncio does for the same errnos, though silently, in keeping with the rest of the module.

The test suite runs with SIGPIPE ignored, which is why this went unnoticed. The new tests install a handler that records the signal and cover both a client gone before it is accepted and one gone before a broadcast.

Bug: https://bugs.gentoo.org/982689


thesamesam:
cc @parona-source

parona-source:
Looks to be working so far.

Edit: Still working, survived the rest of the update with portage jobs running in the background.

PR #1707, opened 2026-09-20
tests: keep Atom out of subTest() kwargs

mattst88:
test_equality_with_foreign_type() passed an Atom as a subTest() value. pytest 9 records subTest() kwargs in the report it emits, and under pytest-xdist that report is serialized with execnet, which dispatches on the exact type and handles only builtin types. Atom is not one, so execnet raises DumpError, failing the test:
```
  execnet.gateway_base.DumpError: can't serialize <class 'portage.dep.Atom'>
```
pytest 9.1 works around this by applying saferepr() to the kwargs. Pass the repr instead, so the test does not depend on the pytest version.


thesamesam (approved):
Huh, OK.

eli-schwartz (approved):
Thanks for investigating this.

PR #1710, opened 2026-09-22
Fix leaked resources that broke testMoveEntWithCorruptIndex

mattst88:
`testMoveEntWithCorruptIndex` turned every warning into an error, so it failed when the garbage collector found resources that earlier tests on the same xdist worker had leaked: an event loop, a Packages.gz source file, a sandbox log temp file, and an HTTP server socket.

The leaks are old, but since 28e396959 ("tests: interleave modules under pytest-xdist") tests from other modules run just before this one much more often, so CI hits the failure more.

The first four commits fix the leaks. The last one limits the test's error filter to the gpkg warning it checks for.

The EbuildMetadataPhase commit also fixes a real bug: the sandbox violation check could stat another phase's SANDBOX_LOG.


thesamesam (approved):
Thanks!

PPN-SD:
Hi @mattst88, since this PR is merged I get some warnings about the sandbox:

```
sandbox:setup_sandbox  could not unlink old log file: /tmp/sandbox3ysswa0x: Operation not permitted
sandbox:main  failed to setup sandbox
```

I can’t figure out why, but some ebuilds (in local repos) can no longer even respond to a direct command:
```
ebuild --debug libnfs-8.0.0.ebuild clean prepare
sandbox:setup_sandbox  could not unlink old log file: /tmp/sandboxzm2yk6sk: Operation not permitted

sandbox:main  failed to setup sandbox
```

I use portage-9999 in Incus containers only for package testing, so there may be several factors involved that I haven’t fully identified yet. What I do know is that I wasn’t seeing these warnings before.

PPN-SD:
Sorry, I just saw https://github.com/gentoo/portage/pull/1713

PR #1711, opened 2026-09-22
Make ebuild-ipc cheaper

mattst88:
Every ebuild phase runs `ebuild-ipc` at least once to report its exit status, and `has_version` and `best_version` run it on every call. Most of that time went to importing portage.

- Phases now write their exit status to a pipe that `AbstractEbuildProcess` passes in (`PORTAGE_EBUILD_EXIT_FD`). The orphan-process timer (bug #278895) and the check for a phase that exits without a status (bug #190128) still work. If the fd is not set, bash falls back to `ebuild-ipc`.
- ebuild-ipc.py uses only the standard library now. It checks that the daemon is alive by polling a pipe (`PORTAGE_IPC_ALIVE_FD`) instead of taking the build directory lock, and it takes `.ipc/lock` with a plain `flock()`. Its messages are no longer translated.
- bash calls `ebuild-ipc.py` directly with `python -S`, not through the `ebuild-pyhelper` wrapper, and the `bin/ebuild-ipc` symlink is gone.
- `EbuildIpcDaemon` used to cut off any reply larger than the pipe buffer. It now finishes sending such replies from the event loop.
- New tests run the client against a real daemon.

Behavior change: a client started without `PORTAGE_IPC_ALIVE_FD` reports "daemon process not detected". Anything that spawns `ebuild-ipc` outside `AbstractEbuildProcess` has to pass that fd.

`ebuild <pkg> clean install`, best of 6 wall-clock runs, no sandbox:

| | master | this branch |
| - | - | - |
| empty EAPI 8 ebuild                             | 2.62s  | 1.69s |
| same, with 100 has_version calls in src_compile | 12.08s | 3.79s |


thesamesam:
There was some interesting work on this a while ago at https://github.com/mid-kid/portage/commits/binpkg-optim/?author=mid-kid but wasn't upstreamed or finished. You may find any remaining bits there interesting too.

bin/ebuild-ipc.py
```diff
@@ -1,9 +1,12 @@
 #!/usr/bin/env python
-# Copyright 2010-2025 Gentoo Authors
+# Copyright 2010-2026 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 #
 # This is a helper which ebuild processes can use
 # to communicate with portage's main python process.
+#
+# It runs for every has_version and best_version call, so it uses the
+# standard library only: importing portage took most of the time of a run.
```
thesamesam:
```suggestion
# standard library only: importing portage would take most of the time of a run.
```

bin/isolated-functions.sh
```diff
@@ -33,6 +33,22 @@ __pipestatus() {
 
 shopt -s extdebug
 
+# The client imports nothing from portage, so it needs neither PYTHONPATH
```
thesamesam, line 36: The 'runs with -S' part is redundant, so maybe just:
```suggestion
# Doesn't use anything except stdlib, so no PYTHONPATH etc.
```

lib/_emerge/EbuildIpcDaemon.py
```diff
@@ -57,24 +67,13 @@ def _input_handler(self):
...
-                # ebuild process if it is waiting for a reply
-                # when we try to kill the ebuild process.
-                reply_hook = getattr(cmd_handler, "reply_hook", None)
-                if reply_hook is not None:
-                    reply_hook()
+                # The command may have a hook to execute after its reply
+                # has been sent. This hook is used by the 'exit' command
+                # to kill the ebuild process. For some reason, the
+                # ebuild-ipc helper hangs up the ebuild process if it is
+                # waiting for a reply when we try to kill the ebuild
```
thesamesam: I wonder if this is true still?

thesamesam (commented):
Some small comments, Zac will likely have better insight.

thesamesam:
I had a running merge when I upgraded Portage, and got:
```
>>> Original instance of package unmerged safely.
 * Updating .desktop files database ...
 [ ok ]
 * Updating icons cache ...
 [ ok ]
ebuild-ipc: daemon process not detected
 * ERROR: app-office/texstudio-4.9.6::gentoo failed (postinst phase):
 *   has_version: invalid atom: app-text/ghostscript-gpl
 *
 * Call stack:                                                                                                                                                                                                     *          ebuild.sh, line  143:  Called pkg_postinst
 *        environment, line 2231:  Called optfeature 'conversion tools' 'app-text/ghostscript-gpl'
 *        environment, line 2189:  Called has_version 'app-text/ghostscript-gpl'
 *   phase-helpers.sh, line  947:  Called ___best_version_and_has_version_common 'app-text/ghostscript-gpl'
 *   phase-helpers.sh, line  928:  Called die
 * The specific snippet of code:
 *                      die "${FUNCNAME[1]}: invalid atom: ${atom}"
 *
 * If you need support, post the output of `emerge --info '=app-office/texstudio-4.9.6::gentoo'`,
 * the complete build log and the output of `emerge -pqv '=app-office/texstudio-4.9.6::gentoo'`.
 * The complete build log is located at '/var/log/portage/build/app-office/texstudio-4.9.6:20260922-043701.log'.
 * For convenience, a symlink to the build log is located at '/var/tmp/portage/app-office/texstudio-4.9.6/temp/build.log'.
 * The ebuild environment file is located at '/var/tmp/portage/app-office/texstudio-4.9.6/temp/environment'.
 * Working directory: '/var/tmp/portage/app-office/texstudio-4.9.6/empty'
 * S: '/var/tmp/portage/app-office/texstudio-4.9.6/work/texstudio-4.9.6'
ebuild-ipc: daemon process not detected
 * The ebuild phase 'postinst' has exited unexpectedly. This type of
 *
```

I'd say this should be fine as long as it doesn't happen when Portage is included in the merge plan in the same process (the usual scenario).

mattst88:
Great catch! I think I've fixed that with the latest push.

bin/isolated-functions.sh
```diff
@@ -33,6 +33,22 @@ __pipestatus() {
...
+# nor a wrapper to set one up, and it runs with -S.
+__ebuild_ipc() {
+	"${PORTAGE_PYTHON:-/usr/bin/python}" -S "${PORTAGE_BIN_PATH:?}/ebuild-ipc.py" "$@"
```
zmedico: I wonder if we should `cd "${PORTAGE_PYM_PATH}"` here in order to ensure a safe working directory. In the past we have had problems with python import interference because python adds the working directory to sys.path.

bin/ebuild-ipc.py
```diff
@@ -42,279 +45,260 @@ def debug_signal(_signum, _frame):
...
             self.ipc_out_fifo = os.path.join(self.fifo_dir, ".ipc", "out")
             self.ipc_lock_file = os.path.join(self.fifo_dir, ".ipc", "lock")
+            alive_fd = os.environ.get("PORTAGE_IPC_ALIVE_FD")
+            self._alive_fd = None if alive_fd is None else int(alive_fd)
+            # Only a portage that predates PORTAGE_IPC_ALIVE_FD runs us
+            # without it: one that was already running when a newer
+            # portage was installed, whose phases then run these files.
+            # Its daemon and build directory are locked by portage.locks,
+            # which uses lockf() wherever lockf() works.
+            self._legacy = self._alive_fd is None
```
zmedico, line 96: I don't understand what would trigger legacy mode. Doesn't our upgrade process, which completely copies the running portage, prevent the kind of mixing that would trigger the legacy mode?

eli-schwartz: This is specifically to work around Sam's issue reported in https://github.com/gentoo/portage/pull/1711#issuecomment-5771406034

(Another portage running in another terminal.)

zmedico: I see. Thanks!

bin/isolated-functions.sh
mattst88: Python only puts the working directory on sys.path for `-c`, `-m`, or an interactive session. For a script, sys.path[0] is the script's directory (PORTAGE_BIN_PATH), so the working directory can't shadow the stdlib modules ebuild-ipc.py imports.

The old wrapper did also replace PYTHONPATH, which an ebuild could export and use to shadow the stdlib, so I now run the client with `-E`. That also ignores PYTHONUTF8, so I pass `-X utf8` too.

bin/ebuild-ipc.py
mattst88: Applied, thanks.

bin/isolated-functions.sh
mattst88: Dropped it. The comment now explains `-E` and `-X utf8` instead (see Zac's thread below).

lib/_emerge/EbuildIpcDaemon.py
mattst88: Not really, so I rewrote it. Only helpers from a portage without PORTAGE_EBUILD_EXIT_FD still send `exit`, and its hook just starts the timer that kills a phase that doesn't exit by itself. Starting that timer after the reply is sent keeps the phase from being killed while ebuild-ipc is still waiting for the reply.

thesamesam (commented):
Thanks! OK with me if Zac is happy.

If you get a chance, the filter environment thing from the branch I linked may be a good next thing to look at.

bin/isolated-functions.sh
zmedico: Thanks!

zmedico (approved):

mattst88:
> If you get a chance, the filter environment thing from the branch I linked may be a good next thing to look at.

will do!

PR #1712, opened 2026-09-22
vartree: don't execute phases w/o default definition if not defined, …

thesamesam:
…and no hooks

Some phases like pkg_postinst don't have a default definition, so if they're not in DEFINED_PHASES, there's no point in executing them unless there is a user-defined hook. We already apply an optimisation like this for pkg_pretend.

Candidate phases are:
* pkg_setup
* pkg_preinst
* pkg_postinst

I've omitted pkg_{pre,post}rm because it seems awkward to get the metadata for those, as they're for the replaced (or removed) package.

Check for hooks earlier and save all hooked phases in PORTAGE_HOOKED_PHASES and read that back before we execute such a phase if we were planning on otherwise skipping it. Stash the result of P_H_P in EbuildExecutor before running pkg_setup so we only call it once.

This assumes a hook isn't defined in bashrc based on EBUILD_PHASE, masking that there is one until it's too late. It doesn't seem likely anybody would do that.

Note that by stashing in EbuildExecutor and not running _aux_env_search again, ebuild(1) doesn't benefit from this, only emerge(1), like where 126553a76b10b20904f7145697629921e4803593 uses FEATURES=noauto to inhibit its optimisation. We could maybe change that at least for this new optimisation depending on feedback.

Timings:
```
$ sudo -E hyperfine \
	'/home/sam/bin/test-portage ~/git/portage/bin/emerge -v1O editor --buildpkg=n --usepkg=n --ignore-default-opts' \
	'/usr/bin/emerge -v1O editor --buildpkg=n --usepkg=n --ignore-default-opts'
Benchmark 1: /home/sam/bin/test-portage ~/git/portage/bin/emerge -v1O editor --buildpkg=n --usepkg=n --ignore-default-opts
  Time (mean ± σ):      5.024 s ±  0.033 s    [User: 3.785 s, System: 1.085 s]
  Range (min … max):    4.976 s …  5.086 s    10 runs

Benchmark 2: /usr/bin/emerge -v1O editor --buildpkg=n --usepkg=n --ignore-default-opts
  Time (mean ± σ):      6.059 s ±  0.046 s    [User: 4.411 s, System: 1.408 s]
  Range (min … max):    6.000 s …  6.115 s    10 runs

Summary
  /home/sam/bin/test-portage ~/git/portage/bin/emerge -v1O editor --buildpkg=n --usepkg=n --ignore-default-opts ran
    1.21 ± 0.01 times faster than /usr/bin/emerge -v1O editor --buildpkg=n --usepkg=n --ignore-default-opts
```

With Matt's IPC patches applied to both:
```
$ sudo -E hyperfine \
	'/home/sam/bin/test-portage ~/git/portage/bin/emerge -v1O editor --buildpkg=n --usepkg=n --ignore-default-opts' \
	'/usr/bin/emerge -v1O editor --buildpkg=n --usepkg=n --ignore-default-opts'
Benchmark 1: /home/sam/bin/test-portage ~/git/portage/bin/emerge -v1O editor --buildpkg=n --usepkg=n --ignore-default-opts
  Time (mean ± σ):      4.133 s ±  0.025 s    [User: 3.155 s, System: 0.956 s]
  Range (min … max):    4.104 s …  4.171 s    10 runs

Benchmark 2: /usr/bin/emerge -v1O editor --buildpkg=n --usepkg=n --ignore-default-opts
  Time (mean ± σ):      4.480 s ±  0.032 s    [User: 3.302 s, System: 1.154 s]
  Range (min … max):    4.449 s …  4.540 s    10 runs

Summary
  /home/sam/bin/test-portage ~/git/portage/bin/emerge -v1O editor --buildpkg=n --usepkg=n --ignore-default-opts ran
    1.08 ± 0.01 times faster than /usr/bin/emerge -v1O editor --buildpkg=n --usepkg=n --ignore-default-opts
```

Closes: https://bugs.gentoo.org/982327

---

This is much easier to review with `-w`.


zmedico (approved):

lib/_emerge/EbuildExecuter.py
```diff
@@ -72,28 +72,39 @@ def _start(self):
...
+            self.scheduler.scheduleSetup(setup_phase)
+        else:
+            self._setup_exit(None, True)
 
-    def _setup_exit(self, setup_phase):
-        if self._default_exit(setup_phase) != os.EX_OK:
+    def _setup_exit(self, setup_phase, skipped_setup=False):
+        if not skipped_setup and self._default_exit(setup_phase) != os.EX_OK:
             self.wait()
             return
```
zmedico, line 109: If _default_exit is not called then nothing will set self.returncode. We should set `self.returncode = 0` if _default_exit is not called.

zmedico: Nevermind, I was mistaken.

zmedico (approved):

lib/portage/package/ebuild/_config/special_env_vars.py
```diff
@@ -155,6 +155,7 @@
         "PORTAGE_FEATURES",
         "PORTAGE_GID",
         "PORTAGE_GRPNAME",
+        "PORTAGE_HOOKED_PHASES",
```
zmedico: Why does it need to be in environ_whitelist?

lib/_emerge/EbuildExecuter.py
```diff
@@ -72,28 +72,39 @@ def _start(self):
...
-            settings=settings,
-        )
+        self.settings["PORTAGE_HOOKED_PHASES"] = vardb._aux_env_search(
+            pkg.cpv, ["PORTAGE_HOOKED_PHASES"]
+        ).get("PORTAGE_HOOKED_PHASES", "")
+        # pkg_setup has no default definition, so elide it where
+        # unnecessary (inc. no hooks from user).
+        if (
+            "setup" in pkg.cpv._metadata["DEFINED_PHASES"]
+            or "pkg_setup" in self.settings["PORTAGE_HOOKED_PHASES"].split()
```
zmedico, line 91: Why not split DEFINED_PHASES like this:
```
"setup" in pkg.cpv._metadata["DEFINED_PHASES"].splt()
```

lib/_emerge/EbuildExecuter.py
```diff
@@ -72,28 +72,39 @@ def _start(self):
...
                 }
             )
 
-        setup_phase = EbuildPhase(
-            background=self.background,
-            phase="setup",
-            scheduler=scheduler,
-            settings=settings,
-        )
+        self.settings["PORTAGE_HOOKED_PHASES"] = vardb._aux_env_search(
```
mattst88, line 84: `_aux_env_search` searches `environment.bz2` from vdb, but the package isn't installed yet. I think that will cause this to return `{}`

I think this means that if you have a bashrc that defines `pre_pkg_setup`, `post_pkg_preinst`, or `post_pkg_postinst` they'll fail to run.

bin/phase-functions.sh
```diff
@@ -1000,6 +1000,16 @@ __ebuild_phase_funcs() {
 			fi
 			;;
 	esac
+
+	if ! [[ -v PORTAGE_HOOKED_PHASES ]] ; then
+		export PORTAGE_HOOKED_PHASES=
+		local phase
+		for phase in ${all_phases} ; do
+			if declare -F "pre_${phase}" >/dev/null || declare -F "post_${phase}" >/dev/null ; then
```
mattst88, line 1008: This only catches `pre_*`/`post_*` functions. Code that runs from the body of a bashrc based on `${EBUILD_PHASE}` isn't detected, and ::gentoo's own profiles do that for setup, e.g.:

- profiles/releases/23.0/profile.bashrc:36 dies if the profile is merged-usr but the disk is split-usr, or vice versa.

With this change those only run when the package happens to have a `pkg_setup`. User bashrcs that do case `${EBUILD_PHASE}` in postinst) ... have the same problem with preinst/postinst.

Separately, this runs once and then gets saved in the environment in vdb. So a hook the user adds later isn't picked up for packages that were built before it existed.

lib/_emerge/EbuildExecuter.py
```diff
@@ -72,28 +72,39 @@ def _start(self):
...
+            pkg.cpv, ["PORTAGE_HOOKED_PHASES"]
+        ).get("PORTAGE_HOOKED_PHASES", "")
+        # pkg_setup has no default definition, so elide it where
+        # unnecessary (inc. no hooks from user).
+        if (
+            "setup" in pkg.cpv._metadata["DEFINED_PHASES"]
```
mattst88: `def _async_start()` in lib/_emerge/EbuildPhase.py contains a conditional

```python
        if self.phase in ("nofetch", "pretend", "setup"):
```

inside which we print info at the start of every emerge:

```python
            msg.append(f"Package:    {self.settings.mycpv}:{self.settings['SLOT']}")
            if self.settings.get("PORTAGE_REPO_NAME"):
                msg.append(f"Repository: {self.settings['PORTAGE_REPO_NAME']}")
            if maint_str:
                msg.append(f"Maintainer: {maint_str}")
            if upstr_str:
                msg.append(f"Upstream:   {upstr_str}")

            msg.append(f"USE:        {use}")
```

I'm worried that skipping setup in this way is going to cause that to not be printed if the package doesn't define `pkg_setup`.

lib/_emerge/EbuildExecuter.py
```diff
@@ -72,28 +72,39 @@ def _start(self):
...
+            )
 
-        setup_phase.addExitListener(self._setup_exit)
-        self._task_queued(setup_phase)
-        self.scheduler.scheduleSetup(setup_phase)
+            setup_phase.addExitListener(self._setup_exit)
+            self._task_queued(setup_phase)
+            self.scheduler.scheduleSetup(setup_phase)
+        else:
+            self._setup_exit(None, True)
```
mattst88: I'd suggest `self._setup_exit(None, skipped_setup=True)` (I hate bare boolean arguments)

lib/portage/package/ebuild/_config/special_env_vars.py
thesamesam: Leftover from an earlier attempt, I'll drop it.

